PR #10, opened 2019-06-25
Fix vertex buffer handling

bghgary:


Library/Source/NativeEngine.cpp
```diff
@@ -450,12 +447,28 @@ namespace babylon
...
         const bgfx::VertexBufferHandle handle{ static_cast<uint16_t>(info[1].As<Napi::Number>().Uint32Value()) };
-        vertexArray.vertexBuffers.push_back(handle);
+        const uint32_t location = info[2].As<Napi::Number>().Uint32Value();
+        const uint32_t byteOffset = info[3].As<Napi::Number>().Uint32Value();
+        const uint32_t byteStride = info[4].As<Napi::Number>().Uint32Value();
+        const uint32_t numElements = info[5].As<Napi::Number>().Uint32Value();
+        const uint32_t type = info[6].As<Napi::Number>().Uint32Value();
+        const bool normalized = info[7].As<Napi::Boolean>().Value();
+
+        // REVIEW: is _asInt useful for decl.add?
```
syntheticmagus: Is this question answered? If not, should we address this now? Or open an issue for it?

bghgary: We will revisit this when fixing skinning as it will require integer arrays for joints.

PR #13, opened 2019-07-02
Implemented blend mode properly on native side...

syntheticmagus:
...and removed obsolete enum stuff from the JavaScript side.


Library/Source/NativeEngine.cpp
```diff
@@ -714,16 +751,7 @@ namespace babylon
...
-        {
-        case 0:
-            break;
-        case 25:
-            m_engineState |= BGFX_STATE_BLEND_NORMAL;
-            break;
-        default:
-            throw std::exception("Unsupported blend mode");
-        }
+        m_engineState |= ALPHA_MODE[blendMode];
```
bghgary: throw when alphamode returns 0?

syntheticmagus: Resolved by adding implementation of alpha interpolate.

PR #16, opened 2019-07-17
It compiles...with some commented code

CedricGuillemet:
# How to compile
- download Android Studio, install NDK
- in the repo: mkdir build, cd build
- cmake -DCMAKE_TOOLCHAIN_FILE="C:\Users\XXX\AppData\Local\Android\Sdk\ndk\20.0.5594570\build\cmake\android.toolchain.cmake" -DANDROID_NDK="C:\Users\XXX\AppData\Local\Android\Sdk\ndk\20.0.5594570" -DCMAKE_BUILD_TYPE=Release -DANDROID_ABI="arm64-v8a" -DANDROID_PLATFORM="android-21" -DCMAKE_CXX_FLAGS=-fexceptions -DANDROID_STL=c++_static -GNinja ..
- I use ninja but you can use any other build system
- if everything works fine, you will get "[2/2] Linking CXX static library Library\Dependencies\napi\libnapi.a"
- open Android project with Android Studio

# Issues

- I had to comment some template code. Not that many but I already spent some hours trying to fix them without success so feel free to find a solution :)
- It's just a lib, next step is linking. I'm starting an Android App, similar to TestApp
- libcurl compilation disable for now. I'll spend more time when it will be a blocker (link time?)

# What changed?
- mostly template stuff
- "ANDROID" value in CMake
- ~~namespace in cpp with using instead of code block~~ 
- std::exception not taking a string -> todo : create our own exception that takes a string a return it in what()
- Android project added at the same level as TestApp

# TODO
- [ ] output directories name based on ABI


Library/Source/RuntimeImpl.cpp
```diff
@@ -7,189 +7,190 @@
...
+    Task = Task.then(m_dispatcher, m_cancelSource, [this, url]
     {
-        Execute([this, string, sourceUrl](auto&)
-        {
-            Env().Eval(string.data(), sourceUrl.data());
-        });
-    }
-
-    void RuntimeImpl::Execute(std::function<void(RuntimeImpl&)> func)
+        //return LoadUrlAsync<std::string>(GetAbsoluteUrl(url).c_str()); todo
```
CedricGuillemet: I get around 10 errors when uncommenting this line. It seems related to Arcana

Library/Source/XMLHttpRequest.cpp
```diff
@@ -2,175 +2,178 @@
...
+    return Napi::Value::From(Env(), m_responseURL);
+}
 
-    Napi::Value XMLHttpRequest::GetResponseURL(const Napi::CallbackInfo&)
-    {
-        return Napi::Value::From(Env(), m_responseURL);
-    }
+Napi::Value XMLHttpRequest::GetStatus(const Napi::CallbackInfo&)
+{
+    return {};//Napi::Value::From(Env(), arcana::underlying_cast(m_status)); todo
```
CedricGuillemet: Error linked to WinRT. Will fix lated.

Library/Source/XMLHttpRequest.cpp
```diff
@@ -2,175 +2,178 @@
...
+    m_method = info[0].As<Napi::String>().Utf8Value();
+    m_url = m_runtimeImpl.GetAbsoluteUrl(info[1].As<Napi::String>().Utf8Value());
+    SetReadyState(ReadyState::Opened);
+}
 
-    void XMLHttpRequest::Open(const Napi::CallbackInfo& info)
+void XMLHttpRequest::Send(const Napi::CallbackInfo& info)
+{
+    auto lock = m_runtimeImpl.AcquireTaskLock();
+    /* todo
```
CedricGuillemet: Another bunch of errors when uncommenting. Definitely the same errors as the previous Arcana one.

Library/Source/NativeEngine.cpp
```diff
@@ -26,1152 +29,1151 @@ namespace bgfx
...
+    bgfx::setViewClear(0, BGFX_CLEAR_COLOR | BGFX_CLEAR_DEPTH, 0x443355FF, 1.0f, 0);
+    bgfx::setViewRect(0, 0, 0, m_size.Width, m_size.Height);
+}
 
-        // Scratch vector used for data alignment.
-        std::vector<float> m_scratch;
-    };
+void NativeEngine::Impl::Initialize(Napi::Env& env)
+{
+    //EngineDefiner::Define(env, this); todo
```
CedricGuillemet: I don't understand errors with this one. Adding Definition:: before the function name gave me different error

Library/Dependencies/napi/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.12)
 
 project(napi)
 
+if (ANDROID)
```
CedricGuillemet: I tried to put there 3 lines in root cmakelist.txt but it still tries to compile with Chakra? What did I miss?

Library/Source/NativeEngine.cpp
```diff
@@ -92,7 +96,7 @@ namespace babylon
 
                 if (memberType.basetype != spirv_cross::SPIRType::Float)
                 {
-                    throw std::exception("Not supported");
+                    throw std::exception();
```
syntheticmagus: Should we consider replacing this with a babylon::exception{} type that inherits from std::exception and has the behavior we'd been using, sort of like is done in the example in the docs?

http://www.cplusplus.com/reference/exception/exception/exception/

Library/Source/NativeEngine.cpp
```diff
@@ -398,7 +402,7 @@ namespace babylon
 
     void NativeEngine::Impl::Initialize(Napi::Env& env)
     {
-        EngineDefiner::Define(env, this);
+        //EngineDefiner::Define(env, this); todo
```
syntheticmagus: I think this _has_ to be called; if it isn't, Napi won't know what a NativeEngine is when Babylon asks for one. What was the error this was causing?

Library/Source/ResourceLimits.cpp
```diff
@@ -110,4 +110,5 @@ namespace babylon
             1, // generalConstantMatrixVectorIndexing
         }
     };
-}
+
```
syntheticmagus: nit: Revert?

Library/Source/RuntimeImpl.cpp
```diff
@@ -9,187 +9,190 @@
 
 namespace babylon
 {
-    RuntimeImpl::RuntimeImpl(void* nativeWindowPtr, const std::string& rootUrl)
```
syntheticmagus: This looks like the spacing is funky. Leftovers from the namespace thing?

Library/Source/Window.cpp
```diff
@@ -5,6 +5,7 @@
 
 namespace babylon
 {
+
```
syntheticmagus: nit: Revert?

Library/Source/NapiBridge.h
```diff
@@ -82,61 +82,61 @@ namespace babylon
...
-                .AddVoidReturningMethod<&NativeEngineT::SetTextureAnisotropicLevel>("setTextureAnisotropicLevel")
-                .AddVoidReturningMethod<&NativeEngineT::SetTexture>("setTexture")
-                .AddVoidReturningMethod<&NativeEngineT::DeleteTexture>("deleteTexture")
-                .AddVoidReturningMethod<&NativeEngineT::DrawIndexed>("drawIndexed")
-                .AddVoidReturningMethod<&NativeEngineT::Draw>("draw")
-                .AddVoidReturningMethod<&NativeEngineT::Clear>("clear")
-                .AddValueReturningMethod<&NativeEngineT::GetRenderWidth>("getRenderWidth")
-                .AddValueReturningMethod<&NativeEngineT::GetRenderHeight>("getRenderHeight")
-                .Finalize();
+                .NapiBridge<NativeEngineT>::AddVoidReturningMethod<&NativeEngineT::RequestAnimationFrame>("requestAnimationFrame")
```
syntheticmagus: This seems very weird, unless I'm reading this wrong. NapiBridge doesn't _have_ an AddVoidReturningMethod() method, does it? NapiBridge<T>::Definition has that method, not the parent class. Unless I've missed something...

Library/Source/Console.cpp
```diff
@@ -38,6 +40,7 @@ namespace babylon
             OutputDebugStringA(info[index].ToString().Utf8Value().c_str());
         }
         OutputDebugStringA("\r\n");
+#endif
```
syntheticmagus: nit: Else cout? Or stderr? Just so the message goes somewhere. I think standard outputs end up in logcat on Android (though that might be _very_ out-of-date information).

Library/CMakeLists.txt
```diff
@@ -37,8 +39,8 @@ set_property(TARGET bimg PROPERTY FOLDER Dependencies/bgfx)
 set_property(TARGET bx PROPERTY FOLDER Dependencies/bgfx)
 
 set(CMAKE_USE_WINSSL ON CACHE BOOL "Set cURL to use WinSSL by default.")
-add_subdirectory(Dependencies/curl EXCLUDE_FROM_ALL)
-set_property(TARGET libcurl PROPERTY FOLDER Dependencies/curl)
+#add_subdirectory(Dependencies/curl EXCLUDE_FROM_ALL)
+#set_property(TARGET libcurl PROPERTY FOLDER Dependencies/curl)
```
syntheticmagus: Is curl part of the Android OS? It looks like it might not be, which means we'll have to get our curl build to work for Android.

BabylonNative.Android/app/src/main/java/com/android/appviewer/AndroidViewAppActivity.java
```diff
@@ -0,0 +1,201 @@
+package com.android.appviewer;
```
CedricGuillemet: This is a copy from what I did in Base3D. Same for the view. Can we release it as OS?

Library/Source/NativeEngine.cpp
CedricGuillemet: Yes, that's the way to go

Library/Source/NativeEngine.cpp
CedricGuillemet: yep, I put a comment on that line. template madness here. I tried to fix without success

Library/Source/RuntimeImpl.cpp
CedricGuillemet: Might be, I'll do a clang format before merging anyway.

Library/Source/Console.cpp
CedricGuillemet: You are right, Logcat is not outdated info :). I'm taking a look now at how to organize Android platform dependent stuff.

Library/CMakeLists.txt
CedricGuillemet: We will have to! I forgot to update the description for that. It doesn't seem to be easily buildable with cmake but it might also be just some configuration that I'm missing.

PR #21, opened 2019-08-08
clang format script

CedricGuillemet:
.clang-format stub. 
bootstrapped from https://github.com/Microsoft/cpprestsdk/blob/master/.clang-format
3 possibilities:
- set VS to use it. It formats the code as you type.
- add a target in CMake to use clang-format
- make a ps script to format files

A good practice would be to do a clang format pass once your PR is approved and not spend time tracking poorly placed braces, missing spaces ,... during PR reviews

What's your take on that?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,47 @@
+# https://releases.llvm.org/7.0.0/tools/clang/docs/ClangFormatStyleOptions.html
+
+---
+Language: Cpp
+
+BasedOnStyle: WebKit
+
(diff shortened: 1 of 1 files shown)


bghgary:
I've never used clang format, so let's try one of these (say the first one) and see how it goes?

CedricGuillemet:
> Can you add some README instructions on how to install the proper extensions for clang format?

VS find the file by itself. Nothing to do.

PR #30, opened 2019-08-19
Android render

CedricGuillemet:
# New features

 - Android Native activity. Basically looks like a Win32 project (messages, main,..). This second Android project shows how to use libnative without the Java activity and content view. Depending on the needs, user might chose one or the other as a startup example
- width/height parameters for the runtime init. Suggest to change. I added the parameters because I needed them for the activity init.
- bgfx callback : bgfx trace messages are now routed to the console. testapp has to set the correct output depending on the platform . Outputting done for Android and Win32
- debug output of backend rendering api
- screen rotation support
- shader preprocessor: shaders work out of the box on d3d11, opengl, opengles
- library/bgfx/… build is handled by Android studio. No more the need to generate ninja build files

# Issues

~~- face winding seem inverted with opengl. I didn't find any flag in bgfx that declares it. Not a bug issue for now~~
~~- Vertex Attribute names: I had to change it manually because by default they are not remapped (and apparently no remapping functions in glslang)~~
~~- Uniform struct: for VS and FS, uniforms are part of a structure called 'Frame'. There is a size conflict  (and content) during linking. I'll try to make it work by renaming both structure (FrameVS and FrameFS) but another solution is needed~~
~~- uniform types: bgfx only allow vec4/mat4 uniforms. Any other type results in a runtime error and no update of the uniforms. I can't think of another solution of using an AST and parsing the shader source. Any idea?~~
~~- Shaders are not used the same depending on Android and Windows Opengl. For Windows, I made it work by adding #version 430. Not a stopping issue.~~
- Android event management/life cycle/app switching,... Main concern is the context switching. When switching to another app, the window (and its context) are destroyed. At least with the Java activity. I'll check with the native activity to see if it behaves the same. --> will be fixed in another PR


syntheticmagus:
Looks like Git has confused itself somehow about Win32/App.cpp. Can we revert whatever it thinks it did to that file?

TestApp/Source/Android/app/CMakeLists.txt
```diff
@@ -71,6 +69,12 @@ target_link_libraries(BabylonNativeJNI
         ${CMAKE_CURRENT_LIST_DIR}/../../../../build/Library/Dependencies/bgfx.cmake/libastc-codec.a
         ${CMAKE_CURRENT_LIST_DIR}/../../../../build/Library/Dependencies/SPIRV-Cross/libspirv-cross-core.a
         ${CMAKE_CURRENT_LIST_DIR}/../../../../build/Library/Dependencies/SPIRV-Cross/libspirv-cross-glsl.a
+        ${CMAKE_CURRENT_LIST_DIR}/../../../../build/Library/Dependencies/SPIRV-Cross/libspirv-cross-hlsl.a
```
syntheticmagus: This whole block doesn't appear in the CMakeLists.txt file for the Windows targets. Maybe this was already discussed in a prior PR -- if so, sorry for asking redundant questions -- but doesn't CMake do the linking for you so long as you specify the hierarchical relationship?

syntheticmagus: Also, does the Android TestApp need to be in its own separate CMake file? Win32 and UWP are both built from the same one, and I think our hope was to have all the platforms share as much as possible as an example to consumers trying to build cross-plat apps. Is there a way we could get Android to share more with the other platforms?

Library/Source/Console.h
```diff
@@ -16,6 +16,10 @@ namespace babylon
         static void RegisterWarnOutput(MessageLogger output);
         static void RegisterErrorOutput(MessageLogger output);
 
+        static void Log(const char* message);
```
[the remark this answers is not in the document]
CedricGuillemet: Everything is static because the outputs in the console is static. When the runtime outputs something, it gets difficult to retrieve the corresponding console. Moreover, it makes sense for that to be related to the platform, not to the runtime.

Library/Source/NativeEngine.cpp
```diff
@@ -34,6 +35,82 @@ namespace babylon
...
+                , bgfx::TextureFormat::Enum _format
+                , bool _yflip
+            ) 
+            {}
+            virtual void captureEnd() 
+            {}
+            virtual void captureFrame(const void* _data, uint32_t _size)
+            {}
+        };
+        static bgfxCallback _bgfxCallback;
```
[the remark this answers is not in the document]
CedricGuillemet: I agree, it will be better in its own cpp

Library/Source/NativeEngine.cpp
```diff
@@ -377,33 +457,76 @@ namespace babylon
...
+        case bgfx::RendererType::Vulkan:       //!< Vulkan
+            bgfxInitInfo += " - Vulkan";
+            break;
+        }
+        Console::Log(bgfxInitInfo.c_str());
     }
 
     void NativeEngine::Impl::Initialize(Napi::Env& env)
     {
+        InitializeRendering();
```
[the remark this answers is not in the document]
CedricGuillemet: I'll add a comment. Openg Context is constraint to 1 thread. Context init must happen in the same thread that's used for rendering.

Library/Source/NativeEngine.cpp
```diff
@@ -574,7 +697,23 @@ namespace babylon
                 {
                     const uint32_t location = compiler.get_decoration(stageInput.id, spv::DecorationLocation);
                     AppendBytes(vertexBytes, bgfx::attribToId(static_cast<bgfx::Attrib::Enum>(location)));
-                    attributeLocations[stageInput.name] = location;
+                    // name attribute has been preprocessed to make link between bgfx and the shader (position -> a_position)
+                    // but bjs doesn't know the attribute names. So the make the reverse here
```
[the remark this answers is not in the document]
CedricGuillemet: The JS preprocessor rename the attributes and change the type of some uniforms so the shaders can be loaded and bound in bgfx. BUT the js part of the engine is not aware of the new attribute names. So, they are changed back here so vertex arrays can be created in JS. Yes, it's a bit tricky.

TestApp/Source/AndroidNativeActivity/app/src/main/cpp/main.cpp
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * Process the next main command.
+ */
+static void engine_handle_cmd(struct android_app* app, int32_t cmd) {
+    struct engine* engine = (struct engine*)app->userData;
+    switch (cmd) {
+        case APP_CMD_SAVE_STATE:
+            // The system has asked us to save our current state.
+            break;
+        case APP_CMD_INIT_WINDOW:
```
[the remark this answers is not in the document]
CedricGuillemet: yes! when you rotate your phone, the native window is destroyed and you have to reinit the output with a new one.

Library/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -1,19 +1,106 @@
...
+    {
+        void AddShader(glslang::TProgram& program, glslang::TShader& shader, std::string_view source)
+        {
+            const std::array<const char*, 1> sources{ source.data() };
+            shader.setStrings(sources.data(), gsl::narrow_cast<int>(sources.size()));
+            shader.setEnvInput(glslang::EShSourceGlsl, shader.getStage(), glslang::EShClientOpenGL, 100);
+            shader.setEnvClient(glslang::EShClientOpenGL, glslang::EShTargetOpenGL_450);
+            shader.setEnvTarget(glslang::EShTargetSpv, glslang::EShTargetSpv_1_0);
+
+            // TODO: Do this to avoid the work around for dFdy?
```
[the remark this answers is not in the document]
CedricGuillemet: I keep the comment for future debugging

Library/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -1,19 +1,106 @@
...
+        {
+            std::vector<uint32_t> spirv;
+            glslang::GlslangToSpv(*program.getIntermediate(stage), spirv);
+
+            spirv_cross::Parser parser{ std::move(spirv) };
+            parser.parse();
+
+            auto compiler = std::make_unique<spirv_cross::CompilerGLSL>(parser.get_parsed_ir());
+
+            // remove "uniform Frame { .... }" . Keep the uniforms
```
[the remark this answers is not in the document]
CedricGuillemet: mop, just a comment telling what's being done. c++ string manipulation is not as explicit and clear as JS, or Python or... any other language actually :D

Library/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -1,19 +1,106 @@
...
+
+        glslang::TShader vertexShader{ EShLangVertex };
+        AddShader(program, vertexShader, vertexSource);
+
+        glslang::TShader fragmentShader{ EShLangFragment };
+        AddShader(program, fragmentShader, fragmentSource);
+
+        if (!program.link(EShMsgDefault))
+        {
+            Console::Error(program.getInfoDebugLog());
```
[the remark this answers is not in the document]
CedricGuillemet: yes, that seems like the right way to do it.

TestApp/Source/Android/app/CMakeLists.txt
CedricGuillemet: Not that CMake part because it's handled by gradle. It's very specific to Android. Gradle handles running cmake, making the apk, signing ,... Maybe we can change the build to only copy the .so. Not sure there is much value in doing that.

Library/Source/Console.h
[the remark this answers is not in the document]
syntheticmagus: I'm not sure I agree that having the outputs should belong to the platform directly instead of the Runtime. As I'm currently thinking of the problem, messages only make sense when associated with the entity that sent them; where they go is less important than where they came from. Furthermore, a _lot_ of new risks get introduced when systems become static, including lingering state and cross-contamination in the case of multiple instances, all of which can be avoided by keeping things instance-specific. As I currently understand it, I really think the console (perhaps under a different name) should be an object owned by the Runtime which the Runtime's possessions can write to and the Runtime's owner can read from. The owning application might then choose to dump all the console outputs into the same platform-level output location; that seems to me like the overwhelmingly most likely use case. But I don't think that's the Runtime's decision to make; it should only have authority over and insight into its own messages.

Library/Source/ShaderCompilerOpenGL.cpp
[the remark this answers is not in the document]
syntheticmagus: Right, but I don't think the hack referenced in this comment applies to OpenGL. The dFdy workaround was to deal with a difference between WebGL and DX regarding whether image space Y pointed up or down, which mattered in a few shaders that were calculating their own bitangents. That difference shouldn't exist (I don't think) between WebGL and OpenGL, so I suspect the comment shouldn't be here.

TestApp/Source/Android/app/CMakeLists.txt
syntheticmagus: Okay, I think there's definitely something about this build system we're not understanding the same way. This CMakeLists.txt file never calls add_library on the Babylon Native libraries; and, looking through the broader changes, I see no path for the top-level CMakeLists.txt to know that _this_ CMakeLists.txt exists at all. It looks like these two things are attempting to be two completely separate build systems, one for the library and one for the app, instead of a single top-level build system designed to make use of the [Android CMake toolchain](https://developer.android.com/ndk/guides/cmake#the_cmake_toolchain_file) to generate the build, sharing as much as possible with the TestApps from other platforms and including as few Android-specific resources as possible. I haven't dived into this deeply enough to fully understand what kind of platform-specific changes might be necessary, but I would not expect that our only alternative would be to run two separate builds and link all the shared libraries by hand (at which point I'm not sure there's any purpose in using CMake anymore). Does the Android CMake toolchain not give us some other options for how to do this?

CedricGuillemet: That's right! 2 build systems. 1 CMake to build the library (with android tool chain cmake) and 1 cmake called by gradle for the .cpp in the android studio project. I've added in this PR a second Android project that can be compiled entirely with cmake, without Android studio. it might be possible with some cmake install command to make the apk and run some adb command to push the binary on the device (simulator of real device). But for the project with java activity, I would not try :)
Now, the inverse might also be possible. That the gradle build running that cmake also builds the library (maybe you are refering to that mecanism). That would look more like VS solution building the library and the testapp.

Library/Source/NativeEngine.cpp
```diff
@@ -377,19 +377,25 @@ namespace babylon
 
         bx::DefaultAllocator m_allocator;
         uint64_t m_engineState;
-
+        void *m_nativeWindow{};
```
bghgary:
```suggestion
        void* m_nativeWindow{};
```

Library/CMakeLists.txt
```diff
@@ -25,6 +25,12 @@ add_compile_definitions(BGFX_CONFIG_DEBUG_UNIFORM=0)
 add_compile_definitions(BGFX_CONFIG_MULTITHREADED=0)
 add_compile_definitions(BGFX_CONFIG_MAX_VERTEX_STREAMS=32)
 add_compile_definitions(BGFX_CONFIG_MAX_COMMAND_BUFFER_SIZE=12582912)
+if(BABYLON_NATIVE_PLATFORM STREQUAL "Android")
+    add_compile_definitions(BGFX_CONFIG_RENDERER_OPENGLES=30)
+elseif()
+    add_compile_definitions(BGFX_CONFIG_RENDERER_OPENGL=31)
```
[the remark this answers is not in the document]
CedricGuillemet: It's used to set the version of opengl/es. not enabling it per platform.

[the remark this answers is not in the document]
syntheticmagus: Should we just not have an else, then? I don't think we're expecting to support OpenGL at all anywhere but Android, right?

Library/Source/NativeEngine.cpp
```diff
@@ -242,10 +245,17 @@ namespace babylon
 
         bgfx::setViewClear(0, BGFX_CLEAR_COLOR | BGFX_CLEAR_DEPTH, 0x443355FF, 1.0f, 0);
         bgfx::setViewRect(0, 0, 0, m_size.Width, m_size.Height);
-
         EngineDefiner::Define(env, this);
     }
 
+    void NativeEngine::Impl::Initialize(Napi::Env& env)
```
syntheticmagus: I feel like I might have already asked this and just forgotten the answer, but why is this a separate method? Is there a circumstance under which we want to call InitializeRendering() but not Initialize()? Or is there an intention to add additional functionality to Initialize() that's unrelated to rendering?

Library/Source/RuntimeAndroid.cpp
```diff
@@ -16,6 +17,18 @@ namespace babylon
...
+    void RuntimeAndroid::SetWindow(ANativeWindow* nativeWindowPtr)
+    {
+        bgfx::PlatformData pd;
+        pd.ndt          = NULL;
+        pd.nwh          = nativeWindowPtr;
+        pd.context      = NULL;
+        pd.backBuffer   = NULL;
+        pd.backBufferDS = NULL;
+        bgfx::setPlatformData(pd);
+        bgfx::reset(128, 128);
```
syntheticmagus: nit: constexpr named variable for magic number?

TestApp/Source/Android/app/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/src/")
 set(BABYLON_NATIVE_TESTAPP_DIR "${CMAKE_CURRENT_LIST_DIR}/../..")
 add_subdirectory(${BABYLON_NATIVE_TESTAPP_DIR}/../../Library/ ${BABYLON_NATIVE_TESTAPP_DIR}/../../build/Android_${CMAKE_ANDROID_ARCH_ABI}/)
 
+
```
syntheticmagus: nit: revert

Library/CMakeLists.txt
```diff
@@ -196,6 +191,19 @@ if(BABYLON_NATIVE_PLATFORM STREQUAL "Win32" OR BABYLON_NATIVE_PLATFORM STREQUAL
...
+
+if(WINDOWS_STORE)
+    target_link_libraries(Library PRIVATE arcana)
+endif()
+
+if(APPLE)
+    target_link_libraries(Library
+        PRIVATE spirv-cross-msl)
+endif()
+
```
bghgary: Why did this move? Merge mistake?

Library/Source/RuntimeAndroid.cpp
```diff
@@ -16,6 +17,18 @@ namespace babylon
         // android stub
     }
 
+    void RuntimeAndroid::SetWindow(ANativeWindow* nativeWindowPtr)
```
bghgary: What is the reason for a separate function here? Can it not go in the constructor?

Library/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -1,19 +1,105 @@
...
+            size_t pos = glsl.find(frame);
+            if (pos != std::string::npos)
+            {
+                glsl.replace(pos, frame.size(), "");
+                pos = glsl.find(frameEnd, pos);
+                if (pos != std::string::npos)
+                {
+                    glsl.replace(pos, frameEnd.size(), "");
+                }
+            }
```
bghgary: The `uniform Frame` thing is added on the JavaScript side. It seems like maybe we should not do some of the processing on the JavaScript side instead of undoing it here for OpenGL. Or maybe we should not target OpenGL at all and only support Vulkan.

Library/Source/ShaderCompilerVulkan.cpp
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            program.addShader(&shader);
+        }
+
+        std::unique_ptr<spirv_cross::Compiler> CompileShader(glslang::TProgram& program, EShLanguage stage, std::vector<uint32_t>& spirv)
+        {
+            glslang::GlslangToSpv(*program.getIntermediate(stage), spirv);
+
+
```
bghgary: nit: extra newline

azure-pipelines.yml
```diff
@@ -135,3 +135,38 @@ jobs:
...
+      git submodule update --init --recursive
+    displayName: 'Checkout dependencies'
+  - script: |
+      set -eu
+      curl -Ls https://github.com/ninja-build/ninja/releases/download/v1.9.0/ninja-mac.zip -o ninja-mac.zip
+      unzip ninja-mac.zip
+      sudo cp -v ninja /usr/local/bin/
+    displayName: 'Install Ninja'
+  - script: |
+      npm install --prefix . v8-android@7.5.1
```
bghgary: Shouldn't this go in the cmake for napi?

Library/Source/ShaderCompilerOpenGL.cpp
CedricGuillemet: I need the structure for Metal and Vulkan (maybe d3d also). Removing OpenGL is not a good idea as there are still many Android devices that do not support Vulkan properly (or at all). See https://gist.github.com/zeux/7d3ff56568c6b02de3a87cc69899d949

Library/CMakeLists.txt
CedricGuillemet: Yes, this branch is getting old

Library/Source/ShaderCompilerOpenGL.cpp
bghgary: Makes sense. I still think this processing should happen on the JS side. We can pass something to the JS side to tell it what to do with the shader processing.

Library/Source/RuntimeAndroid.cpp
CedricGuillemet: No, when there is a screen rotation/app switching, the window is destroyed and a new one is created.

Library/Source/ShaderCompilerOpenGL.cpp
CedricGuillemet: I agree. We can provide a JS function that returns the rendering API and do string/regex work based on that.

TestApp/Source/Android/app/CMakeLists.txt
CedricGuillemet: Fixed

Library/Source/ShaderCompilerVulkan.cpp
CedricGuillemet: Fixed

Library/Source/RuntimeAndroid.cpp
bghgary: Interesting. We should talk about the contract for this. Right now, it's not clear (at least to me) what you are supposed to do with this function.

azure-pipelines.yml
CedricGuillemet: yes, I think it's better for the users to let cmake do that work. But I don't think I'll do it in this PR tho.

Library/Source/NativeEngine.cpp
CedricGuillemet: Moved init code in Initialize and updated comment.

Library/CMakeLists.txt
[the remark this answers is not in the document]
CedricGuillemet: It easier and faster to debug opengl rendering on Windows than on Android. It should get less surprises now so I remove it.

azure-pipelines.yml
CedricGuillemet: I'll do it in this PR actually.

CedricGuillemet:
Closing PR as it's not meant to be merged in master.

Issue #6, opened 2019-06-11
Feature Backlog


syntheticmagus:
**BROWSER CAPABILITIES**
- [x] Complete implementation of XMLHttpRequest
- [x] Investigate Web workers
- [x] Investigate WebAssembly

**ARCHITECTURE**
- [x] Propose and implement proper InputManager

**NATIVE ENGINE**
- [x] Finalize Babylon Native contract
- [x] Correctly support device loss
- [x] Skyboxes
- [x] Fix stub for URL.createObjectURL in engine
- [x] Fix stub for URL.revokeObjectURL in engine
- [x] Fix stub for Blob in engine
- [x] Investigate BGFX multi-threading
- [x] drawElementsType and drawArrayType should be more robust
- [x] Refactor NativeEngine.ts to share more with other engines
- [x] Implement setViewport
- [x] Fix TODO's in NativeEngine.ts
- [x] Complete implementation of NativeEngine.createTexture
- [x] Implement instancing support
- [x] Implement XHR progress support
- [x] Implement GLB support
- [x] Ensure proper support for Unicode file paths
- [x] Revise `Tools.LoadScript` to not depend on `document`
- [x] Handle backbuffer, depth, and stencil in ViewClearState
- [x] Ensure BindFrameBuffer is setting all the necessary state
- [x] Investigate limiting frame buffer SetUpView to only happen after reset (may not have meaningful perf impact)
- [x] Refactor OpenXR support to minimize/eliminate per-frame allocations

**TEST INFRASTRUCTURE**
- [x] Continuous integration
- [x] Pass all asset generator tests
- [x] Pass all Babylon.js tests
- [x] Create test automation

**DOCUMENTATION**
- [x] Create React Native sample project
- [x] Create Vue sample project
- [x] Create InputManager sample project
- [x] Automate documentation (if API is large enough?)
- [x] Document Babylon Native contract
- [x] Investigate XAML example

**CROSS PLATFORM**
- [x] Support Android
- [x] Support iOS
- [x] Support macOS
- [x] Support XAML by implementing [ISwapChainPanel constructor](https://github.com/BabylonJS/BabylonNative/blob/ab9fd87a212eb574bd3c187f2eb46afaa22c57b5/Library/Include/Babylon/RuntimeUWP.h#L15-L16)
- [x] Ensure proper support for used texture types across platforms

**JAVASCRIPT INFRASTRUCTURE**
- [x] Enable V8 debugging
[...]

PR #41, opened 2019-08-30
Render target texture

syntheticmagus:
First working (limited) implementation of render target textures. Also includes changes to have clearing be controlled by Babylon.


Library/Source/NativeEngine.cpp
```diff
@@ -277,6 +282,109 @@ namespace babylon
...
+
+                    Update();
+                }
+
+                return needToUpdate;
+            }
+
+            void Update() const
+            {
+                // TODO: Backbuffer, depth, and stencil.
```
syntheticmagus: Added a line in https://github.com/BabylonJS/BabylonNative/issues/6

Library/Source/NativeEngine.cpp
```diff
@@ -1081,6 +1198,71 @@ namespace babylon
...
+    {
+        const auto frameBufferData = info[0].As<Napi::External<FrameBufferData>>().Data();
+        m_boundFrameBuffer = frameBufferData;
+
+        // TODO: Consider doing this only on bgfx::reset(); the effects of this call don't survive reset, but as
+        // long as there's no reset this doesn't technically need to be called every time the frame buffer is bound.
+        m_boundFrameBuffer->SetUpView();
+
+        // bgfx::setTexture()? Why?
+        // TODO: View order?
```
syntheticmagus: Added lines for both TODOs in https://github.com/BabylonJS/BabylonNative/issues/6.

Library/Source/NativeEngine.cpp
```diff
@@ -1081,6 +1198,71 @@ namespace babylon
...
+        {
+            auto depthStencilFormat = bgfx::TextureFormat::D32;
+            if (generateStencilBuffer)
+            {
+                depthStencilFormat = bgfx::TextureFormat::D24S8;
+            }
+
+            std::array<bgfx::TextureHandle, 3> textures
+            {
+                bgfx::createTexture2D(width, height, generateMipMaps, 1, TEXTURE_FORMAT[format], BGFX_TEXTURE_RT),
```
CedricGuillemet: Why  2 2D textures here?
Also, you have to check if the format is supported for rendering into
https://bkaradzic.github.io/bgfx/bgfx.html?highlight=istexturevalid#_CPPv4N4bgfx14isTextureValidE8uint16_tb8uint16_tN13TextureFormat4EnumE8uint64_t

Library/Source/NativeEngine.cpp
```diff
@@ -277,6 +282,109 @@ namespace babylon
...
+                color += static_cast<uint8_t>(m_alpha * std::numeric_limits<uint8_t>::max());
+                return color;
+            }
+        };
+
+        struct FrameBufferData final
+        {
+            FrameBufferData(bgfx::FrameBufferHandle frameBuffer, uint16_t width, uint16_t height)
+                : FrameBuffer{ frameBuffer }
+                , ViewId{ NextViewId++ }
```
CedricGuillemet: Check the limit for the viewid (max 256 if I recall correctly the default value). But you can use the same viewid for all the render target rendering if you call bgfx::frame when unbinding the target. The drawcalls will be issued at that time and you are free to use the same viewid.

CedricGuillemet: And no swap chain if view 0 is untouched. At the same time, maybe you can use a table of allocated viewids like bgfx does with allocated framebuffer.

Library/Source/NativeEngine.cpp
syntheticmagus: Good point. I used two color textures here because I was following (somewhat blindly) the pattern from the BGFX bloom and deferred lighting examples (https://github.com/bkaradzic/bgfx/blob/master/examples/21-deferred/deferred.cpp#L470-L489). It seems to work here with two, so I've switched it to that, will go up with the next push. But do we know why the examples use 3?

Library/Source/NativeEngine.cpp
syntheticmagus: Good call, there's a max value in the caps I can query. I've added an assert to do that, will come up with the next push. As for calling frame() every time, do we want to do that? I'm not 100% clear on all the moving pieces involved, but it seems like that solution doesn't quite jive with BGFX's existing utilities for views and draw call batching, so I'm wondering if there's a reason not to.

Library/Source/NativeEngine.cpp
CedricGuillemet: It's a gbuffer. The pixel shader writes to both target (like albedo in 1RT and normal in the other one)

Library/Source/NativeEngine.cpp
CedricGuillemet: 256 viewed for all the application life-time seems dangerous. What to do when you reach that value? I encounter that same issue with my pet project and that was the answer Branimir gave me https://github.com/bkaradzic/bgfx/issues/1866#issuecomment-523726872

Library/Source/NativeEngine.cpp
syntheticmagus: Makes sense. I've added asserts to check whether the format of the textures is supported, as I don't think we currently have a back-up plan for what to do if a format isn't supported.

Library/Source/NativeEngine.cpp
syntheticmagus: Spoke offline, added the recycling mechanism we discussed there.

Library/Source/NativeEngine.cpp
CedricGuillemet: We should create an issue for that. I had some issues on Base3D Android because of that and had to use some formats as replacement.

Library/Source/NativeEngine.cpp
```diff
@@ -222,6 +259,11 @@ namespace babylon
             // ALPHA_SCREENMODE: SRC + (1 - SRC) * DEST, SRC ALPHA + (1 - SRC ALPHA) * DEST ALPHA
             BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_INV_SRC_COLOR, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_INV_SRC_ALPHA),
         };
+
+        constexpr std::array<bgfx::TextureFormat::Enum, 1> TEXTURE_FORMAT
+        {
+            bgfx::TextureFormat::RGBA32F
```
CedricGuillemet: rgba32f is 4 floats per pixel. Is it really the intended format? usually, it's rgba8

Library/Source/NativeEngine.cpp
syntheticmagus: Added a line to #6. Hopefully BGFX's cross-platform capabilities and constraints will shield us from some issues like this; but if it doesn't, the asserts and the checkbox should ensure we don't forget.

Library/Source/NativeEngine.cpp
syntheticmagus: I thought it was a bit weird too, but I chose this one because the format requested by Babylon.js was (in its own terms) `Engine.TEXTUREFORMAT_RGBA`, not `Engine.TEXTUREFORMAT_RGBA_INTEGER`, which was also an option. I may have misinterpreted what those options mean, though; I'll ask (here or offline) @bghgary or one of the other JavaScript-side experts to confirm which format should be used to back `Engine.TEXTUREFORMAT_RGBA`.

syntheticmagus: Asked @deltakosh, turns out Babylon specifies whether a texture is byte or float in a different variable. I've put up another PR to Babylon.js to get that information properly sent to native, changes to native side coming up shortly.

Library/Source/NativeEngine.cpp
```diff
@@ -277,6 +320,155 @@ namespace babylon
...
+                    Update();
+                }
+
+                return needToUpdate;
+            }
+
+            void Update() const
+            {
+                // TODO: Backbuffer, depth, and stencil.
+                bgfx::setViewClear(m_viewId, BGFX_CLEAR_COLOR | (m_depth ? BGFX_CLEAR_DEPTH : 0x0), Color());
```
CedricGuillemet: Note: With opengl/webgl, if you bind a framebuffer and call clear, the buffer is cleared. With bgfx, I think the view is cleared if there is at least 1 drawcall. You may have to call bgfx::touch for a dummy DC.

Library/Source/NativeEngine.cpp
```diff
@@ -277,6 +320,155 @@ namespace babylon
...
+            {
+                assert(m_boundFrameBuffer == nullptr);
+                m_boundFrameBuffer = data;
+
+                // TODO: Consider doing this only on bgfx::reset(); the effects of this call don't survive reset, but as
+                // long as there's no reset this doesn't technically need to be called every time the frame buffer is bound.
+                m_boundFrameBuffer->SetUpView();
+
+                // bgfx::setTexture()? Why?
+                // TODO: View order?
```
CedricGuillemet: drawcalls can be reordered based on states, textures,... This is not a problem for a gbuffer with a depth buffer for solid geometry. But if you render things like GUI elements in a buffer with no depth buffer, this can be an issue.

Library/Source/NativeEngine.cpp
syntheticmagus: Okay. I don't currently have a test case that shows any issues related to that, at least as far as I know. I'd rather tackle that when I have a usage that shows the problem, so that I can know what kind of problem it is and that I've actually fixed it. Do you have a test case that exhibits this such that we can tackle it now, or should we cross that bridge when we come to it?

Library/Source/NativeEngine.cpp
syntheticmagus: Spoke offline, added https://github.com/BabylonJS/BabylonNative/issues/42 to make sure ordering work doesn't get lost.

syntheticmagus: Closed that issue, forgot I'd already added notes about these in #6 previously (https://github.com/BabylonJS/BabylonNative/pull/41#discussion_r319688614).

Library/Source/NativeEngine.cpp
CedricGuillemet: in your js sample, create the BABYLON.RenderTargetTexture, do the call to clear the view (let's say with purple color), don't render any mesh into it. Use the rendertargettexture as diffuse on a plane. The plane should be purple in your browser/playbground. but I think the view will not be cleared with bgfx without the touch and the plane will be black.

CedricGuillemet: it's a minor issue and should not prevent to merge and move forward :)

syntheticmagus: Good call, worked exactly as you said it would, and adding bgfx::touch() cleared it right up, fix confirmed. Thanks!

PR #49, opened 2019-09-17
Jsc apple

CedricGuillemet:


CedricGuillemet:
I get the box rendered. Many things to fix on the javascriptcore side(reference, private objects, ...). I also have  a crash in bgfx and it needs improvements on the metal shader loading. I didn't find anything in the shadercompiler to change the structure names. I'll look at shaderc. Anyway, it's in a state that allows me to spend some time on iOS compilation/CI. 

image

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -1,97 +1,220 @@
...
+    auto value = reinterpret_cast<JSValueRef>(v);
+    auto info = new RefInfo
+    {
+        value,
+        initial_refcount
+    };
+    
+    if (info->count != 0)
+    {
+        //CHECK_JSRT(JsAddRef(value, nullptr)); todo
```
bghgary: we need something to track that this implementation isn't complete. maybe add something to the roadmap or create an issue?

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -1,97 +1,220 @@
...
+struct VFTable
+{
+    napi_env env;
+    std::vector<VFTableEntry> vfTable;
+    std::map<std::string, VFTableProperty> properties;
+};
+
+struct JSC_cbInfo
+{
+  napi_callback cb;
```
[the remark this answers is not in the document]
bghgary: in general, we should try to match the style of the other js_native_api files (i.e. spacing, brace style, etc.)

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -1,97 +1,220 @@
 #include <napi/js_native_api.h>
+#include <napi/js_native_api_types.h>
+#include "JavaScriptCore/JavaScriptCore.h"
+#include "js_native_api_JavaScriptCore.h"
+#include <string>
+#include <vector>
+#include <map>
+#include <napi/napi.h>
+
+struct VFTableEntry
```
bghgary: let's get on a call and go over this. mucking with vtables is difficult to review without context ;)

Library/Source/NativeEngine.cpp
```diff
@@ -768,7 +767,22 @@ namespace babylon
                 {
                     const uint32_t location = compiler.get_decoration(stageInput.id, spv::DecorationLocation);
                     AppendBytes(vertexBytes, bgfx::attribToId(static_cast<bgfx::Attrib::Enum>(location)));
-                    attributeLocations[stageInput.name] = location;
+                    
+                    std::string attributeName = stageInput.name;
```
bghgary: is this string replacement necessary for metal?

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
CedricGuillemet: I created issues for JSC and metal. 
https://github.com/BabylonJS/BabylonNative/issues/54
https://github.com/BabylonJS/BabylonNative/issues/53

Library/Source/NativeEngine.cpp
CedricGuillemet: Yes! same as opengl. Metal shader is text (not binary like d3d) so attributes have to match between bgfx and Babylon.

bghgary: Maybe we can add something to bgfx to allow it to use different names.

CedricGuillemet: Hardcoded names in bgfx https://github.com/bkaradzic/bgfx/blob/c6fc1b069d8f88646c702987ad1594c6e5d033ee/src/renderer_gl.cpp#L44
We can ask Branimir if id would accept a PR to allow for name change

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
bghgary: I thought this was a vtable, but it's actually not, but nevertheless we talked and I will continue the review now.

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -1,353 +1,911 @@
...
+  assert(0);
+  return false;
 }
 
+JSValueRef JSCGetProperty(JSContextRef ctx, JSObjectRef object, JSStringRef propertyName, JSValueRef* exception) {
+  auto iter = FunctionTables.find(object);
+  assert(iter != FunctionTables.end());
+  
+  
+  char propertyStr[1024];
```
bghgary: Seems like you should call JSStringGetMaximumUTF8CStringSize instead of hard coded size

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -1,353 +1,912 @@
...
 
-napi_status napi_get_cb_info(
-    napi_env env,               // [in] NAPI environment handle
-    napi_callback_info cbinfo,  // [in] Opaque callback-info handle
-    size_t* argc,      // [in-out] Specifies the size of the provided argv array
-                       // and receives the actual count of args.
-    napi_value* argv,  // [out] Array of values
-    napi_value* this_arg,  // [out] Receives the JS 'this' arg for the call
-    void** data) {         // [out] Receives the data pointer for the callback.
+napi_status napi_define_class(napi_env env,
```
bghgary: should probably put a comment that this function should be fixed

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -1,353 +1,912 @@
...
-napi_status napi_get_element(napi_env env,
-                             napi_value object,
-                             uint32_t i,
-                             napi_value* result) {
+napi_status napi_get_array_length(napi_env env,
+                  napi_value v,
+                  uint32_t* result) {
+  assert(JSValueIsArray(env->m_globalContext, reinterpret_cast<JSObjectRef>(v)));
+  
+  bool hasLength = JSObjectHasProperty(env->m_globalContext, reinterpret_cast<JSObjectRef>(v), JSStringCreateWithUTF8CString("length"));
```
bghgary: this should be inlined with the assert if the intention is to not check it in release so that it won't execute JSObjectHasProperty in release

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -1,353 +1,911 @@
...
+  *result = escapee;
+  return napi_ok;
+}
+
 napi_status napi_coerce_to_string(napi_env env,
-                                  napi_value v,
-                                  napi_value* result) {
+                  napi_value v,
+                  napi_value* result) {
+  *result = v;
```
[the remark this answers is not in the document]
CedricGuillemet: It's ok until I find a use case that breaks it :)

CedricGuillemet:
I managed to remove the std::map for externals using a dummy class. I tried without success for the other 2 as the objectref are functions and not data

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
CedricGuillemet: Fixed

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
CedricGuillemet: Fixed

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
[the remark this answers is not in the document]
bghgary: does this get called? i would prefer putting an assert here if it's not being called.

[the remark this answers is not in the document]
CedricGuillemet: Yes, it's called.

Library/CMakeLists.txt
```diff
@@ -152,6 +153,11 @@ if(WINDOWS_STORE)
     target_link_libraries(Library PRIVATE arcana)
 endif()
 
+if(APPLE)
```
syntheticmagus: nit: What's the difference between checking APPLE, as you do here, versus checking BABYLON_NATIVE_PLATFORM as is done on line 28? I don't necessarily think changing this has to be done as part of this PR, especially since there are a lot of similar things in other places throughout our CMake code. (I'll open an issue so we can discuss, establish a pattern, and eventually propagate that everywhere.) But if there's a behavioral significance to the APPLE variable versus BGFX_NATIVE_PLATFORM, can you note it in a comment here?

syntheticmagus: https://github.com/BabylonJS/BabylonNative/issues/55

CedricGuillemet: Nice catch! I'll change line 28 in my next PR as I'm doing some IOS stuff now.

Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
CedricGuillemet: I prefer to update the issue on GH.

PR #50, opened 2019-09-18
azure pipelines jobified/macosx CI

CedricGuillemet:
macosx CI
CMake special flag for macosx (took me a while to find that one)


CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 3.12)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
+if (APPLE)
```
bghgary: Maybe move this after the `project` line

PR #51, opened 2019-09-23
Added OpenXR support to Win32

syntheticmagus:


Library/Source/XrPlugin.cpp
```diff
@@ -0,0 +1,215 @@
...
+            if (m_frame != nullptr)
+            {
+                EndFrame();
+            }
+
+            EndSession();
+        }
+    }
+
+    // TODO: Make this asynchronous.
```
syntheticmagus: Add an issue to track this work.

Library/Source/XrPlugin.cpp
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    void XrPlugin::EndSession()
+    {
+        assert(IsSessionActive());
+        assert(m_frame == nullptr);
+
+        m_session->RequestEndSession();
+
+        while (m_session->GetNextFrame() != nullptr);
```
syntheticmagus: After recent changes to Frame, this will actually never terminate because GetNextFrame() does not return nullptr anymore. Check the provided boolean values instead.

Library/Source/XrPlugin.cpp
syntheticmagus: #52

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+                XR_CHECK(xrAcquireSwapchainImage(handle, &acquireInfo, &swapchainImageIndex));
+        
+                XrSwapchainImageWaitInfo waitInfo{ XR_TYPE_SWAPCHAIN_IMAGE_WAIT_INFO };
+                waitInfo.timeout = XR_INFINITE_DURATION;
+                XR_CHECK(xrWaitSwapchainImage(handle, &waitInfo));
+        
+                return swapchainImageIndex;
+            };
+
+            Views.resize(viewCountOutput);
```
bghgary: same issue for allocation

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+        struct RenderResources
+        {
+            std::vector<XrView> Views;
+            std::vector<XrViewConfigurationView> ConfigViews;
+            std::vector<Swapchain> ColorSwapchains;
+            std::vector<Swapchain> DepthSwapchains;
+            std::vector<XrCompositionLayerProjectionView> ProjectionLayerViews;
+            std::vector<XrCompositionLayerDepthInfoKHR> DepthInfoViews;
+        };
+        std::unique_ptr<RenderResources> Resources{};
```
bghgary: std::optional?

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+                    renderResources->ProjectionLayerViews[idx].next = &renderResources->DepthInfoViews[idx];
+                }
+            }
+        }
+    }
+
+    System::Session::Frame::~Frame()
+    {
+        // EndFrame can submit mutiple layers
+        std::vector<XrCompositionLayerBaseHeader*> layers{};
```
bghgary: shouldn't allocate... or just be one pointer since there is only one

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+                XR_ASSERT(xrReleaseSwapchainImage(swapchain.Handle, &releaseInfo));
+            }
+        
+            // Inform the runtime to consider alpha channel during composition
+            // The primary display on Hololens has additive environment blend mode. It will ignore alpha channel.
+            // But mixed reality capture has alpha blend mode display and use alpha channel to blend content to environment.
+            layer.layerFlags = XR_COMPOSITION_LAYER_BLEND_TEXTURE_SOURCE_ALPHA_BIT;
+        
+            layer.space = m_sessionImpl.SceneSpace;
+            layer.viewCount = (uint32_t)renderResources->ProjectionLayerViews.size();
```
bghgary:
```suggestion
            layer.viewCount = static_cast<uint32_t>(renderResources->ProjectionLayerViews.size());
```

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+    {
+        return m_impl->IsInitialized();
+    }
+
+    bool System::TryInitialize()
+    {
+        return m_impl->TryInitialize();
+    }
+
+    std::unique_ptr<System::Session> System::CreateSession(void* graphicsDevice)
```
bghgary: Is this factory method necessary? Seems like the class can just be exposed.

Library/Source/NapiBridge.h
```diff
@@ -18,15 +18,15 @@ namespace babylon
             Definition(const Definition&) = delete;
             Definition& operator=(const Definition&) = delete;
 
-            template<void(ImplT::*method)(const Napi::CallbackInfo& info)>
-            Definition& AddVoidReturningMethod(const char* name)
+            template<void(ImplT::*method)(const Napi::CallbackInfo&)>
```
bghgary: Just revert it since it's going away.

Library/Dependencies/xr/Source/XrPlatformCommon.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <openxr/openxr.h>
+#include <openxr/openxr_platform.h>
+
+#define XR_CHECK(OPERATION) do { XrResult result = OPERATION; if (XR_FAILED(result)) throw std::exception{}; } while (false)
```
bghgary: should probably be a function instead of macro

Library/Source/XrPlugin.cpp
```diff
@@ -0,0 +1,215 @@
...
+    public:
+        static void Initialize(Napi::Env&);
+
+        XrPlugin::XrPlugin(const Napi::CallbackInfo& info);
+        ~XrPlugin();
+
+    private:
+        static Napi::FunctionReference constructor;
+
+        void BeginSession(const Napi::CallbackInfo&); // TODO: Make this asynchronous.
```
bghgary: file an issue?

Library/Source/XrPlugin.cpp
```diff
@@ -0,0 +1,215 @@
...
+        void EndSession(const Napi::CallbackInfo&); // TODO: Make this asynchronous.
+        void EndSession();
+        bool IsSessionActive() const;
+        void BeginFrame(const Napi::CallbackInfo&);
+        void EndFrame(const Napi::CallbackInfo&);
+        void EndFrame();
+        Napi::Value GetActiveFrameBuffers(const Napi::CallbackInfo& info);
+
+        xr::System m_hmd{};
```
bghgary:
```suggestion
        xr::System m_system{};
```

Library/Source/XrPlugin.cpp
```diff
@@ -0,0 +1,215 @@
...
+        std::unique_ptr<xr::System::Session> m_session{};
+        std::unique_ptr<xr::System::Session::Frame> m_frame{};
+        FrameBufferManager& m_frameBufferManager;
+        std::vector<FrameBufferData*> m_activeFrameBuffers{};
+
+        std::map<uintptr_t, std::unique_ptr<FrameBufferData>> m_texturesToFrameBuffers{};
+    };
+
+    Napi::FunctionReference XrPlugin::constructor;
```
bghgary: this can be done in the class declaration

Library/Source/XrPlugin.cpp
```diff
@@ -0,0 +1,215 @@
...
+            env,
+            "XrPlugin",
+            {
+                InstanceMethod("beginSession", &XrPlugin::BeginSession),
+                InstanceMethod("endSession", &XrPlugin::EndSession),
+                InstanceMethod("beginFrame", &XrPlugin::BeginFrame),
+                InstanceMethod("endFrame", &XrPlugin::EndFrame),
+                InstanceMethod("getActiveFrameBuffers", &XrPlugin::GetActiveFrameBuffers)
+            },
+            nullptr);
```
bghgary: defaults nullptr

Library/Source/XrPlugin.cpp
```diff
@@ -0,0 +1,215 @@
...
+        XrPlugin::XrPlugin(const Napi::CallbackInfo& info);
+        ~XrPlugin();
+
+    private:
+        static Napi::FunctionReference constructor;
+
+        void BeginSession(const Napi::CallbackInfo&); // TODO: Make this asynchronous.
+        void EndSession(const Napi::CallbackInfo&); // TODO: Make this asynchronous.
+        void EndSession();
+        bool IsSessionActive() const;
```
bghgary: doesn't seem like it's necessary

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+                throw std::exception{ "SystemId initialization failed with unexpected result type." };
+            }
+
+            // Find the available environment blend modes.
+            uint32_t count;
+            XR_CHECK(xrEnumerateEnvironmentBlendModes(Instance, SystemId, VIEW_CONFIGURATION_TYPE, 0, &count, nullptr));
+            std::vector<XrEnvironmentBlendMode> environmentBlendModes(count);
+            XR_CHECK(xrEnumerateEnvironmentBlendModes(Instance, SystemId, VIEW_CONFIGURATION_TYPE, count, &count, environmentBlendModes.data()));
+
+            // TODO: Support preferentially targeting one blend mode over another.
```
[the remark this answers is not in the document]
syntheticmagus: I don't think so. That TODO came over from the sample code, but there are so many features like that (where Babylon Native doesn't need it yet and we don't have a way to test it anyway, but OpenXR theoretically supports it) that making TODOs and issues for them probably wouldn't be helpful. Unless you have a different preference, I've changed the comment from a TODO to a statement declaring that the feature is not supported yet; we can then come back and add support for it if/when it becomes a part of a use case.

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+            EnvironmentBlendMode = environmentBlendModes[0];
+
+            return true;
+        }
+    };
+
+    struct System::Session::Impl
+    {
+        const System::Impl& HmdImpl;
+        XrSession Session{ XR_NULL_HANDLE };
```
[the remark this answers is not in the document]
syntheticmagus: Discussed offline, concluded this is sufficient for an Impl, since it never leaves this CPP.

Library/Dependencies/xr/Source/System.cpp
```diff
@@ -0,0 +1,621 @@
...
+            Resources->Views.resize(viewCount, { XR_TYPE_VIEW });
+        }
+
+        std::unique_ptr<System::Session::Frame> GetNextFrame()
+        {
+            bool exitRenderLoop;
+            bool requestRestart;
+            ProcessEvents(&exitRenderLoop, &requestRestart);
+
+            return std::make_unique<Frame>(*this, exitRenderLoop, requestRestart);
```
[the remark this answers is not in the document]
syntheticmagus: Added an item in https://github.com/BabylonJS/BabylonNative/issues/6

Library/Source/XrPlugin.cpp
syntheticmagus: https://github.com/BabylonJS/BabylonNative/issues/52

Library/Source/NativeEngineImpl.h
```diff
@@ -0,0 +1,383 @@
...
+
+        bx::DefaultAllocator m_allocator;
+        uint64_t m_engineState;
+        ViewClearState m_viewClearState;
+
+        FrameBufferManager m_frameBufferManager{};
+
+        void* m_nativeWindowPtr{};
+
+        // Scratch vector used for data alignment.
```
syntheticmagus: aerfjyus

PR #58, opened 2019-09-30
CI for Android JNI

CedricGuillemet:
android azure pipelines


azure-pipelines.yml
```diff
@@ -98,3 +98,40 @@ jobs:
       configuration: 'Release'
       msbuildArgs: '/p:AppxPackageSigningEnabled=false'
     displayName: 'Build UWP'
+    
+- job: androidJNI
+  pool:
+    vmImage: 'macOS-10.14'
```
bghgary: is this right?

TestApp/Source/Android/app/build.gradle
```diff
@@ -5,12 +5,31 @@ android {
...
 
     defaultConfig {
-        applicationId 'com.android.appviewer'
+        applicationId 'com.android.testapp'
         minSdkVersion "${platformVersion}"
         targetSdkVersion 28
         externalNativeBuild {
             cmake {
-                arguments '-DANDROID_STL=c++_static'
+                arguments '-DANDROID_STL=c++_static',
```
bghgary: nit: single quote here, but then double quote everything else

azure-pipelines.yml
CedricGuillemet: Yes! More tools are available on the mac image for Android. It can also works on Windows vm but Gb of tools have to be unzipped and installed.

TestApp/Source/Android/app/build.gradle
CedricGuillemet: good catch!

PR #60, opened 2019-10-09
Native Validation Tests

CedricGuillemet:
This is the first PR for the validationTests. 
New ValidationTests app and reorganized project.
The validation_native.js reads the tests from BabylonJS and will not run properly for now. Only a tests subset runs. 
I didn't provide a new Babylon.max.js with my changes. This will be in an upcoming PR.


Apps/ValidationTests/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+        "Win32/App.rc"
+        "Win32/Resource.h"
+        "Win32/small.ico"
+        "Win32/targetver.h")
+endif()
+
+if(WIN32)
+    set(WIN32_EXECUTABLE TRUE)
+    add_executable(ValidationTests WIN32 ${SOURCES} ${RESOURCE_FILES})
+endif()
```
syntheticmagus: Should we throw out a readable message here indicating that validation tests aren't supported on a given platform? Just to make it clear what went wrong in case someone tries.

Apps/ValidationTests/Win32/Resource.h
```diff
@@ -0,0 +1,30 @@
+//{{NO_DEPENDENCIES}}
+// Microsoft Visual C++ generated include file.
```
syntheticmagus: Should this be checked in? We have it checked in in the other place, so I assume so, but just thought I'd check to be sure.

Library/Source/BGFXCallback.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <bx/bx.h>
+#include <bx/string.h>
+#include <bx/platform.h>
+#include <bx/debug.h>
+#include <stdarg.h>
+#include <bgfx/bgfx.h>
+
+namespace Babylon
+{
+    std::vector<uint8_t> BGFXCallback::screenShotBitmap;
```
syntheticmagus: I think you can `static inline` in the header to avoid having to do this.

Library/Source/BGFXCallback.h
```diff
@@ -0,0 +1,36 @@
...
+        virtual void profilerEnd() override;
+        virtual uint32_t cacheReadSize(uint64_t id) override;
+        virtual bool cacheRead(uint64_t id, void* data, uint32_t size) override;
+        virtual void cacheWrite(uint64_t id, const void* data, uint32_t size) override;
+        virtual void screenShot(const char* filePath, uint32_t width, uint32_t height, uint32_t pitch, const void* data, uint32_t size, bool yflip) override;
+        virtual void captureBegin(uint32_t width, uint32_t height, uint32_t pitch, bgfx::TextureFormat::Enum format, bool yflip) override;
+        virtual void captureEnd() override;
+        virtual void captureFrame(const void* _data, uint32_t _size) override;
+
+        static std::vector<uint8_t> screenShotBitmap;
```
syntheticmagus: What's the motivation for having this be static? Could it not be a member instead?

Library/Source/NativeEngine.h
```diff
@@ -384,6 +390,7 @@ namespace Babylon
         bx::DefaultAllocator m_allocator;
         uint64_t m_engineState;
 
+        inline static BGFXCallback m_bgfxCallback{};
```
syntheticmagus: nit: Should static variables have the `m_` name styling?

syntheticmagus: Does this need to be static?

Library/Source/BGFXCallback.h
```diff
@@ -0,0 +1,36 @@
...
+#define alloca(size)   __builtin_alloca(size)
+#endif
+
+#include <bgfx/bgfx.h>
+#include <bgfx/platform.h>
+#include <vector>
+
+namespace Babylon
+{
+    struct BGFXCallback : public bgfx::CallbackI
```
syntheticmagus: nit: Consider renaming to `BgfxCallback` to match our treatment of N-API.

Library/Source/BGFXCallback.cpp
```diff
@@ -0,0 +1,103 @@
...
+        {
+            trace(filePath, line, "BGFX 0x%08x: %s\n", code, str);
+            BX_UNUSED(code, str);
+            abort();
+        }
+    }
+
+    void BGFXCallback::traceVargs(const char* filePath, uint16_t line, const char* format, va_list argList)
+    {
+        char temp[2048];
```
syntheticmagus: nit: Magic number? Split out into `constexpr` variable?

Library/Source/BGFXCallback.cpp
```diff
@@ -0,0 +1,103 @@
...
+        char* out = temp;
+        va_list argListCopy;
+        va_copy(argListCopy, argList);
+        int32_t len = bx::snprintf(out, sizeof(temp), "%s (%d): ", filePath, line);
+        int32_t total = len + bx::vsnprintf(out + len, sizeof(temp) - len, format, argListCopy);
+        va_end(argListCopy);
+        if ((int32_t)sizeof(temp) < total)
```
syntheticmagus: nit: `assert` if this check fails?

Library/Source/NativeWindow.cpp
```diff
@@ -98,6 +109,14 @@ namespace Babylon
         return Napi::Value::From(info.Env(), decodedData);
     }
 
+    void NativeWindow::AddEventListener(const Napi::CallbackInfo& info)
+    {
+    }
```
syntheticmagus: nit: Log out a warning or something if these methods ever get called? Just thinking it might be confusing down the line if we forget these are stubs.

Library/Source/RuntimeImpl.cpp
```diff
@@ -226,7 +228,7 @@ namespace Babylon
         m_env = &env;
         auto hostScopeGuard = gsl::finally([this] { m_env = nullptr; });
 
-        InitializeJavaScriptVariables();
+        InitializeJavaScriptVariables(600, 400);
```
syntheticmagus: What's the motivation for committing this? IIRC, we originally started using the 32x32 so that it would be painfully obvious if it wasn't getting updated properly, since it's supposed to be reset to something else immediately.

Apps/ValidationTests/CMakeLists.txt
CedricGuillemet: I'd rather put the info in the doc. Not sure anyone reads the cmake output when there is no error.

Library/Source/BGFXCallback.h
CedricGuillemet: because the callback function is static.

Library/Source/NativeEngine.h
CedricGuillemet: I didn't find any static member variable to copy the style. What about _ instead of m_? What do you propose?

Library/Source/RuntimeImpl.cpp
CedricGuillemet: I'll change it back to 32x32

Library/Source/BGFXCallback.cpp
CedricGuillemet: I don't see the point to use a constexpr here just for that value.

Library/Source/BGFXCallback.cpp
CedricGuillemet: What do you mean?

Library/Source/BGFXCallback.h
CedricGuillemet: I can try to get rid of those statics

Library/Source/NativeWindow.cpp
CedricGuillemet: I don't think I can do better than adding a comment here.

Library/Source/NativeEngine.h
CedricGuillemet: yes, I need it in void NativeEngine::InitializeWindow. Which is static.

Library/Source/BGFXCallback.h
CedricGuillemet: I removed that static

Apps/ValidationTests/Win32/Resource.h
CedricGuillemet: I need that .h for icon, app strings,...

Apps/ValidationTests/CMakeLists.txt
syntheticmagus: I meant have an else case here to throw a readable error message if somebody tries to build this project on a non-Windows platform.

Library/Source/NativeEngine.h
syntheticmagus: Okay. Probably fine for now, then; but long-term, I think we should move to having as little static state as possible. Maybe `NativeEngine::InitializeWindow` could return a state object, for example, which could hold onto this and have its lifecycle tied to the JavaScript context, or something like that.

Library/Source/BGFXCallback.cpp
syntheticmagus: Just so that that number can be put in a place that's easier to see from the top, and can be named. It's a nit, though, personal preference, not a problem either way.

Library/Source/BGFXCallback.cpp
syntheticmagus: Nevermind, misread the purpose of this. 😄

Library/Source/NativeWindow.cpp
syntheticmagus: Can you add a TODO? I'm currently thinking that once everything's plugin-ified, `NativeEngine` should take a dependency on `Console` anyway so it can log stuff out.

Library/Source/NativeEngine.h
syntheticmagus: No idea; I don't think we have a convention as I don't think we're hoping to have static members long-term. @bghgary, any suggestions for this?

CedricGuillemet: Using _ for now.

bghgary: Typical Hungarian notation uses `s_` but I'm not entirely sure we are following Hungarian notation.

Apps/Playground/CMakeLists.txt
```diff
@@ -0,0 +1,171 @@
+set(SCRIPTS
+    "Scripts/experience.js"
+    "Scripts/playground_runner.js")
+set(BABYLONJS
+    "../Babylonjs/babylon.glTF2FileLoader.js"
+    "../Babylonjs/babylon.max.js"
+    "../Babylonjs/babylonjs.materials.js")
```
bghgary: I don't like the Babylonjs folder name. Can we just call it Scripts?

Apps/Playground/CMakeLists.txt
```diff
@@ -0,0 +1,171 @@
...
+        add_custom_command(
+            OUTPUT "${script}"
+            COMMAND "${CMAKE_COMMAND}" -E copy "${CMAKE_CURRENT_SOURCE_DIR}/${script}" "${CMAKE_CURRENT_BINARY_DIR}/${script}"
+            COMMENT "Copying ${script}"
+            MAIN_DEPENDENCY "${CMAKE_CURRENT_SOURCE_DIR}/${script}")
+    endforeach()
+    
+    add_custom_command(TARGET Playground PRE_BUILD
+                   COMMAND ${CMAKE_COMMAND} -E copy_directory
+                   ${CMAKE_SOURCE_DIR}/Apps/Babylonjs/ "${CMAKE_CURRENT_BINARY_DIR}/Scripts")
```
bghgary: I can't remember if this works right or not, but I seem to remember `copy_directory` doesn't work well if only one file changed.

Apps/Playground/Win32/App.cpp
```diff
@@ -80,7 +80,7 @@ namespace
         }
         auto width = rect.right - rect.left;
         auto height = rect.bottom - rect.top;
-        runtime = std::make_unique<Babylon::RuntimeWin32>(hWnd, rootUrl, width, height);
+        runtime = std::make_unique<Babylon::RuntimeWin32>(hWnd, rootUrl, static_cast<float>(width), static_cast<float>(height));
```
bghgary: Might wanna merge from master which looks like this: https://github.com/BabylonJS/BabylonNative/blob/master/TestApp/Source/Win32/App.cpp#L84-L86

Apps/ValidationTests/Win32/Resource.h
```diff
@@ -0,0 +1,30 @@
...
+// Used by App.rc
+
+#define IDS_APP_TITLE			103
+
+#define IDR_MAINFRAME			128
+#define IDD_TESTAPPWIN32_DIALOG	102
+#define IDD_ABOUTBOX			103
+#define IDM_ABOUT				104
+#define IDM_EXIT				105
+#define IDI_TESTAPPWIN32			107
```
bghgary: In the Resource.h of both Playground and ValidationTests, we still have `#define`s that use TESTAPP in the name. Might consider renaming them to just APP or something.

Apps/ValidationTests/Win32/pch.h
```diff
@@ -0,0 +1,81 @@
+// stdafx.h : include file for standard system include files,
```
bghgary: hmm, i don't think we use pch.h anymore... should remove pch.h and pch.cpp from both projects.

Library/Source/BgfxCallback.h
```diff
@@ -0,0 +1,36 @@
...
+#include <bgfx/platform.h>
+#include <vector>
+
+namespace Babylon
+{
+    struct BgfxCallback : public bgfx::CallbackI
+    {
+        virtual ~BgfxCallback() = default;
+
+        virtual void fatal(const char* filePath, uint16_t line, bgfx::Fatal::Enum code, const char* str) override;
```
bghgary: Do the overrides need to be public? Usually I prefer making them private/protected.

Library/Source/NativeEngine.cpp
```diff
@@ -967,30 +970,22 @@ namespace Babylon
...
-
         bool useMipMap = false;
 
         if (invertY)
         {
             FlipYInImageBytes(gsl::make_span(static_cast<uint8_t*>(image.m_data), image.m_size), image.m_height, image.m_size / image.m_height);
         }
 
-        auto imageDataRef{bgfx::makeRef(image.m_data, image.m_size)};
+        auto imageDataRef{ bgfx::makeRef(image.m_data, image.m_size) };
```
bghgary: nit: why did this change? clang-format should not have added spaces here

Library/Source/NativeEngine.cpp
```diff
@@ -967,30 +970,22 @@ namespace Babylon
...
 
-        auto data = Napi::Int8Array::New(info.Env(), mb.getSize());
+        auto data = Napi::Uint8Array::New(info.Env(), mb.getSize());
         memcpy(data.Data(), static_cast<uint8_t*>(mb.more()), imageData->Image->m_size);
 
         return data;
     }
 
-    Napi::Value NativeEngine::LoadTexture(const Napi::CallbackInfo& info)
+    void NativeEngine::ImageToTexture(TextureData* const textureData, bimg::ImageContainer& image, bool invertY, bool mipMap) const
```
bghgary: nit: Functions typically start with a verb. Consider renaming to ConvertImageToTexture

Library/Source/NativeEngine.cpp
```diff
@@ -1011,9 +1006,52 @@ namespace Babylon
...
+        const auto textureData = info[0].As<Napi::External<TextureData>>().Data();
+        const auto width = static_cast<uint32_t>(info[2].As<Napi::Number>().Uint32Value());
+        const auto height = static_cast<uint32_t>(info[3].As<Napi::Number>().Uint32Value());
+        const auto formatIndex = info[4].As<Napi::Number>().Uint32Value();
+        const auto mipMap = info[5].As<Napi::Boolean>().Value();
+        const auto invertY = info[6].As<Napi::Boolean>().Value();
+
+        const void* data = nullptr;
+
+        switch(formatIndex)
```
bghgary: nit: probably should format the whole document?

Library/Source/NativeEngine.cpp
```diff
@@ -1011,9 +1006,52 @@ namespace Babylon
...
+            data = info[1].As<Napi::Uint8Array>().Data();
+            break;
+        case 1: // RGBA32F
+            data = info[1].As<Napi::Float32Array>().Data();
+            break;
+        default:
+            throw std::exception(); // unsupported format
+        }
+
+        if (data)
```
bghgary: this check is extraneous

Library/Source/NativeEngine.cpp
```diff
@@ -1187,7 +1225,7 @@ namespace Babylon
         const auto textureData = info[0].As<Napi::External<TextureData>>().Data();
         uint16_t width = static_cast<uint16_t>(info[1].As<Napi::Number>().Uint32Value());
         uint16_t height = static_cast<uint16_t>(info[2].As<Napi::Number>().Uint32Value());
-        bgfx::TextureFormat::Enum format = static_cast<bgfx::TextureFormat::Enum>(info[3].As<Napi::Number>().Uint32Value());
```
bghgary: lol, good one

Library/Source/NativeEngine.cpp
```diff
@@ -1327,6 +1365,47 @@ namespace Babylon
             static_cast<uint16_t>(height * backbufferHeight));
     }
 
+    Napi::Value NativeEngine::GetFramebufferData(const Napi::CallbackInfo& info)
+    {
+        bgfx::FrameBufferHandle fbh = BGFX_INVALID_HANDLE;
+        bgfx::requestScreenShot(fbh, "GetImageData");
+
+        while (s_bgfxCallback.m_screenShotBitmap.empty())
```
bghgary: This looks unsafe. Does `requestScreenShot` eventually call back on the same thread as this code? What probably should be happening is that the BgfxCallback::screenShot callback should post a message to the Babylon dispatcher somehow. Maybe NativeEngine should register with BgfxCallback so that BgfxCallback knows what objects need the callback. Currently, there might be race conditions here?

Library/Source/RuntimeImpl.h
```diff
@@ -50,8 +50,9 @@ namespace Babylon
         arcana::task<void, std::exception_ptr> Task = arcana::task_from_result<std::exception_ptr>();
         std::scoped_lock<std::mutex> AcquireTaskLock();
 
+        void* GetNativeWindow() const { return m_nativeWindowPtr; }
     private:
-        void InitializeJavaScriptVariables();
+        void InitializeJavaScriptVariables(uint16_t width, uint16_t height);
```
bghgary:
```suggestion
        void InitializeJavaScriptVariables(size_t width, size_t height);
```

NativeWindow::Create takes size_t

Library/Source/XMLHttpRequest.cpp
```diff
@@ -130,8 +130,19 @@ namespace Babylon
     {
         if (m_responseType.empty() || m_responseType == XMLHttpRequestTypes::ResponseType::Text)
         {
-            return m_runtimeImpl.LoadUrlAsync<std::string>(m_url).then(arcana::inline_scheduler, m_runtimeImpl.Cancellation(), [this](const std::string& data) {
-                m_responseText = std::move(data);
+            return m_runtimeImpl.LoadUrlAsync<std::string>(m_url).then(arcana::inline_scheduler, m_runtimeImpl.Cancellation(), [this](const std::string& data)
+            {
+                // check UTF-8 BOM encoding
```
bghgary: Do we only need to remove UTF-8 BOM? Are there other BOMs we need to consider?

Library/Source/XMLHttpRequest.cpp
```diff
@@ -130,8 +130,19 @@ namespace Babylon
...
+                if (data.size() >= 3 && data[0] == int8_t(0xEF) && data[1] == int8_t(0xBB) && data[2] == int8_t(0xBF))
```
bghgary: what's the `int8_t`s for?

Apps/Playground/CMakeLists.txt
CedricGuillemet: Not sure. IMHO 'Scripts' is a fine name for subfolders in ValidationTests and Playground where it contains...well..scripts. But for Babylonjs engine, I'd like a more expressive name. Like 'BabylonEngine', 'Engine' or 'Babylon'. It's more than just a bunch of small scripts.

Library/Source/XMLHttpRequest.cpp
CedricGuillemet: unsigned char(string) -> char conversion(constant)

Library/Source/XMLHttpRequest.cpp
CedricGuillemet: I don't know! It's the first case I've found and tested. I don't know if there are other BOM, nor how I can test it. If you have some doc on that, I'd be happy to learn more.

Library/Source/NativeEngine.cpp
CedricGuillemet: I can do that in a separate PR.

Apps/Playground/CMakeLists.txt
CedricGuillemet: Replaced with a file copy

Library/Source/NativeEngine.cpp
CedricGuillemet: No, a race condition here is not possible. Everything happens in the same thread. The screenshot callback is ran at the end of a bgfx::frame call. It happens a couple frames after the request. That's why there is a loop to wait for the data.

bghgary: This interface doesn't seem like it's intended to do what you are doing. Is there a way to render to a texture first and then call `readTexture`? If `requestScreenShot` is the best (only?) way to get frame buffer data, then I think I would abuse the filename parameter and store a callback instead. Then, there would be no need for the copy and storing of the screenshot.

Thoughts?

Library/Source/XMLHttpRequest.cpp
bghgary: I think LoadUrlAsync should be doing binary to text conversion. Looking at the implementation, my guess is that cURL doesn't do binary to string conversion after getting the data from the network. What if the data is Unicode or some other encoding?

Library/Source/XMLHttpRequest.cpp
bghgary: I see. You can do `\xEF` instead to avoid this, but see my other comment. We probably shouldn't be checking this at all and instead use something to convert binary to text to cover all the cases.

CedricGuillemet: Good idea! I'll take a closer look at curl api

Library/Source/NativeEngine.cpp
CedricGuillemet: best I think is to use a JS promise and pass the buffer as a JS napi bytearray. It would also be better when we try to do rendering in another thread.

bghgary: Okay, but still have to deal with the callback issue.

CedricGuillemet: I'll do a bgfx PR to add a void* userParam to the requestScreenshot method

CedricGuillemet: https://github.com/bkaradzic/bgfx/pull/2025

Apps/Playground/CMakeLists.txt
bghgary: I guess I'm expecting to see scripts other than Babylon in here. Anything that is in common between the apps go here? If it's strictly Babylon, then maybe BabylonScripts as the name?

CedricGuillemet: If it's common between apps and not part of Babylon, then it goes to another directory. Do you have a use case in mind?

bghgary: Do we consider 3rd party scripts (like ammo or Draco) a Babylon.js script?

CedricGuillemet: I'd consider ammo/draco to be part of the engine because they are a dependency of babylon

bghgary: Okay, are you okay BabylonScripts? I'm okay with Babylon also. Note that at some point we probably should create this script by pulling down from NPM and building a custom version using tree shaking.

Library/Source/XMLHttpRequest.cpp
bghgary: I added an item to issue #96 for this.

Apps/Playground/CMakeLists.txt
bghgary: Should use `copy_if_different`.

Library/Source/NativeEngine.cpp
bghgary: I'm assuming branimir wants you to update bgfx code to make `char* filename` to a `void* data`? Can we do the filename hack for now and then update this code once bgfx is updated?

PR #67, opened 2019-10-17
Remove NAPI bridge and refactor to compensate

syntheticmagus:
Fairly substantial refactor of NativeEngine and a few other files, including introducing a new pattern for global state owned by the RuntimeImpl. This PR does not propagate these new patterns everywhere (though we could do so), but attempts to only do enough to allow the clean removal of NapiBridge.h.


Library/Source/NativeEnginePlugin.h
```diff
@@ -0,0 +1,6 @@
+#include <napi/napi.h>
+
+namespace babylon
+{
+    Napi::Function CreateNativeEngineConstructor(Napi::Env&);
```
bghgary: i think we can just collapse this with NativeEngine.h

TestApp/Scripts/experience.js
```diff
@@ -46,6 +46,8 @@ function CreateInputHandling(scene) {
     });
 }
 
+var nativeEngine = new _native.Engine();
```
bghgary: Either add a comment or fix the js side.

Library/Source/RuntimeImpl.cpp
```diff
@@ -28,12 +47,15 @@ namespace babylon
...
-
-    void RuntimeImpl::UpdateRenderTarget()
-    {
-        m_dispatcher.queue([this] { m_engine->UpdateRenderTarget(); });
+        m_dispatcher.queue([width, height, this]
+        {
+            // If this query is showing up on profiles as being noticeably inefficient,
+            // a pointer to the window could simply be cached off as runtime state. That
+            // would add a little efficiency at the cost of a little clutter, so that
+            // tradeoff should be made only if profiles show that it makes sense.
```
bghgary: I don't think you need this comment. If this is even an issue, then it should be obvious what to do.

Library/Source/RuntimeImpl.cpp
```diff
@@ -175,6 +196,23 @@ namespace babylon
...
+    {
+        auto& env = *m_env;
+
+        auto jsNative = Napi::Object::New(env);
+        env.Global().Set(JS_NATIVE_NAME, jsNative);
+
+        auto jsRuntime = Napi::External<RuntimeImpl>::New(env, this);
+        jsNative.Set(JS_RUNTIME_NAME, jsRuntime);
+
+        auto jsWindow = NativeWindow::Create(env, m_nativeWindowPtr, 1024, 768);
```
bghgary: where did 1024 and 768 come from?

Library/Source/RuntimeImpl.h
```diff
@@ -5,19 +5,31 @@
...
-    class NativeEngine;
+    class NativeWindow;
 
-    class RuntimeImpl
+    class RuntimeImpl final
     {
+        static constexpr auto JS_NATIVE_NAME = "_native";
+        static constexpr auto JS_RUNTIME_NAME = "runtime";
+        static constexpr auto JS_WINDOW_NAME = "window";
+        static constexpr auto JS_ENGINE_CONSTRUCTOR_NAME = "Engine";
```
bghgary: nit: I don't like having fields not under a `<access>:` section. Do these even need to be in the header at all?

Library/Source/NativeWindow.cpp
```diff
@@ -0,0 +1,79 @@
...
+        constexpr auto JS_CLASS_NAME = "NativeWindow";
+
+        Napi::HandleScope scope{ env };
+
+        Napi::Function constructor = DefineClass(
+            env,
+            JS_CLASS_NAME,
+            {
+                InstanceAccessor("width", &NativeWindow::GetWidth, nullptr),
+                InstanceAccessor("height", &NativeWindow::GetHeight, nullptr)
```
bghgary: These aren't called from JS, so probably should remove until we need them.

PR #68, opened 2019-10-17
mipmap generation for PoT textures

CedricGuillemet:
rgb,rgba, bgra supported formats for PoT textures
Need more formats/NPoT?
tested with Sponza, FlightHelmet and RenderDoc

### DIFF
--- a/Library/Source/NativeEngine.cpp
+++ b/Library/Source/NativeEngine.cpp
@@ -862,6 +862,7 @@ namespace babylon
                 imageDataRef = imageDataRefMipMap;
                 useMipMap = true;
             }
+            // TODO: log an warning message: "Could not generate mipmap for texture"
         }
 
         textureData->Texture = bgfx::createTexture2D(
--- a/Library/Source/NativeEngine.cpp
+++ b/Library/Source/NativeEngine.cpp
@@ -151,6 +151,28 @@ namespace babylon
             }
         }
 
+        void GenerateMipmap(uint8_t* __restrict source, int size, int channels, uint8_t* __restrict dest)
+        {
+            int mipsize = size / 2;
+
+            for (int y = 0; y < mipsize; ++y)
+            {
+                for (int x = 0; x < mipsize; ++x)
+                {
+                    for (int c = 0; c < channels; ++c)
+                    {
+                        int index = channels * ((y * 2) * size + (x * 2)) + c;
+                        int sum_value = 4 >> 1;
+                        sum_value += source[index + channels * (0 * size + 0)];
+                        sum_value += source[index + channels * (0 * size + 1)];
+                        sum_value += source[index + channels * (1 * size + 0)];
+                        sum_value += source[index + channels * (1 * size + 1)];
+                        dest[channels * (y * mipsize + x) + c] = (uint8_t)(sum_value / 4);
+                    }
+                }
+            }
+        }
+
         enum class WebGLAttribType
         {
             BYTE = 5120,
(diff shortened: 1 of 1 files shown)


Library/Source/NativeEngine.cpp
```diff
@@ -151,6 +151,65 @@ namespace babylon
...
+        const bgfx::Memory* GenerateMipMaps(const bimg::ImageContainer& image)
+        {
+            auto imageDataRef{ bgfx::makeRef(image.m_data, image.m_size) };
+
+            bool widthIsPowerOf2 = ((image.m_width & (~image.m_width + 1)) == image.m_width);
+            bool supportedFormat = image.m_format == bimg::TextureFormat::RGB8 || image.m_format == bimg::TextureFormat::RGBA8 || image.m_format == bimg::TextureFormat::BGRA8;
+            if (image.m_width == image.m_height && image.m_width && widthIsPowerOf2 && supportedFormat)
+            {
+                auto channelCount = (image.m_format == bimg::TextureFormat::RGB8) ? 3 : 4;
+                unsigned int mipMapCount = static_cast<unsigned int>(log2(image.m_width));
```
bghgary: consider using std::log2

bghgary: do you need to cast here? looks like image.m_width is a `uint32_t`? maybe use `uint32_t` instead of `unsigned int`?

Library/Source/NativeEngine.cpp
```diff
@@ -151,6 +151,65 @@ namespace babylon
...
+                unsigned int mipmapImageSize = image.m_size;
+                for (auto i = 1; i <= mipMapCount; i++)
+                {
+                    mipmapImageSize += image.m_size >> (2 * i);
+                }
+                uint8_t* destination = new uint8_t [mipmapImageSize] ;
+                auto imageDataRef = bgfx::makeRef(destination, mipmapImageSize, [](void* _ptr, void* _userData) 
+                    { 
+                        delete [] _ptr; 
+                    });
```
bghgary: use bgfx::alloc instead

Library/Source/NativeEngine.cpp
```diff
@@ -799,14 +858,26 @@ namespace babylon
         textureData->Images.push_back(bimg::imageParse(&m_allocator, buffer.Data(), static_cast<uint32_t>(buffer.ByteLength())));
         auto& image = *textureData->Images.front();
 
+        bool useMipMap = false;
+        auto imageDataRef{ bgfx::makeRef(image.m_data, image.m_size) };
+        if (mipMap)
+        {
+            auto imageDataRefMipMap = GenerateMipMaps(image);
+            if (imageDataRefMipMap)
```
bghgary: might be good to log something when mipmaps are not generated when they should

Library/Source/NativeEngine.cpp
CedricGuillemet: log2 returns float or double

bghgary: Oops, yeah, forgot about the math part. lol

Library/Source/NativeEngine.cpp
CedricGuillemet: No console/log function is accessible to the C++ for now. It's interesting to create a console object and log(const char*) function but not in the scope of this PR. Adding a comment for now.

syntheticmagus: That'll likely be a part of the future change that will perpetuate the new pattern (used for NativeEngine and NativeWindow as of my inbound PR) to the other parts of the runtime-owned state. That change should make the console readily accessible from within any of our native JavaScript objects.

Library/Source/NativeEngine.cpp
```diff
@@ -151,6 +151,59 @@ namespace babylon
             }
         }
 
+        void GenerateMipmap(const uint8_t* const __restrict source, const int size, const int channels, uint8_t* __restrict dest)
```
syntheticmagus: Is `__restrict` a thing we can have here? I'd rather have these buffers be gsl::spans backed by vectors, just for the safety and clarity, as opposed to raw pointers; but that's more of a nit.

Library/Source/NativeEngine.cpp
```diff
@@ -151,6 +151,59 @@ namespace babylon
...
+                auto channelCount = (image.m_format == bimg::TextureFormat::RGB8) ? 3 : 4;
+                auto mipMapCount = static_cast<uint32_t>(std::log2(image.m_width));
+                auto mipmapImageSize = image.m_size;
+                for (auto i = 1; i <= mipMapCount; i++)
+                {
+                    mipmapImageSize += image.m_size >> (2 * i);
+                }
+                auto imageDataRef = bgfx::alloc(mipmapImageSize);
+                uint8_t* destination = imageDataRef->data;
+                uint8_t* source = static_cast<uint8_t*>(image.m_data);
```
syntheticmagus: You could still make gsl::spans with this (gsl::make_span), but it's probably not worth a check-in on its own.

Library/Source/NativeEngine.cpp
CedricGuillemet: It's definitely something useful here :) The performance difference is huge, especially with this case.
Do using  gsl::span prevent memory aliasing with uint8_t ?

bghgary: Is the performance huge from the __restrict or raw pointers?

bghgary: The main issue with `__restrict` (now that I've read more on it), is that it's not part of C++ standard. Different compilers use different symbols. If we really want this, it needs to be part of a macro definition.

gsl::span appears to add range checks to each access which would also be bad for performance here.

PR #71, opened 2019-10-22
Napi v8 forward integration

syntheticmagus:


Library/Dependencies/napi/include/napi/napi.h
```diff
@@ -1672,92 +1769,331 @@ namespace Napi {
...
+    napi_env _env;
+    napi_async_work _work;
+    ObjectReference _receiver;
+    FunctionReference _callback;
+    std::string _error;
+    bool _suppress_destruct;
+  };
+#endif // NODE_ADDON_API_DISABLE_NODE_SPECIFIC
+
+  #if (NAPI_VERSION > 3)
```
bghgary: is this supposed to be indented?

Library/Dependencies/napi/source/js_native_api_v8.cc
```diff
@@ -2879,7 +2941,6 @@ napi_status napi_run_script(napi_env env,
 
   v8::Local<v8::Context> context = env->context();
 
-  // $TODO: pass in sourceUrl
```
bghgary: move this to the other napi_run_script?

PR #72, opened 2019-10-22
MVP hack fix : texture filtering + addressing modes + Anisotropic + viewport

CedricGuillemet:


Library/Source/NativeEngine.cpp
```diff
@@ -1188,14 +1254,44 @@ namespace babylon
...
+    void NativeEngine::SetViewPort(const Napi::CallbackInfo& info)
+    {
+        const auto x = info[0].As<Napi::Number>().FloatValue();
+        const auto y = info[1].As<Napi::Number>().FloatValue();
+        const auto width = info[2].As<Napi::Number>().FloatValue();
+        const auto height = info[3].As<Napi::Number>().FloatValue();
+
+        const auto backbufferWidth = bgfx::getStats()->width;
+        const auto backbufferHeight = bgfx::getStats()->height;
+        auto newViewId = m_viewidSet.Get();
```
bghgary: I don't understand this. Why do we need to keep ids around for view ports? Can we have more than one?

Library/Source/NativeEngine.h
```diff
@@ -294,8 +301,6 @@ namespace babylon
 
     private:
         enum BlendMode {}; // TODO DEBUG
```
bghgary: This can be removed too.

Library/Source/NativeEngine.cpp
bghgary: Also, it seems like the recycle set is being passed to the FrameBufferManager also. Not sure this is right? It will affect CreateFrameBuffer and XR stuff.

Library/Source/NativeEngine.cpp
```diff
@@ -1188,14 +1254,44 @@ namespace babylon
...
+        m_viewportIds.push_back(newViewId);
+        m_currentBackbufferViewId = newViewId;
+        bgfx::setViewFrameBuffer(m_currentBackbufferViewId, BGFX_INVALID_HANDLE);
+        bgfx::setViewClear(m_currentBackbufferViewId, BGFX_CLEAR_COLOR | BGFX_CLEAR_DEPTH, 0x443355FF, 1.0f, 0);
+        if (bgfx::getCaps()->originBottomLeft)
```
bghgary: seem like most of this code is the same except for the y flip? maybe only branch for that one line?

TestApp/Scripts/experience.js
```diff
@@ -26,6 +27,99 @@ function CreateSpheresAsync() {
     return Promise.resolve();
 }
 
+function CreatePlane(width, height, uvScale) {
```
bghgary: Use BABYLON.Mesh.CreatePlane?

TestApp/Scripts/experience.js
```diff
@@ -68,10 +164,35 @@ CreateBoxAsync().then(function () {
 //BABYLON.SceneLoader.AppendAsync("https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/CesiumMan/glTF/CesiumMan.gltf").then(function () {
     BABYLON.Tools.Log("Loaded");
 
-    scene.createDefaultCamera(true);
-    scene.activeCamera.alpha += Math.PI;
+    if (viewports) {
+        // camera positions set for sponza
+        var camera1 = new BABYLON.FreeCamera("camera1", new BABYLON.Vector3(3, 2, -6), scene);
```
bghgary: how do you switch cameras?

CedricGuillemet: You can change the active camera this way:
scene.activeCamera = ...

Library/Source/NativeEngine.cpp
CedricGuillemet: a bgfx view only store 1 viewRect/1 framebuffer. To use more viewRects, we use more views. Multiple view can access the same frameBuffer. To not mixup view ids during a frame, we have to mutualize the ids with the FrameBufferManager.

Library/Source/NativeEngine.cpp
```diff
@@ -421,19 +423,29 @@ namespace babylon
...
+            {
+                bgfx::setVertexBuffer(index, vertexBuffer.handle, vertexBuffer.startVertex, UINT32_MAX, vertexBuffer.declHandle);
+            }
         }
     }
 
     Napi::Value NativeEngine::CreateIndexBuffer(const Napi::CallbackInfo& info)
     {
         const Napi::TypedArray data = info[0].As<Napi::TypedArray>();
+        if (!data.ElementLength())
```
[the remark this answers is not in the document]
CedricGuillemet: From the JS side, you can create an empty index array without any issue. But bgfx will assert. Someone has to do the check.

Library/Source/NativeEngine.cpp
```diff
@@ -1042,35 +1063,79 @@ namespace babylon
...
+            BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIP_POINT,                            /** mag = nearest and min = linear and mip = nearest */
+            BGFX_SAMPLER_MAG_POINT,                                                     /** mag = nearest and min = linear and mip = linear */
+            BGFX_SAMPLER_MAG_POINT,                                                     /** mag = nearest and min = linear and mip = none */
+            BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIN_POINT,                            /** mag = nearest and min = nearest and mip = none */
+            BGFX_SAMPLER_MIP_POINT | BGFX_SAMPLER_MIP_POINT,                            /** mag = linear and min = nearest and mip = nearest */
+            BGFX_SAMPLER_MIN_POINT,                                                     /** mag = linear and min = nearest and mip = linear */
+            0,                                                                          /** mag = linear and min = linear and mip = none */
+            BGFX_SAMPLER_MIN_POINT };                                                   /** mag = linear and min = nearest and mip = none */
+        filter = std::min(filter, filterCount - 1);
```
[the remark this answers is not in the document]
CedricGuillemet: This function is called from JS and we have no guarantee that it will be called with a correct value. I don't want the app the crash because the user sets an invalid value as a filter value.

Library/Source/NativeEngine.cpp
```diff
@@ -1042,35 +1063,79 @@ namespace babylon
     void NativeEngine::SetTextureSampling(const Napi::CallbackInfo& info)
     {
         const auto textureData = info[0].As<Napi::External<TextureData>>().Data();
-        const auto filter = static_cast<Filter>(info[1].As<Napi::Number>().Uint32Value());
-
-        // STUB: Stub.
+        auto filter = static_cast<uint32_t>(info[1].As<Napi::Number>().Uint32Value());
+
+        constexpr uint32_t filterCount = 12;
+        constexpr std::array<uint32_t, filterCount> bgfxFiltering = {
```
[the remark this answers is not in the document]
CedricGuillemet: Same as comment before. Adding check in JS is god but not enough as a guarantee.

Library/Source/NativeEngine.cpp
```diff
@@ -421,19 +423,29 @@ namespace babylon
...
+        {
+            bgfx::setIndexBuffer(vertexArray.indexBuffer.handle);
+        }
 
         const auto& vertexBuffers = vertexArray.vertexBuffers;
         for (uint8_t index = 0; index < vertexBuffers.size(); ++index)
         {
             const auto& vertexBuffer = vertexBuffers[index];
-            bgfx::setVertexBuffer(index, vertexBuffer.handle, vertexBuffer.startVertex, UINT32_MAX, vertexBuffer.declHandle);
+            if (vertexBuffer.handle.idx != bgfx::kInvalidHandle) // TODO : why does this happen?
```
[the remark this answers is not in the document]
CedricGuillemet: Apparently, we hit the limit (4096) of vertex array count in CreateSpheresAsync. When that limit is reached, any subsequent call will return an invalid handle.

Library/Source/NativeEngine.cpp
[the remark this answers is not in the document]
syntheticmagus: nit: I know it means the same thing, but I greatly prefer to avoid Boolean operators for non-Boolean types. I don't know whether we've explicitly established a convention on this, but for cases like these I'd much prefer to directly check `data.ElementLength() == 0`. As a human reading the code, I just find that much more difficult to misunderstand.

Library/Source/NativeEngine.cpp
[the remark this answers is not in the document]
syntheticmagus: If that happens, will this just fail silently? Should we throw or assert or something?

TestApp/Scripts/babylon.max.js
```diff
@@ -36001,6 +36004,7 @@ var NativeEngine = /** @class */ (function (_super) {
     NativeEngine.prototype.setViewport = function (viewport, requiredWidth, requiredHeight) {
         // TODO: Implement.
```
syntheticmagus: nit: Remove `TODO`. Consider replacing with `TODONE`.

Library/Source/NativeEngine.cpp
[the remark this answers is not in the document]
CedricGuillemet: You'll get the bgfx warnings as outputdebugstring and potentially in the console if we attach bgfx warning callback to Console.

Library/Source/NativeEngine.cpp
syntheticmagus: Sounds like the assumptions I made when I wrote the `FrameBufferManager` -- namely, a one-to-one relationship between views and framebuffers -- were incorrect. The `FrameBufferManager` mostly existed to hide the `RecycleSet` as an implementation detail; so if we're having to expose the `RecycleSet` to other things, that might be a sign we need need to reconsider how that system is architected. That might be a task for a different PR, though.

TestApp/Scripts/experience.js
bghgary: At the worst case, you can use VertexData.CreatePlane and strip out the indices before using it.

TestApp/Scripts/experience.js
bghgary: Right, but should we add this code to experience.js if there is no way to change the cameras?

Library/Source/NativeEngine.cpp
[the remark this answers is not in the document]
bghgary: I think you should check this from the JS side and throw a JS exception.

Library/Source/NativeEngine.cpp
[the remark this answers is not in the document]
bghgary: This one is very different. The JS side doesn't need a mapping at all. It currently has enum values for Base3D (Spectre) on the JS side.

Library/Source/NativeEngine.cpp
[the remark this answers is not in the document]
bghgary: Do we need to increase the limit? At least should remove the TODO

Library/Source/NativeEngine.cpp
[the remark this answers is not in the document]
bghgary: I guess I don't like using `bgfx::InvalidHandle` as a valid value that can be returned. I would prefer to do this on the JS side.

TestApp/Scripts/experience.js
CedricGuillemet: Until we have the validation tests, do we keep JS script that helps developing/testing features?  If so, in experience.js or another file?

bghgary: We sort of already have the validation tests in a separate folder: https://github.com/BabylonJS/BabylonNative/tree/master/TestApp/ValidationTests

bghgary: Maybe create a new file in that folder.

TestApp/Scripts/experience.js
bghgary: I see. Perhaps you can load one of the glTF asset generator models to test this? https://github.com/KhronosGroup/glTF-Asset-Generator/tree/master/Output/Positive/Mesh_PrimitiveMode

CedricGuillemet: and I also need scaled uv... I'll move that function into its own test js

bghgary: Sounds good.

Library/Source/NativeEngine.cpp
```diff
@@ -1023,6 +1045,7 @@ namespace babylon
             format,                                         // Self-explanatory
             0x0,                                            // Flags
             allPixels);                                     // Memory
+        return Napi::Value::From(info.Env(), (textureData->Texture.idx != bgfx::kInvalidHandle));
```
bghgary: Probably should do this everywhere
```suggestion
        return Napi::Value::From(info.Env(), isValid(textureData->Texture));
```

PR #74, opened 2019-10-23
Clean up Chakra implemention of N-API

bghgary:
- Fixed all TODOs and HACKs with the exception of the fatal error one.
- Implemented missing functionality (hasOwnProperty, dataViews, etc.)
- Remove dead code
- Reorder to match v8 implementation
- Add argument checks to match v8 implementation

### DIFF
--- a/Library/Dependencies/napi/source/js_native_api_chakra.cc
+++ b/Library/Dependencies/napi/source/js_native_api_chakra.cc
@@ -487,30 +487,6 @@ napi_status napi_get_last_error_info(napi_env env,
   return napi_ok;
 }
 
-// TODO
-//void napi_fatal_error(const char* location, size_t location_len, const char* message, size_t message_len) {
-//  const char* location_string = location;
-//  const char* message_string = message;
-//
-//  if (location_len != NAPI_AUTO_LENGTH) {
-//    char* location_nullterminated = static_cast<char*>(
-//      malloc((location_len + 1) * sizeof(char)));
-//    strncpy(location_nullterminated, location, location_len);
-//    location_nullterminated[location_len] = 0;
-//    location_string = location_nullterminated;
-//  }
-//
-//  if (message_len != NAPI_AUTO_LENGTH) {
-//    char* message_nullterminated = static_cast<char*>(
-//      malloc((message_len + 1) * sizeof(char)));
-//    strncpy(message_nullterminated, message, message_len);
-//    message_nullterminated[message_len] = 0;
-//    message_string = message_nullterminated;
-//  }
-//
-//  node::FatalError(location_string, message_string);
-//}
-
 napi_status napi_create_function(napi_env env,
                                  const char* utf8name,
                                  size_t length,
--- a/Library/Dependencies/napi/source/js_native_api_chakra.cc
+++ b/Library/Dependencies/napi/source/js_native_api_chakra.cc
@@ -1,95 +1,105 @@
-/******************************************************************************
-* Experimental prototype for demonstrating VM agnostic and ABI stable API
-* for native modules to use instead of using Nan and V8 APIs directly.
-*
-* The current status is "Experimental" and should not be used for
-* production applications.  The API is still subject to change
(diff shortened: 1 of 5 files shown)


Library/Source/NativeEngine.cpp
```diff
@@ -184,7 +184,7 @@ namespace babylon
                 auto channelCount = (image.m_format == bimg::TextureFormat::RGB8) ? 3 : 4;
                 auto mipMapCount = static_cast<uint32_t>(std::log2(image.m_width));
                 auto mipmapImageSize = image.m_size;
-                for (auto i = 1; i <= mipMapCount; i++)
+                for (auto i = 1u; i <= mipMapCount; i++)
```
syntheticmagus: nit: This seems a little weird. Should we just make the iterator type size_t or uint32_t or something instead, to match what we tend to do elsewhere?

syntheticmagus: Also line 197.

Library/Dependencies/napi/source/js_native_api_chakra.cc
```diff
@@ -1,337 +1,124 @@
...
-STDAPI_(JsErrorCode) JsCopyString(
-    _In_ JsValueRef value,
-    _Out_opt_ char* buffer,
-    _In_ size_t bufferSize,
-    _Out_opt_ size_t* length)
-{
-    const wchar_t* stringValue;
-    size_t stringLength;
-    auto errorCode = JsStringToPointer(value, &stringValue, &stringLength);
+constexpr unsigned int CP_LATIN1 = 28591;
```
syntheticmagus: nit: Use the `UINT` typedef, just to guarantee it matches `MultiByteToWideChar`? I don't know if `UINT` can never not be the same as `unsigned int`, though I'm sure it would matter much in either case.

Library/Dependencies/napi/source/js_native_api_chakra.cc
```diff
@@ -772,164 +464,85 @@ const char* error_messages[] = {
...
-  static_last_error.error_code = status;
-  static_last_error.engine_error_code = jsError;
-  static_last_error.engine_reserved = engine_reserved;
-  return status;
-}
-
-// $HACK
-//napi_status napi_fatal_exception(napi_env env, napi_value err) {
-//  CHECK_ARG(err);
+// TODO
```
syntheticmagus: Is there an issue to track this?

Library/Dependencies/napi/source/js_native_api_chakra.cc
```diff
@@ -3662,3 +2429,15 @@ napi_status napi_add_finalizer(napi_env env,
                                napi_ref* result) {
   return napi_ok;
 }
+
+napi_status napi_adjust_external_memory(napi_env env,
+                                        int64_t change_in_bytes,
+                                        int64_t* adjusted_value) {
+  CHECK_ARG(env, adjusted_value);
+
+  // TODO(jackhorton): Determine if Chakra needs or is able to do anything here
```
syntheticmagus: Does this need an issue? Also, who's jackhorton?

Library/Source/NativeEngine.cpp
bghgary: Meh, I started with uint32_t. I have no preference. I'll change it.

Library/Dependencies/napi/source/js_native_api_chakra.cc
bghgary: This is from the original source.

Library/Dependencies/napi/source/js_native_api_chakra.cc
bghgary: #6 Correctly propagate and handle JavaScript errors

bghgary: I've removed this part. We need to implement napi_fatal_error somewhere but it doesn't need to be JS API specific.

bghgary:
Hmm, CI is stuck, but ADO says it completed.

PR #79, opened 2019-10-28
Finish runtime impl refactor

syntheticmagus:


TestApp/Source/macOS/ViewController.mm
```diff
@@ -13,33 +13,25 @@ - (void)viewDidLoad {
...
     NSURL * resourceUrl = [main resourceURL];
 
     NSWindow* nativeWindow = [[self view] window];
-    runtime = std::make_unique<babylon::RuntimeApple>((__bridge void*)nativeWindow, [[NSString stringWithFormat:@"file://%s", [resourceUrl fileSystemRepresentation]] UTF8String]);
+    runtime = std::make_unique<babylon::RuntimeApple>(
+        (__bridge void*)nativeWindow, 
+        [[NSString stringWithFormat:@"file://%s", [resourceUrl fileSystemRepresentation]] UTF8String],
+        [](const char* message, babylon::LogLevel level)
+        {
+            LogMessage(message, level);
```
CedricGuillemet: Call NSLog here, no need for another function. same above.

Library/Source/Console.h
```diff
@@ -8,23 +8,17 @@ namespace babylon
     class Console final : public Napi::ObjectWrap<Console>
```
CedricGuillemet: Is it possible to log to console from C++?

CedricGuillemet: Can we just add a ref to the LogCallBack in RuntimeImpl and call it from NativeEngine m_runtimeImpl.m_logCallBack()...)

syntheticmagus: We could do that, but I'm not sure I understand why we would. In my mind at least, the `NativeEngine`'s status as a C++ object is incidental; philosophically, it really _belongs_ to JavaScript. The JavaScript's approach to logging is through the `Console` object, so by default I'd expect the `NativeEngine`'s C++ implementation to do so as well. Would going through the `Console` have a problem that going through `RuntimeImpl` would avoid?

bghgary: I'm torn on this. While I agree the Console class is intended for JavaScript, going through JavaScript is quite a bit more round about. How about we add logging as a feature to the Runtime contract and then Console can just call the Runtime?

Library/Source/RuntimeImpl.cpp
```diff
@@ -36,12 +39,12 @@ namespace babylon
...
 
-    RuntimeImpl::RuntimeImpl(void* nativeWindowPtr, const std::string& rootUrl)
+    RuntimeImpl::RuntimeImpl(void* nativeWindowPtr, const std::string& rootUrl, LogCallback&& logCallback)
         : m_nativeWindowPtr{ nativeWindowPtr }
         , m_thread{ [this] { ThreadProcedure(); } }
         , m_rootUrl{ rootUrl }
-    {
-    }
+        , m_logCallback{ logCallback }
+    {}
```
bghgary: nit: I don't like `{}`. I prefer `{` and `}` on separate lines. multiple places.

Library/Source/RuntimeUWP.cpp
```diff
@@ -6,8 +6,13 @@
 
 namespace babylon
 {
-    RuntimeUWP::RuntimeUWP(ABI::Windows::UI::Core::ICoreWindow* window, const std::string& rootUrl)
-        : Runtime { std::make_unique<RuntimeImpl>(window, rootUrl) }
+    RuntimeUWP::RuntimeUWP(ABI::Windows::UI::Core::ICoreWindow* window, LogCallback&& callback)
+        : RuntimeUWP{ window, ".", std::move(callback) }
```
bghgary: The string parameter was original defaulting to `{}`. Now, it is `"."`. Which is the right one?

Library/Source/XMLHttpRequest.h
```diff
@@ -93,13 +93,11 @@ namespace babylon
     class XMLHttpRequest final : public Napi::ObjectWrap<XMLHttpRequest>
     {
     public:
-        static void Initialize(Napi::Env& env, RuntimeImpl& runtimeImpl);
+        static Napi::FunctionReference InitializeAndCreateConstructor(Napi::Env& env);
```
bghgary: nit: I think you should call this `CreateConstructor`

TestApp/Source/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -61,34 +61,36 @@ Java_com_android_appviewer_AndroidViewAppActivity_finishEngine(JNIEnv* env, jobj
...
-        babylon::Runtime::RegisterWarnOutput(AndroidWarnMessage);
-        babylon::Runtime::RegisterErrorOutput(AndroidErrorMessage);
-
         ANativeWindow *window = ANativeWindow_fromSurface(env, surface);
 
-        runtime = std::make_unique<babylon::RuntimeAndroid>(window, "file:///data/local/tmp");
+        runtime = std::make_unique<babylon::RuntimeAndroid>(
+            window, 
+            "file:///data/local/tmp",
+            [](const char* message, babylon::LogLevel level)
```
bghgary: Can you put `LogMessage` directly here without the lambda? I think it will auto convert when there is no capture.

Library/Source/RuntimeUWP.cpp
syntheticmagus: I don't think that value is supposed to be meaningful for UWP, so I just changed this to match `RuntimeAndroid` for consistency's sake.

Library/Source/Console.h
CedricGuillemet: I see. One way to do it is to extend the return values from native function. For example, the loadTexture function returns a bool telling if it succeeded or not. More enums are needed like mipmap wanted but not generated, ...
Also possible, a GetLastError function in the contract (a la OpenGL) to get more infos on a failing function call.

Library/Source/RuntimeUWP.cpp
syntheticmagus: Not sure how best to test this, and now agree that it doesn't really belong in this PR anyway, so just reverting.

PR #80, opened 2019-10-29
CMake refactor

syntheticmagus:


CMakeLists.txt
```diff
@@ -10,39 +10,24 @@ if (APPLE)
     set(CMAKE_NO_SYSTEM_FROM_IMPORTED TRUE)
 endif()
 
-set(SUPPORTED_PLATFORMS "Win32" "UWP" "Android" "Apple")
-if (ANDROID)
+set(SUPPORTED_PLATFORMS "Android" "Apple" "UWP" "Win32")
+if(ANDROID)
     set(BABYLON_NATIVE_PLATFORM "Android" CACHE STRING "Target platform for Babylon Native.")
```
CedricGuillemet: It seems now BABYLON_NATIVE_PLATFORM  is not used except for testing SUPPORTED_PLATFORMS. Do we still want to set it?
Just a check if (!APPLE && !WIN32 && ....) for the FATAL_ERROR unrecognized platform would be fine

bghgary: It's used in a bunch of places in child `CMakeLists.txt`. Do we need to use the CACHE anymore?

syntheticmagus: Regarding the use of CACHE, apparently not. Who knew?

CMakeLists.txt
```diff
@@ -10,39 +10,21 @@ if (APPLE)
...
-    set(BABYLON_NATIVE_PLATFORM "Win32" CACHE STRING "Target platform for Babylon Native.")
-endif()
-
-if(NOT(BABYLON_NATIVE_PLATFORM IN_LIST SUPPORTED_PLATFORMS))
-    message(FATAL_ERROR "Unrecognized platform: ${BABYLON_NATIVE_PLATFORM}")
-endif()
-
-if(BABYLON_NATIVE_PLATFORM STREQUAL "UWP" AND NOT(WINDOWS_STORE))
-    message(FATAL_ERROR "\nUWP builds must target Windows Store. Make sure to designate\n\t\"-D CMAKE_SYSTEM_NAME=WindowsStore -D CMAKE_SYSTEM_VERSION=10.0\"\nwhen configuring for UWP.")
+    message(FATAL_ERROR "Unrecognized platform: ${CMAKE_SYSTEM_NAME}")
```
bghgary: Is `${CMAKE_SYSTEM_NAME}` defined all the time?

syntheticmagus: I don't know of any cases where it wouldn't be defined. From [this doc](https://cmake.org/cmake/help/v3.0/variable/CMAKE_SYSTEM_NAME.html), it looks like it's guaranteed to have a default definition at the very least for Windows, macOS, and Linux systems.

PR #82, opened 2019-10-31
Shader uniform mvp

CedricGuillemet:


Library/Source/NativeEngine.cpp
```diff
@@ -753,81 +753,104 @@ namespace babylon
...
+    {
+        const auto uniformData = info[0].As<Napi::External<UniformInfo>>().Data();
+        const auto array = info[1].As<arrayType>();
+
+        size_t elementLength = array.ElementLength();
+
+        m_scratch.clear();
+        for (size_t index = 0; index < elementLength; index += size)
+        {
+            const float values[] = { static_cast<float>(array[index])
```
bghgary: nit: I don't see this brace style anywhere. Might be good to do it like this:

```C++
const float values[] = {
    static_cast<float>(array[index]),
    (size > 1) ? static_cast<float>(array[index + 1]) : 0.f,
    (size > 2) ? static_cast<float>(array[index + 2]) : 0.f,
    (size > 3) ? static_cast<float>(array[index + 3]) : 0.f
};
```

bghgary: ditto on the next function

Library/Source/NativeEngine.cpp
```diff
@@ -753,81 +753,104 @@ namespace babylon
...
+    template<int size> void NativeEngine::SetMatrixN(const Napi::CallbackInfo& info)
     {
         const auto uniformData = info[0].As<Napi::External<UniformInfo>>().Data();
         const auto matrix = info[1].As<Napi::Float32Array>();
 
         const size_t elementLength = matrix.ElementLength();
-        assert(elementLength == 16);
+        assert(elementLength == size * size);
+
+        std::array<float, 16> matrixValues{ };
```
bghgary: nit:

```suggestion
        std::array<float, 16> matrixValues{};
```

Library/Source/NativeEngine.cpp
```diff
@@ -753,81 +753,104 @@ namespace babylon
...
+
+        size_t index = 0;
+        for (int line = 0; line < size; line++)
+        {
+            for (int col = 0; col < size; col++)
+            {
+                matrixValues[line * 4 + col] = matrix[index++];
+            }
+        }
```
bghgary: Is there a performance issue here when matrix is 4x4? We used to directly use the data, but now we copy the data with a double for loop? Maybe we should specialize the 4x4 case?

CedricGuillemet:
Added the fix for this issue https://github.com/BabylonJS/BabylonNative/issues/85

Library/Source/NativeEngine.cpp
```diff
@@ -770,10 +770,12 @@ namespace babylon
         m_scratch.clear();
         for (size_t index = 0; index < elementLength; index += size)
         {
-            const float values[] = { static_cast<float>(array[index])
+            const float values[] = {
```
bghgary: nit: brace on newline (sorry, the code I suggest was also wrong style)

Library/Source/NativeEngine.cpp
```diff
@@ -783,10 +785,12 @@ namespace babylon
     template<int size> void NativeEngine::SetFloatN(const Napi::CallbackInfo& info)
     {
         const auto uniformData = info[0].As<Napi::External<UniformInfo>>().Data();
-        const float values[] = { info[1].As<Napi::Number>().FloatValue(),
-                                (size > 1) ? info[2].As<Napi::Number>().FloatValue() : 0.f,
-                                (size > 2) ? info[3].As<Napi::Number>().FloatValue() : 0.f,
-                                (size > 3) ? info[4].As<Napi::Number>().FloatValue() : 0.f };
+        const float values[] = {
```
bghgary: ditto

Issue #84, opened 2019-10-31
XR is broken in master


syntheticmagus:
Call stack:

```
>	TestApp.exe!bx::debugBreak() Line 38	C++
 	TestApp.exe!bgfx::CallbackStub::fatal(const char * _filePath, unsigned short _line, bgfx::Fatal::Enum _code, const char * _str) Line 70	C++
 	TestApp.exe!bgfx::fatal(const char * _filePath, unsigned short _line, bgfx::Fatal::Enum _code, const char * _format, ...) Line 438	C++
 	TestApp.exe!bgfx::setViewFrameBuffer(unsigned short _id, bgfx::FrameBufferHandle _handle) Line 4714	C++
 	TestApp.exe!babylon::NativeEngine::SetViewPort(const Napi::CallbackInfo & info) Line 1349	C++
 	TestApp.exe!Napi::ObjectWrap<babylon::NativeEngine>::InstanceVoidMethodCallbackWrapper::__l2::<lambda>() Line 3409	C++
 	TestApp.exe!Napi::details::WrapCallback<void * <lambda>(void)>(Napi::ObjectWrap<babylon::NativeEngine>::InstanceVoidMethodCallbackWrapper::__l2::void * <lambda>(void) callback) Line 61	C++
 	TestApp.exe!Napi::ObjectWrap<babylon::NativeEngine>::InstanceVoidMethodCallbackWrapper(napi_env__ * env, napi_callback_info__ * info) Line 3411	C++
 	TestApp.exe!`anonymous namespace'::ExternalCallback::Callback(void * callee, bool isConstructCall, void * * arguments, unsigned short argumentCount, void * callbackState) Line 172	C++
```

Looks like the ViewID mechanism is no longer functioning as it used to.

CedricGuillemet:
setViewPort will not work properly with bound framebuffer
https://github.com/BabylonJS/BabylonNative/blob/master/Library/Source/NativeEngine.cpp#L1348
What's the script you were using to get that issue?

CedricGuillemet:
https://github.com/BabylonJS/BabylonNative/pull/86

syntheticmagus:
The issue arises when you enable the `xr` capability on line 6 of `experience.js`. You have to be set up for WinMR with OpenXR for this to work properly.

thomlucc:
Fixed

PR #86, opened 2019-11-04
more generic on viewids and framebuffers

CedricGuillemet:
viewIds and Framebuffer assignment is a bit broken at the moment.
Here is a proposal for a fix.
The 2 main ideas are:
- viewids are requested when a buffer is bound or a new view port is set. The life time for a viewId is only a frame.
- the back buffer becomes a FrameBufferData that is bound when no other FrameBufferData is active.

Set viewport explicitely asks for a new viewId and binds the already bound FrameBufferData.
I'm not a big fan of applying a viewId to the ViewClearState but didn't find a better because of the UpdateFlags that's being called from the JS side
https://github.com/BabylonJS/BabylonNative/compare/master...CedricGuillemet:frameBuffersAndViews?expand=1#diff-21d665b1d21081b0ae44c5cce05bdf0dR116

image

I've spotted a new bug while coding the test script : https://github.com/BabylonJS/BabylonNative/issues/87


syntheticmagus:
#84 is still a problem with this change. In particular, I'm hitting the new assert on [line 35](https://github.com/BabylonJS/BabylonNative/pull/86/files#diff-21d665b1d21081b0ae44c5cce05bdf0dR35) because `m_nextId` reaches 256, which seems surprising. I'm not sure exactly what changed to bring this about; I can try to debug into it more deeply later if that would be helpful.

CedricGuillemet:
You get the assert because the viewId counter is not reset.
Adding a call to m_viewidSet.Clear(); should fix the issue.
https://github.com/BabylonJS/BabylonNative/blob/master/Library/Source/NativeXr.cpp#L834

I can do the change when I'm able to run the xr experience:
I ran testapp with xr set to true in experience.js but I get an Xr exception when the xr experience is created. The exception happens here : https://github.com/BabylonJS/BabylonNative/blob/master/Library/Dependencies/napi/source/env_chakra.cc#L78

Do you get the same? Is there something that I need to install? Is it broken?

Library/Source/NativeEngine.cpp
```diff
@@ -1357,13 +1325,7 @@ namespace babylon
...
             callbackPtr->Call({});
 
-            // recycle viewIds used as viewports
-            for (auto id : m_viewportIds)
-            {
-                m_viewidSet.Recycle(id);
-            }
-            m_viewportIds.clear();
-            m_currentBackbufferViewId = 0;
+            m_viewidSet.Clear();
```
syntheticmagus: It's still very strange to me that the `NativeEngine` owns this and the `FrameBufferManager` just has a reference to it. This also introduces some weirdness in cases where `NativeEngine` doesn't own the frame loop, such as in XR. If this is the only place where the `NativeEngine` interacts with the `RecycleSet`, would it be better to give ownership of the `RecycleSet` back to the `FrameBufferManager` and add a `FrameBufferManager::Reset()` method?

Library/Source/NativeEngine.h
```diff
@@ -22,35 +22,28 @@ namespace babylon
...
-                return m_nextId++;
-            }
-            else
-            {
-                T next = m_queue.front();
-                m_queue.pop();
-                return next;
-            }
+            m_nextId ++;
+            assert(m_nextId < bgfx::getCaps()->limits.maxViews);
```
syntheticmagus: I do not believe this change belongs in the `RecyleSet`. `RecycleSet` is a generic structure that knows nothing about bgfx or ViewIDs; it just keeps a unique and recyclable set of numbers.

Library/Source/NativeEngine.h
```diff
@@ -22,35 +22,28 @@ namespace babylon
...
+            return m_nextId;
         }
 
-        void Recycle(T id)
+        void Clear()
         {
-            assert(id < m_nextId);
-            m_queue.push(id);
+            m_nextId = m_firstId;
```
syntheticmagus: With this change, I don't think the `RecycleSet` is adding value anymore. The reason it was created was to abstract away the maintenance of state; but if we're just throwing out all the state every frame anyway, the `FrameBufferManager` can probably just do that directly; it can then have asserts that check ViewIDs and thinks like that without creating any errant dependencies. In short, if this is the behavior we're using for ViewIDs now, I think the `RecycleSet` should probably be deleted.

CedricGuillemet: I removed RecycleSet and changed some method names. It's a bit more clear now.

Library/Source/NativeEngine.h
```diff
@@ -166,11 +130,17 @@ namespace babylon
         ~FrameBufferData()
         {
             bgfx::destroy(FrameBuffer);
-            m_idSet.Recycle(ViewId);
         }
 
-        void SetUpView()
+        void UseViewId(uint16_t viewId)
```
bghgary:
```suggestion
        void SetViewId(uint16_t viewId)
```

Library/Source/NativeEngine.h
```diff
@@ -224,12 +185,25 @@ namespace babylon
...
+        }
+
+        uint16_t GetNewViewId()
+        {
+            m_nextId++;
+            assert(m_nextId < bgfx::getCaps()->limits.maxViews);
+            return m_nextId;
+        }
+
+        void ResetViewIds()
```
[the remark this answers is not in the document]
syntheticmagus: Or just `Reset`? It's basically flushing the entire state of the `FrameBufferManager`, I think.

Library/Source/NativeEngine.h
bghgary: Or maybe `UpdateViewId`

Library/Source/NativeEngine.h
```diff
@@ -166,11 +130,17 @@ namespace babylon
...
         {
             bgfx::destroy(FrameBuffer);
-            m_idSet.Recycle(ViewId);
         }
 
-        void SetUpView()
+        void UseViewId(uint16_t viewId)
         {
+            ViewId = viewId;
+            ViewClearState.SetViewId(ViewId);
```
bghgary: If ViewClearState already has the ViewId, do we really need another one on this class?

Library/Source/NativeEngine.h
```diff
@@ -124,8 +84,13 @@ namespace babylon
             bgfx::touch(m_viewId);
         }
 
+        void SetViewId(uint16_t viewId)
```
bghgary:
```suggestion
        void UpdateViewId(uint16_t viewId)
```

Library/Source/NativeEngine.h
CedricGuillemet: Yes. You can clear the backbuffer, set a new viewport, and call clear again with a different state.

Issue #20, opened 2019-08-08
Curl and local file access


CedricGuillemet:
To access local files in the APK (scripts/*.js), Android asks to use an asset manager. You can't open the file directly and basically, it's not even a file in the filesystem but a bunch of bytes in a zip. For now, I've added a function, per platform that does the job if curl doesn't succeed at loading the resource.

CedricGuillemet:
Added a resource loading callback PR https://github.com/BabylonJS/BabylonNative/pull/88

CedricGuillemet:
CURL issues centralized here : https://github.com/BabylonJS/BabylonNative/issues/96

PR #88, opened 2019-11-04
resourceLoadingCallback for Android APK resources

CedricGuillemet:
Resource callback to be able to load files from APK on Android and app for iOS.


CedricGuillemet:
Issue https://github.com/BabylonJS/BabylonNative/issues/20

bghgary:
We had a discussion about this and we want to try a pluggable system instead of having to add callbacks every time we need something platform specific. This PR should be replaced with the pattern from #101.

CedricGuillemet:
Sounds good. I'll use this mecanism until the plugin system is in place for this PR : https://github.com/BabylonJS/BabylonNative/pull/30
I'll update it before merging with master.

PR #95, opened 2019-11-06
Windows Mixed Reality color space workaround

syntheticmagus:
So here's a weird thing. After talking with @sebavan and fiddling with the color spaces a bit, we found that if we lie to the OpenXR runtime about the color space of the swapchain it's receiving, we can get the rendering output we expect from the latest runtimes---neither too dark nor too light. I'm pretty sure this isn't something we want to be in the business of doing long-term, and maybe not even short-term, but I thought this draft PR might be a good place to discuss it and figure out where we want to go from here.

### DIFF
--- a/Library/Source/NativeXr.cpp
+++ b/Library/Source/NativeXr.cpp
@@ -8,16 +8,22 @@
 
 namespace
 {
-    std::pair<bgfx::TextureFormat::Enum, uint64_t> XrTextureFormatToBgfxFormat(xr::TextureFormat format)
+    bgfx::TextureFormat::Enum XrTextureFormatToBgfxFormat(xr::TextureFormat format)
     {
         switch (format)
         {
-        case xr::TextureFormat::RGBA8:
-            return{ bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT };
-        case xr::TextureFormat::RGBA8_NONLINEAR:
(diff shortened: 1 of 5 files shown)


Library/Source/NativeXr.cpp
```diff
@@ -8,16 +8,16 @@
...
     {
         switch (format)
         {
         case xr::TextureFormat::RGBA8:
-            return bgfx::TextureFormat::RGBA8;
-        case xr::TextureFormat::RGBA8S:
-            return bgfx::TextureFormat::RGBA8S;
+            return{ bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT };
+        case xr::TextureFormat::RGBA8_NONLINEAR:
+            return{ bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT /*| BGFX_TEXTURE_SRGB*/ };
```
syntheticmagus: If I uncomment `BGFX_TEXTURE_SRGB` then the rendering becomes wrong. I have not yet figured out why.
"

bghgary:
This work around is actually the right thing to do. I've put a comment in the code that states the reason.

https://github.com/BabylonJS/BabylonNative/pull/95/files#diff-520daa33edf6728737870eac92c60e87R16-R17

bghgary:
@syntheticmagus reviewed over the shoulder.

PR #100, opened 2019-11-14
X86/x64/arm64 jobs

### LINKED ISSUE #81: x86 Chakra build is broken
Looks like something became incompatible as a result of the `Promise` work done for the OpenXR integration.

CedricGuillemet:
Fixes #81


CedricGuillemet:
- win32_x86 uwp_arm64 build OK!
 - are win32_arm64 and uwp_x86 worth fixing? should I just remove them from the yml?
 - added cmake --version build step for macos/ios and the version is 3.15.4.

CedricGuillemet:
I closed related issue : https://github.com/BabylonJS/BabylonNative/issues/81

bghgary:
> win32_arm64 

I don't think we care about this. If someone asks, then maybe.

> uwp_x86

We should support this one.

PR #108, opened 2019-11-19
Move console outside library

syntheticmagus:


bghgary:
The code in the test apps are too complicated. I was thinking it was one function call with a callback, just like the runtime constructor.

syntheticmagus:
@bghgary Right this isn't the final form. I just created this to poke the CI so I could figure out the cross-plat problems. 😄

PR #110, opened 2019-11-21
Move console outside library

syntheticmagus:


Library/Include/Babylon/Console.h
```diff
@@ -0,0 +1,81 @@
+#pragma once
```
CedricGuillemet: I'd like to see a separation between the BabylonNative trunk and the plugins. Maybe add a plugin folder and a subdirectory per plugin?

TestApp/Source/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -61,30 +84,20 @@ Java_com_android_appviewer_AndroidViewAppActivity_finishEngine(JNIEnv* env, jobj
...
 Java_com_android_appviewer_AndroidViewAppActivity_surfaceCreated(JNIEnv* env, jobject obj, jobject surface)
 {
     if (!runtime)
     {
         ANativeWindow *window = ANativeWindow_fromSurface(env, surface);
 
-        runtime = std::make_unique<Babylon::RuntimeAndroid>(window, "file:///data/local/tmp", LogMessage);
+        runtime = std::make_unique<Babylon::RuntimeAndroid>(window, "file:///data/local/tmp");
+
+        runtime->Dispatch([](Babylon::Env& env)
```
CedricGuillemet: I'm wondering if it's possible to have a simpler function like 'AddPlugin' that would be called like this:
```
runtime->AddPlugin("Console", logHandler);
```
And do the internal cooking with env.Global in Runtime.

Library/Source/RuntimeImpl.cpp
```diff
@@ -214,9 +212,9 @@ namespace Babylon
         global.Set("setTimeout", NativeWindow::GetSetTimeoutFunction(jsWindow).Value());
         global.Set("atob", NativeWindow::GetAToBFunction(jsWindow).Value());
 
-        auto jsConsole = Console::Create(env, m_logCallback);
-        jsNative.Set(JS_CONSOLE_NAME, jsConsole.Value());
-        global.Set(JS_CONSOLE_NAME, jsConsole.Value());
+        // auto jsConsole = Console::Create(env, m_logCallback);
+        // jsNative.Set(JS_CONSOLE_NAME, jsConsole.Value());
+        // global.Set(JS_CONSOLE_NAME, jsConsole.Value());
```
bghgary: remove?

TestApp/Source/Android/app/src/main/cpp/BabylonNativeJNI.cpp
bghgary: I was thinking it's more like a helper static function instead of an instance function:

```
    AddConsole(runtime, logHandler);
```

Maybe under a namespace also.

Library/Include/Babylon/Console.h
```diff
@@ -0,0 +1,81 @@
...
+
+#include <napi/napi.h>
+
+#include <sstream>
+
+namespace Babylon
+{
+    template<typename LogHandlerT>
```
bghgary: This pattern of passing in an instance to the log handler seems weird to me. The caller must guarantee that the instance live longer than this class. I think what we should do is pass in 3 handlers template args for Log, Warn, and Error and default Warn and Error to Log.

CedricGuillemet:
Also, as a user I'd like to include console.h in my project .cpp, call AddPlugin and that's it.
I should only care about the handler if there is not a default one for my platform or if I want to do more.
I think it would be great to add default handlers in console.h like:
```
#ifdef ANDROID
struct LogHandler
{
    void Log(const char* message, Console::LogLevel level) const
    {
        switch (level)
        {
        case Console::LogLevel::Log:
            __android_log_write(ANDROID_LOG_INFO, "BabylonNative", message);
            break;
        case Console::LogLevel::Warn:
            __android_log_write(ANDROID_LOG_WARN, "BabylonNative", message);
            break;
        case Console::LogLevel::Error:
            __android_log_write(ANDROID_LOG_ERROR, "BabylonNative", message);
            break;
        }
    }
};
LogHandler defaultLogHandler{};
#endif
#ifdef APPLE
...
#endif
```

Library/Include/Babylon/Console.h
syntheticmagus: I agree that that's the direction I want it to go, but I'm not sure I want to make that a part of this PR. Depending on how many of these plugins we expect there to be, I kind of feel like there should be some kind scalable story about how to cleanly add them to a project, and I don't really have a good feel for what that should look at right now. In short, I totally agree something like this should be a thing, but I don't think it should be a part of this PR.

bghgary: File an issue after merging then?

Library/Include/Babylon/Console.h
```diff
@@ -1,16 +1,20 @@
...
-    template<typename LogHandlerT>
-    class Console final : public Napi::ObjectWrap<Console<LogHandlerT>>
+    class Console final : public Napi::ObjectWrap<Console>
     {
     public:
-        using ParentT = Napi::ObjectWrap<Console<LogHandlerT>>;
+        static inline constexpr char* JS_CONSTRUCTOR_NAME{ "Console" };
+        static inline constexpr char* JS_METHOD_NAME_LOG{ "log" };
+        static inline constexpr char* JS_METHOD_NAME_WARN{ "warn" };
+        static inline constexpr char* JS_METHOD_NAME_ERROR{ "error" };
```
bghgary: These shouldn't be exposed.

Library/Include/Babylon/Console.h
```diff
@@ -22,25 +26,31 @@ namespace Babylon
             Error,
         };
 
-        static Napi::ObjectReference Create(Napi::Env env, LogHandlerT& handler)
+        using ParentT = Napi::ObjectWrap<Console>;
+        using CallbackT = std::function<void(const char*, LogLevel)>;
+
+        static Napi::ObjectReference AddConsoleToEnv(Napi::Env env, CallbackT callback)
```
bghgary: Perhaps `CreateInstance` for plugins that assigns itself and `CreateConstructor` for plugins that create a type.

Library/Include/Babylon/Console.h
```diff
@@ -29,17 +25,17 @@ namespace Babylon
         using ParentT = Napi::ObjectWrap<Console>;
         using CallbackT = std::function<void(const char*, LogLevel)>;
 
-        static Napi::ObjectReference AddConsoleToEnv(Napi::Env env, CallbackT callback)
+        static Napi::ObjectReference CreateInstance(Napi::Env env, CallbackT callback)
```
bghgary: I would recommend removing the return until we need it. Returning it possibly implies that the caller has to do something with it.

Library/Include/Babylon/Console.h
syntheticmagus: #119

PR #112, opened 2019-11-26
metal shader compilation, metal layer thread management, some missing…

CedricGuillemet:
I had to split BaseThreadProcedure in 2 functions (init & run) because bgfx init must happen in main thread for UIkit/Apple. Might change (again) when I take a look seriously at swift on iOS.
While testing, I've found an issue with bgfx. It doesn't show up with d3d11. I did see something weird with opengl and vertex streams/views messing up but found a workaround. With metal, I get the same issue but metal validation being more strict, it crashes with an exception. I'll take a closer look next monday/tuesday.


Library/Source/RuntimeImpl.cpp
```diff
@@ -237,15 +241,16 @@ namespace Babylon
             });
         };
 
-        Env env{ GetModulePath().u8string().data(), std::move(executeOnScriptThread) };
-
-        m_env = &env;
-        auto hostScopeGuard = gsl::finally([this] { m_env = nullptr; });
+        m_env = std::make_unique<Env>(GetModulePath().u8string().data(), std::move(executeOnScriptThread));
```
syntheticmagus: So... If I'm understanding this correctly, is the motivation behind this change the desire to have the Env be created on the main thread (instead of the JavaScript thread) on Apple only? This seems kind of bizarre. Can you add some context?

TestApp/Scripts/experience.js
```diff
@@ -48,6 +48,7 @@ function CreateInputHandling(scene) {
 
 var engine = new BABYLON.NativeEngine();
 var scene = new BABYLON.Scene(engine);
+engine.getHostDocument = function () { return {}; }
```
syntheticmagus: Not sure if I missed something. Can you explain what this is for?

Library/Source/RuntimeImpl.h
```diff
@@ -71,7 +73,7 @@ namespace Babylon
         // occasionally need access to the env as well; m_env provides this
         // access when the env is available, reverting to nullptr once the env
         // is destroyed.
-        Babylon::Env* m_env{};
+        std::unique_ptr<Babylon::Env> m_env{};
```
syntheticmagus: Having this thing not be directly bound to the thread on which it lives feels wrong. Does Apple need this? Is there a different way we could work around it?

TestApp/Scripts/experience.js
CedricGuillemet: This is a hack I did to get things going. I've talked to @sebavan to fix it properly.

Library/Source/RuntimeImpl.cpp
CedricGuillemet: metal context init needs to happen on the main thread. So env must be created in the thread init.

Library/Source/RuntimeImpl.h
CedricGuillemet: I don't think so. And there is more to come. Resize event from metal must be in sync with rendering. If I try to render to a smaller UI window, I get a crash. In the same trend, on Android, I get issues if the rendering thread is actively working while the activity/surfaceView is suspended.

CedricGuillemet:
Closing PR. I'll split in 3 : babylonjs update, mac shader, thread lock for window size update.

Issue #70, opened 2019-10-18
RuntimeImpl order of destruction


syntheticmagus:
The order of destruction is not quite working as intended in RuntimeImpl::BaseThreadProcedure right now. As it's currently written, the JavaScript environment (`env`) is correctly tied to the lifetime of the thread on which it was created, which is _strictly less_ than the lifespan of the RuntimeImpl itself. However, the `dispatcher` and `task` chain which represent the thread are actually members of the RuntimeImpl; therefore, they outlive the thread and are destroyed after the the thread has terminated; in other words, they strictly outlive `env`. The problem with this is that the capturing lambdas we use inside the task system end up holding onto unsafe references (i.e., `Napi::FunctionReference`) to objects that are owned by `env`. These objects are of course destroyed when the `env` dies, and the `dispatcher` and `task` chain's attempts to destroy them again therefore cause the program to crash on close.

I can currently think of two ways to address this. Option 1 is to manually clear the `task` chain and `dispatcher` in the shutdown phase of the thread, thereby causing them to be _empty_ before the `env` is destroyed (this will require changes to Arcana). Option 2 is to make the `dispatcher` and `task` chain not be members of the `RuntimeImpl`, but instead make them beholden to the thread in the same way `env` is, perhaps by encapsulating all this thread state in an object which can be used to guarantee destruction order. Option 2 is probably the more correct of these alternatives; it'll just require some minor rearchitecting of the `RuntimeImpl` to achieve the desired result without losing the ability to enqueue work before the thread has actually started.

PR #117, opened 2019-12-09
Added minor safeguards to enforce the proper lifecycle.

syntheticmagus:
Addressing #70.

The issue described is something that appeared in early versions and reared its head again in an intermediate state in October, but it no longer repros now (nor in the non-intermediate states I synced back to from October). However, I believe the potential for lifecycle problems is valid, and the problem can be prevented with a few relatively straightforward safeguards, so the following are prophylactic changes to address the identified vulnerability.

### DIFF
--- a/Library/Source/RuntimeImpl.cpp
+++ b/Library/Source/RuntimeImpl.cpp
@@ -101,6 +101,7 @@ namespace Babylon
 
     void RuntimeImpl::Dispatch(std::function<void(Env&)> func)
     {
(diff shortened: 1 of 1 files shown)


syntheticmagus:
Closing because the issue doesn't repro any more, and prophylactic code changes are a weird idea to begin with. If the bug ever resurfaces, we can resurrect these changes and test to make sure they actually do what they're supposed to do.

PR #120, opened 2019-12-10
Update code style with clang-format

bghgary:
Some notes:
* I only did this for the Library folder. I don't think we need to do it for the other files. Ideally, we should only do it for Library/Include and Library/Source.
* Use style based on LLVM.
* @syntheticmagus and I decided that `{}` initializers should not have spaces inside to match LLVM defaults.

There are some weird bugs with clang-format. I either used `// clang-format off`, rearranged the code, or just didn't bother. Here are some notable ones:
* `{` style for initializers. They should be on their own line, but I couldn't find a way to work around it.
* Continuations with `.then` is really wonky if there are multiple continuations. The style `AlignAfterOpenBracket` doesn't seem to work for multiple continuations.


CedricGuillemet:
I'd keep the clangformat style file in the root.
Otherwise, everything is fine.

bghgary:
> I'd keep the clangformat style file in the root.

I don't have a strong preference, but the way `.clang-format` works right now is kind of bad if you have submodules that don't have `.clang-format`. Maybe I can add a `.clang-format` to `Dependencies` to prevent it from leaking into submodules and move the repo one to the root.

PR #122, opened 2019-12-10
Mac jsc update

CedricGuillemet:
JSC missing functions
metal shader compilation now using glslang instead of poor string replacements
UI pointer and resize hooks

Same as previous mac jsc update without thread lock/main thread init/... as I need to do that for Android first.


Library/Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -325,7 +328,16 @@ napi_status napi_coerce_to_string(napi_env env,
...
+  static JSClassRef classDef = nullptr;
+  if (!classDef)
+  {
+      JSClassDefinition classDefinition = kJSClassDefinitionEmpty;
+      classDefinition.className = "dummyObject";
+      classDef = JSClassCreate(&classDefinition);
+  }
+  JSObjectRef jsObject = JSObjectMake(env->m_globalContext, classDef, nullptr);
+  *result = reinterpret_cast<napi_value>(jsObject);
+
```
bghgary: nit: formatting issues (napi uses two spaces and same line brace, etc.)

PR #123, opened 2019-12-11
Copying some changes from long lasting android branch to master

CedricGuillemet:
I'm pushing some changes from my 'experimental' androidRender branch to master.


Library/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -1,19 +1,164 @@
...
+
+            auto compiler = std::make_unique<spirv_cross::CompilerGLSL>(parser.get_parsed_ir());
+            
+            compiler->build_combined_image_samplers();
+            
+            spirv_cross::CompilerGLSL::Options options = compiler->get_common_options();
+
+#ifdef ANDROID
+            options.version = 300;
+            options.es = true;
```
bghgary: Is it necessary to do something different for Android? All the shaders are for WebGL 1/2, so I would expect to use the same version?

Library/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -1,19 +1,164 @@
...
+                glsl.replace(pos, fragDef.size(), "");
+            }
+
+            // frag
+            static const std::string fragColor = "glFragColor";
+            pos = glsl.find(fragColor);
+            if (pos != std::string::npos)
+            {
+                glsl.replace(pos, fragColor.size(), "gl_FragColor");
+            }
```
bghgary: What is happening with this and why is it necessary?

CedricGuillemet: That's because bgfx changes a couple things on FS outputs : https://github.com/bkaradzic/bgfx/blob/master/src/renderer_gl.cpp#L5842

Library/Source/ShaderCompilerOpenGL.cpp
CedricGuillemet: Yes, syntax for samplers and textures are a bit different.

CedricGuillemet:
The struct Frame is needed in order to convert from GLSL to spv. So, I can't check and not add it in the nativeShaderProcessor.

Library/Source/RuntimeImpl.cpp
```diff
@@ -156,6 +156,15 @@ namespace Babylon
                 curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, callback);
                 curl_easy_setopt(curl, CURLOPT_WRITEDATA, &data);
 
+#if (ANDROID)
+                /*
+                 * /!\ warning! this is a security issue
```
syntheticmagus: Is this going to be a blocking issue in the near future? Should we call this out specifically to interested parties (who might have the expertise to investigate it)?

Library/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -1,19 +1,164 @@
...
+            parser.parse();
+
+            auto compiler = std::make_unique<spirv_cross::CompilerGLSL>(parser.get_parsed_ir());
+            
+            compiler->build_combined_image_samplers();
+            
+            spirv_cross::CompilerGLSL::Options options = compiler->get_common_options();
+
+#ifdef ANDROID
+            options.version = 300;
```
syntheticmagus: Can we specify these values via compiler defines or something, perhaps in CMake? Not sure if that would be better, but they seem like magic numbers here.

CedricGuillemet: Not that magic, it's asking for #version 300 es

PR #125, opened 2019-12-12
Setting Z Offset

4Liamk:
Setting and getting zOffset in the NativeEngine.  There did not seem to be much required to make this happen, if there are other things that need to change, let me know.

I was looking through the issues and found this.  Maybe someone has already done it in another pull request, but it seemed easy enough so I went for it.

### DIFF
--- a/Library/Source/NativeEngine.cpp
+++ b/Library/Source/NativeEngine.cpp
@@ -703,21 +703,17 @@ namespace Babylon
             }
         }
 
-        // TODO: zOffset
-        const auto zOffset = info[1].As<Napi::Number>().FloatValue();
+        m_zOffset = info[1].As<Napi::Number>().FloatValue();
     }
 
     void NativeEngine::SetZOffset(const Napi::CallbackInfo& info)
     {
-        const auto zOffset = info[0].As<Napi::Number>().FloatValue();
-
-        // STUB: Stub.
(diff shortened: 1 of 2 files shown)


4Liamk:
I believe you would want this value propagated into the actual rendering pipeline.  If you can tell me where it needs to end up, I think I can get it there.

It looks like BGFX::SetState should handle passing this own.  Would that be via another parameter or should it be encoded in the int64 state that gets passed in somehow?

4Liamk:
Thanks!
I've been poking around the codebase and I'm pretty sure this is intended to set the Depth Bias in DirectX which is the closest things I can find to the zoffset in OpenGL.  https://docs.microsoft.com/en-us/windows/win32/direct3d9/depth-bias 

I would say the biggest question for me is how the variable should be passed into bgfx, once it is in there, I can probably find where this should be set.

CedricGuillemet:
There is no way to set zoffset in bgfx. This can be fixed with a uniform and some shader changes.

bghgary:
@4Liamk I'm thinking we should close this PR until we know what is the right way to implement setting z offset in bgfx. Are you okay with this?

bghgary:
@4Liamk Closing for now. We can reopen once we know what we should do with z offset.

PR #127, opened 2019-12-12
Device context in main thread

CedricGuillemet:
thread init and run split in 2 functions.
It's not possible to hook a MetalLayer in a thread other than the main thread.

### DIFF
--- a/Library/Source/NativeEngine.cpp
+++ b/Library/Source/NativeEngine.cpp
@@ -275,23 +275,24 @@ namespace Babylon
             bgfx::TextureFormat::RGBA32F};
     }
 
-    Napi::FunctionReference NativeEngine::InitializeAndCreateConstructor(Napi::Env& env)
+    void NativeEngine::InitializeDeviceContext(void *nativeWindowPtr, uint32_t width, uint32_t height)
     {
-        auto& window = RuntimeImpl::GetNativeWindowFromJavaScript(env);
-
         // Initialize bgfx.
         bgfx::Init init{};
-        init.platformData.nwh = window.GetWindowPtr();
+        init.platformData.nwh = nativeWindowPtr;
         bgfx::setPlatformData(init.platformData);
         init.type = bgfx::RendererType::Direct3D11;
-        init.resolution.width = static_cast<uint32_t>(window.GetWidth());
-        init.resolution.height = static_cast<uint32_t>(window.GetHeight());
+        init.resolution.width = width;
+        init.resolution.height = height;
         init.resolution.reset = BGFX_RESET_FLAGS;
         bgfx::init(init);
         bgfx::setViewClear(0, BGFX_CLEAR_COLOR | BGFX_CLEAR_DEPTH, 0x443355FF, 1.0f, 0);
         bgfx::setViewRect(0, 0, 0, init.resolution.width, init.resolution.height);
         bgfx::touch(0);
+    }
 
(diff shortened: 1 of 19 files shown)


Library/Source/NativeEngine.cpp
```diff
@@ -275,23 +275,24 @@ namespace Babylon
             bgfx::TextureFormat::RGBA32F};
     }
 
-    Napi::FunctionReference NativeEngine::InitializeAndCreateConstructor(Napi::Env& env)
+    void NativeEngine::InitializeDeviceContext(void *nativeWindowPtr, uint32_t width, uint32_t height)
```
bghgary:
```suggestion
    void NativeEngine::InitializeDeviceContext(void* nativeWindowPtr, uint32_t width, uint32_t height)
```

Library/Source/NativeEngine.cpp
```diff
@@ -275,23 +275,24 @@ namespace Babylon
...
+        init.resolution.width = width;
+        init.resolution.height = height;
         init.resolution.reset = BGFX_RESET_FLAGS;
         bgfx::init(init);
         bgfx::setViewClear(0, BGFX_CLEAR_COLOR | BGFX_CLEAR_DEPTH, 0x443355FF, 1.0f, 0);
         bgfx::setViewRect(0, 0, 0, init.resolution.width, init.resolution.height);
         bgfx::touch(0);
+    }
 
+    Napi::FunctionReference NativeEngine::InitializeAndCreateConstructor(Napi::Env& env)
```
bghgary: I would rename to just this:
```suggestion
    Napi::FunctionReference NativeEngine::CreateConstructor(Napi::Env& env)
```

Library/Source/RuntimeUWP.cpp
```diff
@@ -15,6 +15,7 @@ namespace Babylon
         : Runtime{std::make_unique<RuntimeImpl>(window, rootUrl)}
         , m_window{window}
     {
+        NativeEngine::InitializeDeviceContext(window, 32, 32);
```
bghgary: I know this is what it was doing before, but ideally, we really should figure out the size of the native window here and possibly all the platforms, like what Win32 one does.

Library/Source/RuntimeWin32.cpp
```diff
@@ -18,9 +18,10 @@ namespace Babylon
         RECT rect;
         if (GetWindowRect(hWnd, &rect))
         {
-            float width = static_cast<float>(rect.right - rect.left);
-            float height = static_cast<float>(rect.bottom - rect.top);
-            UpdateSize(width, height);
+            auto width = rect.right - rect.left;
+            auto height = rect.bottom - rect.top;
+            NativeEngine::InitializeDeviceContext(hWnd, width, height);
+            UpdateSize(static_cast<float>(width), static_cast<float>(height));
```
bghgary: This call is extraneous.

Library/Source/NativeEngine.cpp
bghgary: The name DeviceContext is a bit overloaded, especially with DirectX. Maybe just `Initialize`?

CedricGuillemet: InitializeContext?
Initialize is a too vague

bghgary: InitializeWindow?

bghgary: Though I don't think Initialize is very vague. We are initializing bgfx inside.

PR #129, opened 2020-01-07
update glslang

### LINKED ISSUE #128: Update to latest version of glslang
See https://github.com/KhronosGroup/glslang/issues/1976 for context. It would be good to update <strike>and remove [`ENABLE_HLSL` define](https://github.com/BabylonJS/BabylonNative/blob/master/Library/CMakeLists.txt#L30) to avoid future layout issue</strike>(`ENABLE_HLSL` should still be set to off).

CedricGuillemet:
Fixes #128

<strike>Quick test with an updated glslang.
I get a parsing error here https://github.com/BabylonJS/BabylonNative/blob/master/Library/Source/ShaderCompilerD3D.cpp#L24
@bghgary Any idea before I put more effort in?</strike>
Resolved.


Library/CMakeLists.txt
```diff
@@ -27,8 +27,10 @@ set(ENABLE_SPVREMAPPER OFF CACHE BOOL "Enables building of SPVRemapper")
 set(ENABLE_AMD_EXTENSIONS OFF CACHE BOOL "Enables support of AMD-specific extensions")
 set(ENABLE_GLSLANG_BINARIES OFF CACHE BOOL "Builds glslangValidator and spirv-remap")
 set(ENABLE_NV_EXTENSIONS OFF CACHE BOOL "Enables support of Nvidia-specific extensions")
-set(ENABLE_HLSL OFF CACHE BOOL "Enables HLSL input support")
 set(ENABLE_OPT OFF CACHE BOOL "Enables spirv-opt capability if present")
+if(WIN32 OR WINDOWS_STORE)
+    set(ENABLE_HLSL ON CACHE BOOL "Enables HLSL input support")
+endif()
```
bghgary: This should not be here. ENABLE_HLSL should be set to OFF always as it is for input and our input is always GLSL.

bghgary: Also, some of these settings have been removed and are no longer necessary.

bghgary:
> @bghgary Any idea before I put more effort in?

I have some bandwidth. Let me take care of this one. I found where in the glslang code it is throwing this error and I will see if I can figure out how to stop it. If I can't, then we may have to add a `layout(binding=X)` to the shader processor in Babylon.js

bghgary:
This was a bug in glslang apparently, so we have to add the binding tags to the shader processor.
https://github.com/KhronosGroup/glslang/commit/deec1933e907872eaff2647720b12634890540e2

PR #130, opened 2020-01-07
PoC asset loading plugin

CedricGuillemet:
This is a Proof of concept for plugins and more particularly asset loading/http requests. 
Remarks/critics/ideas/... **hugely** welcome. I guess you have a better/more clear view of the plugin architecture :)

I added a Plugins folder at the root. Then 1 subfolder per plugin.
Console is a simple example. 
```
#include <Console/Console.h>
```

then

```
runtime->Dispatch([](Babylon::Env& env)
        {
            Babylon::Console::CreateInstance(env, [](const char* message, auto)
            {
                OutputDebugStringA(message);
            });
        });
```

Things get more complicated with HTTP requests:
- There might be some CPP involved. Do we create a plugin CMake to make a library or let Apps do it?
- XMLHttpRequest has a depedency on Arcana. How can we avoid it? 
- Is there a better place in the directory tree for these files ?

An example of use on Android is:
```
#include <XMLHttpRequest/Android/XMLHttpRequestAndroid.h>
```

then in the App:

```
runtime->Dispatch([](Babylon::Env& env)
        {
            Babylon::XMLHttpRequestAndroid::CreateInstance(env, asset_manager);
        });
```

**It doesn't build, don't even try.**


CedricGuillemet:
@bghgary @syntheticmagus 

I did some changes and got some issues I need help with.

Changes:
- PluginHost interface as RuntimeImpl is not accessible from the outside. I think the name is selfexplicit https://github.com/BabylonJS/BabylonNative/pull/130/files#diff-c6bfa4c2e51a9f3aa875a6237c8234baR1
- Loading JS using JS https://github.com/BabylonJS/BabylonNative/pull/130/files#diff-91affc66b12cbf8cc748309b61bca936R82 To do the full loop, I need to use JS for loading so the plugin is properly called
- plugin folder and Console and HTTPRequest plugins

Issues:
- This assert is trigger https://github.com/BabylonJS/BabylonNative/blob/65457e46a3f968e753467cc1aac1ff4c788f1c96/Library/Dependencies/napi/source/js_native_api_chakra.cc#L39 . GetLastError tells me the buffer is not big enough. But the size comes from ::WideCharToMultiByte . It happens when loading Babylon.max.js Maybe it has something to do with the 8+Mb file size?
- I get a crash when calling eventHandlerRef.Call({}); from SetReadyState. I don't know yet why it's failing

Library/Dependencies/napi/source/js_native_api_chakra.cc
```diff
@@ -31,7 +31,7 @@ JsErrorCode JsCopyString(_In_ JsValueRef value, _Out_opt_ char* buffer, _In_ siz
   CHECK_JSRT_ERROR_CODE(JsStringToPointer(value, &stringValue, &stringLength));
 
   if (length != nullptr) {
-    *length = stringLength;
+    *length = ::WideCharToMultiByte(codePage, 0, stringValue, static_cast<int>(stringLength), nullptr, 0, nullptr, nullptr);
```
bghgary: Ideally, don't call `WideCharToMultiByte` twice if both length and buffer are specified.

CedricGuillemet:
Closing PR

Issue #102, opened 2019-11-15
Android issues


CedricGuillemet:
Ordered by complexity

- [x] the project compiles for x86 and x86_64 but has never been tested on the simulator
- [x] ~~V8 .so are poorly handled in the APK for the Native Activity project. It needs to be fixed before it can run~~
- [ ] V8 must be installed manually for local builds. This should be done automatically in CMake
- [x] Screen ratio is not updated with screen rotation
- [x] Add support for SSL
- [x] Shader compliance and parsing between BJS and bgfx is not great for OpenGL. Expect issues when dealing with more complex shaders
- [x] Front facing is inverted (you see the inside of the cube). A bgfx repo update will be done shortly and should be done on Master @CedricGuillemet 
- [x] Asset loading is using a callback mecanism that supports APK assets loading. It will be replaced by a plugin extension (https://github.com/BabylonJS/BabylonNative/issues/101)
- [x] getHostDocument is missing in native engine definition ( https://github.com/BabylonJS/BabylonNative/pull/30/files#diff-61eebfcda2bd45c2180b579c8f4fcd42R51 ) => PR : https://github.com/BabylonJS/Babylon.js/pull/7286

CedricGuillemet:
# the project compiles for x86 and x86_64 but has never been tested on the simulator 

- Add a virtual device in AVD (any Android 5.0+), build for the target, run in simulator. 
- The cube should appear with touch support to rotate the cube.
- Test with other virtual devices and various Android version

CedricGuillemet:
# V8 .so are poorly handled in the APK for the Native Activity project. It needs to be fixed before it can run

- Install V8 for Android and AndroidNativeActivity 
- make sur Android project builds and runs on your device
- Open AndroidNativeActivity with Android Studio. 
- Build for your device
- Run app
- check the logcat, you should see:
```
FATAL EXCEPTION: main
Unable to load native library: /data/app/com.android.testapp-2/lib/arm64/libnative-activity.so
```
There is a mismatch in the .so directory. It should be arm64-v8a instead of arm64
I think it's an issue between gradle config and cmake. 
-

CedricGuillemet:
# V8 .so are poorly handled in the APK for the Native Activity project. It needs to be fixed before it can run

TestApp uses V8. I tried to compile an Android version a while ago without success. But found this working npm : https://www.npmjs.com/package/react-native-v8
[...]

PR #134, opened 2020-01-17
Integration android aar

CedricGuillemet:
Android AAR Integration project with CI


bghgary:
@ryantrem Do you mind taking a look?

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -0,0 +1,170 @@
...
+Java_BabylonNative_Wrapper_surfaceCreated(JNIEnv* env, jobject obj, jobject surface)
+{
+    if (!runtime)
+    {
+        ANativeWindow *window = ANativeWindow_fromSurface(env, surface);
+
+        int32_t width  = ANativeWindow_getWidth(window);
+        int32_t height = ANativeWindow_getHeight(window);
+
+        runtime = std::make_unique<Babylon::RuntimeAndroid>(window, Root, width, height);
```
ryantrem: I guess for now the runtime's lifetime is tied to the view's lifetime, and we'll break this out later when we support multi-view? Is that tracked in an issue?

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -0,0 +1,170 @@
...
+JNIEXPORT void JNICALL Java_BabylonNative_Wrapper_surfaceChanged(JNIEnv* env, jobject obj, jint width, jint height, jobject surface);
+JNIEXPORT void JNICALL Java_BabylonNative_Wrapper_loadScript(JNIEnv* env, jobject obj, jstring path);
+JNIEXPORT void JNICALL Java_BabylonNative_Wrapper_eval(JNIEnv* env, jobject obj, jstring source, jstring sourceURL);
+JNIEXPORT void JNICALL Java_BabylonNative_Wrapper_setTouchInfo(JNIEnv* env, jobject obj, jfloat dx, jfloat dy, jboolean down);
+};
+
+std::unique_ptr<Babylon::RuntimeAndroid> runtime{};
+std::unique_ptr<InputManager::InputBuffer> inputBuffer{};
+
+static AAssetManager* g_assetMgrNative = nullptr;
```
ryantrem: Move these statics into the anonymous namespace?

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+JNIEXPORT void JNICALL
+Java_BabylonNative_Wrapper_initEngine(JNIEnv* env, jobject obj,
+                                                   jobject assetMgr)
+{
+    auto asset_manager = AAssetManager_fromJava(env, assetMgr);
+    g_assetMgrNative = asset_manager;
+    jboolean iscopy;
```
ryantrem: What is this?

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -0,0 +1,170 @@
...
+                          });
+
+        inputBuffer = std::make_unique<InputManager::InputBuffer>(*runtime);
+        InputManager::Initialize(*runtime, *inputBuffer);
+
+        runtime->LoadScript("Scripts/babylon.max.js");
+        runtime->LoadScript("Scripts/babylon.glTF2FileLoader.js");
+        //runtime->LoadScript("Scripts/experience.js");
+
+        runtime->UpdateSize(width, height);
```
ryantrem: Why do we need this Update when we just passed the width/height to the constructor. Should it be doing this internally?

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+JNIEXPORT void JNICALL
+Java_BabylonNative_Wrapper_surfaceChanged(JNIEnv* env, jobject obj, jint width, jint height, jobject surface)
+{
+    if (runtime)
+    {
+        ANativeWindow *window = ANativeWindow_fromSurface(env, surface);
+        //runtime->UpdateSurface(width, height, window);
```
ryantrem: Why is this commented out? Is something here not working?

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
```diff
@@ -0,0 +1,113 @@
...
+    private boolean mViewReady = false;
+    private ViewDelegate mViewDelegate;
+
+    public BabylonView(Context context, ViewDelegate viewDelegate) {
+        super(context);
+        init(viewDelegate);
+        BabylonNative.Wrapper.initEngine(context.getResources().getAssets());
+    }
+
+    private void init(ViewDelegate viewDelegate) {
```
ryantrem: I would personally just move this into the constructor to make it clear that all this happens only at construction time.

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
```diff
@@ -0,0 +1,113 @@
...
+    public void onResume() {
+        BabylonNative.Wrapper.activityOnResume();
+    }
+
+    /**
+     * This method is part of the SurfaceHolder.Callback interface, and is
+     * not normally called or subclassed by clients of BabylonView.
+     */
+    public void surfaceCreated(SurfaceHolder holder) {
+        BabylonNative.Wrapper.surfaceCreated(getHolder().getSurface());
```
ryantrem: How is the lifetime of the surface determined? Does the ANativeWindow pointer somehow keep it alive? Is there any thread synchronization we need to ensure the surface is not destroyed while the render loop is using it?

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
```diff
@@ -0,0 +1,113 @@
...
+    public void eval(String source, String sourceURL) {
+        BabylonNative.Wrapper.eval(source, sourceURL);
+    }
+
+    public void onPause() {
+        setVisibility(View.GONE);
+        BabylonNative.Wrapper.activityOnPause();
+    }
+
+    public void onResume() {
```
ryantrem: Do we need a corresponding setVisibility call here?

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -0,0 +1,170 @@
...
+                                                   jobject assetMgr)
+{
+    auto asset_manager = AAssetManager_fromJava(env, assetMgr);
+    g_assetMgrNative = asset_manager;
+    jboolean iscopy;
+}
+
+JNIEXPORT void JNICALL
+Java_BabylonNative_Wrapper_finishEngine(JNIEnv* env, jobject obj)
+{
```
ryantrem: Is something supposed to be done here? I'm guessing there is some unfinished cleanup work? Is that tracked in an issue, and if so, can we reference it here for context?

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * This method is part of the SurfaceHolder.Callback2 interface, and is
+     * not normally called or subclassed by clients of BabylonView.
+     */
+    @Deprecated
+    @Override
+    public void surfaceRedrawNeeded(SurfaceHolder holder) {
+        // Since we are part of the framework we know only surfaceRedrawNeededAsync
+        // will be called.
```
ryantrem: I don't really understand this comment. The default behavior of surfaceRedrawNeededAsync is to just call surfaceRedrawNeeded, so I guess that means we are saying nothing happens on surfaceRedrawNeeded/surfaceRedrawNeededAsync?

ryantrem:
Are the updates to readme.md intentionally excluded?

Integration/AndroidAAR/BabylonNative/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+
+set(BABYLON_NATIVE_PACKAGE_DIR "${CMAKE_CURRENT_LIST_DIR}/..")
+add_subdirectory(${BABYLON_NATIVE_PACKAGE_DIR}/../../Library/ ${BABYLON_NATIVE_PACKAGE_DIR}/../../build/Android_${CMAKE_ANDROID_ARCH_ABI}/)
+
+add_library(BabylonNativeJNI SHARED
+            src/main/cpp/BabylonNativeJNI.cpp
+            ${BABYLON_NATIVE_PACKAGE_DIR}/../../Library/Dependencies/napi/source/js_native_api_v8.cc
+            ${BABYLON_NATIVE_PACKAGE_DIR}/../../Library/Dependencies/napi/source/env_v8.cc
+            ${BABYLON_NATIVE_PACKAGE_DIR}/../../Library/Dependencies/napi/source/env.cc
+            ${BABYLON_NATIVE_PACKAGE_DIR}/../../TestApp/Source/Shared/InputManager.cpp)
```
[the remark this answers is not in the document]
CedricGuillemet: Yes! It will be refactored later. The checked in jar are used by gradle. I don't know exactly what it's been used for.

Integration/AndroidAAR/BabylonNative/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/src/")
+
+set(BABYLON_NATIVE_PACKAGE_DIR "${CMAKE_CURRENT_LIST_DIR}/..")
+add_subdirectory(${BABYLON_NATIVE_PACKAGE_DIR}/../../Library/ ${BABYLON_NATIVE_PACKAGE_DIR}/../../build/Android_${CMAKE_ANDROID_ARCH_ABI}/)
+
+add_library(BabylonNativeJNI SHARED
+            src/main/cpp/BabylonNativeJNI.cpp
+            ${BABYLON_NATIVE_PACKAGE_DIR}/../../Library/Dependencies/napi/source/js_native_api_v8.cc
```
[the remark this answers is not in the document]
CedricGuillemet: v8 is not supported in cmake for Android. This needs more work to pull the npm, untar,... There is an issue : https://github.com/BabylonJS/BabylonNative/issues/102#issuecomment-554454182

Integration/AndroidAAR/BabylonNative/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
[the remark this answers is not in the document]
CedricGuillemet: I've updated the root .gitignore.

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
CedricGuillemet: unused boolean removed now.

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+std::unique_ptr<Babylon::RuntimeAndroid> runtime{};
+std::unique_ptr<InputManager::InputBuffer> inputBuffer{};
+
+static AAssetManager* g_assetMgrNative = nullptr;
+static const char* Root = "file://";
+namespace
+{
+    // this is the way to load apk embedded assets.
+    static std::vector<char> GetAssetContents(const char* filename)
```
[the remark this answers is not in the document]
CedricGuillemet: Removed. Note this PR should not be used for building aar for now. it's only a placeholder. Until we have the plugin system, loading assets from apk will not work. That's why the androidRender branch is still under development.

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
CedricGuillemet: Any opinion on that point @bghgary ?
Anonymous namespace is not used in other testapp.
I'm fine with ou without.

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
CedricGuillemet: Comment updated

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
CedricGuillemet: Adding reset calls instead of letting the system do the job.

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
CedricGuillemet: sounds good.

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
CedricGuillemet: Good point! I've put a mecanism to pause/resume rendering. I have to find a good way to test it.

Integration/AndroidAAR/BabylonNative/src/main/java/BabylonNative/BabylonView.java
CedricGuillemet: It's done in the activity based on another method (onWindowFocusChanged). I will check when working on views/surface if it can be simplified.

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
CedricGuillemet: uncommented.

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
CedricGuillemet: Removed

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
CedricGuillemet: Can you please open an issue with all your needs on views.

Integration/AndroidAAR/BabylonNative/build.gradle
```diff
@@ -0,0 +1,77 @@
...
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles 'consumer-rules.pro'
+
+        externalNativeBuild {
+            cmake {
+                abiFilters 'arm64-v8a', 'armeabi-v7a', 'x86', 'x86_64'
+                arguments '-DANDROID_STL=c++_static',
```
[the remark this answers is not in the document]
CedricGuillemet: BabylonNative.so is built with c++static. The libc++_shared.so libraries are dynamically linked by v8

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
bghgary: I don't have a strong preference, but I would prefer not to mix static and anonymous namespace.

CedricGuillemet: static removed.

Integration/AndroidAAR/BabylonNative/.gitignore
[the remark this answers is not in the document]
CedricGuillemet: Nop. Android Studio generates files nested in the hierarchy

.gitignore
```diff
@@ -1 +1,16 @@
...
+.idea
+Library/Dependencies/v8/include
+.gradle
+.DS_Store
+build/
+captures/
+.externalNativeBuild
+*.apk
+*.ap_
+*.log
```
bghgary: I don't think the Android ignore list should be at the root. At the minimum, put this in the Android specific folder.

Integration/AndroidAAR/BabylonNative/CMakeLists.txt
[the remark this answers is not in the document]
CedricGuillemet: It's more complicated than it looks (see the issue linked). I think it can be done with less work on the CI because it's using a Mac image to build Android. On Windows, it needs tools to uncompress the archive.

CedricGuillemet:
@syntheticmagus Can you also bundle jni .so and babylon.js files in the JAR?

.gitignore
CedricGuillemet: That means the gitignore have to be updated for every Android Studio project (testapp + integration for now). Are you ok with that?

Integration/AndroidAAR/BabylonNative/CMakeLists.txt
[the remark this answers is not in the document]
ryantrem: Should we add a comment here with a link to the issue tracking this?

Integration/AndroidAAR/BabylonNative/src/main/cpp/BabylonNativeJNI.cpp
ryantrem: https://github.com/BabylonJS/BabylonNative/issues/147

Integration/AndroidAAR/BabylonNative/CMakeLists.txt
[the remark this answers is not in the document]
CedricGuillemet: Comment added.

.gitignore
bghgary: There has to be a better way to do this. Ignoring all .cxx, .log, .so, etc. seems a bit extreme.

syntheticmagus:
Sorry for the delay, @CedricGuillemet, just getting to this. Yes, you can bundle .so and JS files into a JAR file, and it appears to be able to find everything as expected. I have a minimal demo where I bundled a Java class, a small C++ library, and a JS resource file into a single JAR and was able to use them all in an Android app by just drag-and-dropping them into my Android Studio project's libs directory.

CedricGuillemet:
No approval in 2 months. Closing.

PR #138, opened 2020-01-24
XR Input

syntheticmagus:
This is by no means a fully-featured capability yet, but it should be in a good state for what it is, and it's now taken on something of the character of a bug fix. As of about two weeks ago, our XR capabilities broke in master because of changes made to Babylon.js, many of which had already been accounted for in this branch. My original plan had been to not check in this branch until I'd added a few more features (primarily buttons and controller models), but considering the circumstances I decided to alter that plan.


Library/Source/NativeXr.cpp
```diff
@@ -93,11 +95,46 @@ namespace
...
+        constexpr char* TARGET_RAY_MODE{ "tracked-pointer" };
+
+        auto jsInputSource = Napi::Object::New(env);
+        jsInputSource.Set("handedness", Napi::String::New(env, HANDEDNESS_STRINGS[static_cast<size_t>(inputSource.Handedness)]));
+        jsInputSource.Set("targetRayMode", TARGET_RAY_MODE);
+        SetXRInputSourceSpaces(jsInputSource, inputSource);
+
+        // auto profiles = Napi::Array::New(env);
+        // Napi::Value string = Napi::String::New(env, "generic-trigger-squeeze-touchpad-thumbstick");
+        // profiles.Set(0, string);
```
bghgary: what's this?

syntheticmagus: WIP feature from when I was waiting to consume a new Babylon.js; it's how Babylon will eventually know what controller model to display. Doesn't show it yet, but the implementation doesn't hurt anything and needs to be there anyway, so I just went ahead and uncommented this and finished it in my most recent check-in.

syntheticmagus:
https://github.com/BabylonJS/BabylonNative/issues/140

For the `TODO`s.

PR #139, opened 2020-01-24
Removed remaining static constructors from NAPI object wrappers.

syntheticmagus:


syntheticmagus:
Currently a draft only because the XR portions haven't been tested yet, so this will need to be merged #138 and go in after that.

Issue #142, opened 2020-01-30
Order of destruction among inter-dependent N-API objects


syntheticmagus:
As [currently written](https://github.com/BabylonJS/BabylonNative/tree/2a964b59978cb734996d85597b1f4b88cdef8140), `NativeEngine` has a direct dependency on `NativeWindow` and _must_ be destroyed first. Specifically, `NativeEngine` holds onto [a ticket](https://github.com/BabylonJS/BabylonNative/blob/2a964b59978cb734996d85597b1f4b88cdef8140/Library/Source/NativeEngine.h#L393) from `NativeWindow`'s [ticketed collection](https://github.com/BabylonJS/BabylonNative/blob/2a964b59978cb734996d85597b1f4b88cdef8140/Library/Source/NativeWindow.h#L36); if the collection is destroyed before that ticket is destroyed, teardown will fail. The only reason we haven't seen this manifested in crashes before is because the problem was being hidden by multiple other teardown problems, which I'll link to here presently. (#94, #143)

Since Chakra (perhaps like other JS engines) doesn't guarantee destruction order for the objects it owns, this raises some tricky questions about how we get our N-API objects to tear themselves down properly. We can theoretically have dependency links into, out of, and around JavaScript. Can we come up with some way to guarantee these will be shut down in a particular order?

syntheticmagus:
Alternatively, should we just work around that problem by forbidding our plugin objects from taking strong references to each other? That would match JavaScript's ownership model (or lack thereof) better, but it might make adding relationships among native objects a bit funkier.

syntheticmagus:
Current plan-of-record is to adopt the "no strong references" model described above. Seems to be working, haven't seen any issues related to order-of-destruction since #157.

PR #146, opened 2020-01-31
Minimal change to get Chakra consistently shutting down properly

syntheticmagus:
Contains some...questionable options. Not sure if this should be checked in or not, but it's probably worth discussing at least. There are fixes in here for three separate problems that were causing Chakra to either fail to shut down or to shut down improperly. I'll call each of the problems/fixes out in comments and explain whether/why iffy about the alternative presented.

### DIFF
--- a/Library/Dependencies/napi/source/env_chakra.cc
+++ b/Library/Dependencies/napi/source/env_chakra.cc
@@ -60,11 +60,9 @@ namespace Babylon
 
         ~napi_env_local()
         {
-            JsErrorCode errorCode = JsDisposeRuntime(m_jsRuntime);
-            if (errorCode != JsErrorCode::JsNoError)
-            {
-                Log(L"Failed to dispose runtime with error code (%u)\n", errorCode);
-            }
+            ThrowIfFailed(JsSetCurrentContext(JS_INVALID_REFERENCE));
+
+            ThrowIfFailed(JsDisposeRuntime(m_jsRuntime));
         }
 
     private:
--- a/Library/Source/NativeEngine.cpp
+++ b/Library/Source/NativeEngine.cpp
@@ -33,6 +33,37 @@ namespace Babylon
 {
     namespace
     {
+        class {
+        public:
+            void RecordBgfxInit()
+            {
+                assert(!m_isBgfxInitialized);
+                m_isBgfxInitialized = true;
+                m_bgfxInitializationId += 1;
+            }
+
+            void RecordBgfxShutdown()
+            {
+                assert(m_isBgfxInitialized);
+                m_isBgfxInitialized = false;
+            }
+
+            bool IsBgfxInitialized() const
+            {
+                return m_isBgfxInitialized;
+            }
+
+            BgfxInitializationId GetBgfxInitializationId() const
+            {
+                assert(m_isBgfxInitialized);
+                return m_bgfxInitializationId;
+            }
+
+        private:
+            bool m_isBgfxInitialized{ false };
+            BgfxInitializationId m_bgfxInitializationId{ 0 };
+        } BgfxStatus;
+
         template<typename AppendageT>
         inline void AppendBytes(std::vector<uint8_t>& bytes, const AppendageT appendage)
         {
@@ -275,6 +306,19 @@ namespace Babylon
             bgfx::TextureFormat::RGBA32F};
     }
 
+    ProgramData::~ProgramData()
+    {
+        // Because this class can be owned by JavaScript, it is possible for it to sometimes outlive 
+        // the bgfx initialization on which it depends. This is okay, as bgfx::shutdown will clean it
+        // up anyway; however, to avoid attempting to clean it up twice, we must make sure we only
+        // call bgfx::destroy() if the originating bgfx initialization is still active at the time
+        // this is destroyed.
+        if (BgfxStatus.IsBgfxInitialized() && BgfxStatus.GetBgfxInitializationId() == m_bgfxInitId)
+        {
+            bgfx::destroy(Program);
+        }
+    }
+
     void NativeEngine::InitializeWindow(void* nativeWindowPtr, uint32_t width, uint32_t height)
     {
         // Initialize bgfx.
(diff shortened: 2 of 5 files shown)


Library/Dependencies/napi/source/env_chakra.cc
```diff
@@ -60,11 +60,9 @@ namespace Babylon
 
         ~napi_env_local()
         {
-            JsErrorCode errorCode = JsDisposeRuntime(m_jsRuntime);
-            if (errorCode != JsErrorCode::JsNoError)
-            {
-                Log(L"Failed to dispose runtime with error code (%u)\n", errorCode);
-            }
+            ThrowIfFailed(JsSetCurrentContext(JS_INVALID_REFERENCE));
```
syntheticmagus: First and least controversial change, allows Chakra to shut down correctly. I have _not_ yet confirmed that this resolves the memory leak we were seeing before as there's a problem I just found with refresh; I'll update this comment when I fix that and confirm (or otherwise).

Library/Source/NativeEngine.cpp
```diff
@@ -33,6 +33,37 @@ namespace Babylon
 {
     namespace
     {
```
syntheticmagus: Definitely the weirdest "fix" in here, and the one I'm least sure about. bgfx doesn't seem to have a mechanism for asking it whether it's alive or not, but some JavaScript-owned objects (relevantly the `ProgramData`) can end up being destroyed after `NativeEngine` has already been destroyed and `bgfx::shutdown` has consequently already been called. They could also theoretically survive long enough to last into another bgfx initialization (if that's possible), which would be disastrous as their handles would now point to things inside bgfx that were really created and owned by somebody else. This global state holder was created to keep track of this. I'm not a fan of it, however, so if anybody has a recommendation of what would be better, please let me know.

Library/Source/NativeWindow.h
```diff
@@ -15,11 +15,26 @@ namespace Babylon
         static Napi::FunctionReference GetAToBFunction(Napi::ObjectReference& nativeWindow);
 
         NativeWindow(const Napi::CallbackInfo& info);
+        ~NativeWindow();
 
         void Resize(size_t newWidth, size_t newHeight);
 
         using OnResizeCallback = std::function<void(size_t, size_t)>;
-        using OnResizeCallbackTicket = arcana::ticketed_collection<OnResizeCallback>::ticket;
```
syntheticmagus: The third and middle-est change removes the dependency on Arcana's `ticketed_collection` and just makes a small bespoke alternative instead. This was to get around the order-of-destruction problem described in #142, which we should still consider how to deal with long-term but may not be as tricky as I thought at first. In this particular case, all that's necessary to allow the `NativeWindow` and `NativeEngine` to be destroyed in arbitrary orders is for the `OnResizeCallbackTicket` to simply do nothing if the collection it came from was destroyed before it was. `arcana::ticketed_collection` doesn't support this, and adding that is a work-in-progress due to the fact that `ticketed_collection` is supposed to be thread-safe. But thread-safety isn't a concern here as these are both Babylon Native plugins that should only talk to each other on the same thread, so we can get away with doing something simpler. (And we can always move back to the Arcana type when it's ready, if we want to.)

Library/Dependencies/napi/source/env_chakra.cc
syntheticmagus: Fixed the problems with refresh. Tested to confirm that refresh works, and has consistent and sane memory behavior, on Win32-Chakra, Win32-V8, and UWP-Chakra. UWP-V8 builds but does not run, which seems to be no change from master.

Library/Source/NativeEngine.cpp
CedricGuillemet: There are 2 resources to free here : the bgfx shader handle and the JS object. I'd try to keep a list of ProgramData (register in ctor, unregister in dtor). When bgfx::shutdown is called, I'd run thru all the registered ProgramData and free their bgfx handle. I'd then let JS engine free the ProgramData when ever it needs to. I think the same mechanism happens with TextureData. I prefer to be pro active in destruction than trying to determine a state(m_isBgfxInitialized/m_bgfxInitializationId) where I can or can't do something. The more states, the more potential bugs.

syntheticmagus: Definitely agree that that's cleaner, and it seems the ticketed mechanism can be adapted for this purpose. I've refactored to try this other approach.

Library/Source/NativeEngine.cpp
```diff
@@ -620,11 +625,10 @@ namespace Babylon
 
         programData->Program = bgfx::createProgram(vertexShader, fragmentShader, true);
 
-        auto finalizer = [](Napi::Env, ProgramData* data) {
-            delete data;
-        };
-
-        return Napi::External<ProgramData>::New(info.Env(), programData, finalizer);
+        auto* programDataPtr = programData.get();
```
syntheticmagus: `rawProgramData`

Library/Source/ticketed_collection.h
```diff
@@ -0,0 +1,82 @@
...
+
+            ticket(T&& value, MapT& collection)
+                : m_isCollectionStillAlive{ new bool{true} }
+                , m_collection{ collection }
+            {
+                m_collection[m_isCollectionStillAlive] = std::move(value);
+            }
+
+            bool* m_isCollectionStillAlive{};
+            MapT& m_collection;
```
bghgary: We discussed this offline. I would recommend using a double pointer here instead.

Library/Dependencies/napi/source/env_chakra.cc
syntheticmagus: #148

Library/Source/ticketed_collection.h
```diff
@@ -0,0 +1,82 @@
+#pragma once
+
+#include <map>
+
+namespace Babylon
+{
+    // NOTE: This type is not thread-safe.
+    template<typename T>
+    class ticketed_collection
```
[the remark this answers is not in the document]
syntheticmagus: Just saw this is done on line 27. Moving it above the destructor for clarity.

PR #149, opened 2020-02-07
Move JavaScript creation responsibilities outside of N-API

syntheticmagus:


Library/Dependencies/napi/include/napi/env.h
```diff
@@ -2,15 +2,11 @@
...
-namespace Babylon
+namespace Napi
 {
-    class Env : public Napi::Env
-    {
-    public:
-        explicit Env(const char* executablePath, std::function<void(std::function<void()>)> executeOnScriptThread);
-        Env(const Env&) = delete;
-        ~Env();
+    template<typename ...Ts> Napi::Env Attach(Ts... args);
```
bghgary: If we are expecting to push this back to node, I'm not sure the usage of variadic template arguments will be accepted.

Library/Source/RuntimeImplV8.cpp
```diff
@@ -0,0 +1,78 @@
...
+            create_params.array_buffer_allocator = v8::ArrayBuffer::Allocator::NewDefaultAllocator();
+            return v8::Isolate::New(create_params);
+        }
+
+        void DestroyIsolate(v8::Isolate* isolate)
+        {
+            // todo : GetArrayBufferAllocator not available?
+            //delete isolate->GetArrayBufferAllocator();
+            isolate->Dispose();
+        }
```
bghgary: not sure these helper function are useful anymore. perhaps inline them in BaseThreadProcedure?

Library/Dependencies/napi/include/napi/env.h
syntheticmagus: Talked offline. [It wouldn't be Node's first use of variadics](https://github.com/nodejs/node-addon-api/blob/7f56a78ff7eab77f1316696aef1de3304b70e639/napi-inl.h#L900-L904), so I think we'll be okay. If we're not, though, it shouldn't be too difficult to switch to another option (single template type, separate headers, or even [choke] arglist) when the objection is raised.

PR #152, opened 2020-02-12
Create JsRuntime

syntheticmagus:
Beginning the real decoupling process by removing all direct dependencies on `RuntimeImpl` and replacing them with dependencies on the new `JsRuntime`. This has resulted in some temporary code duplication as the Babylon Native API hasn't been changed yet, only its internal structure, so some things like URL loading end up semi-duplicated as ownership of capabilities is being moved from one place to another.


Library/Source/NetworkUtils.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include <string>
+#include <vector>
+
+namespace Babylon
+{
+    std::string GetAbsoluteUrl(const std::string& url, const std::string& rootUrl);
+
+    arcana::task<std::string, std::exception_ptr> LoadUrlToStringAsync(const std::string url);
+    arcana::task<std::vector<uint8_t>, std::exception_ptr> LoadUrlToBytesAsync(const std::string url);
```
bghgary: nit: how about `LoadTextAsync` and `LoadBinaryAsync` for the function names?

Library/Source/RuntimeImpl.cpp
```diff
@@ -77,25 +55,27 @@ namespace Babylon
 
     void RuntimeImpl::LoadScript(const std::string& url)
     {
-        auto lock = AcquireTaskLock();
-        auto whenAllTask = arcana::when_all(LoadUrlAsync<std::string>(GetAbsoluteUrl(url).data()), Task);
-        Task = whenAllTask.then(*m_dispatcher, m_cancelSource, [this, url](const std::tuple<std::string, arcana::void_placeholder>& args) {
+        std::scoped_lock lock{ m_taskMutex };
```
bghgary: nit: spacing everywhere

```suggestion
        std::scoped_lock lock{m_taskMutex};
```

Library/Source/XMLHttpRequest.cpp
```diff
@@ -27,14 +29,15 @@ namespace Babylon
                 InstanceMethod("removeEventListener", &XMLHttpRequest::RemoveEventListener),
                 InstanceMethod("open", &XMLHttpRequest::Open),
                 InstanceMethod("send", &XMLHttpRequest::Send),
-            });
+            }, const_cast<char*>(rootUrl));
```
bghgary: this is weird.
```suggestion
            }, reinterpret_cast<void*>(rootUrl));
```

Library/Source/XMLHttpRequest.cpp
```diff
@@ -130,30 +133,27 @@ namespace Babylon
...
         if (m_responseType.empty() || m_responseType == XMLHttpRequestTypes::ResponseType::Text)
         {
-            return m_runtimeImpl.LoadUrlAsync<std::string>(m_url).then(arcana::inline_scheduler, m_runtimeImpl.Cancellation(), [this](const std::string& data)
+            return LoadUrlToStringAsync(m_url).then(arcana::inline_scheduler, arcana::cancellation::none(), [this](const std::string& data)
             {
-                // check UTF-8 BOM encoding
-                if (data.size() >= 3 && data[0] == '\xEF' && data[1] == '\xBB' && data[2] == '\xBF')
-                {
-                    m_responseText = data.substr(3);
-                }
```
bghgary: I think this code got deleted in the move?

Library/Source/XMLHttpRequest.cpp
```diff
@@ -27,14 +29,15 @@ namespace Babylon
...
 
-        return Napi::Persistent(func);
+        env.Global().Set(JS_XML_HTTP_REQUEST_CONSTRUCTOR_NAME, func);
     }
 
     XMLHttpRequest::XMLHttpRequest(const Napi::CallbackInfo& info)
         : Napi::ObjectWrap<XMLHttpRequest>{info}
-        , m_runtimeImpl{RuntimeImpl::GetRuntimeImplFromJavaScript(info.Env())}
+        , m_runtime{JsRuntime::GetFromJavaScript(info.Env())}
+        , m_rootUrl{reinterpret_cast<const char*>(info.Data())}
```
bghgary:
```suggestion
        , m_rootUrl{static_cast<const char*>(info.Data())}
```

Library/Source/RuntimeImpl.cpp
syntheticmagus: Did it, though I'm generally not a huge fan of more terse styles. Is there an established convention we're trying to align with on that?

bghgary: LLVM default for clang format

Issue #118, opened 2019-12-09
Window resizing issues


CedricGuillemet:
Because of this crash on macos: https://github.com/BabylonJS/BabylonNative/issues/151
now bgfx::frame is called when resizing, replacing a call to bgfx::touch
Crash is fixed and no strange behavior appearing on win32/Android.

- [x] On UWP a flickering was experienced before, now the refresh happens 0.5s after resizing, without flickering.

On iOS, there is a Metal exception when a bgfx::touch(0) is not performed once a frame.

bghgary:
Blocked on the rendering rework (#345)

bghgary:
No, it possibly has made it behave differently, but there are still issues.

bghgary:
I need to check if there are still issues.

CedricGuillemet:
Resize and resume tested OK on Android, win32, uwp, ios and mac. Closing.

PR #153, opened 2020-02-13
Preparation for merging AR

CedricGuillemet:
Please note the temporary Hack for asset loading.
Lot of code removed for clarity after iterations with Ryan.


Library/Source/NativeEngine.cpp
```diff
@@ -407,7 +411,7 @@ namespace Babylon
         {
             bgfx::reset(w, h, BGFX_RESET_FLAGS);
             bgfx::setViewRect(0, 0, 0, w, h);
-            bgfx::touch(0);
+            bgfx::frame();
```
syntheticmagus: Is this workaround something we should check in? If so, should we create an issue to track it and change it back once we figure out the right way of making this work on Metal?

Library/Source/RuntimeImpl.cpp
```diff
@@ -59,6 +65,23 @@ namespace Babylon
         });
     }
 
+    void RuntimeImpl::UpdateWindow(float width, float height, void* nativeWindowPtr)
```
syntheticmagus: Is this a new direct dependency? I'm not sure we want to have this file (or what it will eventually turn into) depend on bgfx.

Library/Source/NativeEngine.cpp
CedricGuillemet: Issue updated here : https://github.com/BabylonJS/BabylonNative/issues/118

Library/Source/RuntimeImpl.cpp
CedricGuillemet: yes, new contract function for Android.

PR #154, opened 2020-02-13
Fix to make updateSize work, at least in one way, across platforms

syntheticmagus:


syntheticmagus:
Abandoning, will just use #define instead.

PR #157, opened 2020-02-20
Plugin Refactor

syntheticmagus:
This is the refactor to convert Babylon Native to relying on plugins as its preeminent variety of component. This change is intended to improve scalability by "flattening" the structure of the repository, moving certain responsibilities outside the library and allowing a looser coupling of components. This change is by no means finished; but with the build system working and the apps of the four primary Windows builds confirmed to run, it is now at least ready to be looked at (and, if desired, downloaded and tried).


Apps/Playground/UWP/App.cpp
```diff
@@ -165,15 +163,34 @@ concurrency::task<void> App::RestartRuntimeAsync(Windows::Foundation::Rect bound
         });
     });
 
+    // Initialize NativeWindow plugin.
+    size_t width = static_cast<size_t>(bounds.Width * m_displayScale);
+    size_t height = static_cast<size_t>(bounds.Height * m_displayScale);
+    auto* windowPtr = reinterpret_cast<ABI::Windows::UI::Core::ICoreWindow*>(CoreWindow::GetForCurrentThread());
+    m_runtime->Dispatch([windowPtr, width, height](Napi::Env env)
+    {
+        Babylon::NativeWindow::Initialize(env, windowPtr, width, height);
```
ryantrem: Thinking about how this will work later with multi-view... maybe we would initialize the NativeWindow with just the env, and then call a function on it like RegisterView that takes the windowPtr, width, and height? Is it worth trying to get the right API now even if we only support one window/view?

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -12,16 +12,21 @@
...
 
 #define MAX_LOADSTRING 100
 
     // Global Variables:
 HINSTANCE hInst;                                // current instance
 WCHAR szTitle[MAX_LOADSTRING];                  // The title bar text
 WCHAR szWindowClass[MAX_LOADSTRING];            // the main window class name
-std::unique_ptr<Babylon::RuntimeWin32> runtime{};
+std::unique_ptr<Babylon::AppRuntime> runtime{};
+std::string rootUrl{};
```
bghgary: why is this global now?

Core/AppRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+target_include_directories(AppRuntime INTERFACE "Include")
+
+target_link_libraries(AppRuntime
+    PRIVATE arcana
+    PUBLIC JsRuntime)
+
+target_compile_definitions(AppRuntime
+    PRIVATE NOMINMAX)
+
+    set_property(TARGET AppRuntime PROPERTY FOLDER Core)
```
bghgary:
```suggestion
set_property(TARGET AppRuntime PROPERTY FOLDER Core)
```

Core/AppRuntime/Include/Babylon/AppRuntime.h
```diff
@@ -0,0 +1,32 @@
...
+{
+    class TaskChain;
+
+    class AppRuntime final : public JsRuntime
+    {
+    public:
+        AppRuntime(const char* rootUrl = nullptr);
+        ~AppRuntime();
+
+        const std::string& GetRootUrl() const;
```
bghgary: We should talk about the convention for getters in our code.

```suggestion
        const std::string& RootUrl() const;
```

Core/AppRuntime/Source/AppRuntimeChakra.cpp
```diff
@@ -17,8 +18,14 @@ namespace Babylon
...
 
-    void RuntimeImpl::BaseThreadProcedure()
+    void AppRuntime::CreateEnvironmentAndRun()
     {
+#if WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP
+        HRESULT hr = CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);
+        assert(SUCCEEDED(hr));
+        auto coInitScopeGuard = gsl::finally([] { CoUninitialize(); });
+#endif
+
```
bghgary: This seems like it should be outside of this class and in AppRuntimeWin32 or something.

Core/AppRuntime/Source/TaskChain.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "Common.h"
+
+#include <arcana/threading/task.h>
+#include <napi/env.h>
+
+namespace Babylon
+{
+    class TaskChain
```
ryantrem: Consider naming this `WorkQueue` (or `JSWorkQueue`) (I feel like that is a more standard name for this concept).

Core/AppRuntime/Source/TaskChain.h
```diff
@@ -0,0 +1,36 @@
...
+        void RunTaskChain(Napi::Env);
+
+    private:
+        std::optional<Napi::Env> m_env{};
+
+        std::mutex m_appendMutex{};
+        std::mutex m_blockingTickMutex{};
+        std::mutex m_suspendMutex{};
+        std::condition_variable m_suspendConditionVariable{};
+        bool m_suspended{ false };
```
bghgary: nit: might consider auto formatting all the files

Core/JsRuntime/Source/JsRuntime.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    JsRuntime::JsRuntime(DispatchFunctionT&& dispatchFunction)
+        : m_dispatchFunction{ dispatchFunction }
+    {
+    }
+
+    void JsRuntime::Initialize(Napi::Env env, DispatchFunctionT dispatchFunction)
+    {
+        JsRuntime* runtime = new JsRuntime(std::move(dispatchFunction));
+        runtime->AddJavaScriptReference(env, true);
```
bghgary: I think it's clearer if you passed a lambda instead of true here:

```suggestion
        runtime->AddJavaScriptReference(env, [](Napi::Env, JsRuntime* jsRuntime) { delete jsRuntime; });
```

Core/AppRuntime/Source/TaskChain.cpp
```diff
@@ -0,0 +1,69 @@
+#include "TaskChain.h"
+
+namespace Babylon
+{
+    TaskChain::TaskChain(std::function<void()> threadProcedure)
```
ryantrem: It seems weird that this has to be passed in, and the assumption is that this function eventually calls `RunTaskChain`. Why can't `TaskChain` start it's own work queue thread, and the `TaskChain` consumer just call `Append` to get work injected into the thread?

Core/JsRuntime/Source/JsRuntime.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Babylon
+{
+    namespace
+    {
+        static constexpr auto JS_RUNTIME_NAME = "runtime";
+        static constexpr auto JS_WINDOW_NAME = "window";
+    }
+
+    JsRuntime::JsRuntime(DispatchFunctionT&& dispatchFunction)
```
bghgary: This shouldn't be a &&. There is no reason to force people to move the dispatch function into the constructor.

```suggestion
    JsRuntime::JsRuntime(DispatchFunctionT dispatchFunction)
```

Use std::move like you are doing in the other constructor to move. The calling code can choose whether to move or not.

Dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,131 @@
+set(CMAKE_CXX_STANDARD 17)
```
bghgary: I'm not sure I understand the point of this file. Are we expecting that anyone who wants to use any dependencies add this as a subdirectory in CMakeLists.txt?

Apps/Playground/UWP/App.cpp
```diff
@@ -165,15 +163,34 @@ concurrency::task<void> App::RestartRuntimeAsync(Windows::Foundation::Rect bound
...
+    });
+
+    // Initialize NativeEngine plugin.
+    Babylon::InitializeNativeEngine(*m_runtime, windowPtr, width, height);
```
ryantrem: I'm a bit confused why the windowPtr, width, and height need to be passed to both NativeWindow and NativeEngine.

Dependencies/babylon_native_utils/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+set(SOURCES
+    "Include/Babylon/NetworkUtils.h"
+    "Include/Babylon/ticketed_collection.h"
```
bghgary: Maybe TicketedCollection.h to be consistent in naming schemes?

Dependencies/babylon_native_utils/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+    "Source/NetworkUtils.cpp")
+
+add_library(BabylonNativeUtils ${SOURCES})
+
+target_include_directories(BabylonNativeUtils PRIVATE "Include/Babylon")
```
bghgary: Ugh, this naming scheme is killing me. The directory name is snake case but the library is upper camel case?

Dependencies/babylon_native_utils/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+target_include_directories(BabylonNativeUtils INTERFACE "Include")
+
+target_link_libraries(BabylonNativeUtils
+    PUBLIC arcana
```
bghgary: Is it expected that arcana is already added?

Apps/Playground/UWP/App.cpp
syntheticmagus: This is actually an analogue to the JavaScript concept of the [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window), so this particular aspect of it likely won't change for multiview scenarios as I think that's a separate issue.

Apps/Playground/UWP/App.cpp
syntheticmagus: This, in part, is just me porting over what the existing API was rather than making too many invasive API changes at this point in the refactor. However, the reason the API was that way is because of macOS and iOS. Those three values are required (in the current implementation) for the initialization of bgfx, and my understanding is that the Apple operating systems require bgfx to be initialized on the main thread, where it doesn't have access to the `NativeWindow` because the `Napi::Env` can't be accessed on that thread and may not even exist yet anyway.

To be clear, this is something I'd definitely like to think about revising, but I'm trying to keep cleanup work on the plugins themselves to a minimum in this change. I'm currently hoping to limit this PR as much as possible to restructuring, after which we can tackle the individual plugins in relative isolation.

Apps/Playground/Win32/App.cpp
```diff
@@ -100,27 +99,54 @@ namespace
...
+            Babylon::NativeWindow::Initialize(env, hWnd, width, height);
+        });
+
+        // Initialize NativeEngine plugin.
+        Babylon::InitializeNativeEngine(*runtime, hWnd, width, height);
+
+        // Initialize XMLHttpRequest plugin.
+        runtime->Dispatch([rootUrl](Napi::Env env)
+        {
+            Babylon::XMLHttpRequest::Initialize(env, rootUrl.data());
```
syntheticmagus: This is dangerous and probably doesn't work at all, must have dodged tests: the char* being passed to `Initialize` points to a copy of rootUrl local to this lambda, which will get cleaned up long before any `XMLHttpRequest` objects are created. `XMLHttpRequest::Initialize` should make a copy of the text and put it somewhere safe.

Apps/ValidationTests/Win32/App.cpp
syntheticmagus: Workaround for dangerous design decision that makes `XMLHttpRequest::Initialize` take a `const char*` that must outlive the `JsRuntime`. Touched on this in the comment I made in `.../Win32/App.cpp`. When that issue is fixed, this variable should also stop being global.

Core/AppRuntime/Include/Babylon/AppRuntime.h
syntheticmagus: Since it's `const` anyway, maybe we should just make the member public.

Core/AppRuntime/Source/TaskChain.cpp
syntheticmagus: It might be possible to do it that way; I'll have to think about it more. This is a fairly direct port of the existing mechanism, and the main reasoning behind that mechanism is because of stack variables. In particular, V8 wants its initialization and state to be stored on the stack, and it goes to great lengths to make it difficult to do anything else. Thus, to use V8 correctly, we need to actually call `TaskChain::RunTaskChain` from inside the scope that contains the V8 state, rather than calling a task at the start to set up V8 and presumably calling another later to tear it down.

Core/JsRuntime/Source/JsRuntime.cpp
syntheticmagus: Since this is a protected function, that externalizes a little bit of implementation knowledge from this class that I'm not sure needs to be externalized. At present, the fact that `JsRuntime` is stored up on the JavaScript as a `Napi::External` which may or may not have a custom deleter is an implementation detail; inheritors need to know whether or not JavaScript owns the JsRuntime, but they don't need to know why that matters or what's different as a result. We've only got one inheritor at present, and it's just the inheritance contract anyway, so I don't think it's too big of a deal either way; but unless you feel strongly otherwise, my _slight_ preference would be to leave this argument as a boolean to keep the fact that this is stored as a `Napi::External` out of the contract. Again, though, I don't feel _too_ strongly about it either way.

Core/JsRuntime/Source/JsRuntime.cpp
syntheticmagus: The reasoning here was that this is protected, and that in this case we actually do want to force the function to be moved in. I don't think there's ever a case where we want this function to be passed by copy because there's no reason for anybody else to have this lambda. In such a circumstance, is it better to still leave the choice up to the calling code?

Dependencies/CMakeLists.txt
syntheticmagus: Spoke offline. This file allows the Dependencies folder to participate in the CMake pattern shared by the other major folders (Core, Plugins, and Apps). Most of the code in here is currently just ported directly from the old build system, but hopefully we'll improve it as we go to make it as clean and airtight as possible over time.

Dependencies/babylon_native_utils/CMakeLists.txt
syntheticmagus: Yes. The pattern being established here is that plugins (and modules following the plugin pattern) do not supply their own dependencies, but instead assume they exist as a part of a larger build system that bears the responsibility of making sure all dependencies are loaded in the right order. This allows the inclusions to scale horizontally and, when applicable, share dependencies amongst themselves.

CedricGuillemet:
Not much to say on the implementation. I'm focusing more on the architecture.
- "BabylonNativeUtils" folder : try to avoid too generic terms for 'utils' (like 'data'). For some LoL : https://twitter.com/skaven_/status/1126645815332593664 I'd go for 'helper'
- ScriptLoader is not a plugin but a helper. In the sense it doesn't extend the JS engine -> move it to dependencies/BabylonNativeUtils ?
- 2 files named NativeEngine.h this is dangerous and can lead to hours of cmake debugging because of order of include paths
- can the jsruntime be a plugin? it looks logical to me to have a ChakraRuntime/v8runtime/... and I'd like to do tests with other JS engine more optimized for binary size 
- Scriptloader has a static depedency on the network utils. So, I'm a bit lost here. I believed the Scriptloader would depend on a plugin that would correspond to my platform (asset loader + curl on Android, UIKit on mac/ios, ...). Imagine a customer that doesn't want http requests but the scriptloader.
- at first I was a bit scared by things like this:
```
        runtime->Dispatch([width, height](Napi::Env env)
        {
            auto& window = Babylon::NativeWindow::GetFromJavaScript(env);
            window.Resize(static_cast<size_t>(width), static_cast<size_t>(height));
        });
```
basically, we just ask here to do a resize that, functionaly speaking is that:
```
runtime->resize(width, height);
```
Same for plugin registration. As this will only be maintained by us and experienced c++ programers (most users will only use prebuilt packages) I don't think it's an issue. But the syntax looks like over engineering IMHO. The dispatch is a needed mecanism under the hood but exposing it to users who want to integrate it in their app is overkill.

Core/JsRuntime/Source/JsRuntime.cpp
bghgary: Ahh, didn't realize this is protected. Then it doesn't matter that much to me.

Apps/Playground/Android/app/CMakeLists.txt
```diff
@@ -29,22 +29,24 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/src/")
 
 set(BABYLON_NATIVE_TESTAPP_DIR "${CMAKE_CURRENT_LIST_DIR}/../..")
-add_subdirectory(${BABYLON_NATIVE_TESTAPP_DIR}/../../Library/ ${BABYLON_NATIVE_TESTAPP_DIR}/../../build/Android_${CMAKE_ANDROID_ARCH_ABI}/)
+add_subdirectory(${BABYLON_NATIVE_TESTAPP_DIR}/../../Dependencies/)
+add_subdirectory(${BABYLON_NATIVE_TESTAPP_DIR}/../../Core/)
+add_subdirectory(${BABYLON_NATIVE_TESTAPP_DIR}/../../Plugins/)
+add_subdirectory(${BABYLON_NATIVE_TESTAPP_DIR}/../../build/Android_${CMAKE_ANDROID_ARCH_ABI}/)
```
syntheticmagus: This might not be necessary anymore.

syntheticmagus: Actually binary dir. https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -89,17 +107,18 @@ Java_BabylonNative_Wrapper_surfaceChanged(JNIEnv* env, jobject obj, jint width,
     if (runtime)
     {
         ANativeWindow *window = ANativeWindow_fromSurface(env, surface);
-        runtime->UpdateWindow(width, height, window);
+        // TODO: Add UpdateWindow for Android.
+        // runtime->UpdateWindow(width, height, window);
```
syntheticmagus: Don't forget to add this functionality back!

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -110,7 +129,12 @@ Java_BabylonNative_Wrapper_eval(JNIEnv* env, jobject obj, jstring source, jstrin
     {
         jboolean iscopy;
         std::string url = env->GetStringUTFChars(sourceURL, &iscopy);
-        runtime->Eval(env->GetStringUTFChars(source, &iscopy), url);
+        std::string src = env->GetStringUTFChars(source, &iscopy);
+        // TODO: Put this on ScriptLoader?
```
syntheticmagus: Go ahead and do as part of this PR?

Plugins/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -1,15 +1,31 @@
...
 #include <Babylon/JsRuntime.h>
 
 namespace Babylon
 {
+    namespace
+    {
+        constexpr auto JS_ROOT_URL_NAME = "RootUrl";
+    }
+
+    void InitializeXMLHttpRequest(JsRuntime& runtime, const char* rootUrl)
```
bghgary:
```suggestion
    void InitializeXMLHttpRequest(JsRuntime& runtime, std::string rootUrl)
```

PR #160, opened 2020-02-26
README updates for Babylon Native public preview

syntheticmagus:


README.md
```diff
@@ -2,90 +2,276 @@
...
+The following major features are currently currently in unsupported preview. Note 
+that this list is not exhaustive.
+
+- Developing Babylon Native on the following platforms: Linux.
+- Building and running the libraries and demo apps for the following platforms: 
+Android.
+- Mixed reality powered by OpenXR on the following platforms: Win32, UWP.
+
+The following major features are not yet supported or implemented, even as previews, 
+but are expected to be supported in the future. Note that this list is not exhaustive.
```
bghgary: Maybe also add:

* Debugging JavaScript with V8

README.md
```diff
@@ -2,90 +2,276 @@
...
-- Open the generated solution `BabylonNative.sln`.
-  ```
-  C:\BabylonNative\Build>start BabylonNative.sln
-  ```
-- Once the solution opens, right click the Playground project, click "Set as StartUp Project" and click the play button at the top to run the test app which should render a box
-
-### Android / iOS / MacOS
-
-*Planned but not yet implemented*
+Welcome to Babylon! This quick overview will help you get started developing in the
```
bghgary: > Welcome to Babylon!

Seems like this should either be right at the top or not present.

README.md
```diff
@@ -2,90 +2,276 @@
...
+```
+mkdir Build
+cd Build
+```
+
+### **Building on Windows 10, Targeting Windows Desktop (Win32)**
+
+**Required Tools:** [Visual Studio 2019](https://visualstudio.microsoft.com/vs/) with 
+C++ development tools, [Python 3.0](https://www.python.org/) or newer (required by 
+dependencies)
```
bghgary: Have we confirmed Python is actually necessary?

README.md
```diff
@@ -2,90 +2,276 @@
...
+build directory. Build directory location is up to you, but we recommend just creating 
+a `Build` directory within your clone of the Babylon Native repository (Babylon 
+Native's `.gitignore` file is already set up to ignore this `Build` directory).
+
+```
+mkdir Build
+cd Build
+```
+
+### **Building on Windows 10, Targeting Windows Desktop (Win32)**
```
bghgary: What about architectures? ARM64, etc.?

README.md
syntheticmagus: There are numerous Python scripts in the bgfx submodule, but I haven't actually confirmed what happens if you try to build without Python. I could check that now, or we could do it later; I've created an issue (#161) to track it either way. Thoughts?

README.md
syntheticmagus: Oops, forgot about those. Included them now.

README.md
bghgary: Not that big a deal either way.

README.md
```diff
@@ -2,90 +2,300 @@
...
+```
+
+By default, the "Playground" demo app should be set as the Visual Studio start-up
+project. Build and run this app by pressing the green "Play" button or by pressing
+`F5` on your keyboard.
+
+### **Building on macOS, Targeting macOS**
+
+**Required Tools:** [Xcode 11](https://developer.apple.com/xcode/) or newer, 
+[Python 3.0](https://www.python.org/) or newer (required by dependencies)
```
CedricGuillemet: Python is bundled by default on MacOS. For cmake, use brew.

bghgary: Can you suggest a change? For now, I will merge and we can update it after.

PR #163, opened 2020-03-02
update to get marble tower running on mac and ios

CedricGuillemet:
- improved JSC (properties, exceptions that get display in Safari, arraybuffer)
- Correct binding of textures for metal
- fixed view size in ios viewcontroller

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -549,7 +549,12 @@ namespace Babylon
                 const spirv_cross::ShaderResources resources = compiler.get_shader_resources();
                 assert(resources.uniform_buffers.size() == 1);
                 const spirv_cross::Resource& uniformBuffer = resources.uniform_buffers[0];
-                const spirv_cross::SmallVector<spirv_cross::Resource>& samplers = resources.separate_samplers;
+#if (BGFX_CONFIG_RENDERER_METAL)
+                // with metal, we bind images and not samplers
+                const spirv_cross::SmallVector<spirv_cross::Resource>& samplers = resources.separate_images;
+#else
+                 const spirv_cross::SmallVector<spirv_cross::Resource>& samplers = resources.separate_samplers;
+#endif
                 size_t numUniforms = compiler.get_type(uniformBuffer.base_type_id).member_types.size() + samplers.size();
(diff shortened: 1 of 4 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -549,7 +549,12 @@ namespace Babylon
                 const spirv_cross::ShaderResources resources = compiler.get_shader_resources();
                 assert(resources.uniform_buffers.size() == 1);
                 const spirv_cross::Resource& uniformBuffer = resources.uniform_buffers[0];
-                const spirv_cross::SmallVector<spirv_cross::Resource>& samplers = resources.separate_samplers;
+#if (BGFX_CONFIG_RENDERER_METAL)
+                // with metal, we bind images and not samplers
+                const spirv_cross::SmallVector<spirv_cross::Resource>& samplers = resources.separate_images;
+#else
+                 const spirv_cross::SmallVector<spirv_cross::Resource>& samplers = resources.separate_samplers;
```
syntheticmagus: nit: Spacing.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -549,7 +549,12 @@ namespace Babylon
                 const spirv_cross::ShaderResources resources = compiler.get_shader_resources();
                 assert(resources.uniform_buffers.size() == 1);
                 const spirv_cross::Resource& uniformBuffer = resources.uniform_buffers[0];
-                const spirv_cross::SmallVector<spirv_cross::Resource>& samplers = resources.separate_samplers;
+#if (BGFX_CONFIG_RENDERER_METAL)
```
syntheticmagus: Is this define guaranteed to be available here? (i.e., is this an implementation detail of bgfx.cmake, or part of its public contract?) Could we instead consider embedding this behavior into a helper function inside `ShaderCompiler.h`, with implementations swapped out per platform by our existing mechanism?

CedricGuillemet: tbh, I find this change pretty ugly but it works and I didn't find a better solution yet. I hope to get it removed once work on validation tests begins. If no solution is found then it will be hidden in the platform shader cpp.

PR #166, opened 2020-03-03
Added link-time CMake behavior

syntheticmagus:
Added mechanism to allow us to execute a CMake "lambda" when linking to a library. This will allow us to follow a consistent pattern for libraries that require special behaviors when being linked, like N-API when it uses V8. Also made N-API with V8 use this mechanism.


Dependencies/napi/CMakeLists.txt
```diff
@@ -90,10 +50,51 @@ endif()
 
 target_include_directories(napi PUBLIC "include")
 
-if(NAPI_JAVASCRIPT_ENGINE STREQUAL "Chakra")
-    target_link_libraries(napi PUBLIC "chakrart.lib")
-elseif(NAPI_JAVASCRIPT_ENGINE STREQUAL "V8")
-    target_link_libraries(napi PUBLIC v8)
-endif()
+# TODO: Make this block less weird.
```
bghgary: what do you mean by this?

syntheticmagus:
```suggestion
```

syntheticmagus: Outdated. This block used to be weirder, but it's now less weird, and make it not weird at all is no longer an ambition of this change.

PR #169, opened 2020-03-05
Apple libcurl replacement

CedricGuillemet:
ios build down from 22min to 3
macos down from 10min to 2


Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -7,10 +7,23 @@
...
 #import <Shared/InputManager.h>
+#import "Babylon/XMLHttpRequestApple.h"
 
 std::unique_ptr<Babylon::AppRuntime> runtime{};
 std::unique_ptr<InputManager::InputBuffer> inputBuffer{};
 
+void XMLHTTPRequestPlugin(Napi::Env env)
+{
+    napi_env napienv = env;
+    JSGlobalContextRef globalContext = *(JSGlobalContextRef*)napienv;
```
bghgary: this seems a bit dangerous. if the napi_env struct changes, this will be wrong. i don't understand this syntax, what is globalContext used for here?

Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -7,10 +7,23 @@
 #import <Babylon/ScriptLoader.h>
 #import <Babylon/XMLHttpRequest.h>
 #import <Shared/InputManager.h>
+#import "Babylon/XMLHttpRequestApple.h"
 
 std::unique_ptr<Babylon::AppRuntime> runtime{};
 std::unique_ptr<InputManager::InputBuffer> inputBuffer{};
 
+void XMLHTTPRequestPlugin(Napi::Env env)
```
bghgary: The function name is not consistent with other plugins, though it's not completely consistent within the other plugins either. Is it possible to put this in a namespace so that you can do `Babylon::InitializeXMLHttpRequest`?

Dependencies/BabylonNativeUtils/Source/NetworkUtils.mm
```diff
@@ -0,0 +1,54 @@
+#include "NetworkUtils.h"
+#import <Foundation/Foundation.h>
+
+namespace Babylon
+{
+    std::string GetAbsoluteUrl(const std::string& url, const std::string& rootUrl)
+    {
+        NSString *urlStr = [NSString stringWithCString:url.c_str() encoding:[NSString defaultCStringEncoding]];
```
bghgary: nit all over
```suggestion
        NSString* urlStr = [NSString stringWithCString:url.c_str() encoding:[NSString defaultCStringEncoding]];
```

Dependencies/BabylonNativeUtils/Source/NetworkUtils.mm
```diff
@@ -0,0 +1,54 @@
...
+        __block arcana::task_completion_source<DataT, std::exception_ptr> taskCompletionSource{};
+        NSString *urlStr = [NSString stringWithCString:url.c_str() encoding:[NSString defaultCStringEncoding]];
+        NSURLRequest *request = [NSURLRequest requestWithURL:[NSURL URLWithString:urlStr]];
+        NSURLSession *session = [NSURLSession sharedSession];
+        NSURLSessionDataTask *task = [session dataTaskWithRequest:request completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+            if (!error) {
+                DataT dataT{};
+                dataT.resize(data.length);
+                [data getBytes:dataT.data()
+                        length:data.length];
```
bghgary: Does this handle text conversion correctly?

CedricGuillemet: in XMLHttpRequest, the conversion UTF8-BOM to UTF8 is handled correctly thanks to the NSString conversion. In NetworkUtils, there is no conversion. It read raw bytes from the network.

Apps/Playground/iOS/LibNativeBridge.mm
CedricGuillemet: Yes it is. I can try with a function that returns the JSGlobalContextRef from Napi::Env for Apple platform. The JS context is used for JSC object definition.

Dependencies/BabylonNativeUtils/Source/NetworkUtils.mm
CedricGuillemet: That's the standard for ObjC.

Dependencies/BabylonNativeUtils/Source/NetworkUtils.mm
bghgary: Isn't that going to cause issues?

Plugins/XMLHttpRequest/Include/Babylon/XMLHttpRequestApple.h
```diff
@@ -0,0 +1,50 @@
+// XMLHttpRequest from https://github.com/Lukas-Stuehrk/XMLHTTPRequest
+// with modifications (addEventListener, arraybuffer)
+// MIT License
```
bghgary: NOTICE at the root of the repo also needs a blurb

Dependencies/BabylonNativeUtils/Source/NetworkUtils.mm
CedricGuillemet: Yes. If you try to load and evaluate an UTF8-BOM .js you'll get a parsing error. On all the platforms we support.

bghgary: Can we not do the same thing as what XMLHttpRequest is doing?

CedricGuillemet: It would be easier if we can share code between webrequest/xmlhttprequest and NetworkUtils. Life is more difficult with Android. As we have to call Java from jni, the network code has to be part of the testapp/integration aar. So this will revisited.

Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -7,10 +7,22 @@
...
+void InitializeXMLHttpRequest(Napi::Env env)
+{
+    JSGlobalContextRef globalContext = Napi::GetJavaScriptCoreGlobalContext(env);
+    [[XMLHttpRequest new] extend:globalContext :^(CompletionHandlerFunction completetionHandlerFunction) {
+        runtime->Dispatch([completetionHandlerFunction](Napi::Env env) {
+            completetionHandlerFunction();
+        });
+     }
+     ];
+}
```
bghgary: Why are these functions not part of the plugin? Shouldn't the initialization of plugin be in the plugin?

CedricGuillemet: I need runtime. I could pass it as a lambda but it's already intracted enough

bghgary: Passing in runtime is expected for many plugins. Why is this one different?

Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -56,7 +59,10 @@ - (void)init:(void*)CALayerPtr width:(int)inWidth height:(int)inHeight
     Babylon::InitializeNativeEngine(*runtime, windowPtr, width, height);
     
     // Initialize XMLHttpRequest plugin.
-    Babylon::InitializeXMLHttpRequest(*runtime, runtime->RootUrl());
+    runtime->Dispatch([](Napi::Env env)
+    {
+        InitializeXMLHttpRequest(env);
+    });
```
syntheticmagus: I think we want to move away from this pattern and more toward just passing the runtime itself to the plugin initializers, which can then do their own dispatching. Creating a new pattern for this is really a part of the post-plugin refactor cleanup work, which hasn't been done yet, but that's the direction I think it'll go.

Dependencies/BabylonNativeUtils/CMakeLists.txt
```diff
@@ -1,13 +1,25 @@
-set(SOURCES
-    "Include/Babylon/NetworkUtils.h"
-    "Include/Babylon/TicketedCollection.h"
-    "Source/NetworkUtils.cpp")
+if(APPLE)
```
syntheticmagus: nit: I'm a little iffy about the platform-specific guards we have in here. I know they're inevitable in certain cases, but we really ended up with a ton of them before the plugin refactor (many of which we still have), and for me it felt pretty messy. Can we look for alternatives for how to minimize these, at least within scenarios where we have duplicated code, as below?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -616,7 +616,7 @@ namespace Babylon
                 const spirv_cross::ShaderResources resources = compiler.get_shader_resources();
                 assert(resources.uniform_buffers.size() == 1);
                 const spirv_cross::Resource& uniformBuffer = resources.uniform_buffers[0];
-#if (BGFX_CONFIG_RENDERER_METAL)
+#if __APPLE__
```
syntheticmagus: Why this change? The comment below indicates that the guarded behavior is specific to Metal, not Apple; is that incorrect?

Plugins/XMLHttpRequest/CMakeLists.txt
```diff
@@ -1,8 +1,21 @@
-set(SOURCES 
-    "Include/Babylon/XMLHttpRequest.h"
-    "Source/XMLHttpRequest.cpp"
-    "Source/XMLHttpRequest.h"
-    "Source/XMLHttpRequest${BABYLON_NATIVE_PLATFORM}.cpp")
+if(APPLE)
```
syntheticmagus: If we're splitting the implementation down here, too, could the networking code just live here?

CedricGuillemet: You mean remove NetworkUtils from BabylonNativeUtils and put it in XMLHttpRequest project?

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: BGFX_CONFIG_RENDERER_METAL not defined. __APPLE__ is the only define that I've found to make the guard.

Dependencies/BabylonNativeUtils/CMakeLists.txt
CedricGuillemet: best alternative is to have specific plugin/project per platform IMHO. instead of guards we would have more cmakelists.

Apps/Playground/iOS/LibNativeBridge.mm
CedricGuillemet: I've spent 2-3hours trying to do the init on the htttprequest side without success. I get a crash I can't explain in the lambda cascaded calls. This makes no sense as the code should just be moved from an .mm to another.

bghgary: Can you put this in a branch? I can try it out also.

CedricGuillemet: I pushed into the same branch. I did the change for iOS.

Plugins/XMLHttpRequest/CMakeLists.txt
syntheticmagus: Kind of. I was more thinking about just having the network code itself be here as a non-exposed thing, similar to how UWP does it. I'd forgotten that the script loader needs network stuff too, though, so it probably wouldn't work now that I think of it.

Dependencies/napi/include/napi/env.h
```diff
@@ -2,11 +2,19 @@
...
     template<typename ...Ts> Napi::Env Attach(Ts... args);
 
     void Detach(Napi::Env);
 
     Napi::Value Eval(Napi::Env env, const char* source, const char* sourceUrl);
-}
\ No newline at end of file
+
+#ifdef __APPLE__
+    JSGlobalContextRef GetJavaScriptCoreGlobalContext(Napi::Env env);
```
syntheticmagus: This seems really weird. You're using Objective-C, right? Can you get the context [this way](https://developer.apple.com/documentation/javascriptcore/jscontext/1451545-currentcontext) instead?

CedricGuillemet: [JSContext currentContext] returns nil. At that point, we are not executing JS. That method would return a valid context inside a napi function. Not at init time.

syntheticmagus: Then can't we just call it from inside JS? Since we have the `Env` and should already be on the right thread, just create a JS function that calls straight back into native, then call that. That'll get JSC on the stack, which should then cause that function to return a legitimate value. Would that work?

syntheticmagus: I really don't think we can put platform- and engine-specific concepts into `env.h` itself.

CedricGuillemet: > Then can't we just call it from inside JS?
Do you mean registering the JSC class from a napi function that's being called?

bghgary: There doesn't appear to be an easy way to get the JS context using `currentContext`. The next best thing would be to expose cleanly from JsRuntime. Exposing it from Napi::Env is probably not a good idea. Let's chat about it offline.

PR #172, opened 2020-03-11
Remove setting ARCHIVE_OUTPUT_DIRECTORY with an invalid ${ARCH} property

bghgary:


bghgary:
UWP is certainly working for me.

PR #176, opened 2020-03-13
Fixes for XMLHttpRequest on UWP

bghgary:
The XMLHttpRequest class was broken for UWP when accessing local files. This fixes it.


bghgary:
CI is not running for some reason. Closing to try again.

PR #179, opened 2020-03-13
Fixes for XMLHttpRequest on UWP

bghgary:
The XMLHttpRequest class was broken for UWP when accessing local files. This fixes it.

### DIFF
--- a/Plugins/XMLHttpRequest/Source/XMLHttpRequest.h
+++ b/Plugins/XMLHttpRequest/Source/XMLHttpRequest.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <Babylon/JsRuntime.h>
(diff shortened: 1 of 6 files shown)


Plugins/XMLHttpRequest/Source/XMLHttpRequest.h
```diff
@@ -121,7 +121,26 @@ namespace Babylon
...
+            {
+                m_runtime.Dispatch([callable{std::forward<CallableT>(callable)}](Napi::Env){
+                    callable();
+                });
+            }
+
+        private:
+            JsRuntime& m_runtime;
+        } m_runtimeScheduler;
+
```
syntheticmagus: This is super neat. Should we just make that operator part of `JsRuntime` so we can use it this way by default?

PR #183, opened 2020-03-19
Multithreaded texture loading

bghgary:
See https://github.com/BabylonJS/Babylon.js/pull/7862 for the Babylon.js side.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -235,6 +220,25 @@ namespace Babylon
...
+            uint32_t rowPitch = image->m_size / image->m_height;
+
+            std::vector<uint8_t> buffer(rowPitch);
+
+            for (size_t row = 0; row < rowCount / 2; row++)
+            {
+                auto frontPtr = bytes + (row * rowPitch);
+                auto backPtr = bytes + ((rowCount - row - 1) * rowPitch);
+
+                std::memcpy(buffer.data(), frontPtr, rowPitch);
```
syntheticmagus: nit: Not really a problem, but I think we could reduce the number of copy operations done here by a third by doing the entire "copy to buffer" operation at once, basically just grabbing the top half of the image and putting that in a buffer in a single operation, then reversing the bottom into the top, then reversing the buffer into the bottom. Created #187 to track this, as it's probably not worth doing in this PR.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -258,28 +262,68 @@ namespace Babylon
...
+            return Napi::Reference<T>{env, ref};
+        }
 
-            textureData->Texture = bgfx::createTexture2D(image->m_width, image->m_height, generateMips, 1, Cast(image->m_format), 0, mem);
-            textureData->Width = image->m_width;
-            textureData->Height = image->m_height;
+        inline Napi::FunctionReference AttachFunctionRef(napi_env env, napi_ref ref)
+        {
+            return Napi::FunctionReference{env, ref};
         }
```
syntheticmagus: This stuff seems kind of funky. This is for getting around the non-copyable N-API function thing, right? Can we create a type that encapsulates this behavior and hides it from us?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -432,6 +478,11 @@ namespace Babylon
         return m_frameBufferManager;
     }
 
+    void NativeEngine::Dispose(const Napi::CallbackInfo& info)
+    {
+        m_cancelSource.cancel();
```
syntheticmagus: Why not in the destructor?

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: Certainly, but then maybe just fix Napi::Reference to allow copying. I went back and forth on this a bunch. Let me try some things

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: I just realized we don't even call Dispose in the Playground app. It probably should be in both. Dispose is more immediate.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: I added a Napi::SharedReference class to do this for now. Maybe we can discuss with NAPI people to see if this works for them. If not, we can still continue to use this in our code.

PR #184, opened 2020-03-19
clean Android CMake

CedricGuillemet:
Some files were compile in napi project (Dependencies/napi/source/js_native_api_v8.cc Dependencies/napi/source/env_v8.cc Dependencies/napi/source/env.cc) and also in JNI. No error at link time as only missing symbols are searched in provided libraries.
Also removed some unused define, removed include dirs used by napi.
Only root subdirectory used instead of 3 for ease of integration.


Apps/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
-add_subdirectory(Playground)
+if(NOT ANDROID)
+    add_subdirectory(Playground)
+endif()
```
bghgary: This is because we don't build Android from CMake?

CedricGuillemet: we don't buidl android playground from cmake. we use gradle.

bghgary: okie!

PR #185, opened 2020-03-21
X11port

CedricGuillemet:
PR for X11 port. Should also work with other kernels. Remaining work needed for GLSL compiler but this PR is enough for getting a compiled binary and test portability features.

- I didn't commit my hacky changes in GLSL compiler
- Needs Clang 8 or 9

image


CedricGuillemet:
@bghgary hosted Azure package repo doesn't have a recent curl (7.58 vs 7.65 in BN repo). curl_url api is introduced in 7.62+. We can keep curl subrepo or recode that function without relying on those api.

Core/AppRuntime/Source/AppRuntimeJavaScriptCore.cpp
```diff
@@ -1,3 +1,4 @@
+#include "WorkQueue.h"
```
bghgary, line 1: Why do you need to include WorkQueue? Also, the first include should be the AppRuntime.

Dependencies/UrlLib/CMakeLists.txt
```diff
@@ -6,6 +6,11 @@ if(ANDROID)
 elseif(APPLE)
     set(ADDITIONAL_SOURCES
         "Source/Apple/UrlRequest.mm")
+elseif(UNIX)
+    set(ADDITIONAL_SOURCES
+        "Source/Unix/UrlRequest.cpp")
+    set(ADDITIONAL_LIBRARIES
+        PUBLIC "curl")
```
bghgary, line 13:
```suggestion
        PRIVATE curl)
```

Plugins/NativeEngine/Source/BgfxCallback.cpp
```diff
@@ -5,6 +5,7 @@
 #include <bx/debug.h>
 #include <stdarg.h>
 #include <bgfx/bgfx.h>
+#include <memory.h> // memcpy
```
bghgary, line 8: This seems like it should be bgfx

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -142,20 +142,20 @@ namespace Babylon
...
             bgfx::setViewFrameBuffer(ViewId, FrameBuffer);
-            ViewClearState.Update();
+            viewClearState.Update();
             bgfx::setViewRect(ViewId, 0, 0, Width, Height);
         }
 
         bgfx::FrameBufferHandle FrameBuffer{bgfx::kInvalidHandle};
         bgfx::ViewId ViewId{};
-        ViewClearState ViewClearState;
+        ViewClearState viewClearState;
```
bghgary, line 158: Maybe rename the type instead of the variable. Otherwise, the variable conventions are mixed.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -54,18 +72,58 @@ namespace UrlLib
...
+
+                auto result = curl_easy_perform(curl);
+                if (result != CURLE_OK)
+                {
+                    throw std::exception();
+                }
+
+                curl_easy_cleanup(curl);
+            }
+        }
```
bghgary, line 117: this is synchronous. we should try to implement this correctly now.

CedricGuillemet: https://github.com/BabylonJS/BabylonNative/issues/232

Apps/CMakeLists.txt
```diff
@@ -2,6 +2,6 @@ if(NOT ANDROID)
     add_subdirectory(Playground)
 endif()
 
-if(WIN32 AND NOT WINDOWS_STORE) # TODO: Chakra only?
+if(WIN32 OR UNIX AND NOT APPLE AND NOT ANDROID AND NOT WINDOWS_STORE) # TODO: Chakra only for win32?
```
syntheticmagus, line 5: nit: Change TODO to, "Default JS engine for platform only?"

Dependencies/napi/CMakeLists.txt
```diff
@@ -44,6 +44,12 @@ if(NAPI_JAVASCRIPT_ENGINE STREQUAL "Chakra")
     target_compile_definitions(napi PUBLIC USE_EDGEMODE_JSRT)
 endif()
 
+if (UNIX AND NOT ANDROID AND NOT APPLE)
+    target_include_directories(napi PUBLIC "/usr/include/webkitgtk-4.0")
+    find_library(JSCORE_LIBRARY JavaScriptCore)
+    target_link_to_dependencies(napi PUBLIC ${JSCORE_LIBRARY})
+endif()
```
syntheticmagus, line 51: Doesn't this break the `javascript_engine` injection mechanism below? I think this should only be done if `javascript_engine` isn't already a target, and it should probably be a third case in the conditional that begins on line 58.

Dependencies/UrlLib/CMakeLists.txt
```diff
@@ -6,6 +6,11 @@ if(ANDROID)
 elseif(APPLE)
     set(ADDITIONAL_SOURCES
         "Source/Apple/UrlRequest.mm")
+elseif(UNIX)
+    set(ADDITIONAL_SOURCES
+        "Source/Unix/UrlRequest.cpp")
+    set(ADDITIONAL_LIBRARIES
+        PRIVATE curl)
```
syntheticmagus, line 13: Installed using `apt`, I guess?

Core/AppRuntime/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ add_library(AppRuntime ${SOURCES})
 target_include_directories(AppRuntime PRIVATE "Include/Babylon")
 target_include_directories(AppRuntime INTERFACE "Include")
 
+if(UNIX AND NOT APPLE AND NOT ANDROID)
+    target_include_directories(AppRuntime INTERFACE "/usr/include/webkitgtk-4.0/")
```
syntheticmagus, line 17: Installed using `apt`?

Dependencies/UrlLib/CMakeLists.txt
syntheticmagus: Yep, just read the README part.

Core/AppRuntime/CMakeLists.txt
syntheticmagus: Yep, stated in README, just saw.

Dependencies/napi/CMakeLists.txt
CedricGuillemet: I moved that block into a new conditional branch. Let  me know if I missed something.

PR #186, opened 2020-03-23
Android remove curl

CedricGuillemet:
removed dep on curl
removed openssl sub repo
use Java for async loading from url
tested with marble tower
build time from 24min to 5min (CI)
arm64 ABI goes down from 16Mb to 11Mb (32% less !!)


Apps/Playground/Android/app/src/main/cpp/BabylonNativeNetwork.cpp
```diff
@@ -0,0 +1,140 @@
...
+    {
+        std::vector<uint8_t> content;
+        if (AndroidLoadAsset(url, content))
+        {
+            contentHandler(content);
+        }
+        else
+        {
+            DispatchLooperFunction<DispatchPayloadDirect>(g_mainThreadLooper, [url, contentHandler = std::move(contentHandler)](){
+                JNIEnv * env = AcquireEnvFromJVM(g_jvm);
```
ryantrem: Which thread would this be getting called from? Is it the JS thread? If so, this part of the change will overlap with a more generalized solution that I believe @bghgary was going to implement for getting the JVM attached to the JS thread as needed (we need to do this for the ARCore integration).

bghgary: That's right. @CedricGuillemet put a comment on DipatchLooperFunction letting me know the code to put in arcana.

ryantrem: I wasn't talking about the looper scheduler, I was talking about attaching the JVM to the JS thread.

bghgary:
CI not running. Closing to retry.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeNetwork.cpp
bghgary: Oh, right. Yeah, we should use the new code that in the arcore branch.

CedricGuillemet:
Replaced by @bghgary 's work on xmlhttprequest. Closing.

PR #188, opened 2020-03-24
Misc Android clean up

bghgary:
- Clean up .gitignore files
- Update Android app name
- Remove unused Java imports
- Auto format Java files
- Fix various compiler warnings in NativeEngine


Apps/Playground/Android/app/src/main/AndroidManifest.xml
```diff
@@ -33,18 +32,21 @@
       <intent-filter >
         <action android:name="android.intent.action.VIEW" />
         <category android:name="android.intent.category.DEFAULT" />
-        <data android:scheme="file" />
+        <data android:scheme="file"
+            android:host="*" />
```
ryantrem: What is this?

bghgary: it's a filter for allowed paths which according to the schema is required

PR #190, opened 2020-03-25
Android remove curl

bghgary:
removed dep on curl
removed openssl sub repo
use Java for async loading from url
tested with marble tower
build time from 24min to 5min (CI)
arm64 ABI goes down from 16Mb to 11Mb (32% less !!)


bghgary:
Hmm, CI is still not running.

PR #192, opened 2020-03-27
Fix napi jsc string handling

bghgary:


Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -206,14 +214,13 @@ napi_status napi_get_value_string_utf8(napi_env env,
                      size_t bufsize,
                      size_t* result) {
   JSStringRef stringRef = JSValueToStringCopy(env->m_globalContext, reinterpret_cast<JSValueRef>(value), nullptr);
-  if (!buf && result) {
-    // get only string length
-    *result = JSStringGetLength(stringRef);
```
bghgary: `JSStringGetLength` gets the unicode length, so this is wrong.

Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -206,14 +214,13 @@ napi_status napi_get_value_string_utf8(napi_env env,
...
+  if (!buf) {
+    *result = GetUTF8String(stringRef).size();
     return napi_ok;
   }
-  size_t length = JSStringGetUTF8CString(stringRef, buf, bufsize);
+  size_t size = JSStringGetUTF8CString(stringRef, buf, bufsize);
```
bghgary: This is still wrong. It's off by 1 since JSStringGetUTF8CString returns the size including the null terminator whereas the API expects result to be without the null terminator.

PR #193, opened 2020-03-30
Android JS Engine selection

CedricGuillemet:
npm installation of JS engine
V8/JSC switch in Gradle
android pipelines for v8 and JSC
build instructions update in an upcoming PR

### DIFF
--- a/Apps/Playground/Android/app/CMakeLists.txt
+++ b/Apps/Playground/Android/app/CMakeLists.txt
@@ -24,6 +24,16 @@ set(BABYLON_NATIVE_PLATFORM "Android")
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
 
+if (${ANDROID_JSENGINE_LIBNAME} STREQUAL "jsc")
+    set(NAPI_JAVASCRIPT_ENGINE "JavaScriptCore")
+else()
+    set(NAPI_JAVASCRIPT_ENGINE "V8")
+endif()
+
+add_library(javascript_engine SHARED IMPORTED GLOBAL)
+set_target_properties(javascript_engine PROPERTIES IMPORTED_LOCATION "${ANDROID_JSENGINE_LIBPATH}/${ANDROID_ABI}/lib${ANDROID_JSENGINE_LIBNAME}.so")
+target_include_directories(javascript_engine INTERFACE "${ANDROID_JSENGINE_INCPATH}")
+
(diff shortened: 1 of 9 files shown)


azure-pipelines.yml
```diff
@@ -195,24 +195,43 @@ jobs:
...
-      mkdir Dependencies/v8/include
-      mkdir Apps/Playground/Android/app/src/main/jniLibs
-      cp -r node_modules/v8-android/dist/include Dependencies/v8
-      cp -r arm64-v8a Apps/Playground/Android/app/src/main/jniLibs
-      cp -r armeabi-v7a Apps/Playground/Android/app/src/main/jniLibs
-      cp -r x86 Apps/Playground/Android/app/src/main/jniLibs
-      cp -r x86_64 Apps/Playground/Android/app/src/main/jniLibs
-    displayName: 'Install V8 npm'    
+      cd Apps/Playground/Android
+      npm install
```
syntheticmagus: nit: Is it possible/advisable to get Gradle to do this part, too?

Apps/Playground/Android/package-lock.json
```diff
@@ -0,0 +1,18 @@
+{
```
bghgary: update .gitignore to ignore node_modules?

Apps/Playground/Android/app/CMakeLists.txt
```diff
@@ -24,6 +24,16 @@ set(BABYLON_NATIVE_PLATFORM "Android")
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
 
+if (${ANDROID_JSENGINE_LIBNAME} STREQUAL "jsc")
+    set(NAPI_JAVASCRIPT_ENGINE "JavaScriptCore" CACHE STRING "JavaScript engine for N-API.")
+else()
+    set(NAPI_JAVASCRIPT_ENGINE "V8" CACHE STRING "JavaScript engine for N-API.")
+endif()
```
bghgary: nit: might be better to have an else case that throws an error if the libname is not jsc or v8android.

Apps/Playground/Android/package-lock.json
bghgary: also remove Dependenceis/v8 from root .gitignore

azure-pipelines.yml
CedricGuillemet: I tried without success to do it with gradle using shell command. No idea why. I'm adding a issue: https://github.com/BabylonJS/BabylonNative/issues/197

PR #196, opened 2020-03-31
Plugin cleanup and standardization

syntheticmagus:


ryantrem:
I'm a bit confused by the naming. It seems like some things (but not all things) that are a polyfill have the Polyfill suffix, and some things (but not all things) that are a plugin have the Plugin suffix. What is the distinction?

syntheticmagus:
> some things (but not all things) that are a polyfill have the Polyfill suffix

You're referring to `XMLHttpRequest`? There's quite a bit of active ongoing work in and around `XMLHttpRequest`, and it's very likely to change a whole lot in the near future even independent of anything I'm trying to explore with this PR. Because of that, and because this PR is mainly about understanding and establishing some conventions, I decided not to make changing that class part of this PR. Instead, we'll just take whatever consensus we reach here and roll that into the upcoming changes to `XMLHttpRequest`, or whatever replaces it. That's the thinking, anyway.

> and some things (but not all things) that are a plugin have the Plugin suffix. What is the distinction?

By this you mean `ScriptLoader`? Gary and I have talked about this a bit, and we're actually not really sure `ScriptLoader` is actually a plugin since it doesn't actually add any new JavaScript functionality on its own. It's more like a utility class. With that in mind, I don't really know where it should end up (it's currently in the Plugins folder because it depends on similar things, but that may change with the upcoming modifications to `XMLHttpRequest` and our networking code), but I think it's likely it will ultimately get moved out of the Plugins folder altogether.

PR #198, opened 2020-03-31
Various fixes to support react native integration

ryantrem:
This change addresses a few issues when integrating Babylon Native with React Native:
- Adds static graphics deinitialization (which is symmetric with the static graphics initialization). I need this to properly shut down bgfx when react native reloads the JS engine which triggers reinitializing Babylon Native.
- NativeWindow Conditionally polyfills global functions, since in the case of react native some of these (setTimeout) are already provided.


ryantrem:
I'm guessing with this change I would need to add an explicit call to `DeinitializeGraphics` somewhere in the Playground apps, but I'm not sure where yet.

PR #201, opened 2020-04-01
Adding calls to DeinitializeGraphics to Playground apps

ryantrem:
I tried to add these calls in the right places, but it also seems like the existing deinitialization code was only partially done, so I tried to flesh it out where it seemed like it was missing as part of this change as well. I tested on Android, MacOS, Win32, and UWP. I haven't been able to get the iOS playground app to build or run yet.

### DIFF
--- a/Apps/Playground/UWP/App.cpp
+++ b/Apps/Playground/UWP/App.cpp
@@ -162,9 +162,6 @@ concurrency::task<void> App::RestartRuntimeAsync(Windows::Foundation::Rect bound
         m_runtime = std::make_unique<Babylon::AppRuntime>(std::move(rootUrl));
     }
 
-    // Ensure this is properly uninitialized since it depends on state of the runtime.
-    m_inputBuffer.reset();
-
     // Create the console plugin.
     m_runtime->Dispatch([](Napi::Env env)
     {
--- a/Apps/Playground/UWP/App.cpp
+++ b/Apps/Playground/UWP/App.cpp
@@ -112,6 +112,9 @@ void App::Run()
 // class is torn down while the app is in the foreground.
 void App::Uninitialize()
 {
+    m_inputBuffer.reset();
+    m_runtime.reset();
+    Babylon::DeinitializeGraphics();
 }
 
 // Application lifecycle event handlers.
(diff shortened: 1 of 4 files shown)


Apps/Playground/UWP/App.cpp
```diff
@@ -135,8 +138,16 @@ void App::OnActivated(CoreApplicationView^ applicationView, IActivatedEventArgs^
...
+    if (m_inputBuffer)
+    {
+        m_inputBuffer.reset();
+    }
+
+    if (m_runtime)
+    {
+        m_runtime.reset();
+        Babylon::DeinitializeGraphics();
+    }
```
bghgary, line 150: Can you not just call Uninitialize here?

Apps/Playground/UWP/App.cpp
```diff
@@ -112,9 +112,16 @@ void App::Run()
 // class is torn down while the app is in the foreground.
 void App::Uninitialize()
 {
-    m_inputBuffer.reset();
-    m_runtime.reset();
-    Babylon::DeinitializeGraphics();
+    if (m_inputBuffer)
```
bghgary, line 115: why do you need these guards?

Apps/Playground/UWP/App.cpp
```diff
@@ -138,16 +145,7 @@ void App::OnActivated(CoreApplicationView^ applicationView, IActivatedEventArgs^
...
-    {
-        m_inputBuffer.reset();
-    }
-
-    if (m_runtime)
-    {
-        m_runtime.reset();
-        Babylon::DeinitializeGraphics();
-    }
+    Uninitialize();
```
bghgary, line 148: There is some code below that isn't necessary anymore?

```
    // Ensure this is properly uninitialized since it depends on state of the runtime.
    m_inputBuffer.reset();
```

Apps/Playground/UWP/App.cpp
ryantrem: If I call `Babylon::DeinitializeGraphics` when `Babylon::InitializeGraphics` has not been called, bgfx dies. So I'm just using `m_runtime` as a flag to know whether to do this since as the code currently stands they are always initialized/deinitialized together.

bghgary: ok for now. i suspect we will change this behavior soon anyways.

PR #203, opened 2020-04-03
Initial ARCore integration

ryantrem:
This change adds initial support for XR on Android via ARCore.
- Moves XR.cpp (and some other associated files) into an OpenXR specific directory.
- Adds a new XR.cpp that is backed by ARCore (instead of OpenXR).
- Adds ARCore SDK as a submodule, and updates to cmake and gradle files to consume it.
- Minor updates to Playground app to request camera permissions (this is harder to do inside the XR implementation, so I think making this a requirement of the caller for now is ok).
- Adds a Platform class for Android with some platform specific stuff, particularly for getting a JNIEnv attached to the JS thread (@bghgary added this).

This all provides initial basic functionality, but there are a bunch of things that still need to be done:
- Handling surface resizing/screen rotation.
- Handling suspend resume.
- Handling devices that don't support ARCore.
- Checking for ARCore package installation on the device (this might be another thing that the consuming app is required to do, not sure yet).
- Support for plane detection, hit testing, and anchors (needs to be plumbed through more layers because this is not supported in the pre-existing OpenXR implementation of XR.h).


Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -31,40 +33,53 @@ extern "C" {
...
+        g_runtime = std::make_unique<Babylon::AppRuntime>("");
 
-        ANativeWindow *window = ANativeWindow_fromSurface(env, surface);
+        JavaVM* javaVM{};
+        if (env->GetJavaVM(&javaVM) != JNI_OK)
+        {
+            throw std::runtime_error("Failed to get Java VM");
+        }
+
+        // TODO: This should be cleaned up via env->DeleteGlobalRef
```
bghgary, line 75: should we fix this before merge?

Apps/Playground/Android/build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.2'
+        classpath 'com.android.tools.build:gradle:3.6.1'
```
bghgary, line 10: This causes the build to fail, so don't merge this part.

Apps/Playground/Android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Nov 27 14:44:53 CET 2019
+#Sat Mar 21 13:59:40 PDT 2020
```
bghgary, line 1: Same goes for this file. Don't merge, it will cause CI to fail.

Dependencies/CMakeLists.txt
```diff
@@ -11,6 +11,10 @@ if(WINDOWS_STORE)
     set_property(TARGET arcana PROPERTY FOLDER Dependencies)
 endif()
 
+# -------------------------------- GSL -----------------------------------
+# Dependencies: none
+add_subdirectory(arcana.cpp/Source/Submodules/GSL)
```
bghgary, line 16: why do you need this?

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
ryantrem: My vote is no. I want to stop having to merge cmakelists.txt files :)

Dependencies/CMakeLists.txt
ryantrem: I'm using gsl::finally in xr.cpp to revert OpenGL state changes I make since we are taking over the gl context every frame (sharing it with Babylon rendering). Ideally we would use our own context that wraps the context used by Babylon rendering, but I never was able to get this to work.

bghgary: Okay. I'm a bit concerned about reaching into a dependency's dependency. Not sure what the implications are.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
bghgary: Okay, maybe we can file an issue on it.

Apps/Playground/Android/app/build.gradle
```diff
@@ -49,25 +53,64 @@ android {
...
     }
+
     externalNativeBuild {
         cmake {
             version '3.13.2+'
             path 'CMakeLists.txt'
         }
     }
+
+    ndkVersion "21.0.6113669"
```
bghgary, line 82: This was from me. Need to remove this line also.

bghgary, line 82:
```suggestion
```

PR #208, opened 2020-04-08
Validation tests

CedricGuillemet:
Here is a batch of 4tests that work. Run the ValidationTests project and your error code should be 0. 
- Resulting images are saved (toggled by a JS variable) to help debugging
- As before, if too many pixels are different, an image with highlighted difference is saved in working directory
- I've put some comments for future work. I guess I should open an issue for validation tests progress.
- Image comparison methods moved from engine to testutils.
- Some tests (like "GLTF Mesh Primitive Mode (0)") that were running last time I worked on validation are now failing. There is some kind of regression here.
- The tests run quickly: around 1second per test. 4-5 seconds for this batch. We can hope to get all tests running with under 4minutes.
- If you want to test a scene, add it to config.json. Once every test is ok, we can rely on babylonjs config and remove the file from the repo.
- I've left for now hacks in validation_tests.js (like ```engine.getInputElement```). This will be removed with a future PR on nativeengine.ts


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1061,6 +1014,8 @@ namespace Babylon
         arcana::make_task(arcana::threadpool_scheduler, m_cancelSource,
             [this, dataSpan, generateMips, invertY]() {
                 bimg::ImageContainer* image = bimg::imageParse(&m_allocator, dataSpan.data(), dataSpan.size());
+                // todo: bimg::imageParse will return nullptr when trying to load a texture with an url that is not a valid texture
+                // Like a 404 html page.
```
bghgary, line 1018: that should be okay. it should be caught by the error handler below. is that happening?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1252,6 +1207,11 @@ namespace Babylon
         delete texture;
     }
 
+    void NativeEngine::UpdateRawTexture(const Napi::CallbackInfo& info)
+    {
+        // todo
```
bghgary, line 1212: Why do we need this? If we do need it, it should implemented?

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -75,11 +74,13 @@ namespace
...
             });
 
-            Babylon::Polyfills::Window::Initialize(env);
 
-            // Initialize NativeWindow plugin.
-            auto width = static_cast<float>(rect.right - rect.left);
-            auto height = static_cast<float>(rect.bottom - rect.top);
+            Babylon::Polyfills::Window::Initialize(env);
+            // Initialize NativeWindow plugin to the test size.
+            // TODO: TestUtils::UpdateSize should do it properly but the client size
```
bghgary, line 80: Maybe try to find out why? If not, then we should create an issue.

Apps/ValidationTests/Shared/TestUtils.h
```diff
@@ -112,6 +117,47 @@ namespace Babylon
...
+        {
+            Image() = default;
+            ~Image()
+            {
+                if (m_Image)
+                {
+                    bimg::imageFree(m_Image.get());
+                }
+            }
+            std::unique_ptr<bimg::ImageContainer> m_Image;
```
bghgary, line 130: This is a bit confusing. Why do you need a `unique_ptr` when you are `new`ing the `Image`? `m_Image` will get deleted via the default allocator which calls delete `m_Image`. I don't think you want that?

Apps/ValidationTests/Shared/TestUtils.h
```diff
@@ -54,7 +59,7 @@ namespace Babylon
         {
             const int32_t exitCode = info[0].As<Napi::Number>().Int32Value();
 #ifdef WIN32
-            PostMessageW((HWND)_nativeWindowPtr, WM_QUIT, exitCode, 0);
+            PostMessageW((HWND)_nativeWindowPtr, WM_DESTROY, exitCode, 0);
```
bghgary, line 62: I think you probably want to use WM_CLOSE? https://stackoverflow.com/questions/3155782/what-is-the-difference-between-wm-quit-wm-close-and-wm-destroy-in-a-windows-pr

Apps/ValidationTests/Win32/App.cpp
CedricGuillemet: https://github.com/BabylonJS/BabylonNative/issues/118

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: I've started updating nativeengine.ts. Removing for now.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: image is null and get a null pointer exception. FlipY, GenerateMips or CreateTextureFromImage do not provide test for null pointer. What kind of errors are handled? It's not obvious to me

bghgary: arcana tasks will eat all exceptions and hold on to them for the next task. If the task chain has a check for exceptions (like it does here), it should be handled. But that said, we can probably check for nullptr explicitly for when imageParse fails.

Apps/ValidationTests/Shared/TestUtils.h
```diff
@@ -28,18 +37,17 @@ namespace Babylon
...
-                    ParentT::StaticMethod("updateSize", &TestUtils::UpdateSize),
-                    ParentT::StaticMethod("setTitle", &TestUtils::SetTitle),
-                    ParentT::StaticMethod("writePNG", &TestUtils::WritePNG),
+                    ParentT::InstanceMethod("exit", &TestUtils::Exit),
+                    ParentT::InstanceMethod("updateSize", &TestUtils::UpdateSize),
+                    ParentT::InstanceMethod("setTitle", &TestUtils::SetTitle),
+                    ParentT::InstanceMethod("writePNG", &TestUtils::WritePNG),
+                    ParentT::InstanceMethod("decodeImage", &TestUtils::DecodeImage),
+                    ParentT::InstanceMethod("getImageData", &TestUtils::GetImageData),
+                    ParentT::InstanceMethod("getWorkingDirectory", &TestUtils::GetWorkingDirectory),
```
CedricGuillemet, line 48: getWorkingDirectory is the easiest way I found to get it working for now as I need absolute path for Linux

Polyfills/Window/Source/Window.cpp
```diff
@@ -26,6 +26,11 @@ namespace Babylon::Polyfills::Internal
         auto jsNative = global.Get(JsRuntime::JS_NATIVE_NAME).As<Napi::Object>();
         auto jsWindow = constructor.New({});
 
+        // Need a reference or It's destroyed when loading babylon.material.js
+        // TODO: Find why
+        napi_ref result;
+        napi_create_reference(env, jsWindow, 1, &result);
```
CedricGuillemet, line 32: I filled an issue for this one : https://github.com/BabylonJS/BabylonNative/issues/236

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: This is something I'll address shortly as I need to, basically, do the same with empty index/vertex buffer.

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -55,6 +64,27 @@ namespace
         return { url };
     }
 
+    std::vector<std::string> GetCommandLineArguments()
```
CedricGuillemet, line 67: I think we should move this function and make it multiplatform

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -9,7 +9,10 @@
 #include <Windowsx.h>
 #include <Shlwapi.h>
 #include <filesystem>
+#include <algorithm>
 
+bool doExit = false;
```
CedricGuillemet, line 14: No window on CI and I need to exit properly with an error. It's the simplest way to do it.

PR #212, opened 2020-04-10
Command line arguments to napi

CedricGuillemet:
Expose command line arguments to JS
For upcoming use in validation tests and preview

### DIFF
--- a/Apps/ValidationTests/Win32/App.cpp
+++ b/Apps/ValidationTests/Win32/App.cpp
@@ -19,6 +19,7 @@
 #include <Babylon/Polyfills/Window.h>
 #include <Babylon/ScriptLoader.h>
 #include <Babylon/XMLHttpRequest.h>
+#include <Babylon/CmdArgs.h>
 
 #define MAX_LOADSTRING 100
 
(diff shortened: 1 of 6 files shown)


bghgary:
Also, why expose an object to JavaScript for this? Why not just call a method to get the arguments?

CedricGuillemet:
> I'm not sure this makes sense as a plugin. I think it probably make more sense to expose this directly in ValidationTests? What was the thinking behind making it a plugin?

Mutualize between validation tests and preview tool

CedricGuillemet:
> Also, why expose an object to JavaScript for this? Why not just call a method to get the arguments?

If I used a method, I would have to store argc/argv. I prefer to avoid states/data member when it's possible.

bghgary:
Okay, but it still seems a bit weird to me. The code that gets the command line arguments are not in the plugin, so it's just a container of strings right?

bghgary:
> If I used a method, I would have to store argc/argv. I prefer to avoid states/data member when it's possible.

I'm not sure what you mean by this. If you add the strings to JavaScript, JavaScript owns it. You don't need to hold on to anything.

Plugins/CmdArgs/Source/CmdArgs.cpp
```diff
@@ -0,0 +1,32 @@
...
+            : Napi::ObjectWrap<CmdArgs>{info}
+        {
+        }
+    };
+
+    void InitializeCmdArgs(Napi::Env env, const std::vector<std::string>& args)
+    {
+        CmdArgs::Initialize(env, args);
+    }
+}
```
syntheticmagus, line 32: nit: Newline.

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -64,7 +80,6 @@ namespace
         }
 
         runtime.reset();
-        Babylon::Plugins::NativeEngine::DeinitializeGraphics();
```
syntheticmagus, line 67: Does this need to be called from somewhere? It looks like in this PR it's removed (presumably because it was causing problems) and never put back, unless I missed it somewhere.

CedricGuillemet:
Closing for now. I will use it for BabylonPreview and see if it's needed for ValidationTests/Playground later.

PR #227, opened 2020-04-14
Some small cleanup

bghgary:
- Move ScriptLoader from Plugins to Core
- Move TicketedCollection to arcana as weak_table and remove BabylonNativeUtils
- Rename BabylonNativeCMakeFunctions to CMakeExtensions


Core/Shared/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_subdirectory(JsRuntime)
+add_subdirectory(AppRuntime)
+add_subdirectory(ScriptLoader)
```
syntheticmagus, line 3: Is this file supposed to be here? This feels like it might have been left over from something.

bghgary: Oops. Yeah, this is not supposed to be here.

Issue #230, opened 2020-04-15
Support ARCore


ryantrem:
Initial ARCore support was added in #203, but there is some follow up work that needs to be done to complete the ARCore work:

- [x] Handling surface resizing/screen rotation.
- [x] Handling suspend resume.
- [x] Handling devices that don't support ARCore.
- [x] Checking for ARCore package installation on the device (this might be another thing that the consuming app is required to do, not sure yet).
- [x] Support hit testing
- [x] Support for plane detection
- [x] Support anchors
- [x] Coming up with a pattern for releasing the global `jobject` we create to hold onto the `appContext` (see https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp#L58 for an example).
- [x] Fix garbage frames that are drawn while transitioning to AR
- [x] Move camera prompt to the AR initialization code path.
- [x] Verify projection ratio aspect ratio and maybe apply the same compensation fix from #325

PR #239, opened 2020-04-24
Handle display orientation changes in XR (ARCore specifically)

ryantrem:
This change is to handle display orientation changes for the ARCore-based XR support (as noted in #230). The primary changes are:
1. Creation of the textures moved from the constructor to GetNextFrame (it is done whenever the surface dimensions have changed (querying the surface dimensions is a ~12 microsecond operation, so I think it's ok to just check it every frame)).
1. Adding cleanup code to free the various resources (textures, shaders, etc.) that are allocated especially since some of these are now allocated every time the display orientation changes.
1. Exposing Android APIs to get the current display rotation through the JavaWrappers interfaces.
1. Moving around some more code to consolidate much of it to Session::Impl::GetNextFrame, and also moving the drawing code out of Frame::Impl::~Frame and into Session::Impl::DrawFrame for better encapsulation (Frame::Impl::~Frame still calls Session::Impl::DrawFrame).
1. Recreating frame buffers in NativeXR.cpp when the texture resolution changes.

### DIFF
--- a/Dependencies/xr/Source/ARCore/XR.cpp
+++ b/Dependencies/xr/Source/ARCore/XR.cpp
@@ -43,41 +43,12 @@ namespace xr
 
         bool TryInitialize()
         {
-            // Perhaps call eglGetCurrentSurface to get the render surface *before* XR render loop starts and changes to rendering to an FBO?
             return true;
         }
     };
 
     namespace
     {
-        void log(const char* message)
-        {
-            __android_log_print(ANDROID_LOG_VERBOSE, __FILE__, "%s", message);
-        }
-
-        template<typename TTimeUnit>
-        class DiagnosticTimer final
-        {
-        public:
-            DiagnosticTimer()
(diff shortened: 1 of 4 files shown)


Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -49,6 +50,34 @@ namespace xr
 
     namespace
     {
+        void log(const char* message)
+        {
+            __android_log_print(ANDROID_LOG_VERBOSE, __FILE__, "%s", message);
+        }
+
+        template<typename TTimeUnit>
```
ryantrem, line 58: This was useful for doing some quick and dirty "perf testing" as I went. It doesn't really belong here, but I also think it's probably worth keeping around for dev purposes...

bghgary: Probably put this in macro or put in a separate file somewhere else, so that it's clearly a debug thing?

PR #242, opened 2020-04-29
Created separate NativeXr plugin.

syntheticmagus:
Still have to test this in a headset because of problems with the runtime; but I've confirmed that with this change, the XR capabilities are no more broken than they currently are with the public runtime, and they're broken in the same way.


Apps/Playground/CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ elseif(WINDOWS_STORE)
         ${APPX_ASSETS}
         "UWP/App.cpp"
         "UWP/App.h")
+    set(ADDITIONAL_LIBRARIES ${ADDITIONAL_LIBRARIES} PRIVATE NativeXr)
```
bghgary, line 73:
```suggestion
    set(ADDITIONAL_LIBRARIES PRIVATE NativeXr)
```

Apps/Playground/CMakeLists.txt
```diff
@@ -80,6 +81,7 @@ elseif(WIN32)
         "Win32/Resource.h"
         "Win32/small.ico"
         "Win32/targetver.h")
+    set(ADDITIONAL_LIBRARIES ${ADDITIONAL_LIBRARIES} PRIVATE NativeXr)
```
bghgary, line 84:
```suggestion
    set(ADDITIONAL_LIBRARIES PRIVATE NativeXr)
```

PR #244, opened 2020-04-30
Enable validation tests on CI for Win32

CedricGuillemet:
Run the tests provided in config.json and fails the build when image difference threshold is reached.
I successfully failed it by setting threshold to 0.

### DIFF
--- a/Apps/ValidationTests/Win32/App.cpp
+++ b/Apps/ValidationTests/Win32/App.cpp
@@ -11,8 +11,6 @@
 #include <filesystem>
 #include <algorithm>
 
-volatile bool doExit = false;
-int errorCode{};
 #include <Shared/TestUtils.h>
 
 #include <Babylon/AppRuntime.h>
--- a/Apps/ValidationTests/Win32/App.cpp
+++ b/Apps/ValidationTests/Win32/App.cpp
@@ -29,7 +29,6 @@ HINSTANCE hInst;                                // current instance
 WCHAR szTitle[MAX_LOADSTRING];                  // The title bar text
 WCHAR szWindowClass[MAX_LOADSTRING];            // the main window class name
 std::unique_ptr<Babylon::AppRuntime> runtime{};
-bool _NoWindow{};                               // No window for CI
 
 // 600, 400 mandatory size for CI tests
 static const int TEST_WIDTH = 600;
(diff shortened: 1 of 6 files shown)


Apps/ValidationTests/Win32/App.cpp
```diff
@@ -11,7 +11,7 @@
 #include <filesystem>
 #include <algorithm>
 
-bool doExit = false;
+volatile bool doExit = false;
```
bghgary, line 14: who is setting `doExit` to true?

bghgary: This code is awkward. Maybe declare the doExit variable in TestUtils.h instead?

bghgary: Also, why does it get optimized out if `doExit` is being used? 🤔

CedricGuillemet: No write to that variable detected by the linker in that .cpp and you get an endless loop :)

CedricGuillemet: That's amazing! Defining the boolean in testutils and just testing it in app.cpp, I end up with the same result: boolean is not checked.

bghgary: I'm not that surprised since it's the same to the compiler. That said, using a while loop for this is not a great pattern. Maybe you should use some kind of blocking signal instead?

CedricGuillemet: Sure! Do you have something in mind for that? Is there something in Arcana?

bghgary: There is, but I don't think you need it. Try [std::condition_variable](https://en.cppreference.com/w/cpp/thread/condition_variable)?

CedricGuillemet: condition_variable is fine on win32 but on x11, XNextEvent is a blocking call and it's easier to use a boolean for exiting. I prefer to have the same exit mecanism for all the platforms. So, I'll stick with an atomic<bool> for now.

CedricGuillemet: I did another try at creating the window...and it worked! I should have kept the window manager error the first time I got it.

PR #248, opened 2020-05-05
Validation tests Linux CI

CedricGuillemet:
Validationt tests on Linux (JSC + OpenGL) with results artifact publishing.

### DIFF
--- a/Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
+++ b/Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
@@ -44,7 +44,7 @@ namespace Babylon
             options.version = 310;
             options.es = true;
 #else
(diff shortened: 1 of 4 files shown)


bghgary:
Something might be wrong with the merge? The changes don't look right.

CedricGuillemet:
@bghgary Better with a rebase

Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -44,7 +44,7 @@ namespace Babylon
             options.version = 310;
             options.es = true;
 #else
-            options.version = 430;
+            options.version = 330;
```
bghgary, line 47: This is likely going to be an issue when we fix #300.

PR #252, opened 2020-05-07
Move app context jobject lifetime management to Globals

ryantrem:
This change makes it the responsibility of `Globals` to manage the lifetime of the `jobject` referring to the app context. With the current implementation, it does this by first getting the application context from an arbitrary context, and creating a global ref from this jobject. The application context itself has the lifetime of the application, so in this case Globals just never cleans up this jobject (it will always exist, and reference an always existing application context).

### DIFF
--- a/Dependencies/AndroidExtensions/Source/Globals.cpp
+++ b/Dependencies/AndroidExtensions/Source/Globals.cpp
@@ -55,10 +55,10 @@ namespace android::global
         RequestPermissionsResultEvent g_requestPermissionsResultEvent{};
     }
 
-    void Initialize(JavaVM* javaVM, jobject appContext)
+    void Initialize(JavaVM* javaVM, jobject context)
     {
         g_javaVM = javaVM;
-        g_appContext = appContext;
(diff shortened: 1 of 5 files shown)


Dependencies/AndroidExtensions/Source/Globals.cpp
```diff
@@ -55,10 +55,10 @@ namespace android::global
         RequestPermissionsResultEvent g_requestPermissionsResultEvent{};
     }
 
-    void Initialize(JavaVM* javaVM, jobject appContext)
+    void Initialize(JavaVM* javaVM, jobject context)
     {
         g_javaVM = javaVM;
-        g_appContext = appContext;
+        g_appContext = GetEnvForCurrentThread()->NewGlobalRef(android::content::Context{context}.getApplicationContext());
```
bghgary, line 61: Need to release this?

ryantrem: Nowhere. It's not obvious to me where we would do this, and this is one jobject referencing an object that already will never get deleted, so I think it is fine.

PR #257, opened 2020-05-11
Dynamic vertex and index buffers

syntheticmagus:
The Babylon.js side of this PR is still in progress, and I won't be able to thoroughly test this until the shader compilation bug is fixed later this week, but the C++ side of this change should already be quite representative of what the final proposed change will be. The core modification behind this change comes from learning that the hack we were doing during vertex buffer creation -- creating a "placeholder" vertex layout -- didn't work with dynamic vertex buffers, which tried to use the stride from that layout and ended up with errors. Switched instead to a lazy initialization of sorts, which removes the placeholder layout by delaying actual buffer initialization until we have the real layout. Also changed index buffers to follow the same pattern.


CedricGuillemet:
Please check the bjs validation tests config to see if a PG that uses this feature can be added to our config.json

syntheticmagus:
@CedricGuillemet There's not one specifically for text, but the text is a fairly thin layer built on top of the Solid Particle System, which is just pretty much just dynamic mesh. Apparently those features are also leveraged by the glTF import system when it has to generate something (for instance, if a mesh doesn't come packaged with normals), so Gary just pointed me to some of the Asset Generator files. Bottom line, I don't think there's a Validation Test specifically for this frontline feature, but the underlying code paths should all be exercised by the Asset Generator Tests

Apps/Playground/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(BABYLONSCRIPTS
     "../BabylonScripts/babylon.max.js"
     "../BabylonScripts/babylonjs.materials.js"
     "../BabylonScripts/ammo.js"
+    "../BabylonScripts/meshwriter.min.js"
```
ryantrem, line 6: So this is not part of Babylon.js itself, correct? It *looks* like in the React Native case we would just install the [meshwriter NPM package](https://www.npmjs.com/package/meshwriter), does that sound right?

Apps/Playground/Scripts/experience.js
```diff
@@ -158,6 +159,18 @@ CreateBoxAsync().then(function () {
             });
         }, 5000);
     }
+
+    if (text) {
+        var Writer = BABYLON.MeshWriter(scene, { scale: 1.0, defaultFont: "Arial" });
+        text1 = new Writer(
```
ryantrem, line 165: Where is `text1` used?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -294,6 +294,140 @@ namespace Babylon
             texture->Width = width;
             texture->Height = height;
         }
+
+        class IndexBufferData
```
ryantrem, line 298: `final`

Apps/Playground/CMakeLists.txt
syntheticmagus: That sounds right to me. I haven't tried it, but I just copied this file out of the Babylon.js extensions repository, so it presumably should be the same as the one on NPM.

Apps/Playground/Scripts/experience.js
syntheticmagus: It isn't; copy-paste leftover. I'll remove it.

ryantrem: I’m confused then... how do you set the transform of the text? How do you remove the text from the scene?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -294,6 +294,140 @@ namespace Babylon
...
+                }
+            }
+
+            const bool Dynamic{};
+
+        private:
+            uint16_t m_handleIndex{bgfx::kInvalidHandle};
+        };
+
+        class VertexBufferData final
```
bghgary, line 352: This bool with a variant return seems annoying. Would it be cleaner to use two different classes VertexBufferData and DynamicVertexBufferData and `new` a variant of these?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -294,6 +294,140 @@ namespace Babylon
...
+            }
+
+            void EnsureFinalized(Napi::Env env, const bgfx::VertexLayout& layout)
+            {
+                if (m_handleIndex != bgfx::kInvalidHandle)
+                {
+                    return;
+                }
+
+                const bgfx::Memory* bytes = bgfx::copy(m_bytes.data(), static_cast<uint32_t>(m_bytes.size()));
```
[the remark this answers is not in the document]
syntheticmagus: My initial version was that way, but there are times when `Update()` will get called before `EnsureFinalized()`, which is an allowed behavior. When that happens, the intermediate value stored by the `VertexBufferData` needs to be discarded and replaced with a new one because we can't directly call `bgfx::update()` since the handle isn't valid yet. (This is the first conditional case in `Update()`.) `bgfx::Memory` objects don't have a contract for getting rid of them except to pass them into `bgfx` so that they can be used, then freed; so I didn't see a clean way to proactively make one and decide later that I didn't want to actually give it to `bgfx`. This does end up being an extra copy operation, but only in order to avoid using `bgfx::Memory` in a way it doesn't seem to be designed to support.

Apps/Playground/Scripts/experience.js
syntheticmagus: I should have clarified. You _can_ do all those things using the former `text1` variable or equivalent; you can get an ordinary Babylon mesh from it with which you can do whatever you want. (The "usage summary" on the NPM page you linked to above shows an example of this.) I just don't do any of that here, so the variable was never used.

Plugins/NativeEngine/Source/NativeEngine.cpp
syntheticmagus: It's a bit clunky, but without this we'd have to pass down a Boolean `isDynamic` parameter (or equivalent) from the JavaScript to the Create and Record methods for both vertex and index buffers; otherwise we won't know what kind of `Napi::External` to to interpret our argument as. I did something like that initially, but it was even clunkier than this and added unnecessary complexity to the JS-native boundary, and the few other ideas I considered were clunkier still, so I ended up using a `std::variant` as the least ugly of many ugly options.

Plugins/NativeEngine/Source/NativeEngine.cpp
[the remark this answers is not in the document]
bghgary: You can use bgfx::makeRef to control the lifetime?

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: I'm not sure you understand what I mean or I'm missing something. If you pass std::variant<VertexBufferData, DynamicVertexBufferData>* as the data to NAPI, the you can tell which one it is when you receive it?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -296,6 +297,143 @@ namespace Babylon
...
+                {
+                    // Buffer was already created, do a real update operation.
+                    const bgfx::Memory* memory = bgfx::copy(bytes.Data() + offset, byteLength);
+                    bgfx::update(handle, 0, memory);
+                } });
+        }
+
+        void SetAsBgfxVertexBuffer(uint8_t index, uint32_t startVertex, bgfx::VertexLayoutHandle layout) const
+        {
+            DoForHandleTypes([index, startVertex, layout](auto handle) { bgfx::setVertexBuffer(index, handle, startVertex, UINT32_MAX, layout); }, [index, startVertex, layout](auto handle) { bgfx::setVertexBuffer(index, handle, startVertex, UINT32_MAX, layout); });
```
syntheticmagus, line 430: I'm not even going to pretend to understand why clang format thought this was a good idea.

It looks like there's at least some support for less appalling [Allman-style](https://en.wikipedia.org/wiki/Indentation_style#Allman_style) lambda formatting in clang format 11. I think VS code can do clang format 11, and there's an add-on to make VS 2019 do it. Should we explore that further?

syntheticmagus: This went away when I had to rephrase the code for clang.

PR #258, opened 2020-05-12
NAPI JSC updates

bghgary:
Fixes shutdown issues in BabylonReactNative repo.

- Retain JSC global context.
- Track references and delete them on env detach.
- Implement weak references properly.


Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -205,7 +202,8 @@ namespace {
       JSObjectSetPrototype(env->context, constructor, prototype);
       
       JSValueRef exception{};
-      JSObjectSetProperty(env->context, prototype, JSString("constructor"), constructor, kJSPropertyAttributeReadOnly | kJSPropertyAttributeDontDelete, &exception);
+      JSObjectSetProperty(env->context, prototype, JSString("constructor"), constructor,
```
ryantrem, line 205: Nit: Is this line break intentional? I thought it was perfectly readable on one line, so wasn't sure if this was some automatic tooling.

Dependencies/napi/source/js_native_api_JavaScriptCore.cc
```diff
@@ -464,6 +505,86 @@ namespace {
...
+    , _count{count} {
+  }
+  
+  napi_status init(napi_env env) {
+    // track the ref values to support weak refs
+    auto pair{env->active_ref_values.insert(_value)};
+    if (pair.second) {
+      ExternalInfo* info{};
+      CHECK_NAPI(ExternalInfo::Wrap(env, _value, &info));
+      info->AddFinalizer([iter{pair.first}](ExternalInfo* info) {
```
ryantrem, line 520: So an interator for an `unordered_set` remains valid when the collection is mutated?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -467,10 +467,16 @@ namespace Babylon
...
     {
         auto callback = info[0].As<Napi::Function>();
-        arcana::make_task(m_runtimeScheduler, m_cancelSource, [this, callbackRef = Napi::Persistent(callback)]() {
+
+        if (m_requestAnimationFrameCalback.IsEmpty() ||
+            m_requestAnimationFrameCalback.Value() != callback) {
+            m_requestAnimationFrameCalback = Napi::Persistent(callback);
+        }
+
+        arcana::make_task(m_runtimeScheduler, m_cancelSource, [this]() {
```
ryantrem, line 476: Use Dispatch directly.

Dependencies/napi/source/js_native_api_JavaScriptCore.cc
bghgary: I did this manually in this case. I try to keep NAPI conventions which seems to limit the line length.

Dependencies/napi/source/js_native_api_JavaScriptCore.cc
bghgary: Good catch. I was originally using a different container that was okay, but unordered_set is not okay.

https://en.cppreference.com/w/cpp/container/unordered_set
> Operations | Invalidated
> -- | --
> All read only operations, swap, std::swap | Never
> clear, rehash, reserve, operator= | Always
> insert, emplace, emplace_hint | Only if causes rehash
> erase | Only to the element erased
> 
>

PR #259, opened 2020-05-12
gcc pipeline Linux

CedricGuillemet:
- GCC 9 compilation pipeline for Linux (doesn't build with GCC8 because of a template issue with ScriptLoader::impl)
 - Readme updated to reflect the change

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,23 @@ or
 make
 ```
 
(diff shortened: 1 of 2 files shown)


CedricGuillemet:
@syntheticmagus The issue is related to gcc supported c++ features. If no one complains about the gcc version then no need to open an issue. We will see in the future if any one needs an older version.

PR #262, opened 2020-05-14
Update device input system to match the latest contract on the Babylon.js side

ryantrem:
With this change the native contract for `DeviceInputSystem` matches the latest JS contract for `DeviceInputSystem`. The notable changes are:
- Devices are now identified by the DeviceType/DeviceSlot pair (rather than a string, e.g. "Pointer-0").
- There is now an event that fires whenever any input state changes.

Note that this still doesn't work in the sense of "overriding" the JS class. It seems it is not enough to just define a class with the same name and try to put it in the global scope. So as is, this class can be used/tested in the React Native repo by just not importing the JS `DeviceInputSystem` (in which case the native provides the implementation), but as soon as you import the JS version, then the native one does not replace the JS one. I need to think more about how to handle this, and I'll address it in a different change.

FYI @polygonalsun.

### DIFF
--- a/Plugins/NativeInput/Source/NativeInput.h
+++ b/Plugins/NativeInput/Source/NativeInput.h
@@ -26,6 +26,9 @@ namespace Babylon::Plugins
         using DeviceStatusChangedCallback = std::function<void(DeviceType deviceType, int32_t deviceSlot)>;
         using DeviceStatusChangedCallbackTicket = arcana::weak_table<DeviceStatusChangedCallback>::ticket;
 
+        using InputStateChangedCallback = std::function<void(DeviceType deviceType, int32_t deviceSlot, uint32_t inputIndex, std::optional<int32_t> previousState, std::optional<int32_t> currentState)>;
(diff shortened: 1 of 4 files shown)


Plugins/NativeInput/Source/NativeInput.h
```diff
@@ -23,16 +37,29 @@ namespace Babylon::Plugins
...
     private:
-        std::vector<int32_t>& GetOrCreateInputMap(const std::string& deviceId, const std::vector<uint32_t>& inputIndices);
-        void RemoveInputMap(const std::string& deviceId);
+        using InputMapKey = std::pair<DeviceType, int32_t>;
+
+        struct InputMapKeyHash
+        {
+            size_t operator()(const InputMapKey& inputMapKey) const
+            {
+                return std::hash<uint64_t>{}((static_cast<uint64_t>(inputMapKey.first) << (sizeof(uint32_t) * 8)) + static_cast<uint32_t>(inputMapKey.second));
```
bghgary, line 50: > `<< (sizeof(uint32_t) * 8))`

Is this really what you mean?

PR #263, opened 2020-05-15
Update ARCore native session intialization

Alex-MSFT:
This change adds proper ARCore compatibility checks, ARCore initialization, and moves the camera permission check logic to the actual AR Session start time in the Android implementation of xr.cpp, as described in #230.   

Key changes:

- Added arcana tasking to nativexr.cpp & xr.cpp to support async logic during IsSessionSupported, and BeginSession.

- Added globals.cpp reference and JNI wrappers to the current activity to support checking and requesting permissions

- Added AR Core compatibility check to IsSessionSupported, updated IsSessionSupported to properly handler "immersive-vr", "immersive-ar", and "inline" as 3 separate supported session types.

- Changed constructor of the native session from a plain constructor to an async factory function that returns a shared_ptr (can change back to unique_ptr once we have identified the blocking issue in arcana.)

- Added AR Core installation check+request and camera permission check+request to the codepath in BeginSession.

- Updated playground to support AR and VR as separate options.


Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -642,6 +663,158 @@ namespace xr
...
+
+    bool CheckARCoreInstallStatus(bool requestInstall)
+    {
+        ArInstallStatus install_status;
+        ArStatus installStatus = ArCoreApk_requestInstall(
+                GetEnvForCurrentThread(), GetMainActivity(), requestInstall, &install_status);
+        return installStatus == AR_SUCCESS && install_status == AR_INSTALL_STATUS_INSTALLED;
+    }
+
+    arcana::task<void, std::exception_ptr> CheckAndInstallARCore(const Babylon::JsRuntimeScheduler& runtimeScheduler)
```
bghgary, line 724: I don't understand why all of these functions take a JsRuntimeScheduler. Do you actually need to be on the JsRuntime thread?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -147,8 +149,8 @@ namespace Babylon
         NativeXr();
         ~NativeXr();
 
-        void BeginSession(Napi::Env env); // TODO: Make this asynchronous.
-        void EndSession();                // TODO: Make this asynchronous.
+        arcana::task<void, std::exception_ptr> BeginSession(Napi::Env env, const JsRuntimeScheduler& runtimeScheduler);
```
bghgary, line 152: You don't need to do this in this PR, but it would be good to clean up the code for this. BeginSession should probably be renamed to BeginSessionAsync and EndSession should probably become async as well. Also, these functions should actually do the work asynchronously instead of a while loop.

Dependencies/xr/Source/ARCore/XR.cpp
BrodoBaggins: The final constructor of session impl is the only one that must run in the context of the jsRuntime, as we must load shaders on the main JS thread.

I had these all using an inline_scheduler, but it was not running subsequent tasks when coming back from the nested tasks.  I figured since we are already passing down the jsRuntime I could just throw them on the main JsRuntime thread, as we already have a handle to that context.

Alex-MSFT: The final constructor of session impl is the only one that must run in the context of the jsRuntime, as we must load shaders on the main JS thread.

I had these all using an inline_scheduler, but it was not running subsequent tasks when coming back from the nested tasks. I figured since we are already passing down the jsRuntime I could just throw them on the main JsRuntime thread, as we already have a handle to that context.

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: I'll look at that after this PR.  I briefly looked at the init code in OpenXR which is the reason why this is a busy while loop (on the Android implementation it just returns true, so breaks on the first pass), but since I was primarily working on Android and this change was already getting big I deferred that change.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -642,6 +655,158 @@ namespace xr
...
+            });
+
+            return tcs.as_task();
+        }
+
+        // VR and inline sessions are not supported at this time.
+        return arcana::task_from_result<std::exception_ptr>(false);
+    }
+
+    bool CheckARCoreInstallStatus(bool requestInstall)
```
ryantrem, line 708: This is a global function. I would put this in an anonymous namespace like the other helper functions (e.g. `LoadShader`) so that it is not visible to anything outside this file, and can't conflict with other globals.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -642,6 +655,158 @@ namespace xr
...
+                else
+                {
+                    // ARCore installed successfully, complete the promise.
+                    installTcs.complete();
+                }
+            })};
+
+            // Kick off the install request, and set the task for our caller to wait on.
+            CheckARCoreInstallStatus(true);
+            task = installTcs.as_task().then(runtimeScheduler, arcana::cancellation::none(), [resumeTicket = std::move(resumeTicket)](){});
```
ryantrem, line 743: All this continuation does is destroy the ticket, and the underlying ticketed_collection is already thread safe, so you should be able to just use `inline_scheduler` here.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -642,6 +655,158 @@ namespace xr
...
+                            std::ostringstream message;
+                            message << "Camera permission not acquired successfully";
+                            permissionTcs.complete(arcana::make_unexpected(make_exception_ptr(std::runtime_error{message.str()})));
+                        }
+                    })
+            };
+
+            // Kick off the permission check request, and set the task for our caller to wait on.
+            GetMainActivity().requestPermissions(PERMISSION_NAME, PERMISSION_REQUEST_ID);
+            task = permissionTcs.as_task().then(runtimeScheduler, arcana::cancellation::none(), [ticket = std::move(permissionTicket)](){});
```
ryantrem, line 790: Same comment as above, I would expect you could use `inline_scheduler` here since the purpose is just to extend the lifetime of the ticket.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -642,6 +655,158 @@ namespace xr
...
+        }
+
+        return task;
+    }
+
+    arcana::task<std::shared_ptr<System::Session>, std::exception_ptr> System::Session::CreateAsync(const Babylon::JsRuntimeScheduler& runtimeScheduler, System& system, void* graphicsDevice)
+    {
+        // First perform the ARCore installation check, request install if not yet installed.
+        return CheckAndInstallARCoreAsync(runtimeScheduler).then(runtimeScheduler, arcana::cancellation::none(), [&runtimeScheduler, &system, graphicsDevice]()
```
ryantrem, line 799: It doesn't matter what thread `CheckAndInstallARCoreAsync` or `CheckCameraPermissionAsync` are called on right? It only matters where the caller continues, since the caller is the one that transitions back to JS by completing a promise. Assuming this is correct, then I would think that this code could just use `inline_scheduler` and the caller could use the JS scheduler to get back on the JS thread before completing the promise.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -239,7 +240,9 @@ namespace Babylon
             }
         }
 
-        m_session = std::make_unique<xr::System::Session>(m_system, bgfx::getInternalData()->context);
+        return xr::System::Session::CreateAsync(runtimeScheduler, m_system, bgfx::getInternalData()->context).then(runtimeScheduler, arcana::cancellation::none(), [this](std::shared_ptr<xr::System::Session> session) {
```
ryantrem, line 243: Since the final continuation is just doing an assignment, can it just use `inline_scheduler`?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1186,15 +1189,34 @@ namespace Babylon
...
+                            if (result.has_error())
+                            {
+                                try
+                                {
+                                    std::rethrow_exception(result.error());
+                                }
+                                catch (const std::exception& e)
+                                {
+                                    deferred.Reject(Napi::Error::New(env, e.what()).Value());
+                                    throw e;
```
ryantrem, line 1211: We shouldn't re-throw this should we? The promise has been rejected, so the JS side should be able to handle the exception. If we re-throw here, it doesn't go anywhere that can be handled, right?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1186,15 +1189,34 @@ namespace Babylon
...
-                auto& session = *XRSession::Unwrap(jsSession);
-
-                session.m_xr.BeginSession(info.Env());
+                auto jsSession = Napi::Persistent(info.Env().Global().Get(JS_CLASS_NAME).As<Napi::Function>().New({info[0]}));
+                auto& session = *XRSession::Unwrap(jsSession.Value());
 
                 auto deferred = Napi::Promise::Deferred::New(info.Env());
-                deferred.Resolve(jsSession);
+                session.m_xr.BeginSession(info.Env(), runtimeScheduler)
+                    .then(runtimeScheduler,
```
ryantrem, line 1199: I think this is really the only place I would expect we'd need to use the `JsRuntimeScheduler`.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1186,15 +1189,34 @@ namespace Babylon
                 env.Global().Set(JS_CLASS_NAME, func);
             }
 
-            static Napi::Promise CreateAsync(const Napi::CallbackInfo& info)
+            static Napi::Promise CreateAsync(const Napi::CallbackInfo& info, const JsRuntimeScheduler& runtimeScheduler)
```
ryantrem, line 1192: I think rather than passing in a scheduler to this function (seems weird) I would just have `XRSession` own it's own `JsRuntimeScheduler`.

Dependencies/xr/Source/ARCore/XR.cpp
ryantrem: So I don't think at the xr.cpp layer we need to have any knowledge of the `JsRuntimeScheduler`.

Dependencies/xr/Source/ARCore/XR.cpp
ryantrem: Oh, I just saw this comment, oops. I left a bunch of comments on this topic as well. Let's try `inline_scheduler` and if it isn't working for some reason I'm happy to take a look with you.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: The OpenGL initialization in the constructor of the session can only occur on the main JS thread as each OpenGL context is only valid on the thread it is created on (which is the JS runtime thread).  So I would either need to go back to breaking this out into two calls for creation & initialization or pass the jsruntime scheduler down.

When I switch over to using the inline_scheduler it successfully runs the final tasks in CheckCameraPermissionsAsync and CheckAndInstallARCore, but does not properly picking up the continuation in the calling method.  We can debug it Monday.

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: You're right, removed.

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: Done

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -235,6 +242,27 @@ namespace xr
...
+            ArFrame_destroy(frame);
+            ArSession_destroy(session);
+
+            glDeleteTextures(1, &cameraTextureId);
+            glDeleteProgram(shaderProgramId);
+
+            DestroyDisplayResources();
+        }
+
+        void StartARSession()
```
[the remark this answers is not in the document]
Alex-MSFT: I had this separated out when I had a separate initializer.  Will move this back.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -642,6 +655,158 @@ namespace xr
         return m_impl->TryInitialize();
     }
 
+    arcana::background_dispatcher<64> scheduler;
```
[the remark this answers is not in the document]
Alex-MSFT: Switched this one over to std::thread

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Done

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -164,6 +164,27 @@ namespace java::net
...
+namespace android::app
+{
+    Activity::Activity(jobject object)
+        : Object{"android/app/Activity", object}
+    {
+    }
+
+    void Activity::requestPermissions(const char* permissionName, int permissionRequestID)
+    {
+        jclass cls{m_env->FindClass("android/Manifest$permission")};
```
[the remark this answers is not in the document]
Alex-MSFT: Done

Dependencies/AndroidExtensions/Include/AndroidExtensions/Globals.h
```diff
@@ -6,12 +6,13 @@
 
 namespace android::global
 {
-    void Initialize(JavaVM* javaVM, jobject appContext);
+    void Initialize(JavaVM* javaVM, jobject appContext, jobject mainActivity);
```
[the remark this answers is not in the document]
Alex-MSFT: Changed to currentActivity, and added setCurrentActivity, which will handle setting the activity.

Dependencies/AndroidExtensions/Source/Globals.cpp
```diff
@@ -55,10 +56,11 @@ namespace android::global
         RequestPermissionsResultEvent g_requestPermissionsResultEvent{};
     }
 
-    void Initialize(JavaVM* javaVM, jobject context)
+    void Initialize(JavaVM* javaVM, jobject context, jobject mainActivity)
     {
         g_javaVM = javaVM;
         g_appContext = GetEnvForCurrentThread()->NewGlobalRef(android::content::Context{context}.getApplicationContext());
+        g_mainActivity = GetEnvForCurrentThread()->NewGlobalRef(mainActivity);
```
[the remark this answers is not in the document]
Alex-MSFT: Change dto GetCurrentActivity, and added SetCurrentActivity to support swapping out activity context.

Dependencies/xr/Include/XR.h
```diff
@@ -131,12 +141,10 @@ namespace xr
                 std::unique_ptr<Impl> m_impl{};
             };
 
+            static arcana::task<std::shared_ptr<Session>, std::exception_ptr> CreateAsync(const Babylon::JsRuntimeScheduler& runtimeScheduler, System& system, void* graphicsDevice);
             Session(System& system, void* graphicsDevice);
             ~Session();
 
-            Session(Session&) = delete;
-            Session& operator=(Session&&) = delete;
-
```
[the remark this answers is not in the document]
Alex-MSFT: 1.) Yes they should, but we discovered a bug in arcana where it would implicitly try to copy the ptr before executing the move, and we could not trick arcana into using the const ptr implementation of the get value on the expected.  We decided to just change this to shared_ptr for now as m_impl is already a unique_ptr and does all the real work, and investigate the arcana issue in the future.

2.) I would like to get rid off the public Session constructor, but this is implicitly needed to be public because make_shared enforces that you use a public constructor.

3.) These are implicitly deleted by the use of unique_ptr for the m_impl class.

Dependencies/xr/Include/XR.h
```diff
@@ -149,11 +157,9 @@ namespace xr
         System(const char* = "OpenXR Experience");
         ~System();
 
-        System(System&) = delete;
-        System& operator=(System&&) = delete;
-
```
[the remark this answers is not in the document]
Alex-MSFT: These are implicitly deleted by the use of unique_ptr for the m_impl class, so cleaning these up to get rid of compiler warnings.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -642,6 +737,69 @@ namespace xr
...
+    {
+        // First perform the ARCore installation check, request install if not yet installed.
+        return CheckAndInstallARCoreAsync(runtimeScheduler).then(runtimeScheduler, arcana::cancellation::none(), [&runtimeScheduler, &system, graphicsDevice]()
+        {
+            // Next check for camera permissions, and request if not already granted.
+            return CheckCameraPermissionAsync(runtimeScheduler).then(runtimeScheduler, arcana::cancellation::none(), [&system, graphicsDevice]()
+            {
+                // Finally if the previous two tasks succeed, start the AR session.
+                return std::make_shared<System::Session>(system, graphicsDevice);
+            });
```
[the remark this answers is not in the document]
Alex-MSFT: Sure, done

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Worked with Ryan offline, and got rid of the need to use another scheduler.  There was a bug in our use of ticketed collection using a mutex instead of a recursive_mutex that was causing my other issues, and we moved the initialization code to the first instance of GetNextFrame to get around the OpenGL issues.

Dependencies/xr/CMakeLists.txt
```diff
@@ -46,6 +46,10 @@ else()
     target_link_libraries(xr PRIVATE openxr_loader)
 endif()
 
+target_link_libraries(xr
+    PUBLIC arcana
+    PUBLIC JsRuntime)
```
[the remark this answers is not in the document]
Alex-MSFT: Worked with Ryan offline, and got rid of the need to use another scheduler.  There was a bug in our use of ticketed collection using a mutex instead of a recursive_mutex that was causing my other issues, and we moved the initialization code to the first instance of GetNextFrame to get around the OpenGL issues.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Worked with Ryan offline, and got rid of the need to use another scheduler. There was a bug in our use of ticketed collection using a mutex instead of a recursive_mutex that was causing my other issues, and we moved the initialization code to the first instance of GetNextFrame to get around the OpenGL issues.

Dependencies/AndroidExtensions/Source/Globals.cpp
```diff
@@ -7,6 +7,7 @@ namespace android::global
     {
         JavaVM* g_javaVM{};
         jobject g_appContext{};
+        jobject g_currentActivity = nullptr;
```
ryantrem, line 10: `jobject g_currentActivity` should initialize this to nullptr and be consistent with the other variable initialization.

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -164,6 +164,36 @@ namespace java::net
...
+    Activity::Activity(jobject object)
+        : Object{"android/app/Activity", object}
+    {
+    }
+
+    void Activity::requestPermissions(const char* permissionName, int permissionRequestID)
+    {
+        jstring systemPermissionName{ManifestPermission::getPermissionName(permissionName)};
+
+        jobjectArray permissionArray{m_env->NewObjectArray(
```
ryantrem, line 189: Do we need to free this object array after the `CallVoidMethod`?

Dependencies/xr/Include/XR.h
```diff
@@ -131,11 +140,11 @@ namespace xr
...
             };
 
-            Session(System& system, void* graphicsDevice);
+            static arcana::task<std::shared_ptr<Session>, std::exception_ptr> CreateAsync(System& system, void* graphicsDevice);
             ~Session();
 
-            Session(Session&) = delete;
-            Session& operator=(Session&&) = delete;
+            // Do not use, call CreateAsync instead.
+            Session(System& system, void* graphicsDevice);
```
ryantrem, line 147: Can this not be private?

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -164,6 +164,36 @@ namespace java::net
     }
 }
 
+namespace android
+{
+    jstring ManifestPermission::getPermissionName(const char* permissionName)
```
ryantrem, line 169: Seems like this should be private, and we should just expose functions like `CAMERA()` that call this function with the right hard coded constant strings (names of Java class fields). This would be a better 1:1 mapping of the underlying Java classes.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -24,11 +27,15 @@
...
 #include <gtc/type_ptr.hpp>
 #include <gtx/quaternion.hpp>
+#include <arcana/threading/task_schedulers.h>
 
 using namespace android::global;
 
 namespace xr
 {
+    const char* PERMISSION_NAME{"CAMERA"};
+    const int PERMISSION_REQUEST_ID = 8435;
```
ryantrem, line 37: Maybe put a comment on this :)

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1188,13 +1191,31 @@ namespace Babylon
...
+                auto jsSession = Napi::Persistent(info.Env().Global().Get(JS_CLASS_NAME).As<Napi::Function>().New({info[0]}));
+                auto& session = *XRSession::Unwrap(jsSession.Value());
 
                 auto deferred = Napi::Promise::Deferred::New(info.Env());
-                deferred.Resolve(jsSession);
+                session.m_xr.BeginSession(info.Env())
+                    .then(session.m_runtimeScheduler,
+                        arcana::cancellation::none(),
+                        [deferred, jsSession = std::move(jsSession), env = info.Env()](const arcana::expected<void, std::exception_ptr>& result) {
+                            if (result.has_error())
```
ryantrem, line 1202: Nit: The indentation/formatting here seems weird. Seems like it would be more readable to have something like:
```c++
session.m_xr.BeginSession(info.Env()).then(session.m_runtimeScheduler, arcana::cancellation::none(),
    [deferred, jsSession = std::move(jsSession), env = info.Env()](const arcana::expected<void, std::exception_ptr>& result) {
        if (result.has_error())
```

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
```diff
@@ -30,6 +30,11 @@ namespace java::net
     class URLConnection;
 }
 
+namespace android::app
+{
+    class Activity;
```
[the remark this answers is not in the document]
bghgary: Some of the declarations below are circular. I put all of the forward declared to be consistent.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -118,6 +118,12 @@ extern "C"
         }
     }
 
+    JNIEXPORT void JNICALL
+    Java_BabylonNative_Wrapper_setCurrentActivity(JNIEnv* env, jclass clazz, jobject currentActivity)
+    {
+        android::global::SetCurrentActivity(currentActivity);
+    };
```
bghgary, line 125: nit:
```suggestion
    }
```

Dependencies/AndroidExtensions/Source/Globals.cpp
bghgary: > jobject g_mainActivity{};

Does this not initialize to nullptr by default?

BrodoBaggins: Fixed

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1188,13 +1191,31 @@ namespace Babylon
...
-
-                session.m_xr.BeginSession(info.Env());
+                auto jsSession = Napi::Persistent(info.Env().Global().Get(JS_CLASS_NAME).As<Napi::Function>().New({info[0]}));
+                auto& session = *XRSession::Unwrap(jsSession.Value());
 
                 auto deferred = Napi::Promise::Deferred::New(info.Env());
-                deferred.Resolve(jsSession);
+                session.m_xr.BeginSession(info.Env())
+                    .then(session.m_runtimeScheduler,
+                        arcana::cancellation::none(),
```
bghgary, line 1200: I would highly recommend adding a cancellation token in this class that is canceled when the class is destructed, especially when going back to the js runtime. Once the js runtime dies, we shouldn't have continuations running afterwards.

bghgary: Sorry, just realized I put this on the static function. I meant to have put this on one of the member functions. I still recommend passing in a cancellation token for the static function though.

Alex-MSFT: I'll put one on the nativeXR session as a private, and pass it in like I'm using the session's runtime scheduler here, and in the destructor set the cancellation.

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
Alex-MSFT: Added call to DeleteLocalRef

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
syntheticmagus: 1) Sounds good to me. Can you add an issue to keep track of this, at least to this repo and perhaps to Arcana.cpp too? This workaround creates ownership ambiguity at the contract level, so let's just not lose track of it.

2) Makes sense.

3) That seems to be true for the copy constructor, but not for the move operator; from my quick tests, `struct Foo { std::unique_ptr<int> Bar{}; }` is still a moveable type. If we don't need the `Session` to be moveable, can we add back former line 138, or replace it with `Session(Session&&) = delete;`?

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
syntheticmagus: Same as above.

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
Alex-MSFT: We actually do need the session to be movable now to move the result passed back from CreateAsync to the NativeXR session, so I'm going to leave #3 as is.

I managed to get the constructor private by using the explicit std::shared_ptr function rather than make_share, so #2 should be fixed now.

Dependencies/xr/Include/XR.h
Alex-MSFT: Fixed by using shared_ptr directly instead of make_shared.

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
Alex-MSFT: done

PR #264, opened 2020-05-15
[ARCore] Clear color/depth texture each frame, and explicitly present to the display

ryantrem:
While adding XR support into the React Native integration layer, I found that the ARCore XR implementation did not work in this context. After some debugging, I found a couple issues:
1. Babylon intentionally does not clear the color texture each frame because whether or not it should be cleared is a detail of a specific XR implementation. For example, OpenXR provides a color texture that already contains the current camera frame, so you just draw your scene on top of it without clearing. In the case of ARCore, we use two separate textures - one for the camera frame, and one for the Babylon scene. So in this case, the ARCore XR implementation should clear the color texture each frame before Babylon draws the scene onto it.
1. When Babylon is rendering to an offscreen frame buffer (the color render texture mentioned above), it is not presenting to the display, and so there is no call to `eglSwapBuffers`. This seems expected (since `eglSwapBuffers` basically means present to the display), so the ARCore XR implementation now explicitly calls `eglSwapBuffers` since it is an implementation detail that it presents to the current surface.

When running in the regular Babylon Native Playground app, the above two are not needed because bgfx is internally clearing and presenting, but this behavior should not be different between the regular playground and the react native playground. We need to figure out why this is happening and fix it, but I want to do this separately from this change. What that means is that if I leave the added call to `eglSwapBuffers`, it breaks the regular playground app but fixes the react native playground app. If I remove it, then it works in the regular playground app, but breaks the react native playground app. I don't really care which of the two are temporarily broken until we figure out the underlying issue, so if you care, speak up.

For clearing the texture, there is an OpenGL function to clear a texture directly, but it is not available on Android (as far as I can tell). So instead, I create another frame buffer that the color texture can be bound to specifically for the purpose of clearing it (by binding the frame buffer and calling `glClear`). According to "the internet" this is a cheap operation. 😁

### DIFF
--- a/Dependencies/xr/Source/ARCore/XR.cpp
+++ b/Dependencies/xr/Source/ARCore/XR.cpp
@@ -483,6 +483,8 @@ namespace xr
                 glDrawArrays(GL_TRIANGLE_STRIP, 0, VERTEX_COUNT);
 
                 // Present to the screen
+                // NOTE: For a yet to be determined reason, bgfx is also doing an eglSwapBuffers when running in the regular Android Babylon Native Playground playground app.
+                //       The "double" eglSwapBuffers causes rendering issues, so until we figure out this issue, comment out this line while testing in the regular playground app.
                 eglSwapBuffers(eglGetCurrentDisplay(), eglGetCurrentSurface(EGL_DRAW));
 
                 glUseProgram(0);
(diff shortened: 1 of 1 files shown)


Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -456,6 +481,10 @@ namespace xr
 
                 // Draw the quad
                 glDrawArrays(GL_TRIANGLE_STRIP, 0, VERTEX_COUNT);
+
+                // Present to the screen
+                eglSwapBuffers(eglGetCurrentDisplay(), eglGetCurrentSurface(EGL_DRAW));
```
Alex-MSFT, line 486: Can you add a note that this breaks the non-react-native implementation, so that we can easily find and swap it out when working locally?

PR #269, opened 2020-05-20
warning as error

CedricGuillemet:
Disable warnings for 3rd party dependencies.
Warning as error with pedantic and extra warnings

 - [x] The repo for base-n is my own fork. I'll change it to the original one once my PR is merged.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1187,7 +1187,7 @@ namespace Babylon
         const auto texture = info[0].As<Napi::External<TextureData>>().Data();
         const auto value = info[1].As<Napi::Number>().Uint32Value();
 
-        texture->AnisotropicLevel = value;
+        texture->AnisotropicLevel = static_cast<uint8_t>(value);
```
ryantrem, line 1190: For cases where the static cast is “narrowing” to a smaller or signed data type, we should consider using gel::narrow instead of static_cast. If we ever have a narrowing bug, this would make it much easier to diagnose.

CedricGuillemet: WIP

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1187,7 +1187,7 @@ namespace Babylon
         const auto texture = info[0].As<Napi::External<TextureData>>().Data();
         const auto value = info[1].As<Napi::Number>().Uint32Value();
 
-        texture->AnisotropicLevel = value;
```
ryantrem, line 1190: I think we’d want narrow rather than narrow_cast. narrow_cast just does a static_cast, but narrow does runtime checks for data loss. @bghgary thoughts?

bghgary: The only difference between narrow and narrow_cast is whether it throws or not if the value doesn't fit. My opinion is that since this is an implementation detail of the engine, we don't need a runtime check. Is there something that asserts?

Also, we should try to be more consistent in our usage of static_cast vs gsl casts.

ryantrem: @bghgary do you mean runtime asserts? Ultimately that would just end up with a runtime exception anyway, right? We can't do compile time asserts since we don't know these values at compile time.

bghgary: Yes, but I mean debug asserts, so that it doesn't impact perf.

ryantrem: 🤷🏼‍♂️

CedricGuillemet: Using static_cast for now. replacing some with gsl::narrow_cast will be done in another PR.

PR #270, opened 2020-05-20
V8 for Win32 fixes

syntheticmagus:


Apps/Playground/CMakeLists.txt
```diff
@@ -184,4 +184,5 @@ set_property(TARGET Playground PROPERTY FOLDER Apps)
 source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR}/../BabylonScripts PREFIX Scripts FILES ${BABYLONSCRIPTS})
 source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${SCRIPTS})
 source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${SOURCES})
+set_property(TARGET Playground PROPERTY VS_DEBUGGER_WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```
bghgary, line 187: Why do you need this?

Core/AppRuntime/Source/AppRuntimeWin32.cpp
```diff
@@ -7,12 +7,20 @@
...
+    }
+
     void AppRuntime::RunPlatformTier()
     {
         HRESULT hr = CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);
         assert(SUCCEEDED(hr));
         auto coInitScopeGuard = gsl::finally([] { CoUninitialize(); });
 
-        RunEnvironmentTier();
+        char filename[FILENAME_BUFFER_SIZE];
```
bghgary, line 21: do we need to do this for uwp?

syntheticmagus: It seems not, presumably because of the way UWP apps are run differently in the debugger versus Win32 apps.

Apps/Playground/CMakeLists.txt
syntheticmagus: Don't, forgot to take it out. Should be gone now.

Core/AppRuntime/Source/AppRuntimeWin32.cpp
bghgary: It could be because it's sandboxed.

Core/AppRuntime/Source/AppRuntimeWin32.cpp
```diff
@@ -7,12 +7,20 @@
...
+        DWORD result = GetModuleFileNameA(nullptr, filename, FILENAME_BUFFER_SIZE);
```
bghgary, line 22:
```suggestion
        DWORD result = GetModuleFileNameA(nullptr, filename, std::size(filename));
```

Perhaps don't need a constexpr for buffer size anymore.

PR #273, opened 2020-05-22
Set matrix for view to not be view matrix

syntheticmagus:
Apparently it wants to be a world-space matrix instead of a view-space one.


ryantrem:
Assuming this has not been tested with Android/ARCore yet?

ryantrem:
Tested on Android/ARCore and seems to be working as expected.

Issue #219, opened 2020-04-10
Add a JavaScript way of loading scripts


bghgary:
Perhaps replace Tools.LoadScript with a custom implementation for Babylon Native.

syntheticmagus:
My understanding from a long time ago was that the plan for this was to get the text of the desired JS file via network request, then just `eval` it. Does this solution still seem viable? If so, is this a change we should make to `Tools.LoadScript` on the Babylon.js side?

CedricGuillemet:
Yes, I'm starting to think it would be better to do the work only on the js side. I'll try to do that instead.

CedricGuillemet:
Fixed with this PR : https://github.com/BabylonJS/Babylon.js/pull/8998 Closing.

Issue #214, opened 2020-04-10
Investigate web workers / web assembly


bghgary:
Make sure web workers work automatically in all the JS engines that we support.

One way to test is:
- [ ] Try a glTF with Draco and see if it works (will test both web workers and web assembly) (draco compression will fallback if these things are not available, so make sure it is actually being used)

CedricGuillemet:
should be done in conjunction of https://github.com/BabylonJS/BabylonNative/issues/219
In order to load draco.js script from js script.

CedricGuillemet:
Win32/Chakra, this test is false:
https://github.com/BabylonJS/Babylon.js/blob/5fb3378450f974fbb0d105c8e48f49b55fc9ce7f/src/Meshes/Compression/dracoCompression.ts#L305

bnolan:
Is this under feature under development? Our project has several web workers and we're working out how we can port them to BN.

bghgary:
Not currently, no. We do intend on adding this though.

bnolan:
Ok awesome. We'll update this issue if we can contribute this code. Any gotchas we should look out for while implementing this? We'll basically spin up a std::thread and put an instance of the javascript engine in it and then use a mutex to push messages back and forward.

bghgary:
I'm not exactly sure how to implement yet. A simple `std::thread` might be right or maybe it should be using a threadpool. As for message passing, a concurrent queue is more efficient. I suspect it will be useful to use [arcana.cpp](https://github.com/microsoft/arcana.cpp) to do some of this, which, for example, has a [concurrent queue implementation](https://github.com/microsoft/arcana.cpp/blob/master/Source/Shared/arcana/threading/blocking_concurrent_queue.h).

bghgary:
Closing for now. Reopen if we want to support these.

PR #274, opened 2020-05-26
script loader

CedricGuillemet:
This is WIP for https://github.com/BabylonJS/BabylonNative/issues/219 and https://github.com/BabylonJS/BabylonNative/issues/214 because PG needs a way to load .js

Using polyfills as loading JS use DOM.
Once this PR is ine, I'll do a PR in bjs to override ```BABYLON.Tools.GetAbsoluteUrl``` and ```BABYLON.Tools.LoadScript``` from nativeengine init as it seems the best way to do it after talking to @sebavan .

the NativeScript polyfill uses the ScriptLoader. I'm running into 1 c++ issue:
I need to call onSuccess and onError JS function to loadScript. I tried many things with lamba, std::move, std::unique,... to push to Napi::FunctionReference to that function without success. I used a new/delete.  https://github.com/BabylonJS/BabylonNative/pull/274/files#diff-827872a1f803d4a7fc75b6c63e938a2cR42

### DIFF
--- a/Core/ScriptLoader/Include/Babylon/ScriptLoader.h
+++ b/Core/ScriptLoader/Include/Babylon/ScriptLoader.h
@@ -23,7 +23,7 @@ namespace Babylon
 
         ~ScriptLoader();
 
-        void LoadScript(std::string url);
+        void LoadScript(std::string url, Napi::FunctionReference* onSuccess = nullptr, Napi::FunctionReference* onError = nullptr);
         void Eval(std::string source, std::string url);
 
     private:
--- a/Polyfills/Script/Source/Script.cpp
+++ b/Polyfills/Script/Source/Script.cpp
@@ -1,0 +1,56 @@
+#include "Script.h"
+#include <Babylon/JsRuntime.h>
+#include <functional>
+#include <sstream>
+
+#include <arcana/threading/task.h>
+#include <arcana/threading/task_schedulers.h>
+
+
+namespace Babylon::Polyfills::Internal
+{
+    void Script::CreateInstance(Napi::Env env, DispatchFunctionT dispatchFunction)
+    {
+        Napi::HandleScope scope{env};
+
+        Napi::Function func = ParentT::DefineClass(
+            env,
+            "Script",
+            {
+                ParentT::InstanceMethod("loadScript", &Script::LoadScript),
+            });
+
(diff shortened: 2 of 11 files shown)


bghgary:
> Once this PR is ine, I'll do a PR in bjs to override BABYLON.Tools.GetAbsoluteUrl and BABYLON.Tools.LoadScript from nativeengine init as it seems the best way to do it after talking to @sebavan .

This isn't going to work. Overriding is fine for legacy js files, but it doesn't work with ES6 modules. We will have to come up with a different way. See #268.

Polyfills/Script/Source/Script.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+    void Script::CreateInstance(Napi::Env env, DispatchFunctionT dispatchFunction)
+    {
+        Napi::HandleScope scope{env};
+
+        Napi::Function func = ParentT::DefineClass(
+            env,
+            "Script",
+            {
+                ParentT::InstanceMethod("loadScript", &Script::LoadScript),
```
syntheticmagus, line 20: Does this use case go away if we implement JS script loading as discussed in my comment on #219?

Core/ScriptLoader/Include/Babylon/ScriptLoader.h
```diff
@@ -23,7 +23,7 @@ namespace Babylon
 
         ~ScriptLoader();
 
-        void LoadScript(std::string url);
+        void LoadScript(std::string url, Napi::FunctionReference* onSuccess = nullptr, Napi::FunctionReference* onError = nullptr);
```
syntheticmagus, line 26: This seems a little odd to me.
1. Having `Napi::FunctionReference`s as the callbacks doesn't seem right for a Core component which targets C++ consumers. Could these just be `std::function`s?
2. Is this call allowed to fail? I think that might not be allowed; and if not, can we just leave out the `onError` callback?
3. Also, if `LoadScript` isn't allowed to fail (or, more accurately, if it's not allowed to _survive_ failure), I think the same effective thing as an `onSuccess` callback could be achieved by giving `ScriptLoader` its own `Dispatch()` method, which would simply use the same mechanisms as are already within `ScriptLoader` to execute a given `function` linearly alongside all other `ScriptLoader` calls, similar to pre-existing patterns.

CedricGuillemet:
I will replace this plugin with JS code in bjs. Closing.

PR #276, opened 2020-05-28
async screenshot

### LINKED ISSUE #155: Revisit bgfx screenshot request
Update bgfx for userparam instead of file path string.
see discussion here : https://github.com/BabylonJS/BabylonNative/pull/60#discussion_r374420778

CedricGuillemet:
fixes #155 
as screenshot and bgfx callback are in order, I can use a std::vector for FIFO screenshot/callback association.


Plugins/NativeEngine/Source/BgfxCallback.cpp
```diff
@@ -79,12 +87,31 @@ namespace Babylon
...
+                *bitmap++ = ptr[px * 4 + 2];
+                *bitmap++ = ptr[px * 4 + 1];
+                *bitmap++ = ptr[px * 4 + 0];
+                *bitmap++ = ptr[px * 4 + 3];
+            }
         }
+
+        JsRuntime& runtime{ JsRuntime::GetFromJavaScript(env) };
+
+        runtime.Dispatch([this, array](Napi::Env) {
```
bghgary, line 110:
```suggestion
        runtime.Dispatch([this, array{std::move(array)}](Napi::Env) {
```

Plugins/NativeEngine/Source/BgfxCallback.h
```diff
@@ -26,5 +28,8 @@ namespace Babylon
         void captureEnd() override;
         void captureFrame(const void* _data, uint32_t _size) override;
         void trace(const char* _filePath, uint16_t _line, const char* _format, ...);
+
+        std::mutex m_ssCallbackAccess;
+        std::vector<Napi::FunctionReference> m_screenshotCallbacks;
```
bghgary, line 33: Consider using arcana::blocking_concurrent_queue

CedricGuillemet: I tried using arcana::blocking_concurrent_queue but to be able to get an env for creating the uint8array, I had to pop from the queue. Once poped, I can't send the FunctionReference to the lambda because of unmovable members in std::function. I didn't want to create a temporary array that I would pass to the lambda and then do a uint8array in the dispatched lambda. So, I used a std::queue with a mutex.

bghgary: You can use Napi::ArrayBuffer::New with an external buffer instead. That way you can create the buffer using normal C++ and then only create the Napi types inside the dispatch.

CedricGuillemet: I have to unswizzle the buffer and the data pointer that bgfx gives me is unmapped outside the screenshot function.

PR #284, opened 2020-06-04
curl Multinterface/async transfer

### LINKED ISSUE #232: [Linux] Curl and multi interface
Making curl transfers seemed easy at first:
- create a multi handle
- create curl easy using ```curl_easy_init```
- add those easy handle to the multi handle
The issue I got is how to know when the transfer is complete. Reading official doc and samples, it looks like you have to poll, get messages and check created easy handles.
Other way to do is to monitor the sockets using multi_socket but it appears to be polling as well.
Do you remember anything about that @syntheticmagus  ?

CedricGuillemet:
Is it safe to have ```arcana::task_completion_source``` as a class member?
Is there a way to exit the thread loop/know when the runtime is stopping?

fixes #232


Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -76,6 +75,80 @@ namespace UrlLib
...
+                    int numfds;
+                    int msgsLeft;
+                    curl_multi_wait(m_multiHandle, NULL, 0, 1000, &numfds);
+                    {
+                        std::lock_guard<std::mutex> guard(m_mutex);
+                        curl_multi_perform(m_multiHandle, &stillRunning);
+                    }
+                    
+                    // See how the transfers went
+                    CURLMsg *m = NULL;
```
bghgary, line 124:
```suggestion
                    CURLMsg* m = nullptr;
```

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -76,6 +75,80 @@ namespace UrlLib
...
+                    while ((m = curl_multi_info_read(m_multiHandle, &msgsLeft))) 
+                    {
+                        if (m->msg == CURLMSG_DONE) 
+                        {
+                            CURL *handle = m->easy_handle;
```
bghgary, line 129:
```suggestion
                            CURL* handle = m->easy_handle;
```

Fix everywhere

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -76,6 +75,80 @@ namespace UrlLib
...
+                            UrlRequest::Impl *request;
+                            curl_easy_getinfo(handle, CURLINFO_PRIVATE, &request);
+                            if (m->data.result == CURLE_OK) 
+                            {
+                                long codep;
+                                curl_easy_getinfo(handle, CURLINFO_RESPONSE_CODE, &codep);
+
+                                // codep 0 for file access
+                                if (codep == 0|| codep == 200)
```
bghgary, line 138:
```suggestion
                                if (codep == 0 || codep == 200) 
```

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -76,6 +75,80 @@ namespace UrlLib
...
+            }
+
+        private:
+            CURLM* m_multiHandle;
+            // can't add new handle when curl is inside a callback. To be certain, a mutex is used when performing calls.
+            std::mutex m_mutex;
+
+            void Loop()
+            {
+                while (true)
```
bghgary, line 112: You should probably use an arcana dispatcher for this instead of polling for things to do. Then you can use arcana to create tasks on the curl thread and pass data to it (like your question about task completion source). Maybe we can go over this offline.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -76,6 +75,80 @@ namespace UrlLib
...
+        {
+            CurlMulti()
+            : m_multiHandle(curl_multi_init())
+            {
+                assert(m_multiHandle != NULL);
+                curl_multi_setopt(m_multiHandle, CURLMOPT_MAX_TOTAL_CONNECTIONS, 200);
+                curl_multi_setopt(m_multiHandle, CURLMOPT_MAX_HOST_CONNECTIONS, 6L);
+                std::thread([this](){
+                    Loop();
+                }).detach();
```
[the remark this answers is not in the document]
CedricGuillemet: Added cancellationsource

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
CedricGuillemet: Curl multi interface handles everything once it's in its thread. The 'task' is added with ```curl_multi_add_handle```.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -76,6 +75,80 @@ namespace UrlLib
...
 
     private:
+        struct CurlMulti
+        {
+            CurlMulti()
+            : m_multiHandle(curl_multi_init())
+            {
+                assert(m_multiHandle != NULL);
+                curl_multi_setopt(m_multiHandle, CURLMOPT_MAX_TOTAL_CONNECTIONS, 200);
+                curl_multi_setopt(m_multiHandle, CURLMOPT_MAX_HOST_CONNECTIONS, 6L);
```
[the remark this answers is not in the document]
CedricGuillemet: What do you mean?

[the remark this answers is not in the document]
bghgary: [`curl_multi_setopt`](https://curl.haxx.se/libcurl/c/curl_multi_setopt.html) returns an error code. Probably should check the result?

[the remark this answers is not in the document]
bghgary: Meh, nevermind. It's just for setting options. Generally speaking, I usually flag functions that return error codes but are left unchecked.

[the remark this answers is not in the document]
bghgary: Some of the calls below like `curl_multi_wait` and `curl_multi_perform` can maybe use a check though.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -37,22 +37,21 @@ namespace UrlLib
 
         arcana::task<void, std::exception_ptr> SendAsync()
         {
-            return arcana::make_task(arcana::threadpool_scheduler, m_cancellationSource, [this]()
+            m_taskCompletionSource = {};
```
bghgary, line 40: If multiple requests are active at the same time, does it work?

CedricGuillemet: I've tested a loop with 10 simultaneous HTTPS requests and it worked fine.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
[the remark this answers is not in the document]
CedricGuillemet: Done.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -105,19 +184,14 @@ namespace UrlLib
...
-
-                curl_easy_cleanup(curl);
-                m_statusCode = UrlStatusCode::Ok;
+                curl_easy_setopt(curl, CURLOPT_PRIVATE, this);
+                s_curlMulti.AddHandle(curl);
             }
         }
 
+        static inline CurlMulti s_curlMulti{};
         arcana::cancellation_source m_cancellationSource{};
```
bghgary, line 193: Hmm, this isn't used anywhere.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
bghgary: I think what we need to do is associate some information with the CURL handle. This includes the cancellation source and the task completion source. Can be a separate issue though.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -76,6 +75,86 @@ namespace UrlLib
...
+                            if (m->data.result == CURLE_OK) 
+                            {
+                                long codep;
+                                curl_easy_getinfo(handle, CURLINFO_RESPONSE_CODE, &codep);
+
+                                // codep 0 for file access
+                                if (codep == 0 || codep == 200) 
+                                {
+                                    request->m_statusCode = UrlStatusCode::Ok;
+                                }
```
bghgary, line 147: What about failures? What happens if the url returns 404 or something else?

bghgary (approved):
Can check in as in, but maybe file an issue to track some of the remaining work in the implementation.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
CedricGuillemet: I've open this issue : https://github.com/BabylonJS/BabylonNative/issues/427

PR #289, opened 2020-06-08
Implement XRAnchor and ARCore polyfill

Alex-MSFT:
This change implement [XRAnchor ](https://immersive-web.github.io/anchors/) and ARCore polyfills as noted in (#230) 

Main changes:
- Introduced XRAnchor and implemented native polyfill for ArCore
- Fixed some implementation details to make XRReferenceSpaces actually used for XRFrame.GetPose()
- Fixed the constructor of XRReferenceSpace to actually store its internal transform, so that we can actually compare anchor space to the default reference space.


Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -671,33 +673,134 @@ namespace xr
...
+            // First translate the passed in pose to something usable by ArCore.
+            ArPose* arPose;
+            float rawPose[7];
+            PoseToRaw(rawPose, pose);
+            ArPose_create(session, rawPose, &arPose);
+
+            // Create the actual anchor. If a trackable was passed in (from a hit test result) create the
+            // anchor against the tracakble. Otherwise create it against the session.
+            ArAnchor* arAnchor;
+            auto trackableObj = (ArTrackable*) trackable;
```
ryantrem, line 716:
```suggestion
            auto trackableObj = reinterpret_cast<ArTrackable*>(trackable);
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -671,33 +673,134 @@ namespace xr
...
+
+            // Store the anchor the vector tracking currently allocated anchors, and pass back the result.
+            arCoreAnchors.push_back(arAnchor);
+            return {pose, arAnchor};
+        }
+
+        void UpdateAnchor(xr::Anchor& anchor)
+        {
+            // First check if the anchor still exists, if not then mark the anchor as no longer valid.
+            auto arAnchor = (ArAnchor*) anchor.NativeAnchor;
```
ryantrem, line 737:
```suggestion
            auto arAnchor = reinterpret_cast<ArAnchor*>(anchor.NativeAnchor);
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -671,33 +673,134 @@ namespace xr
...
+            if (arAnchor == nullptr)
+            {
+                anchor.IsValid = false;
+                return;
+            }
+
+            ArTrackingState trackingState;
```
ryantrem, line 744:
```suggestion
            ArTrackingState trackingState{};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -671,33 +673,134 @@ namespace xr
...
+            }
+        }
+
+        void DeleteAnchor(xr::Anchor& anchor)
+        {
+            // If this anchor has not already been deleted, then detach it from the current AR session,
+            // and clean up its state in memory.
+            if (anchor.NativeAnchor != nullptr)
+            {
+                auto arAnchor = (ArAnchor*) anchor.NativeAnchor;
```
ryantrem, line 768:
```suggestion
                auto arAnchor = reinterpret_cast<ArAnchor*>(anchor.NativeAnchor);
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -748,6 +851,28 @@ namespace xr
 
             ActiveFrameViews[0] = {};
         }
+
+        void PoseToRaw(float rawPose[], Pose& pose)
```
ryantrem, line 855:
```suggestion
        void PoseToRaw(float rawPose[], const Pose& pose)
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -748,6 +851,28 @@ namespace xr
...
+            rawPose[0] = pose.Orientation.X;
+            rawPose[1] = pose.Orientation.Y;
+            rawPose[2] = pose.Orientation.Z;
+            rawPose[3] = pose.Orientation.W;
+            rawPose[4] = pose.Position.X;
+            rawPose[5] = pose.Position.Y;
+            rawPose[6] = pose.Position.Z;
+        }
+
+        void RawToPose(float rawPose[], Pose& pose)
```
ryantrem, line 866:
```suggestion
        void RawToPose(const float rawPose[], Pose& pose)
```

Plugins/NativeInput/Source/NativeInput.cpp
```diff
@@ -11,7 +11,7 @@ namespace Babylon::Plugins
     {
         constexpr auto JS_NATIVE_INPUT_NAME = "_nativeInput";
 
-        constexpr auto POINTER_BASE_DEVICE_ID = "Pointer";
+        //constexpr auto POINTER_BASE_DEVICE_ID = "Pointer";
```
ryantrem, line 14: This looks like it was left over from the earlier contract that used string ids? Remove?

Plugins/NativeInput/Source/NativeInput.cpp
```diff
@@ -85,7 +85,7 @@ namespace Babylon::Plugins
             SetInputState(DeviceType::Touch, pointerId, inputIndex, 0, deviceInputs);
 
             // If all "buttons" are up, then remove the device (e.g. device "disconnected").
-            for (int32_t index = 0; index < deviceInputs.size(); index++)
+            for (uint index = 0; index < deviceInputs.size(); index++)
```
ryantrem, line 88: Why did you need to make this change? Also should it be `uint_32t`?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -536,6 +551,14 @@ namespace Babylon
                 Update(space, true);
             }
 
+            xr::Pose GetNativePose()
+            {
+                auto position = m_position.Value();
+                auto orientation = m_orientation.Value();
+                return {{position.Get("x").ToNumber().FloatValue(), position.Get("y").ToNumber().FloatValue(), position.Get("z").ToNumber().FloatValue()},
+                    {orientation.Get("x").ToNumber().FloatValue(), orientation.Get("y").ToNumber().FloatValue(), orientation.Get("z").ToNumber().FloatValue(), orientation.Get("w").ToNumber().FloatValue()}};
```
ryantrem, line 559:
```suggestion
                return
                {
                    {position.Get("x").ToNumber().FloatValue(), position.Get("y").ToNumber().FloatValue(), position.Get("z").ToNumber().FloatValue()},
                    {orientation.Get("x").ToNumber().FloatValue(), orientation.Get("y").ToNumber().FloatValue(), orientation.Get("z").ToNumber().FloatValue(), orientation.Get("w").ToNumber().FloatValue()}
                };
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -924,24 +953,43 @@ namespace Babylon
                 return info.Env().Global().Get(JS_CLASS_NAME).As<Napi::Function>().New({info[0]});
```
ryantrem, line 953: Nit: Call your new overload?
```suggestion
                return New(info.Env(), info[0]);
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -924,24 +953,43 @@ namespace Babylon
...
+                    {
+                        // TODO: Actually support the different types of reference spaces.
+                        const auto referenceSpaceType = info[0].As<Napi::String>().Utf8Value();
+                        assert(referenceSpaceType == XRReferenceSpaceType::UNBOUNDED ||
+                            referenceSpaceType == XRReferenceSpaceType::VIEWER);
+                    }
+                    else
+                    {
+                        /*auto* transform = XRRigidTransform::Unwrap(info[0].As<Napi::Object>());
+                        assert(transform != nullptr);*/
```
ryantrem, line 976: ?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1116,12 +1245,42 @@ namespace Babylon
...
+            void DeleteNativeAnchor(xr::Anchor& nativeAnchor)
+            {
+                m_frame->DeleteAnchor(nativeAnchor);
             }
 
         private:
             const xr::System::Session::Frame* m_frame{};
             Napi::ObjectReference m_jsXRViewerPose{};
             XRViewerPose& m_xrViewerPose;
+            std::vector<Napi::Value> m_trackedAnchors;
```
ryantrem, line 1283:
```suggestion
            std::vector<Napi::Value> m_trackedAnchors{};
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1053,25 +1172,33 @@ namespace Babylon
...
+
+            void SetXRFrame(XRFrame* frame)
+            {
+                m_frame = frame;
             }
 
         private:
+            // The hit hit result, which contains the pose in default AR Space, as well as the native entity.
+            xr::HitResult m_hitResult;
+            XRFrame* m_frame;
```
ryantrem, line 1189:
```suggestion
            xr::HitResult m_hitResult{};
            XRFrame* m_frame{};
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -950,6 +998,76 @@ namespace Babylon
...
+
+                Napi::Object napiSpace = XRReferenceSpace::New(info.Env(), napiTransform);
+                return napiSpace;
+            }
+
+            // Forward declaration of delete, as this relies on the XRFrame implementation.
+            void Delete(const Napi::CallbackInfo& info);
+
+            // The native anchor which holds the current position of the anchor, and the native ref to the anchor.
+            xr::Anchor m_nativeAnchor;
```
ryantrem, line 1067:
```suggestion
            xr::Anchor m_nativeAnchor{};
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -950,6 +998,76 @@ namespace Babylon
...
+
+            // Our native anchor.
+            XRFrame* m_frame;
```
ryantrem, line 1070:
```suggestion
            XRFrame* m_frame{};
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -950,6 +998,76 @@ namespace Babylon
 
                 return XRReferenceSpace::New(info);
             }
+
+            Napi::ObjectReference m_jsTransform;
```
ryantrem, line 1002:
```suggestion
            Napi::ObjectReference m_jsTransform{};
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1163,25 +1332,82 @@ namespace Babylon
...
+
+            Napi::Value GetTrackedAnchors(const Napi::CallbackInfo& info)
+            {
+                // Create a JavaScript set to store all currently tracked anchors.
+                Napi::Object anchorSet = info.Env().Global().Get("Set").As<Napi::Function>().New({});
+
+                // Loop over the list of tracked anchors, and add them to the set.
+                for (Napi::Value napiValue : m_trackedAnchors)
+                {
+                    anchorSet.Get("add").As<Napi::Function>().Call(anchorSet, {napiValue});
```
ryantrem, line 1368: Can this just be:
```c++
anchorSet.Get("add").As<Napi::Function>()(anchorSet, {napiValue});
```

Dependencies/xr/Include/XR.h
```diff
@@ -133,7 +146,10 @@ namespace xr
                 Frame(System::Session::Impl&);
                 ~Frame();
 
-                void GetHitTestResults(std::vector<Pose>&, Ray) const;
+                void GetHitTestResults(std::vector<HitResult>&, Ray) const;
+                Anchor CreateAnchor(Pose, void*) const;
```
ryantrem, line 150: Seeing `void*` here is a bit confusing, particularly since there is no parameter name. Since this is basically an opaque pointer to a platform native type, maybe we should give this type a name, e.g.
```c++
using NativeTrackable = void; // -or-
using NativeTrackablePtr = void*; // -or-
typedef struct NativeTrackable_ NativeTrackable; // ARCore does something like this, but I don't really get what it's doing.
```

Plugins/NativeInput/Source/NativeInput.cpp
Alex-MSFT: I had to comment this out as it was causing a warning during BabylonReactNative compile (oddly not during regular BabylonNative compile), and since we have warnings as errors turned on this was breaking the build.  I didn't remove it because I was not sure if it was planned to be used in the future, but with your vote I'll go ahead and take it out.

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: Good point!

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: This caused a build break on as the transform variable is unused when compiling retail.  I will just remove the code.

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: I will give that a try, it was failing when I had tried just:
`anchorSet.Get("add").As<Napi::Function>()({napiValue});`

As it was not referencing back to anchorSet, and landed on the .call with the ref to the object, but looking at the operator what you suggested should work as well.

Dependencies/xr/Include/XR.h
Alex-MSFT: Good idea the vanilla void* makes it pretty confusing in the context of xr.cpp.

Plugins/NativeInput/Source/NativeInput.cpp
Alex-MSFT: When I brought the code into BabylonReactNative it threw this as a warning.  Since we have warnings as errors turned on this was a build break only in the context of BabylonReactNative.  The type for vector.size seems to be size_type, which is dependent on the allocator, but is always unsigned.  Looks like it usually defaults to size_t so I'll use that instead.

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: Ahh nevermind the operator doesn't handle the receiver properly, will have to leave this using .Call.

    Value operator ()(const std::initializer_list<napi_value>& args) const;
vs:
    Value Call(napi_value recv, const std::initializer_list<napi_value>& args) const;

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: Actually on second look this would not cover the case where we pass in a napi::string, which is different from an object.

Dependencies/xr/Include/XR.h
```diff
@@ -46,6 +46,12 @@ namespace xr
         } Orientation;
     };
 
+    struct HitResult
+    {
+        Pose Pose{};
+        void* NativeEntity{};
```
ryantrem, line 52: Define a NativeEntityPtr for this as well?

PR #291, opened 2020-06-10
refactor includes to be framework compliant for Apple

CedricGuillemet:
When using a Framework on Apple platform, the header must be included this way by the user:
```
#include <frameworkname/header.h>
```
This brings issues when let's say ```JsRuntime.h``` includes ```<napi/env.h>```
So, any plugin header definition shall not include anything except STL.
This introduces 2 things:
- .cpp will have to include napi headers when needed (in plugins or in the app)
- napi.h is now a meta header that includes various napi related things. In the framework header, it is replaced by another napi.h with modified path to support the framework hierarchy (in an upcoming PR)


bghgary:
I don't understand what a "Framework" is. Can you maybe explain that a bit? Standard coding practices for public headers is that they are self-contained, i.e. including a header doesn't require including another header first.

bghgary:
I'm also a bit wary of the change to `napi.h`. These headers come from node.js. Changing them will make it harder to take updates.

CedricGuillemet:
A good definition of Frameworks can be found here : https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WhatAreFrameworks.html
Yes, the work on headers is annoying. Another possible solution would be to hide napi::env as an opaque void*. I don't know all the issues it would bring. I think we will have the exact same issue with Android package. With a .dylib or .a we will have to provide headers.

bghgary:
Let's chat when offline to make sure we are on the same page and come up with a solution.

CedricGuillemet:
Packages not a priority for now. Closing.

PR #293, opened 2020-06-12
Integration + iOS.Framework

CedricGuillemet:
This is part 2 of ios packaging. Part 1 is here https://github.com/BabylonJS/BabylonNative/pull/291
The goal of this PR is to add a new target that builds a Framework usable for Xcode directly (drag and drop .framework in the source tree) or with a cocoapod.

The artifact contains this:
image
- Headers from the Include directories of plugins, core, polyfills, and napi
- .js resources
- a big BabylonNative static library that is the concatenation of all .a (plugins, dependencies, ...)
Xcode automaticaly recognise the format of this archive so end user doesn't have to set path for includes, name of lib,...

When building the framework, napi.h is replaced with a special napi.h with relative to framework path. See Part1.

The archive is published as a build artifact. Part3 will be cocoapods and binary feed.
It's not possible to use this framework before the Part1 is merged.

Once you have downloaded/installed the framework in the Xcode project, just add some includes.
image

Some words on cocoapods, I'll elaborate more in the next PR. A pod, basically, needs to files :  a podspec and a pod. The podspec specifies what the pod is composed of. It's used for publishing, and consuming the pod. a pod file contains the list and path to podspec that will be integrated in a named project.

BabylonNative.podspec:
```
Pod::Spec.new do |s|  
    s.name              = 'BabylonNative'
    s.version           = '0.0.4'
    s.summary           = 'Babylon Native.'
    s.homepage          = 'https://github.com/CedricGuillemet'

    s.author            = { 'ceguille' => 'ceguille@microsoft.com' }
    s.license           = { :type => 'MIT' }

    s.platform          = :ios
    s.source            = { :http => 'file:///Users/cedricguillemet/dev/podtest/BabylonNative/buildios/Integration/Debug-iphonesimulator/BabylonNative.framework.zip'}
    s.resource = '**/*.js'
    s.ios.deployment_target = '12.0'
    s.ios.vendored_frameworks = 'BabylonNative.framework'
end  
```
Two most important things here:
the s.source and s.resource. s.resource is needed so xcode can merge the resource in the app bundle.
s.source here points to a local directory that I'm using for my tests.  This path can points to a GitHub release (so we need to update it when we do a GitHub). My task now is to see if it can point to a CI Feed: when merging into master, a build is triggered and it's the artifact from that master version that I'd like to put in the feed. User would have to do a 'pod update' to update to the latest good version of BabylonNative.
It also possible to point to the GitHub and trigger a local build when installing pod.
This file will be at the root of the repo.

End user will only have to update its own pod file like this:
```
target 'name_of_app' do
   pod 'BabylonNative', :git => "https://github.com/BabylonJS/BabylonNative.git"
end
```
Once the pod is installed, load the workspace that cocoapods has generated.

For reactnative integration, it's possible to write its own podspec that will not use the resources from the framework (and use npm instead for Babylonjs).

It's possible to release the framework as an npm package. You then need a podspec to points to that framework (s.source = ".... /node_modules/BabylonNative/). You pod would point to your local podspec.

user would do :
```
npm install
pod install
```

### DIFF
--- a/Integration/iOS.Framework/CMakeLists.txt
+++ b/Integration/iOS.Framework/CMakeLists.txt
@@ -1,0 +1,145 @@
+set(BABYLONSCRIPTS
+    "../../Apps/BabylonScripts/babylon.glTF2FileLoader.js"
+    "../../Apps/BabylonScripts/babylon.max.js"
+    "../../Apps/BabylonScripts/babylonjs.materials.js"
+    "../../Apps/BabylonScripts/ammo.js"
+    "../../Apps/BabylonScripts/meshwriter.min.js"
+    "../../Apps/BabylonScripts/recast.js")
+
+set(SCRIPTS
+    "../../Apps/Playground/Scripts/experience.js"
+    "../../Apps/Playground/Scripts/playground_runner.js")
+
+set(SOURCES
+    "Source/Framework.cpp"
+    "Include/napi.h")
+
+set(RESOURCE_FILES ${STORYBOARD} ${BABYLONSCRIPTS} ${SCRIPTS})
+
+add_library(BabylonNative ${SOURCES} ${RESOURCE_FILES})
+warnings_as_errors(BabylonNative)
+
+target_compile_definitions(BabylonNative PRIVATE UNICODE)
+target_compile_definitions(BabylonNative PRIVATE _UNICODE)
+
+target_include_directories(BabylonNative PUBLIC "Source" ".")
+file(GLOB_RECURSE appruntime_headers ${CMAKE_CURRENT_SOURCE_DIR}/../Core/AppRuntime/Include/Babylon/*.h)
+file(GLOB_RECURSE jsruntime_headers ${CMAKE_CURRENT_SOURCE_DIR}/../Core/JsRuntime/Include/Babylon/*.h)
+file(GLOB_RECURSE scriptloader_headers ${CMAKE_CURRENT_SOURCE_DIR}/../Core/ScriptLoader/Include/Babylon/*.h)
+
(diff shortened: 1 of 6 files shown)


ryantrem:
Instead of pointing the podspec to a separate BabylonNative NPM package, we could include the framework *directory* in the @babylonjs/react-native package, and point `source` to `.`, and then just reference the framework folder (if I understand correctly). An example of this can be found here: https://medium.com/@aliakhtar_16369/distribute-framework-with-cocoapods-locally-and-remotely-4a39f95b3077 if you scroll down to the *Consume Framework Using Local CocoaPods* section.

CedricGuillemet:
'we could include the framework directory in the @babylonjs/react-native package', do you mean adding the prebuilt framework in the Github repo? Who would be responsible for building and pushing the changes? Having the framework in an npm allows the CI to build and push a package automatically while still having a podspec/pod in @babylonjs/react-native that points to a local directory (subdir in node_modules).

Integration/iOS.Framework/CMakeLists.txt
```diff
@@ -0,0 +1,146 @@
...
+            $<TARGET_FILE:NativeInput>
+            $<TARGET_FILE:napi>
+            $<TARGET_FILE:UrlLib>
+            $<TARGET_FILE:bgfx>
+            $<TARGET_FILE:bimg>
+            $<TARGET_FILE:bx>
+            $<TARGET_FILE:glslang>
+            $<TARGET_FILE:SPIRV>
+            $<TARGET_FILE:spirv-cross-core>
+            $<TARGET_FILE:spirv-cross-hlsl>
```
syntheticmagus, line 88: Is this used by iOS?

CedricGuillemet:
Packages not a priority for now. Closing.

PR #303, opened 2020-06-23
Fix regressions from new BabylonJS upgrade

Alex-MSFT:
BabylonJS internals updated Vector3 and Quaternion's member variable property names from x, y , z, w to _x, _y, _z, _w.  So we need to change the results from our rigid transform to match.  It should be noted that this actually does not match the WebXR spec, so it is possible this will regress again in the future.

getFrontPosition has been removed from the default camera, and is now exclusive to TargetCamera and requires an explicit target be set.  Removing usage of this in the playground app.


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -531,16 +531,16 @@ namespace Babylon
...
             {
                 auto position = m_position.Value();
-                position.Set("x", space.Pose.Position.X);
-                position.Set("y", space.Pose.Position.Y);
-                position.Set("z", space.Pose.Position.Z);
-                position.Set("w", 1.f);
+                position.Set("_x", space.Pose.Position.X);
+                position.Set("_y", space.Pose.Position.Y);
+                position.Set("_z", space.Pose.Position.Z);
+                position.Set("_w", 1.f);
```
syntheticmagus, line 537: I think this switch from `.x` to `._x` was part of the work @deltakosh did [to make more things properties](https://github.com/BabylonJS/Babylon.js/commit/3a9b56c6db9db2fa7673ee6494cdb72bc75f1b44#diff-e80c36351a868a7dc9c5ab90c2dabb09) so dirtiness (for things like matrices) would be easier to manage for performance's sake. Consequently, if this vector has a callback for when its properties change, setting the underlying numeric representation probably won't cause the callback to be called when it should. If this fixes the issues you were seeing, I think that's an indication that there isn't such a callback and that this is an alright stopgap for now; but can you add a comment inline and a GitHub issue to track this behavior so we don't forget to make this work correctly with properties in the future?

Alex-MSFT: Added a comment

PR #249, opened 2020-05-07
ARKit/OpenXR

CedricGuillemet:
No matter what I did, I'm unable to run it on my 6th gen ipad. It seems I'm the only 1 for now. If someone can take this over, it'd be awesome! The metal rendering part works as I was able to test it on the simulator. 
Getting camera pose and camera texture is done in the delegate
Rendering works like the Android version. The shader blends both texture and display the result using a triangle.

What's done

- ARKit init and session delegate
- metal pipeline and rendering
- cmake update for openxr ios

What's left to do

- test and debug on real device
- GetHitTestResults function

Note: The metal layer is needed for the swapchain. I used an extern butI think it's possible to use the polyfills::window instead. I don't know if it's worth the effort.

Issue #278, opened 2020-05-28
Support ARKit


ryantrem:
#305 includes the initial implementation of ARKit support.  This allows for entering/exiting an AR session and displaying a Babyon scene in an AR View.

Work remaining:
- [x] Hit testing support
- [x] Anchor support
- [x] Plane detection support
- [x] Handling orientation changes
- [x] Properly handle view re-use, and confirm correctness with React-Native.
- [x] Properly handle Landscape orientations in updateFoV
- [x] Simplify updateDisplayOrientedPose in xr.mm to correctly handle translating from camera space to world space.

PR #305, opened 2020-06-23
Initial implementation of ARKit support

Alex-MSFT:
This is a continuation of Cedric's earlier PR here: https://github.com/BabylonJS/BabylonNative/pull/249 
Taking his changes and deploying it on device resulted in some changes to the code, but the basic principle is the same as Android.  The shader blends both textures and displays the result using two triangles.

What's done
- ARKit init and session delegate
- metal pipeline and rendering
- cmake update for openxr iOS

What's left to do
- GetHitResults
- Create/Update/Delete anchor.
- Properly handle orientation changes
- Properly handle the view management to re-use the view passed in by the consumer.
- Do another pass on updateFoV in XR.mm to properly handle Landscape orientations.  This seems like discrepancy between the behavior of projectionMatrixForOrientation and displayTransformForOrientation
- Do another pass on updateDisplayOrientedPose, this seems like a hacky way to change between coordinate systems, and could probably be vastly simplified, but will take some digging.


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+        float DepthFarZ{ DEFAULT_DEPTH_FAR_Z };
+
+        Impl(System::Impl& systemImpl, void* graphicsContext, void* window)
+            : SystemImpl{ systemImpl }
+        {
+            MetalDevice = id<MTLDevice>(graphicsContext);
+            UIView* MainView = (UIView*)window;
+            
+            // Create the XR View to stay within the safe area of the main view.
+            // TODO: Should re-use the view passed into the engine rather than creating a sub-view.
```
CedricGuillemet, line 463: Using the same view results in this pipeline and bgfx using the same layer. So, when the swap chain is flipped, it flicks between bgfx and the XR display. That's why a new view is created on top of the previous one.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+    };
+}
+
+/**
+ Implementation of the ARSessionDelegate interface for more info see: https://developer.apple.com/documentation/arkit/arsessiondelegate
+ */
+@implementation SessionDelegate
+{
+    std::vector<xr::System::Session::Frame::View>* activeFrameViews;
+    CVMetalTextureCacheRef textureCache;
```
CedricGuillemet, line 38: Is everything properly destroyed when we exit XR mode? Did you check to enter/exit XR mode multiple times and see if the memory comsumption is stable?

Alex-MSFT: I have not checked this yet, and I suspect it does not. I will check this today.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: I believe on Android we re-use the existing view @ryantrem Do you remember what we had to do to stop bgfx from using the same view?  This becomes important when integrating into react-native where there is an assumption that we will use the view handed to us, which might not cover the full screen.  We might be able to get away with using a sub-view that matches the dimensions of its parent, but I will defer to Ryan on the requirements in React-Native.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+            void* ysrcPlane = CVPixelBufferGetBaseAddressOfPlane(arCameraBuffer, 0);
+            memcpy(ydestPlane, ysrcPlane, bufferWidth * bufferHeight);
+            
+            auto uvdestPlane = CVPixelBufferGetBaseAddressOfPlane(frameBuffer, 1);
+            auto uvsrcPlane = CVPixelBufferGetBaseAddressOfPlane(arCameraBuffer, 1);
+            memcpy(uvdestPlane, uvsrcPlane, bufferWidth * bufferHeight / 2);
+        }
+        @finally
+        {
+            CVPixelBufferUnlockBaseAddress(arCameraBuffer, 0);
```
ryantrem, line 179: Are these two finally blocks reversed? I'd have though the outer one would free the `arCameraBuffer` lock.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Good idea to check this, stop/resume was actually not working at all!

 I now have that working, and have addressed most of the major memory leaks, however we are still seeing ~15MB (down from 50+ MB!) of leaks between each session.  From looking at the profiler these leaks come from the MtkView's drawables not actually getting released after session end.  As far as I can tell from documentation calling [xrView releaseDrawables] should do the trick, but it doesn't seem to be working.  Do you have any experience working with MtkView is there some other process I should follow to fully deallocate it?

Another interesting thing I have learned is you actually cannot destroy an ARSession you can only pause and then later resume it.  So once we have allocated the AR Session we cannot fully reclaim its resources.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+        {
+            return;
+        }
+        
+        @try
+        {
+            // Copy both planes of the AR camera image to the frame buffer.
+            void* ydestPlane = CVPixelBufferGetBaseAddressOfPlane(frameBuffer, 0);
+            void* ysrcPlane = CVPixelBufferGetBaseAddressOfPlane(arCameraBuffer, 0);
+            memcpy(ydestPlane, ysrcPlane, bufferWidth * bufferHeight);
```
ryantrem, line 171: Is this what other ARKit examples do? This looks like it is copying pixels out of GPU memory into main memory and then back to GPU memory.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+            MetalDevice = id<MTLDevice>(graphicsContext);
+            sessionDelegate = [[SessionDelegate new]init:&ActiveFrameViews metalContext:MetalDevice];
+            session.delegate = sessionDelegate;
+            configuration.planeDetection = ARPlaneDetectionHorizontal;
+            configuration.lightEstimationEnabled = false;
+            configuration.worldAlignment = ARWorldAlignmentGravity;
+            [session runWithConfiguration:configuration];
+
+            // build pipeline
+            NSError* error;
```
ryantrem, line 511: Move this down to line 500?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+            [sessionDelegate release];
+            [configuration release];
+            [session release];
+            dispatch_sync(dispatch_get_main_queue(), ^{
+                [xrView removeFromSuperview]; });
+        }
+
+        std::unique_ptr<System::Session::Frame> GetNextFrame(bool&, bool&)
+        {
+            unsigned int width = viewportSize.x;
```
ryantrem, line 569: uint32_t?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+                    renderPassDescriptor.colorAttachments[0].clearColor = MTLClearColorMake(0.0,0.0,0.0,0.0);
+                    
+                    // Set up the clear for the depth texture.
+                    renderPassDescriptor.depthAttachment.texture = reinterpret_cast<id<MTLTexture>>(ActiveFrameViews[0].DepthTexturePointer);
+                    renderPassDescriptor.depthAttachment.loadAction = MTLLoadActionClear;
+                    renderPassDescriptor.depthAttachment.clearDepth = 1.0f;
+                    
+                    // Create and end the render encoder.
+                    id<MTLRenderCommandEncoder> renderEncoder =
+                        [commandBuffer renderCommandEncoderWithDescriptor:renderPassDescriptor];
```
ryantrem:
```suggestion
                    id<MTLRenderCommandEncoder> renderEncoder = [commandBuffer renderCommandEncoderWithDescriptor:renderPassDescriptor];
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+- (CGSize)viewportSize
+{
+    auto frameSize = activeFrameViews->front().ColorTextureSize;
+    return CGSizeMake(frameSize.Width, frameSize.Height);
+}
+
+/**
+ Called every frame during the active ARKit session.  Updates the AR Camera texture, and Camera pose. If a size change is detected also sets the UVs, and FoV values.
+*/
+- (void)session:(ARSession *)session didUpdateFrame:(ARFrame *)frame
```
ryantrem: Maybe update the comment to note that this is part of the `ARSessionDelgate` protocol?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+            auto& frameView = ActiveFrameViews[0];
+            frameView.FieldOfView.AngleDown = -(frameView.FieldOfView.AngleUp = 0.5);
+            frameView.FieldOfView.AngleLeft = -(frameView.FieldOfView.AngleRight = 0.5);
+        }
+
+        ~Impl()
+        {
+            [sessionDelegate release];
+            [configuration release];
+            [session release];
```
ryantrem: Do you need to explicitly release `_cameraTextureY` and `_cameraTextureCbCr`?

Alex-MSFT: These should automatically be managed by the texture cache, I do need to free the MTLTexture which I added in my last iteration.

Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -23,7 +24,7 @@ - (void)dealloc
 {
 }
 
-- (void)init:(void*)CALayerPtr width:(int)inWidth height:(int)inHeight
+- (void)init:(void*)inView width:(int)inWidth height:(int)inHeight
```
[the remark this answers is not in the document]
Alex-MSFT: Done

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,838 @@
...
+
+            [session pause];
+            session.delegate = nil;
+            [sessionDelegate release];
+            [pipelineState release];
+            dispatch_sync(dispatch_get_main_queue(), ^{
+                [xrView removeFromSuperview]; });
+            [xrView releaseDrawables];
+            [xrView dealloc];
+            xrView = nil;
```
ryantrem, line 564: Do you not have to release the `session`?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,782 @@
...
+    {
+        // Find the width, height, and format of the AR image.
+        size_t bufferWidth = CVPixelBufferGetWidth(arCameraBuffer);
+        size_t bufferHeight = CVPixelBufferGetHeight(arCameraBuffer);
+        auto format = CVPixelBufferGetPixelFormatType(arCameraBuffer);
+        
+        // If we don't have two planes the arCamera has not yet been initialized so back out for now.
+        if (CVPixelBufferGetPlaneCount(arCameraBuffer) < 2)
+        {
+            CVPixelBufferUnlockBaseAddress(arCameraBuffer, kCVPixelBufferLock_ReadOnly);
```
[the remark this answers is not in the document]
Alex-MSFT: Ah yes I forgot to remove this when I added the finally block good catch

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: We have to do this unfortunately because if you don't then it holds the memory in the ARFrame indefinitely, and the ARKit session freezes after 12 frames.  It might be a side effect of how we are managing the command buffer from the non-UI thread.

I'll give it another look to see if some of the other changes I have made make this no longer necessary.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Yes fixed

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: I am using a static session & config here that is persistent across sessions.  I'm looking into seeing if that was not actually necessary, and if I can just go back to having it owned by the instance itself and then I would need to call release here.

Alex-MSFT: Confirmed going back to releasing the session after every call would leak memory across sessions.  I think there are some resources that are not tied to the lifetime of the session that might need to be cleaned up separately.  Going to leave it as is for now.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Looks like this is no longer needed! I'm guessing the important change was adding the @autoreleasepool in DrawFrame.

ryantrem:
Will you add the remaining work to #278? Also, do we need to handle pause/resume?

Alex-MSFT:
> Will you add the remaining work to #278? Also, do we need to handle pause/resume?

These should be automatically handled by ARKit, and I have tested it out and at least on the surface it seems to work.  We may want to register handlers for [sessionWasInterrupted ](https://developer.apple.com/documentation/arkit/arsessionobserver/2891620-sessionwasinterrupted?language=objc) and [sessionInterruptionEnded](https://developer.apple.com/documentation/arkit/arsessionobserver/2891619-sessioninterruptionended?language=objc), but as far as I can tell this is optional for cases where we have extra processing not tied to didUpdateFrame that should not run while the app is in the background.

Dependencies/xr/Source/ARKit/XR.mm
ryantrem: I think the subview will work ok. I don't understand why the situation is different between Android and iOS though.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -0,0 +1,677 @@
...
+            // TODO
+        }
+        
+        private:
+            inline static ARSession* session{};
+            inline static ARWorldTrackingConfiguration* configuration{};
+            MTKView* xrView{};
+            bool SessionEnded{ false };
+            id<MTLDevice> MetalDevice{};
+            CAMetalLayer* MetalLayer{};
```
ryantrem, line 589: Seems like there is a mix of casing here. Should these three be `sessionEnded`, `metalDevice`, `metalLayer`?

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Sounds good, I left the task in to verify this work for React-Native in #278

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -556,6 +572,34 @@ bool TryInitialize()
...
+                if(renderPassDescriptor != nil)
+                {
+                    renderPassDescriptor.colorAttachments[0].texture = reinterpret_cast<id<MTLTexture>>(ActiveFrameViews[0].ColorTexturePointer);
+                    renderPassDescriptor.colorAttachments[0].loadAction = MTLLoadActionClear;
+                    renderPassDescriptor.colorAttachments[0].clearColor = MTLClearColorMake(0.0,0.0,0.0,0.0);
+                    //renderPassDescriptor.colorAttachments[0].storeAction = MTLStoreActionStore;
+                    renderPassDescriptor.depthAttachment.texture = reinterpret_cast<id<MTLTexture>>(ActiveFrameViews[0].DepthTexturePointer);
+                    renderPassDescriptor.depthAttachment.loadAction = MTLLoadActionClear;
+                    renderPassDescriptor.depthAttachment.clearDepth = 1.0f;
+                    //renderPassDescriptor.depthAttachment.storeAction = MTLStoreActionDontCare;
```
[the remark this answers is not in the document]
Alex-MSFT: I was going back and forth on it, and I don't think we need it as these match the default action.  Will remove.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Fixed

PR #240, opened 2020-04-24
Implement the WebXR Hit Test Module polyfill for ARCore

Alex-MSFT:
This change introduces the polyfill to fulfill the [WebXR hit test contract](https://immersive-web.github.io/hit-test/) for ARCore devices as described in (#230).

This includes:
1. Implementing the NAPI polyfill classes for: XRRay, XRPose, XRHitTestsource, and XRHitTestResult
2. Implementing the polyfill APIs for: XRSession.RequestHitTestSource, XRFrame.GetHitTestResults
3. Implementing the ARcore routine to perform a hit test and return the results
4. Updating the playground app to expose XRHitTest as an option when XR is enabled.

PR #308, opened 2020-06-25
Implement GetHitTestResults for ARKit

Alex-MSFT:
This change implements the iOS ARKit polyfill for GetHitTestResults as detailed in #278 .  This is the ARKit equivalent of #240.  

This supports hit testing via ARRaycastQuery for iOS 13+ and ARFrame.hitTest for iOS <=12.0.  For versions less than 13.0 only hit testing from the center of the screen is supported.  We will need to come up with a scheme for translating a world space ray to a screen point, as casting from a world space ray is not supported in older versions.  Since we currently don't support setting an offset ray at the API level I will defer this for now.

### DIFF
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -619,7 +619,6 @@ namespace xr {
                 }
             }
         }
-               
         private:
             inline static ARSession* session{};
             inline static ARWorldTrackingConfiguration* configuration{};
@@ -638,13 +637,17 @@ namespace xr {
         HitResult transformToHitResult(simd_float4x4 transform) const {
             auto orientation = simd_quaternion(transform);
             HitResult hitResult{};
-            hitResult.Pose.Orientation = { orientation.vector.x,
+            hitResult.Pose.Orientation = {
+                orientation.vector.x,
                 orientation.vector.y,
                 orientation.vector.z,
-                orientation.vector.w};
-            hitResult.Pose.Position = { transform.columns[3][0],
+                orientation.vector.w
+            };
+            hitResult.Pose.Position = {
+                transform.columns[3][0],
                 transform.columns[3][1],
-                transform.columns[3][2] };
+                transform.columns[3][2]
+            };
             
             return hitResult;
         }
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -609,8 +609,7 @@ namespace xr {
                     } else {
                         // On iOS versions prior to 13, fall back to doing a raycast from a screen point, for now don't support translating the offset ray.
                         auto hitTestResults = [session.currentFrame hitTest:CGPointMake(.5, .5) types:(ARHitTestResultTypeExistingPlane)];
-                        for (ARHitTestResult* result in hitTestResults)
-                        {
+                        for (ARHitTestResult* result in hitTestResults) {
                             filteredResults.push_back(transformToHitResult(result.worldTransform));
                         }
                     }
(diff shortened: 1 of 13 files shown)


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -61,12 +61,11 @@ - (UIInterfaceOrientation)orientation {
...
     if (@available(iOS 13.0, *)) {
-        if (window && window.windowScene) {
-            return window.windowScene.interfaceOrientation;
-        }
+        return window.windowScene.interfaceOrientation;
     }
-
-    return [[UIApplication sharedApplication] statusBarOrientation];
+    else {
+        return [[UIApplication sharedApplication] statusBarOrientation];
```
ryantrem, line 67: I don't think you need this change, but either way.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -574,10 +571,52 @@ void DrawFrame() {
...
+                                             alignment:ARRaycastTargetAlignmentAny];
+                        
+                        // Perform the actual raycast.
+                        auto rayCastResults = [session raycast:raycastQuery];
+                        
+                        // Process the results and push them into the results list.
+                        for (ARRaycastResult* result in rayCastResults) {
+                            filteredResults.push_back(transformToHitResult(result.worldTransform));
+                        }
+                    } else {
```
ryantrem, line 611: If `offsetRay`'s origin is where the camera is (raycasting from the camera), then couldn't we hit test to a correct (according to the Ray) screen coordinate?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -574,10 +571,52 @@ void DrawFrame() {
...
+                        auto hitTestResults = [session.currentFrame hitTest:CGPointMake(.5, .5) types:(ARHitTestResultTypeExistingPlane)];
+                        for (ARHitTestResult* result in hitTestResults) {
+                            filteredResults.push_back(transformToHitResult(result.worldTransform));
+                        }
+                    }
+                }
+            }
         }
-        
+
```
ryantrem, line 621:
```suggestion

```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -589,6 +628,23 @@ void GetHitTestResults(std::vector<HitResult>&, xr::Ray) const {
...
+        /*
+         Helper function to translate a world transform into a hit test result.
+         */
+        HitResult transformToHitResult(simd_float4x4 transform) const {
+            auto orientation = simd_quaternion(transform);
+            HitResult hitResult{};
+            hitResult.Pose.Orientation = { orientation.vector.x,
+                orientation.vector.y,
+                orientation.vector.z,
+                orientation.vector.w};
```
ryantrem, line 643:
```suggestion
            hitResult.Pose.Orientation = {
                orientation.vector.x,
                orientation.vector.y,
                orientation.vector.z,
                orientation.vector.w
            };
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -589,6 +628,23 @@ void GetHitTestResults(std::vector<HitResult>&, xr::Ray) const {
...
+            hitResult.Pose.Position = { transform.columns[3][0],
+                transform.columns[3][1],
+                transform.columns[3][2] };
```
ryantrem, line 646:
```suggestion
            hitResult.Pose.Position = {
                transform.columns[3][0],
                transform.columns[3][1],
                transform.columns[3][2]
            };
```

Dependencies/xr/Source/ARKit/XR.mm
ryantrem: Oh I see now you commented in the PR saying we could do this later.

Alex-MSFT: Yeah I'll do this later, currently we don't actually pass the offset ray in so effectively we're always hit testing from the center point

PR #312, opened 2020-06-30
Various ARKit integration fixes

Alex-MSFT:
This PR fixes a few ARKit implementation issues mainly:

- Memory spiking caused by ARKit tracking of vertical planes, and leaking across frames due to not releasing MTLTexture.
- Babylon UVs drawing image reversed, and coordinate space to be flipped across the X axis.
- Simplify logic in UpdateDisplayOrientedPose.

One thing to call out is the use of @synchronized to prevent race conditions between the renderer and ARKit session delegate, which may not be performant enough though as far as I have seen does not have a material impact on frame rate.    This is necessary because they run on two separate threads, which may not be in sync so neither can exclusively own the lifetime of the camera textures. Please review its usage and let me know if this is unacceptable.


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -77,28 +99,37 @@ - (CGSize)viewportSize {
...
-    if ([self checkAndUpdateCameraUVs:frame]) {
-        // If our camera UVs updated, then also update the FoV to match the updated UVs.
-        [self updateFoV:frame.camera];
+    @autoreleasepool{
+        // Update both metal textures used by the renderer to display the camera image.
+        CVMetalTextureRef newCameraTextureY = [self getCameraTexture:frame.capturedImage plane:0];
+        CVMetalTextureRef newCameraTextureCbCr = [self getCameraTexture:frame.capturedImage plane:1];
+        
+        // Swap the camera textures, do this under synchronization lock to prevent null access.
+        @synchronized(self) {
```
Alex-MSFT, line 112: Please review this usage, and let me know if you think this will be okay or if we should find another way to handle the lifetime of these textures either by moving renderer onto the main UI thread or find a way to properly ref count across the threads.  From what I have read p_thread_mutex_t is slightly more efficient, but once exception handling etc. has been applied the difference is negligible.  

Since we're not using [ARC](https://en.wikipedia.org/wiki/Automatic_Reference_Counting) we need to manually manage the lifetime of camera textures, but since the renderer and the ARKit session run on two separate threads we can't just blindly have the lifetime of the texture be owned by any one thread, so I wanted the render loop manually grab the current frame's texture in a threadsafe manner when it needs it then own freeing it, and the session needs to own freeing unused buffers in the cache as new frames come in from the camera.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -540,39 +580,58 @@ void DrawFrame() {
...
-                    [renderEncoder endEncoding];
+                        // Schedule a present once the framebuffer is complete using the current drawable.
+                        [commandBuffer presentDrawable:drawable];
+                    }
 
-                    // Schedule a present once the framebuffer is complete using the current drawable.
-                    [commandBuffer presentDrawable:drawable];
+                    // Finalize rendering here & push the command buffer to the GPU.
+                    [commandBuffer commit];
+                    [commandBuffer waitUntilCompleted];
```
ryantrem, line 624: Could this cause a perf issue?

Alex-MSFT: Replaced with a completion handler to free the textures after the command buffer is exhausted.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -19,10 +17,10 @@ @interface SessionDelegate : NSObject <ARSessionDelegate>
 
     static XRVertex vertices[] = {
         // 2D positions, UV,        camera UV
-        { { -1, -1 },   { 0, 0 },   { 0, 0} },
-        { { -1, 1 },    { 0, 1 },   { 0, 0} },
-        { { 1, -1 },    { 1, 0 },   { 0, 0} },
-        { { 1, 1 },     { 1, 1 },   { 0, 0} },
+        { { -1, -1 },   { 1, 0 },   { 0, 0} },
```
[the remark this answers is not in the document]
Alex-MSFT: Yes I'm not sure, I noticed the Babylon texture was mirrored, and applied these UVs to fix it.  This also allowed me to fix some of the issues I was seeing in UpdateDisplayOrientedPose where I had to reflect translation across the X axis, and invert the Y and Z orientation.

@CedricGuillemet do you have any insight into why the texture we're getting from BGFX are mirrored like this?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -621,16 +635,17 @@ void DrawFrame() {
...
+                @catch (NSException* exception) {
                     if (cameraTextureY != nil) {
                         [cameraTextureY setPurgeableState:MTLPurgeableStateEmpty];
                     }
                     
                     if (cameraTextureCbCr != nil) {
                         [cameraTextureCbCr setPurgeableState:MTLPurgeableStateEmpty];
                     }
+                    
+                    [exception raise];
```
ryantrem, line 648: Does this blow away the original stack trace (this is common in other languages)? Is replacing this with `@throw` the right way to re-throw an exception while maintaining the original stack trace?

PR #313, opened 2020-06-30
New suspend mechanism

syntheticmagus:
Extricated the suspend mechanism from the old AppLifecycle PR to be in its own PR.


Core/AppRuntime/Include/Babylon/AppRuntime.h
```diff
@@ -15,7 +16,7 @@ namespace Babylon
         AppRuntime();
         ~AppRuntime();
 
-        void Suspend();
+        std::future<void> Suspend();
```
bghgary, line 19: Who uses the std::future?

syntheticmagus: At present no one. Its inclusion comes from your observation back when suspend is actually an asynchronous operation and some mechanism should be provided for determining when the suspension has truly happened. This future provides such a mechanism without exposing Arcana.

syntheticmagus: I did test it -- it was used by the old AppLifecycle code that this change used to be in a PR with -- but I'm fine with leaving it out and putting it back in when/if we need it. It won't be a hard change to make later.

PR #318, opened 2020-07-01
Release build on CI for Android and Apple

CedricGuillemet:
Release build for Android and Apple on CI and update glslang because the unused variable has been fixed.

### DIFF
--- a/Plugins/NativeXr/Source/NativeXr.cpp
+++ b/Plugins/NativeXr/Source/NativeXr.cpp
@@ -972,6 +972,8 @@ namespace Babylon
                         const auto referenceSpaceType = info[0].As<Napi::String>().Utf8Value();
                         assert(referenceSpaceType == XRReferenceSpaceType::UNBOUNDED ||
                             referenceSpaceType == XRReferenceSpaceType::VIEWER);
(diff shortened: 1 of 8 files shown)


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -972,6 +972,8 @@ namespace Babylon
                         const auto referenceSpaceType = info[0].As<Napi::String>().Utf8Value();
                         assert(referenceSpaceType == XRReferenceSpaceType::UNBOUNDED ||
                             referenceSpaceType == XRReferenceSpaceType::VIEWER);
+                        (void)XRReferenceSpaceType::UNBOUNDED;
+                        (void)XRReferenceSpaceType::VIEWER;
```
syntheticmagus, line 976: Is this to prevent these variables from getting compiled out? If so, and if this is the way we actually want to do this, can you specify that in a comment just for clarity's sake?

CedricGuillemet: Actually, we need some UNUSED(x) macro. It's on my todo list.

PR #320, opened 2020-07-02
Render loop experimentation

syntheticmagus:
Preliminary experimentation for decoupling the render loop from RequestAnimationFrame.


syntheticmagus:
Abandoned in favor of newer approaches such as #324.

PR #322, opened 2020-07-06
bgfx Encoder POC

CedricGuillemet:
Unlike Android and Windows, on iOS bgfx init and resize must happen on main thread.
And because of that, rendering must also happen on the main thread (resizing when doing rendering on another thread crashes the app)
I tried with locks here and there without getting 100% reliable behavior. That's why I ended up doing this bgfx encoder Proof of concept.

You can record rendering commands on an encoder. All encoding streams will be effectively send to the graphics driver when calling bgfx::frame.

There are 2 semaphores involved here. The bgfx good practice is to create an encoder, record commands, and repeat. If there is no more encode available, then skip rendering until bgfx is called and flushes the encoders. With 1 semaphore I usually get all encoders (8) used before getting to bgfx::frame. So I had to sync in both ways.

I also had to change the win32 loop. I don't know yet how to do the same with Android. 

The good part of this for some integration is you get your rendering when you call ```Babylon::Plugins::NativeEngine::Render();```
POC, not meant to be merged in this state. Only tested with the cube.

### DIFF
--- a/Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
+++ b/Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
@@ -81,7 +81,7 @@ extern "C"
 
                 Babylon::Plugins::NativeWindow::Initialize(env, window, width, height);
 
-                Babylon::Plugins::NativeEngine::InitializeGraphics(window, width, height);
+                Babylon::Plugins::NativeEngine::InitializeGraphics(window, width, height, true);
                 Babylon::Plugins::NativeEngine::Initialize(env);
 
                 Babylon::Plugins::NativeXr::Initialize(env);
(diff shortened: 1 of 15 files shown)


CedricGuillemet:
A followup on yesterday discussion.

1. BGFX multithread is only needed if multiple threads encode commands. In our case, for any platform, only the JS is encoding commands. -> flag is set back to 0 in CMakeLists.txt, bgfx::renderframe() call is removed
2. There are 2 behaviors: Let the JS thread do the rendering/resize (what's being done now) OR let the main thread do the rendering while the JS thread encode commands -> I've added a boolean flag at init to specify the wanted behavior
3. What defines the behavior is when calling bgfx resize and render, so the logic is NativeEngine::EndFrame(), NativeEngine::Render() and NativeEngine::UpdateSize()
4. with that change, it works exactly the same for win32,UWP, Android, Linux if you set true to the ```threadedRendering``` Flag
5. On iOS, we want to handle the Rendering, so set the flag to false, and call Render/Resize from the MTKView Delegate

CedricGuillemet:
Closing thanks to @syntheticmagus 's work on render loop.

PR #324, opened 2020-07-08
ScheduleRender-based approach to rendering

syntheticmagus:
Possible alternative way to evolve the way we're doing rendering.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.h
+++ b/Plugins/NativeEngine/Source/NativeEngine.h
@@ -465,6 +465,9 @@ namespace Babylon
 
         void UpdateSize(size_t width, size_t height);
 
+        void ScheduleRender();
+        bool m_isRenderScheduled{false};
+
         arcana::cancellation_source m_cancelSource{};
 
         ShaderCompiler m_shaderCompiler;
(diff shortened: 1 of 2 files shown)


Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -465,6 +465,9 @@ namespace Babylon
 
         void UpdateSize(size_t width, size_t height);
 
+        void ScheduleRender();
+        bool m_isRenderScheduled{false};
```
CedricGuillemet, line 469: atomic<bool> ?

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -495,6 +498,6 @@ namespace Babylon
         // Scratch vector used for data alignment.
         std::vector<float> m_scratch{};
         
-        Napi::FunctionReference m_requestAnimationFrameCalback{};
+        std::vector<Napi::FunctionReference> m_requestAnimationFrameCallbacks{};
```
CedricGuillemet, line 501: Does accessing that vector should be protected by a mutex ?

syntheticmagus:
Subsumed into #345.

PR #325, opened 2020-07-09
Compensate aspect ratio AR

CedricGuillemet:
Long story short, the aspect ratio is not good when creating the projectionMatrix because tan function is not linear.
So, I've added code to compensate the ratio deviation. I didn't want to do the change in NativeXR to not break the existing.


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -221,6 +221,25 @@ - (void)updateFoV:(ARCamera*)camera {
     float fov =  atanf(1 / yScale);
     frameView.FieldOfView.AngleDown = -(frameView.FieldOfView.AngleUp = fov);
     frameView.FieldOfView.AngleLeft = -(frameView.FieldOfView.AngleRight = fov * aspectRatio);
+    
+    // Angles for FieldOfView respect the viewport ratio
+    // but tangent is not a linear function and ratio of values(l,r,b,t) computed
```
Alex-MSFT, line 226: Thanks for the fix!  I'm guessing we also need to apply this compensation on the Android side as well.  I'll add checking this to #230.

PR #328, opened 2020-07-15
XR Controllers Crash Fix

syntheticmagus:
Reworked XR input argument check to prevent crashing when OpenXR controllers are provided.

The way XR inputs for OpenXR scenarios are currently implemented is at least incomplete and possibly incorrect altogether; it doesn't take into account relative spaces or many other features, and the current state provides little more than a working demo for a very narrow set of scenarios. Long-term, this implementation will need to be replaced and greatly expanded to support the desired feature set. For now, though, this PR restores the functionality of the old proof-of-concept and fixes the crash that was happening whenever an OpenXR scenario included a controller.


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1319,20 +1317,21 @@ namespace Babylon
 
             Napi::Value GetPose(const Napi::CallbackInfo& info)
             {
-                auto* xrSpace = XRReferenceSpace::Unwrap(info[0].As<Napi::Object>());
-                if (xrSpace != nullptr)
+                if (info[0].IsExternal())
```
Alex-MSFT, line 1320: Thanks for the fix Justin this was my intention with the original change, but I had thought that unwrap would just return null in the event the object was an incorrect type.

syntheticmagus: Gotcha. I actually didn't know for sure what it would do either. Turns out it throws a NAPI error associated with invalid arguments, apparently.

PR #331, opened 2020-07-17
update babylon.max.js for 32bits support in index streams

### LINKED ISSUE #321: Inconsistent GLB loading on device and in emulation.
Loading GLBs can be unpredictable. Some models load fine, others do not. Happens more often in emulation, but also occurs on device. Links to problematic GLBs will be provided.

This GLB looks like this in Sandbox:

image

But looks like this when loaded in the demo app:

image

It looks like this whether loaded in the emulator or deployed on device (Android).

While it looks simple it's actually a bit more complicated than it appears. It was subdivided to get it to 364,544 triangles, with one draw call, so that it would be a certain file size (for a different test).

image

Link to the file is here:

https://raw.githubusercontent.com/RonnieAshlock/TestAssets/master/WarehouseShelf.glb

Other test files found here:

https://github.com/RonnieAshlock/TestAssets/tree/master/GLB

CedricGuillemet:
fixes #321 
related PR on babylonjs : https://github.com/BabylonJS/Babylon.js/pull/8550


ryantrem:
What version of Babylon.js is this? What version of the @babylonjs/core NPM package has this fix?

CedricGuillemet:
built from my babylonjs branch that is not yet merged into master.

syntheticmagus:
That PR has since been merged into Babylon.js. Just for the sake of easy record-keeping (makes it easier to understand the history of `babylon.max.js` when debugging regressions), can we take a version built from Babylon.js master and include in a comment here the commit ID from which the change was built?

ryantrem:
> That PR has since been merged into Babylon.js. Just for the sake of easy record-keeping (makes it easier to understand the history of `babylon.max.js` when debugging regressions), can we take a version built from Babylon.js master and include in a comment here the commit ID from which the change was built?

When a Babylon.js release is created, the built version of `babylon.max.js` is checked in to the babylonjs repo as well. Ideally I think we should just grab those since they are built by the automated build system and less likely to have any human error (like accidentally having a local change or something like that). Maybe this is the same thing @syntheticmagus is saying, not sure.

syntheticmagus:
@ryantrem It's a step beyond what I was saying. I like it, though.

ryantrem:
I think this is the alpha 26 release that has the change: https://github.com/BabylonJS/Babylon.js/blob/61c67dddd783d4d7b1fff039cd7b3b224a26ed32/dist/preview%20release/babylon.max.js

CedricGuillemet:
Closed because of PR https://github.com/BabylonJS/BabylonNative/pull/350

PR #332, opened 2020-07-17
Enable bitcode with cmake CLI

### LINKED ISSUE #296: [iOS] Enable Bitcode
@ryantrem reported this : 
All our cmake stuff on the iOS side disables bitcode. Why do we do this? Bitcode is enabled by default on new iOS projects, including React Native ones, and projects with bitcode enabled can't consume libraries where bitcode was not enabled (so basically as is we force consumers to disable bitcode for their apps).
 
My understanding is that bitcode is an intermediate format that allows Apple to do the final optimization within their store services, including re-optimizing as their optimizers improve without requiring any app updates to be submitted to the store. Seems like something we should support.

CedricGuillemet:
fixes #296


CedricGuillemet:
@ryantrem without XCODE_ATTRIBUTE_ENABLE_BITCODE in cmakelists but with -DENABLE_BITCODE=ON, bitcode is enabled for every libraries.

PR #333, opened 2020-07-17
Unicode fixes for NAPI/JSI

bghgary:
Unicode conversion is a mess in C++. Using codecvt directly results in crashes on certain platforms. But using deprecated codecvt facets API works just fine.

### DIFF
--- a/Dependencies/napi/napi-direct/source/js_native_api_JavaScriptCore.cc
+++ b/Dependencies/napi/napi-direct/source/js_native_api_JavaScriptCore.cc
@@ -8,6 +8,7 @@
 #include <string>
 #include <vector>
 #include <locale>
+#include <codecvt>
 
 struct napi_callback_info__ {
   napi_value newTarget;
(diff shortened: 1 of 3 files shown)


Dependencies/napi/napi-direct/source/js_native_api_JavaScriptCore.cc
```diff
@@ -101,21 +102,9 @@ namespace {
...
-      char16_t* to_next;
-      auto result = facet.in(state, from, from + length, from_next, to, to + length, to_next);
-      if (result != std::codecvt_base::ok) {
-        throw std::invalid_argument{"Invalid unicode string"};
-      }
-      chars.resize(to_next - to);
-
-      return JSStringCreateWithCharacters(chars.data(), chars.size());
+      std::u16string u16str{std::wstring_convert<
+        std::codecvt_utf8_utf16<char16_t>, char16_t>{}.from_bytes(string, string + length)};
```
ryantrem, line 106:
```suggestion
      std::u16string u16str{std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>{}.from_bytes(string, string + length)};
```

PR #335, opened 2020-07-22
Plane detection support for iOS and Android

Alex-MSFT:
This PR implements plane detection support for #230 and #278  according to the spec proposed here: [Plane Detection Explainer](https://github.com/immersive-web/real-world-geometry/blob/master/plane-detection-explainer.md)

The idea is to track planes by checking for updated frames every time we request a new frame on the XR session, and surface it up in the newly added frame.WorldInformation.DetectedPlanes object.  Planes are represented as a set of boundary vertices situated about a center point.

On ARCore this is achieved by using the [GetUpdatedTrackables](https://developers.google.com/ar/reference/c/group/frame#arframe_getupdatedtrackables) API, and on ARKit this is a combination of session(_:didAdd:), session(_:didUpdate:), and session(_didRemove:) protocols [API Reference](https://developer.apple.com/documentation/arkit/arsessiondelegate/2865617-session}.


Dependencies/xr/Include/XR.h
```diff
@@ -78,6 +86,17 @@ namespace xr
         bool IsValid{true};
     };
 
+    using NativePlaneIdentifier = void*;
+    struct Plane
+    {
+        Pose Center{};
+        float* Polygon{};
```
syntheticmagus, line 93: Broadly speaking, I think we should try to avoid raw pointers in general, but especially in contracts as much as possible. Would `std::vector<float>` work here?

Dependencies/xr/Include/XR.h
```diff
@@ -46,6 +53,7 @@ namespace xr
         } Orientation;
     };
 
+    const float FLOAT_COMPARISON_THRESHOLD = .02f;
```
syntheticmagus, line 56: A few points:
- I don't think this is part of the contract and should not be exposed through XR.h.
- Our existing style leads floats with numerals (i.e. `0.02f`) for consistency.
- I believe this could be `constexpr`.
- nit: In my experience, the value used for this purpose is usually called epsilon, signifying, "a number small enough that we don't care about it."

Dependencies/xr/Include/XR.h
```diff
@@ -78,6 +86,17 @@ namespace xr
         bool IsValid{true};
     };
 
+    using NativePlaneIdentifier = void*;
```
syntheticmagus, line 89: Is this ever dereferenced externally (by which I mean in NativeXr.cpp or beyond)? If not, should we really be exposing a pointer here?

Dependencies/xr/Include/XR.h
```diff
@@ -149,6 +168,9 @@ namespace xr
                 ~Frame();
 
                 void GetHitTestResults(std::vector<HitResult>&, Ray) const;
+                void UpdatePlanes(std::map<NativePlaneIdentifier, Plane*>& existingPlanes, std::vector<Plane>& newPlanes, std::vector<Plane*>& removedPlanes) const;
```
syntheticmagus, line 171: `std::vector<Plane>` instead of `Plane*`?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -314,6 +315,25 @@ namespace xr
...
+        bool operator==(const Pose& lhs, const Pose& rhs)
+        {
+            return abs(lhs.Position.X - rhs.Position.X) < FLOAT_COMPARISON_THRESHOLD
+                && abs(lhs.Position.Y - rhs.Position.Y) < FLOAT_COMPARISON_THRESHOLD
+                && abs(lhs.Position.Z - rhs.Position.Z) < FLOAT_COMPARISON_THRESHOLD
+                && abs(lhs.Orientation.X - rhs.Orientation.X) < FLOAT_COMPARISON_THRESHOLD
+                && abs(lhs.Orientation.Y - rhs.Orientation.Y) < FLOAT_COMPARISON_THRESHOLD
+                && abs(lhs.Orientation.Z - rhs.Orientation.Z) < FLOAT_COMPARISON_THRESHOLD
+                && abs(lhs.Orientation.W - rhs.Orientation.W) < FLOAT_COMPARISON_THRESHOLD;
+        }
```
syntheticmagus, line 331: It may be fine in this case, but strictly speaking I'm not sure this is mathematically rigorous. The unit quaternion is over-described to represent 3D orientation, meaning it can (at least at times) represent the same orientation using different numbers. Consequently, this operator may yield false negatives if orientations are generated in such a way that this can occur.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -314,6 +315,25 @@ namespace xr
 
             return task;
         }
+
+        /**
+         * Compare two poses and determine if they are approximately equal.
```
syntheticmagus, line 320: Instead of an operator, should this have a name that clarifies it's an approximate comparison, such as `ApproxEqual()`?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -890,33 +890,33 @@ namespace Babylon
             {
                 xr::Ray nativeRay{{0, 0, 0}, {0, 0, -1}};
                 auto originObject = m_origin.Value();
-                if (originObject.HasOwnProperty("x"))
+                if (originObject.Has("x"))
```
syntheticmagus, line 893: I apologize that this isn't really part of this PR, but this series of checks seems pretty bizarre. I think these checks involve string operations, so this could potentially be surprisingly slow. Is it an expected case that we'll be trying to get rays from objects that don't have XYZ components? That seems like it should be an error.

Dependencies/xr/Include/XR.h
Alex-MSFT: I am using a plane pointer so that we can edit the plane in place.  I don't think you can do std::vector<Plane&> right?

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: I mainly want this function to ignore very small changes to avoid spamming the consumer with micro plane adjustments that occur every frame.  I'll change it to ApproxEqual instead to make it more clear for its intended purpose.

Dependencies/xr/Include/XR.h
Alex-MSFT: Will move to the individual .cpp files.  I had it here so it could be shared between xr.mm and xr.cpp

Dependencies/xr/Include/XR.h
Alex-MSFT: The value we get back from ARCore must be pushed into a dynamically allocated float array, so doing it in a std::vector requires us to both dynamically allocate the array then copy it over to a std::vector, but I can do that if we prefer.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Will do

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: I can remove it.  Yes I added this for safety as on the Babylon.js side this object is user specified and can technically exclude these properties, though it does not conform to the XRRay object spec.

Dependencies/xr/Include/XR.h
Alex-MSFT: This is not dereferenced, but we do use this type as a key in the dictionary.  Is there a way I can expose this without declaring it explicitly as a void*?

Dependencies/xr/Include/XR.h
syntheticmagus: Yes, as long as the vector components are non-const, you can edit them.

Dependencies/xr/Source/ARCore/XR.cpp
syntheticmagus: Gotcha. In that case, if you name it something like `WasMeaningfullyUpdated` or `ShouldUpdateConsumer` (or some other name less awful than both of those), that will remove all ambiguity both on the approximate nature of it and on the intended usage.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +813,107 @@ namespace xr
...
+                // Dynamically allocate the polygon array, and fill it in.
+                int32_t polygonSize;
+                ArPlane_getPolygonSize(session, planeTrackable, &polygonSize);
+                float* polygon = reinterpret_cast<float*>(malloc(sizeof(float) * polygonSize));
+
+                // Get the polygon.
+                ArPlane_getPolygon(session, planeTrackable, polygon);
+
+                // Update the existing plane if it exists, otherwise create a new plane, and add it to our list of planes.
+                auto planeIterator = existingPlanes.find(reinterpret_cast<NativePlaneIdentifier>(trackable));
```
ryantrem, line 855: It would be a bit easier to follow if this used `planeTrackable` instead of `trackable`.
```suggestion
                auto planeIterator = existingPlanes.find(reinterpret_cast<NativePlaneIdentifier>(planeTrackable));
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +813,107 @@ namespace xr
...
+            // Next check for updated planes, and update their pose and polygon or create a new plane if it does not yet exist.
+            ArFrame_getUpdatedTrackables(session, frame, AR_TRACKABLE_PLANE, trackableList);            
+            int32_t size{};
+            ArTrackableList_getSize(session, trackableList, &size);
+            for (int i = 0; i < size; i++)
+            {
+                // Get the plane.
+                ArTrackable* trackable;
+                ArTrackableList_acquireItem(session, trackableList, i, &trackable);
+                auto planeTrackable = reinterpret_cast<ArPlane*>(trackable);
```
ryantrem, line 830: I would put `ArTrackable* trackable` in its own scope so we don't have two variables referring to the same object. This will help force the following code to be a bit easier to follow.
```suggestion
                ArPlane* planeTrackable{};
                {
                    ArTrackable* trackable{};
                    ArTrackableList_acquireItem(session, trackableList, i, &trackable);
                    planeTrackable = reinterpret_cast<ArPlane*>(trackable);
                }
```

Dependencies/xr/Include/XR.h
Alex-MSFT: Worked this out offline, I'm now using an std::vector<float> and using std::vector::resize along with .data() to accomplish this.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +813,107 @@ namespace xr
...
+
+                // Check if this plane has been subsumed. If so skip it as we are about to delete this plane.
+                ArPlane* subsumingPlane = nullptr;
```
ryantrem, line 833:
```suggestion
                ArPlane* subsumingPlane{};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +813,107 @@ namespace xr
...
+                if (planeIterator != existingPlanes.end())
+                {
+                    UpdateExistingPlane(planeIterator->second, rawPose, polygon, polygonSize);
+                }
+                else
+                {
+                    // This is a new plane, create it and initialize its values.
+                    Plane plane{};
+                    plane.NativePlaneId = reinterpret_cast<NativePlaneIdentifier>(trackable);
+                    UpdateExistingPlane(&plane, rawPose, polygon, polygonSize);
```
ryantrem: Maybe this function should be named `UpdatePlane`?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +813,107 @@ namespace xr
...
+                    newPlanes.push_back(plane);
+                }
+            }
+        }
+
+        bool CheckIfPlaneWasUpdated(Plane* existingPlane, float* newPolygon, size_t newPolygonSize, Pose& newCenter)
+        {
+            // First check if the center has changed, or the polygon size has changed.
+            if (existingPlane->Center != newCenter
+                || existingPlane->PolygonSize * 2 != newPolygonSize)
```
ryantrem: What is the `* 2`?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +813,107 @@ namespace xr
...
+        }
+
+        // Cleans up a plane. Called when a plane gets deleted.
+        void CleanupPlane(Plane* plane)
+        {
+            ArTrackable_release(reinterpret_cast<ArTrackable*>(plane->NativePlaneId));
+
+            if (plane->Polygon != nullptr)
+            {
+                free(plane->Polygon);
```
ryantrem: Similar comment as above, why `free` (paired with `malloc`) instead of `delete` (paired with `new`)?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +813,107 @@ namespace xr
...
+                plane->Polygon = nullptr;
+            }
+        }
+
     private:
         bool isInitialized{false};
         bool sessionEnded{false};
         std::vector<ArTrackable*> frameTrackables{};
         std::vector<ArAnchor*> arCoreAnchors{};
+        std::set<float*> planeBuffers{};
```
ryantrem: Could this be `unordered_set`?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1411,6 +1517,72 @@ namespace Babylon
...
+                }
+
+                // Pass the world information object back to the caller.
+                worldInformationObj.Set("detectedPlanes", planeSet);
+                return std::move(worldInformationObj);
+            }
+
+            void UpdatePlanes(const Napi::Env& env, uint32_t timestamp)
+            {
+                std::map<xr::NativePlaneIdentifier, xr::Plane*> existingNativePlaneMap;
```
ryantrem, line 1520: Can this be `xr::Plane&`? It's expected to never be `nullptr`, correct?

ryantrem, line 1520:
```suggestion
                std::map<xr::NativePlaneIdentifier, xr::Plane*> existingNativePlaneMap{};
```

Dependencies/xr/Include/XR.h
ryantrem: Yea I agree, in general it seems like there is too much manual memory management happening here (easy to make mistakes and have leaks). It seems like if we use `std::vector<Plane>`, `std::vector<float>`, etc. we can avoid a lot of the manual allocations/deallocations.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -66,6 +72,11 @@ static simd_float4x4 PoseToTransform(xr::Pose pose) {
  */
 @implementation SessionDelegate {
     std::vector<xr::System::Session::Frame::View>* activeFrameViews;
+    
+    NSLock *planeLock;
```
ryantrem, line 76:
```suggestion
    NSLock* planeLock;
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -402,6 +482,20 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
+                   && abs(lhs.Position.Z - rhs.Position.Z) < FLOAT_COMPARISON_THRESHOLD
+                   && abs(lhs.Orientation.X - rhs.Orientation.X) < FLOAT_COMPARISON_THRESHOLD
+                   && abs(lhs.Orientation.Y - rhs.Orientation.Y) < FLOAT_COMPARISON_THRESHOLD
+                   && abs(lhs.Orientation.Z - rhs.Orientation.Z) < FLOAT_COMPARISON_THRESHOLD
+                   && abs(lhs.Orientation.W - rhs.Orientation.W) < FLOAT_COMPARISON_THRESHOLD;
+        }
+
+        bool operator!=(const Pose& lhs, const Pose& rhs) {
+            return !(lhs == rhs);
+        }
```
ryantrem: Is this the same as the one for ARCore? Should we put this in a shared header (not xr.h, but a shared implementation header) that both files can use?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -449,18 +543,18 @@ bool TryInitialize() {
             });
 
             // If the singleton session and configuraiton have not yet been created, create them here now.
```
ryantrem, line 542: Is this comment out of date now? I assume the "If" is referring to the condition you removed?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -774,6 +869,102 @@ void DeleteAnchor(xr::Anchor& anchor) {
...
+                        polygon[polygonIndex + 1] = geometry.boundaryVertices[i].y;
+                        polygon[polygonIndex + 2] = geometry.boundaryVertices[i].z;
+                    }
+
+                    // Update the existing plane if it exists, otherwise create a new plane, and add it to our list of planes.
+                    auto planeIterator = existingPlanes.find(reinterpret_cast<NativePlaneIdentifier>(updatedPlane.identifier));
+                    if (planeIterator != existingPlanes.end()) {
+                        UpdatePlane(planeIterator->second, updatedPlane, polygon, polygonSize);
+                    }
+                    else {
```
ryantrem, line 895:
```suggestion
                    } else {
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1217,6 +1217,104 @@ namespace Babylon
...
+                // Translate the polygon from a native array to a JS array.
+                auto polygonArray = Napi::Array::New(info.Env(), m_nativePlane.PolygonSize);
+                for (size_t i = 0; i < m_nativePlane.PolygonSize; i++)
+                {
+                    auto polygonPoint = Napi::Object::New(info.Env());
+                    if (m_nativePlane.PolygonFormat == xr::PolygonFormat::XZ)
+                    {
+                        size_t polygonIndex = 2 * i;
+                        polygonPoint.Set("x", m_nativePlane.Polygon[polygonIndex]);
+                        polygonPoint.Set("y", 0);
```
ryantrem, line 1267: I wonder if it would be simpler to just always have XYZ and make it an implementation detail of ARCore that the y value is always 0 (that the planar polygon is always on the XZ plane)?

Dependencies/xr/Include/XR.h
```diff
@@ -90,7 +89,7 @@ namespace xr
     struct Plane
     {
         Pose Center{};
-        float* Polygon{};
+        std::vector<float> Polygon{};
         size_t PolygonSize{0};
```
ryantrem, line 93: Remove?

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Yes, will fix.

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: We get the polygon buffer directly from ARCore, which is formatted as 2D points.  I think it's better to leave it as is than have to copy it again.

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
Alex-MSFT: I don't think order matters.  I can use unordered_map.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Added a new XRHelpers.h which implements this, and removed the individual implementations.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Done, I renamed the equivalent UpdatePlane on the iOS side, I guess I forgot to refactor it here.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: I got rid of this.  This is because the existingPlane->Polygonsize is the number of verts and newPolygonSize was the size of the array.  I agree it's confusing so I removed this and just check for the vector size now.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Got rid of this manual memory management.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Got rid of this code

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: You can't use xr::Plane& in a std::unordered_map as it always tries to copy it.

Alex-MSFT: Well I'll be damned nevermind, this does work for maps, but does not work for vectors.  I will use that here.

Dependencies/xr/Include/XR.h
syntheticmagus: Spoke a little offline, but I think it might be valuable to go a little deeper here, so I sent out a meeting invite for Monday.

Dependencies/xr/Source/Shared/XRHelpers.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+#include <XR.h>
+
+namespace xr
+{
+	/**
```
ryantrem, line 6: This file has all tabs instead of spaces.

Alex-MSFT: VS code why :(

Dependencies/xr/Include/XR.h
Alex-MSFT: This comment is outdated now.

Dependencies/xr/Include/XR.h
```diff
@@ -142,8 +150,21 @@ namespace xr
...
 
+                struct Plane
+                {
+                    using Identifier = size_t;
+                    Identifier ID{ 0 };
+                    Pose Center{};
+                    std::vector<float> Polygon{};
+                    size_t PolygonSize{0};
+                    PolygonFormat PolygonFormat{};
+                };
```
syntheticmagus, line 161:
```suggestion
                struct Plane
                {
                    using Identifier = size_t;
                    const Identifier ID{ NEXT_ID++ };
                    Pose Center{};
                    std::vector<float> Polygon{};
                    size_t PolygonSize{0};
                    PolygonFormat PolygonFormat{};
                
                private:
                    static inline Identifier NEXT_ID{0};
                };
```

Dependencies/xr/Include/XR.h
```diff
@@ -169,6 +190,7 @@ namespace xr
             void RequestEndSession();
             Size GetWidthAndHeightForViewIndex(size_t viewIndex) const;
             void SetDepthsNearFar(float depthNear, float depthFar);
+            void SetPlaneDetectionEnabled(bool enabled)const;
```
syntheticmagus, line 193: This, I think, makes more sense on the `Session` than here since it's not really a setting of the frame, but of the session.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -321,8 +322,10 @@ namespace xr
         const System::Impl& SystemImpl;
         std::vector<Frame::View> ActiveFrameViews{ {} };
         std::vector<Frame::InputSource> InputSources;
+        std::vector<Frame::Plane> Planes;
```
syntheticmagus, line 325:
```suggestion
        std::vector<Frame::InputSource> InputSources{};
        std::vector<Frame::Plane> Planes{};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +795,79 @@ namespace xr
...
+        }
+
     private:
         bool isInitialized{false};
         bool sessionEnded{false};
         std::vector<ArTrackable*> frameTrackables{};
         std::vector<ArAnchor*> arCoreAnchors{};
+        std::vector<float> planePolygonBuffer{};
+        std::unordered_map<ArPlane*, Frame::Plane::Identifier> planeMap{};
+        Frame::Plane::Identifier nextPlaneID = 0;
```
syntheticmagus, line 870: I think you can eliminate this variable if you make the plane IDs set themselves.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -796,11 +795,79 @@ namespace xr
...
+                ArPlane_acquireSubsumedBy(session, planeTrackable, &subsumingPlane);
+                if (subsumingPlane != nullptr)
+                {
+                    ArTrackable_release(reinterpret_cast<ArTrackable*>(planeTrackable));
+                    ArTrackable_release(reinterpret_cast<ArTrackable*>(subsumingPlane));
+                    continue;
+                }
+
+                // Get the center pose.
+                float rawPose[7]{};
```
syntheticmagus, line 833:
```suggestion
                std::array<float, 7> rawPose{};
```

syntheticmagus: nit, it's just a slightly safer type.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -873,6 +941,73 @@ namespace xr
...
+                ArPlane* subsumingPlane = nullptr;
+                ArPlane_acquireSubsumedBy(session, arPlane, &subsumingPlane);
+
+                // Plane has been subsumed, stop tracking it explicitly.
+                if (subsumingPlane != nullptr)
+                {
+                    subsumedPlanes.push_back(planeID);
+                    Planes[planeID].Polygon.clear();
+                    Planes[planeID].PolygonSize = 0;
+                    planeMapIterator = planeMap.erase(planeMapIterator);
```
syntheticmagus, line 965: nit: If this works, it works, but it seems a little weird. I think it's more typical to collect a list of keys to erase than to erase from a collection while you're actively iterating over it.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -866,4 +870,9 @@ namespace xr
     {
         throw std::runtime_error("Anchors not yet implemented for OpenXR.");
     }
+
+    void System::Session::SetPlaneDetectionEnabled(bool) const
+    {
+        // NYI
```
syntheticmagus, line 876:
```suggestion
        assert(!enabled);
```

syntheticmagus: `NYI` is, "Not yet implemented," I'm guessing? You can if you want to, but no need to call that out, I think; there's a lot that's not implemented. 😆

Dependencies/xr/Source/Shared/XRHelpers.h
```diff
@@ -0,0 +1,41 @@
...
+    bool PoseWasMeaningfullyUpdated(const Pose& lhs, const Pose& rhs)
+    {
+        return abs(lhs.Position.X - rhs.Position.X) > FLOAT_COMPARISON_THRESHOLD
+            || abs(lhs.Position.Y - rhs.Position.Y) > FLOAT_COMPARISON_THRESHOLD
+            || abs(lhs.Position.Z - rhs.Position.Z) > FLOAT_COMPARISON_THRESHOLD
+            || abs(lhs.Orientation.X - rhs.Orientation.X) > FLOAT_COMPARISON_THRESHOLD
+            || abs(lhs.Orientation.Y - rhs.Orientation.Y) > FLOAT_COMPARISON_THRESHOLD
+            || abs(lhs.Orientation.Z - rhs.Orientation.Z) > FLOAT_COMPARISON_THRESHOLD
+            || abs(lhs.Orientation.W - rhs.Orientation.W) > FLOAT_COMPARISON_THRESHOLD;
+    }
```
syntheticmagus, line 19: I'm not sure we want to be in the business of making decisions about what's a meaningful update or not; that seems like it should be either the platform's or the app's decision, not XR's.

Alex-MSFT: I want to revisit this later.  For now this is sane for both of the supported platforms and stops us from spamming with re-draws.  I agree that this makes sense to be pushed to the consuming app for more accurate devices.

Dependencies/xr/Include/XR.h
Alex-MSFT: This is already on the session.

syntheticmagus: Ha! Ignore me. 😆

PR #342, opened 2020-07-28
Clear bgfx framebuffers when exiting XR mode on Android

CedricGuillemet:
Android part of this issue: https://github.com/BabylonJS/BabylonReactNative/issues/57

When toggling engine view, resize happens on the bgfx size and similarly to a previous bug, bgfx try to refresh framebuffers. But because the texture it's bound to is already destroy, an assert in bgfx is raised. 

The FrameBuffer data was not properly destroyed (and the bound bgfx framebuffer).

I've added a bgfx::frame to flush and execute the resize commands. On Android, this happens in the js thread. On iOS, I'm doing a PR to make it happen on the UI thread.
So, this has no implication on Android.
On iOS, the resize is also performed on bgfx::frame() which use UI functions to get the layer. If this happens on the JS thread, it crashes. So, doing it in the main thread.

Optional windowPtr in resize. XR needs an updated windowPtr to create its own view when the mtkview changes on iOS (toggle engineView)

I've found an issue with rendertarget life cycle. I'm commented an assert to make it work. https://github.com/BabylonJS/BabylonNative/issues/344


Plugins/NativeWindow/Source/NativeWindow.cpp
```diff
@@ -37,8 +37,13 @@ namespace Babylon::Plugins::Internal
     {
     }
 
-    void NativeWindow::Resize(size_t newWidth, size_t newHeight)
+    void NativeWindow::Resize(size_t newWidth, size_t newHeight, void* newWindowPtr)
```
ryantrem, line 40: Nit: Can we match parameter order with `NativeEngine::Reinitialize` for consistency?
```suggestion
    void NativeWindow::Resize(void* newWindowPtr, size_t newWidth, size_t newHeight)
```

CedricGuillemet: My first thought was to add a setWindowPtr and let the resize with just width,height. newWindowPtr is used on iOS only. I don't like other platform to be forced to call the function with a nullptr. What do you think @syntheticmagus ?

Issue #314, opened 2020-07-01
Move rendering out of RequestAnimationFrame()


syntheticmagus:
Rendering -- meaning, as implemented, the call to `bgfx::frame()` -- is currently done from inside `NativeEngine::RequestAnimationFrame()`, which is called tail-recursively from the JavaScript by the Babylon Engine's `runRenderLoop()` function. This is not a correct implementation as there are many scenarios that will not work using this mechanism: manual rendering using `scene.render()`, rendering at correct framerates when `bgfx::frame` is non-blocking, etc. The logic for rendering must therefore be reworked and the call to `bgfx::frame` moved such that all supported scenarios function as expected.

syntheticmagus:
Fixed in #345.

PR #345, opened 2020-07-30
Native graphics

syntheticmagus:
Preliminary mechanism for splitting out graphics into its own component in order to be able to handle more complicated situations such as off-thread rendering, multiple engines, and even (potentially) multiple JavaScript environments. Currently NOT fully-featured, but a functioning prototype within the Playground app.

The main missing piece (aside from cleaned-up interfaces and sanity checks on the architecture) is a notion for how to correctly manage the lifetime of `NativeGraphics`. A major challenge of the current design is that it gets its task system densely comingled with tasks that do JavaScript work; these tasks capture variables, etc. from within the JavaScript world and thus cannot be allowed to survive if the JavaScript thread has been shut down. However, a significant part of the point of designing this mechanism was to allow the rendering work to not be owned by the JavaScript thread. This creates a bit of a conundrum that I have not yet figured out how to cleanly resolve, but must be resolved in order for it to be possible to destroy `NativeGraphics` cleanly.

Other smaller missing pieces include porting over other functionality from where it used to live on the `NativeEngine`, including reinitialization and resizing. I don't anticipate these features being as complicated -- we can even do things like forbid calling them if there's a living frame being rendered, forcing them to only be called in the space between frames -- but there may be nuances I'm missing. There's also currently no protection against calling `AdvanceFrame()` before you've discarded the last frame you asked for, which may be an architecture flaw; still thinking about that and very open to suggestions.


Apps/Playground/Win32/App.cpp
```diff
@@ -78,29 +80,41 @@ namespace
 
     void Uninitialize()
     {
-        inputBuffer.reset();
+        if (inputBuffer)
```
CedricGuillemet, line 83: Is that ```if``` necessary?

Core/Graphics/Include/Babylon/Graphics.h
```diff
@@ -0,0 +1,42 @@
...
+{
+    class Graphics
+    {
+    public:
+        class Impl;
+
+        ~Graphics();
+
+        template<typename NativeWindowT>
+        static std::unique_ptr<Graphics> InitializeFromWindow(NativeWindowT window, size_t width, size_t height);
```
CedricGuillemet, line 17: If I call ```InitializeFromWindow<HWND>``` in my code, I'll get a linker error I guess because the template specialization is only done for void* but it will compile just fine.

CedricGuillemet:
Once again I'm boring with naming but I don't like 'Graphics' as a name :D . I've only seen that word in marketing text. I think 'Renderer' or 'NativeRenderer' would be a better fit. Or any other word can we find more frequently in 3D engine codebase.

Core/Graphics/Include/Babylon/Graphics.h
syntheticmagus: Correct. This is a technique we created to allow platform-specific types to be eliminated from platform-agnostic header files while still having the build-time type safety (not compile-time, but link-time) that would be lost by just casting things to `void*`. I think we first used this pattern somewhere in XR, but it's currently most prominently used for [`Napi::Attach(...)`](https://github.com/BabylonJS/BabylonNative/blob/master/Dependencies/napi/napi-direct/include/napi/env.h#L7). Intellisense can't figure much out from this, but it remains safe without the overhead of either a parameter struct or multiple headers with a config-time swapping mechanism in the build system.

syntheticmagus: To clarify, specializing this for `void*` is an idiotic thing to do that completely discards the aforementioned benefits of this pattern, and that's not what it's supposed to do long-term. 😄 I just haven't gotten around to doing the actual correct specializations yet; and until I do, the `void*` specialization emulates the prior behavior, which was unsafe but still worked.

syntheticmagus:
`Graphics` in this case is intended to reference the [GPU](https://en.wikipedia.org/wiki/Graphics_processing_unit), for which this object is allegedly Babylon Native's high-level abstraction. This was the best name Gary and I thought of, though it's far from perfect as the actual GPU abstraction extends far beyond this object into bgfx. In reality, what this object actually _does_ is more along the lines of threading, logic flow, and lifecycle management for the app's relationship with the GPU, but I don't know of a good existing name for a concept like that. It's not a renderer, but it's also not quote a GPU abstraction, at least not a full one. Anyway, that's where that name comes from. I still think it's a better name than the other names I've thought about, but let's definitely chat offline if you feel `Renderer` or another alternative would be an improvement.

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -29,6 +30,7 @@ HINSTANCE hInst;                                // current instance
 WCHAR szTitle[MAX_LOADSTRING];                  // The title bar text
 WCHAR szWindowClass[MAX_LOADSTRING];            // the main window class name
 std::unique_ptr<Babylon::AppRuntime> runtime{};
+std::unique_ptr<Babylon::Graphics> graphics{};
```
syntheticmagus, line 33:
```suggestion
std::unique_ptr<Babylon::Graphics> graphics{};
std::unique_ptr<Babylon::AppRuntime> runtime{};
```

syntheticmagus:
This PR should address #314.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+            auto bgfxStats = bgfx::getStats();
+            if (w != bgfxStats->width || h != bgfxStats->height)
+            {
+                bgfx::reset(w, h, BGFX_RESET_FLAGS);
+                bgfx::setViewRect(0, 0, 0, w, h);
+#ifdef __APPLE__
+                bgfx::frame();
+#else
+                    bgfx::touch(0);
```
bghgary, line 186: nit: spacing

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -384,19 +386,22 @@ namespace Babylon
...
-        static void Initialize(Napi::Env);
+        static void Initialize(Napi::Env, bool autoRender);
 
         FrameBufferManager& GetFrameBufferManager();
         void Dispatch(std::function<void()>);
-        void EndFrame();
+
+        void ScheduleRender();
+
+        const bool AutomaticRenderingEnabled{};
```
bghgary, line 403: Is this always false?

bghgary (approved):
Small comments.

Apps/Playground/UWP/App.cpp
```diff
@@ -112,12 +112,19 @@ void App::Run()
 // class is torn down while the app is in the foreground.
 void App::Uninitialize()
 {
-    m_inputBuffer.reset();
+    if (m_inputBuffer)
```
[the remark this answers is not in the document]
syntheticmagus: No, it works, I just copied the existing pattern without thinking. Cedric already pointed this out on Win32 and I fixed it there, but I forgot the same pattern was here too.

Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -66,6 +68,7 @@ - (void)init:(void*)view width:(int)inWidth height:(int)inHeight
 
 - (void)resize:(int)inWidth height:(int)inHeight
 {
+    graphics->UpdateSize(static_cast<size_t>(inWidth), static_cast<size_t>(inHeight));
```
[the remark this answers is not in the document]
syntheticmagus: Given the order of operations for iOS, I don't think it should be possible for either of them to be null here. (It seems impossible that resize should be called before init.) I don't know that for sure, though; and since that isn't the point of this PR, I can just copy the existing pattern for now.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -0,0 +1,191 @@
...
+        pd.ndt = nullptr;
+        pd.nwh = windowPtr;
+        pd.context = nullptr;
+        pd.backBuffer = nullptr;
+        pd.backBufferDS = nullptr;
+        bgfx::setPlatformData(pd);
+        bgfx::reset(static_cast<uint32_t>(width), static_cast<uint32_t>(height));
+    }
+
+    void Graphics::Impl::AddToJavaScript(Napi::Env env)
```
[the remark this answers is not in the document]
syntheticmagus: Kind of. The Graphics component is also designed to support multi-`Env` scenarios, at least in theory; and though we're currently resetting the Graphics component a lot to match existing behavior, long-term the usage pattern will more likely tend (I think) toward creating the Graphics once at the beginning, after which it will survive an arbitrary number of Runtimes.

Plugins/NativeEngine/Source/NativeEngine.h
syntheticmagus: No, it's usually true. A per-`Env` constant is set in the JavaScript type at initialization type; the [default value is true](https://github.com/BabylonJS/BabylonNative/pull/345/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&hide-deleted-files=true#diff-f70275df353ecc0e0b28ddcfc891504cR9). Individual instances of `NativeEngine` will then read this value in their constructors, using that to set the `AutomaticRenderingEnabled` value.

Apps/Playground/UWP/App.cpp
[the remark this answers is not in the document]
bghgary: The existing pattern is because there is a DeinitializeGraphics call. Since that's isn't necessary anymore, the if statements look weird.

Issue #347, opened 2020-07-30
PointCloudSystem crashes BabylonNative when calling BuildMeshAsync


Alex-MSFT:
The PointCloudSystem does not work on BabylonNative due to a failure in CreateIndexBuffer

1.) Add the following 3 lines to the sample app experience.js (or use attached experience.js file)
```
var pcs= new BABYLON.PointsCloudSystem("pcs", 1, scene);
pcs.addPoints(10000);
pcs.buildMeshAsync();
```

2.) Run the sample app, and observe it crashes on an assert in NativeEngine::CreateIndexBuffer

[experience.txt](https://github.com/BabylonJS/BabylonNative/files/5003570/experience.txt)

CedricGuillemet:
Thanks for the repro, Alex. The first issue is 0sized IndexBuffer. I fixed it with a check in the JS code. But I have another issue deeper in the code related to shader compilation...Investigating.

CedricGuillemet:
I've fixed the shader compilation but now some functions in the native engine JS are missing...

PR #350, opened 2020-07-31
point cloud rendering

CedricGuillemet:
- Fix hlsl compatibility with point rendering
- bindbuffer function
- bjs max update

fixes https://github.com/BabylonJS/BabylonNative/issues/347

This PR contains built changes from bjs : https://github.com/BabylonJS/Babylon.js/pull/8650

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -663,6 +663,7 @@ namespace Babylon
                 InstanceMethod("setViewPort", &NativeEngine::SetViewPort),
                 InstanceMethod("getFramebufferData", &NativeEngine::GetFramebufferData),
                 InstanceMethod("getRenderAPI", &NativeEngine::GetRenderAPI),
+                InstanceMethod("bindBuffer", &NativeEngine::BindBuffer),
             });
 
         env.Global().Get(JsRuntime::JS_NATIVE_NAME).As<Napi::Object>().Set(JS_ENGINE_CONSTRUCTOR_NAME, func);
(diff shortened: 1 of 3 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1563,21 +1590,29 @@ namespace Babylon
...
         // TODO: handle viewport
 
+        // TODO: support other fill modes
+        uint64_t fillModeState = 0; //indexed tri list
+        if (fillMode == 2)
+        {
+            fillModeState |= BGFX_STATE_PT_POINTS;
+        }
+
+
```
syntheticmagus, line 1608:
```suggestion

```

nit: Extra line.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -557,6 +560,7 @@ namespace Babylon
 
     private:
         std::vector<uint8_t> m_bytes{};
+        bgfx::VertexLayoutHandle vertexLayoutHandle{};
```
syntheticmagus, line 563:
```suggestion
        bgfx::VertexLayoutHandle m_vertexLayoutHandle{};
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1563,21 +1590,29 @@ namespace Babylon
...
+
+
         for (const auto& it : m_currentProgram->Uniforms)
         {
             const ProgramData::UniformValue& value = it.second;
             bgfx::setUniform({it.first}, value.Data.data(), value.ElementLength);
         }
 
-        bgfx::setState(m_engineState);
+        bgfx::setState(m_engineState | fillModeState);
```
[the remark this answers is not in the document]
CedricGuillemet: bgfx::setstate merges 2 different concepts: rendering states (cullin, z test/write, color mask, ...) and draw primitive (lines, triangles, point,...). I want to keep those concepts sperate as much as I can. The rendering state is m_engineState but we have to set the primitive with SetState. So that call is performed just before submitting. It's not supposed to stick around. Next drawcall will call SetState.

PR #352, opened 2020-08-03
Add removeEventListener definition to XRSession to prevent crash

chrisfromwork:
We have been seeing crashes on hololens when controllers connect and then disconnect. This crash was due to a missing remove event listener definition. This change adds a definitmion for that function.


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1706,7 +1707,7 @@ namespace Babylon
             JsRuntimeScheduler m_runtimeScheduler;
             uint32_t m_timestamp{0};
 
-            std::vector<std::pair<const std::string, Napi::FunctionReference>> m_eventNamesAndCallbacks{};
+            std::vector<std::pair<std::string, std::shared_ptr<Napi::FunctionReference>>> m_eventNamesAndCallbacks{};
```
bghgary, line 1710: why do you need to use shared_ptr?

bghgary, line 1710: Also, why remove const?

chrisfromwork: if the string is declared as const, the "=" operator doesn't get defined for the std::pair. This breaks move operations used in remove_if/erase.

chrisfromwork: i think i was hitting issues with the const std::string, retesting this again it doesn't seem needed. It has been reverted.

chrisfromwork:
Is there any update on this? What can I do to help get this checked in?

bghgary:
Sorry, I'm running behind still.

bghgary:
Merge from master and I'll merge.

PR #353, opened 2020-08-03
Expose AR Feature Points on Android

Alex-MSFT:
This PR introduces feature points for ARCore on Android.  Feature Points are the base understanding of the world geometry for AR Devices which are the foundation of planes and scene meshes.  The newly added API allows for turning on feature point detection and tracking feature point changes across frames.  This allows consuming applications to process them and apply their own custom spatial mesh reconstruction that is consistent across platforms.

The general design of the API is to surface changed feature points directly from the native API up to the JS consumer as a flat array of changed points.  The average frame contains ~0-500 changed feature points depending on the current motion of the camera.

For more info see the equivalent PR in Babylon.js: https://github.com/BabylonJS/Babylon.js/pull/8677


Apps/Playground/Scripts/experience.js
```diff
@@ -156,6 +157,45 @@ CreateBoxAsync().then(function () {
...
                 }
 
+                if (xrFeaturePoints) {
+                    const xrFeaturePointsModule = xr.baseExperience.featuresManager.enableFeature(
+                        BABYLON.WebXRFeatureName.FEATURE_POINTS,
+                        "latest",
+                        {});
+
+                    var pcs= new BABYLON.PointsCloudSystem("pcs", 5, scene);
+                    var featurePointInitFunc = function (particle, i, s) {
```
hendryems, line 167: I probably don't have the right context but what's experience.js file supposed to represent? Is this examples of how client is expected to consume this?

I'm going to assume that's the case - I'm confused about is why one would need to create and initiate the feature points. If this is bootstrapping code just for the purpose of illustrating usage in some sandbox context, then i think comments would be super useful.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -860,6 +862,69 @@ namespace xr
...
+                FeaturePointCloud.resize(numberOfPoints);
+                for (int32_t i = 0; i < numberOfPoints; i++)
+                {
+                    FeaturePointCloud.emplace_back();
+                    auto& featurePoint = FeaturePointCloud.back();
+
+                    int32_t dataIndex = i * 4;
+                    featurePoint.x = pointCloudData[dataIndex];
+                    featurePoint.y = pointCloudData[dataIndex + 1];
+                    featurePoint.z = -1 * pointCloudData[dataIndex + 2];
```
hendryems, line 900:
```
featurePoint.z = -1 * pointCloudData[dataIndex + 2];
```

Curious why the -1; are we doing RH <--> LH conversion here?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -321,11 +321,13 @@ namespace xr
     {
         const System::Impl& SystemImpl;
         std::vector<Frame::View> ActiveFrameViews{ {} };
-        std::vector<Frame::InputSource> InputSources{};
+        std::vector<Frame::InputSource> InputSources;
```
ryantrem, line 324: I'm guessing this change was unintentional?

Apps/Playground/Scripts/experience.js
Alex-MSFT: Experience.js is a sample app we use for testing and to show off how to use some of the features.  The reason we need to initiate the feature here is because this is essentially just a babylonjs script like any consuming application would use. 

I'll add some comments around what is going on here for the point cloud system.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: So we transport Babylon space directly into camera space whereas this is reported directly in world space if you take a look at the documentation for ARCore [getDisplayOrientedPose](https://developers.google.com/ar/reference/java/arcore/reference/com/google/ar/core/Camera#public-pose-getdisplayorientedpose) you will see that it is oriented such that -z is pointing in the direction the camera is looking.  So we need to flip this position to be correct relative to its actual position.

ryantrem, line 900: Why is this value multiplied by -1?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -860,6 +862,69 @@ namespace xr
...
+            }
+            catch (std::exception)
+            {
+                // Release the point cloud to free its memory.
+                ArPointCloud_release(pointCloud);
+                throw;
+            }
+
+            // Release the point cloud to free its memory.
+            ArPointCloud_release(pointCloud);
```
ryantrem, line 925: Can't we just use a finally?
```suggestion
            finally
            {
                // Release the point cloud to free its memory.
                ArPointCloud_release(pointCloud);
            }
```

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: See the above comment from Hendry.  We transport Babylon space directly into camera space whereas this is reported directly in world space if you take a look at the documentation for ARCore getDisplayOrientedPose you will see that it is oriented such that -z is pointing in the direction the camera is looking. So we need to flip this position to be correct relative to its actual position.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: No finally in C++ because well designed APIs use RAII, but the ARCore code was designed for C.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Yes unintentional thanks fixing.

Dependencies/xr/Include/XR.h
```diff
@@ -86,6 +86,17 @@ namespace xr
         bool IsValid{true};
     };
 
+    struct FeaturePoint
+    {
+        using Identifier = size_t;
+
+        float x{};
```
syntheticmagus, line 93: nit: Capitalize public struct fields.

Dependencies/xr/Include/XR.h
```diff
@@ -195,6 +208,7 @@ namespace xr
             Size GetWidthAndHeightForViewIndex(size_t viewIndex) const;
             void SetDepthsNearFar(float depthNear, float depthFar);
             void SetPlaneDetectionEnabled(bool enabled)const;
+            bool SetFeaturePointCloudEnabled(bool enabled)const;
```
syntheticmagus, line 211: If this can non-fatally fail, should we call `TrySet...(bool)`?

syntheticmagus: Also nit on this and line above: space before `const`.

Alex-MSFT: Good point will rename to TrySet...

Apps/Playground/Scripts/experience.js
syntheticmagus: No urgency to this, but at some point in the future the Playground app is likely to evolve into something a little different from what it's being used for right now. That's a ways into the future, but it might be good to get a head start on thinking about how we can keep our test harnesses for new features without necessarily checking them all into this one file. I'll start giving that a little thought; please let me know if anybody has any preferences or cool ideas to try.

PR #357, opened 2020-08-05
Shader alien fixes

syntheticmagus:
Fixes to get the alien head working in Babylon Native.

- Fixed fall-through case in glslang traversers when a uniform is considered neither a vector nor a matrix.
- Added handling logic to avoid location collision when generic attributes and UVs are used at the same time.
- Ingested new Babylon.js (commit 794aca8124c41c215c4f1bc134d0b8bfc2a0f086) with `Engine`-caching fix to allow Babylon.js to properly enumerate attributes on MorphTarget shaders with influences.


Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -136,6 +136,7 @@ namespace Babylon::ShaderCompilerTraversers
 
                 for (const auto& [name, symbol] : traverser.m_uniformNameToSymbol)
                 {
+
```
syntheticmagus, line 139:
```suggestion
```

PR #359, opened 2020-08-06
replaced bindbuffer function with bindvertexarray calls

### LINKED ISSUE #358: Malformed GLTF loading on Android
I just merged with master, and it looks like we are no longer properly loading models on Android.  This issue does not repro for windows builds.  Here are some examples.

Box/BoxAnimated/BoxTextured (just nothing)
image

Cesium Man (amorphous wiggling blob):
MicrosoftTeams-image (2)

Brainstem:
image

CedricGuillemet:
fixes #358 
babylon.max.js version with commit https://github.com/BabylonJS/Babylon.js/pull/8697/commits/7dd3d8607015a4b95aec52039ee558092e981b54
PR on babylonjs : https://github.com/BabylonJS/Babylon.js/pull/8697
Use a vertex array object instead of binding individual streams

tested ok win32, Android and iOS

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -503,7 +503,6 @@ namespace Babylon
                     },
                     &m_bytes);
 
-                m_vertexLayoutHandle = bgfx::createVertexLayout(layout);
                 m_handle = bgfx::createVertexBuffer(memory, layout);
             };
             const auto dynamic = [&layout, this](auto handle) {
@@ -519,7 +518,6 @@ namespace Babylon
                     },
                     &m_bytes);
 
-                m_vertexLayoutHandle = bgfx::createVertexLayout(layout);
                 m_handle = bgfx::createDynamicVertexBuffer(memory, layout);
             };
             DoForHandleTypes(nonDynamic, dynamic);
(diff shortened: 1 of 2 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -768,39 +763,21 @@ namespace Babylon
...
     void NativeEngine::BindVertexArray(const Napi::CallbackInfo& info)
     {
         const auto& vertexArray = *(info[0].As<Napi::External<VertexArray>>().Data());
 
-        vertexArray.indexBuffer.data->SetBgfxIndexBuffer();
+        // a vertex array might not have an index buffer associated with
+        if (vertexArray.indexBuffer.data)
+        {
+            vertexArray.indexBuffer.data->SetBgfxIndexBuffer();
+        }
```
bghgary, line 774: Technically speaking, a VAO (at least for WebGL) doesn't include the index buffer. Maybe we should separate these into separate functions like it is in WebGL.

CedricGuillemet: I'm opening an issue for that. I don't like to destroy and recreate a new vertex array object when call bindbuffers: Too many allocations/deallocations.

CedricGuillemet: https://github.com/BabylonJS/BabylonNative/issues/361

PR #360, opened 2020-08-06
Change XRViewerPose.Update to set the transform of the ViewerPose

Alex-MSFT:
This change changes the logic of XRViewerPose.Update to populate its transform with the pose of the first active view.

The XRViewerPose's transform is supposed to represent the position of the "Viewer" (i.e. phone, headset, or other tracked device) relative to a given reference space.  

This change assuming this pose is equal to that of the first active view.  This works for screen based viewers such as a phone or tablet, but is likely incorrect for devices with multiple views.  We will need to come up with a proper way to represent the viewer for these cases.


chrisfromwork:
Has this change been validated with a HMD?

Alex-MSFT:
> Has this change been validated with a HMD?

I don't have an HMD available to test right now, but this would be the equivalent of the left eye representing the viewpoint.  The current state is that we always report identity as the position of the WebXRCamera, so while not 100% correct this is still much closer to correct than the current state.  This change doesn't affect the perspective of the actual presented view but it is used for things like scene.pick in BabylonJS and reflected in xr.baseExperience.camera.position/orientation.

I set up a meeting next Tuesday to discuss where we want to properly support this for OpenXR where we can have multiple views.

chrisfromwork:
sounds good. I will try and ramp up more in this space before then.

PR #366, opened 2020-08-12
Fix for one of the alien head problems on Android

syntheticmagus:
Added handling in shader traverser for reshaping a level up when indexing into a type-adjusted uniform array. This allows Android to successfully render a single frame of alien head -- specifically the frame with the minimum number of morph target influences. There are other problems, including the placement of attributes at unallowably-high locations, which prevent the alien head from rendering more than a single frame on Android; those, however, are separate in origin and should probably be fixed in separate PRs.


syntheticmagus:
Yeah, this file needs some commentary, and not just for this. Will try to make a PR for that in the next day or so.

Issue #365, opened 2020-08-12
Support secondary view configurations and first person observer flags for OpenXR


chrisfromwork:
Secondary view configurations and first person observer views are required for high quality filming on most AR HMDs. Photo video cameras on headsets are often at physical offsets that don't allow for reuse of a stereo frame when compositing holograms into camera streams. This functionality enables having additional views rendered by the engine in order to create higher quality video recordings of holograms.

This is not a native-only concept. WebXR already has concepts for secondary views and a first person observer flag:
* https://immersive-web.github.io/webxr/#primary-and-secondary-views
* https://immersive-web.github.io/webxr-ar-module/#first-person-observer-view

The expectation is that this functionality won't have an impact on the Babylon.js developer experience. The typical OpenXR logic flow is the following:
1. Enable the 2 extensions [XR_MSFT_secondary_view_configuration](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_MSFT_secondary_view_configuration) and [XR_MSFT_first_person_observer](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_MSFT_first_person_observer)
1. Enable a first person observer view configuration during xrBeginSession
1. On xrWaitFrame, observe whether the first person observer view configuration has been set to active for the frame
1. If the first person observer view is active, obtain its location with a call to xrLocateView
1. If the first person observer view is active, render content to its associated swapchain using the found head pose
1. If the first person observer view is active, submit the first person observer view during xrEndFrame

An example for using this is HoloLens 2's mixed reality capture. When starting mixed reality capture, the declared secondary view configuration will be set to active. When stopping mixed reality capture, the view should return to no longer being active. (Note: HoloLens 2 emulators do not appear to support secondary views for mixed reality capture. Testing these changes will require running code on a physical HoloLens 2 device.)

chrisfromwork:
this should now be supported.

Issue #362, opened 2020-08-10
Left eye rendering seems broken for the Playground app on HoloLens 2


chrisfromwork:
Right eye rendering seems to work as expected. Left eye rendering fails to draw content in the center of the display.

bnolan:
I get this same bug on Odyssey Plus WMR. Right eye works fine, left eye works ok but when you move your head in certain orientations it's like a clipping plane moves through the cube or something and the cube disapears in the left eye, right eye keeps rendering properly.

syntheticmagus:
@chrisfromwork, can you confirm this no longer repros after #386?

chrisfromwork:
Yeah, this now looks a lot better. Thanks! :)

PR #367, opened 2020-08-12
Support secondary view configurations and first person observer flags for OpenXR

chrisfromwork:
This pull request addresses issue: https://github.com/BabylonJS/BabylonNative/issues/365
Using logic in this pull request will benefit from a fix for issue: https://github.com/BabylonJS/BabylonNative/issues/362

This pull request contains the following changes:
1. OpenXR-SDK is updated to release 1.0.11
1. XrEnumerate.h, XrExtensions.h, XrSTruct.h and XrViewConfiguration.h have been added to the xr component from the [OpenXR-MixedReality sample project](https://github.com/microsoft/OpenXR-MixedReality/tree/master/shared/XrUtility). These files have some changes because they are based on the OpenXR-SDK 1.0.10.2 NuGet package.
1. XrPlatform.h has been updated to continue to work with OpenXR 1.0.11
1. OpenXR xr.cpp has been updated to support secondary view configurations:
    * Logic has been updated to enable XR_MSFT_SECONDARY_VIEW_CONFIGURATION and XR_MSFT_FIRST_PERSON_OBSERVER extensions
    * Logic now assesses whether a secondary view configuration has been activated by an external source (such as Mixed Reality Capture). If it is set to active, the secondary view is located and declared to the Babylon.js engine for rendering
    * Logic has been refactored to reuse code for the new secondary view configuration types
1. Logic has been reenabled in NativeXr.cpp to support XRViews not having an assigned eye
1. isFirstPersonObserver property has been added to the NativeXR XRView class constructor definition to match the [immersive-ar spec](https://immersive-web.github.io/webxr-ar-module/#first-person-observer)


tymlipari:
Are there any issues with pulling in Apache licensed files into an MIT licensed project? I don't know what the rules are for those licenses.

chrisfromwork:
> Are there any issues with pulling in Apache licensed files into an MIT licensed project? I don't know what the rules are for those licenses.

I'm not sure on this one. @syntheticmagus and @bghgary do we have a perspective on what helper code we can bring in from other projects? We could theoretically reference OpenXR-MixedReality as a submodule, but it would still need changes to work with the 1.0.9 release we are targetting for OpenXR-SDK. Once a fix goes into 1.0.10 for a json lookup, we may be able to move everything to submodules if its preferred.

bghgary:
We need to follow guidelines here: https://docs.opensource.microsoft.com/. Even though this project is not under Microsoft Governance, we are using Microsoft rules.

For the [OpenXR-MixedReality](https://github.com/microsoft/OpenXR-MixedReality) one, at a minimum, it needs to be updated in the [NOTICE.md](https://github.com/BabylonJS/BabylonNative/blob/master/NOTICE.md) file at the root.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -276,57 +308,113 @@ namespace xr
...
-            uint32_t viewCount;
-            XrCheck(xrEnumerateViewConfigurationViews(instance, systemId, HmdImpl.VIEW_CONFIGURATION_TYPE, 0, &viewCount, nullptr));
-            assert(viewCount == HmdImpl.STEREO_VIEW_COUNT);
-            RenderResources.ConfigViews.resize(viewCount, { XR_TYPE_VIEW_CONFIGURATION_VIEW });
-            XrCheck(xrEnumerateViewConfigurationViews(instance, systemId, HmdImpl.VIEW_CONFIGURATION_TYPE, viewCount, &viewCount, RenderResources.ConfigViews.data()));
-
-            // Create all the swapchains.
+            uint32_t viewCount = static_cast<uint32_t>(viewState.Views.size());
+            RenderResources.ColorSwapchains[viewState.Type] = {};
+            RenderResources.DepthSwapchains[viewState.Type] = {};
```
syntheticmagus, line 357: nit: Will this construct, then copy? If so, is it better to move the vector in, or construct it in-place somehow?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -276,57 +308,113 @@ namespace xr
...
                         0,
-                        XR_SWAPCHAIN_USAGE_SAMPLED_BIT | XR_SWAPCHAIN_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT));
+                        XR_SWAPCHAIN_USAGE_SAMPLED_BIT | XR_SWAPCHAIN_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT,
+                        viewState.Type));
+            }
+        }
+
+        void CleanupSwapchains(ViewConfigurationState& viewState)
+        {
+            // TODO - unclear if this is the appropriate way to cleanup a swap chain. seems wrong
```
syntheticmagus, line 387: nit: When this goes in, add an issue?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -438,6 +527,13 @@ namespace xr
             swapchainCreateInfo.createFlags = createFlags;
             swapchainCreateInfo.usageFlags = usageFlags;
 
+            XrSecondaryViewConfigurationSwapchainCreateInfoMSFT secondaryViewConfigCreateInfo{
+                XR_TYPE_SECONDARY_VIEW_CONFIGURATION_SWAPCHAIN_CREATE_INFO_MSFT};
+            if (HmdImpl.Extensions->SecondaryViewConfigurationSupported) {
```
syntheticmagus, line 532: nit: Style, this document has condition block braces on newlines. We might should consider adding a clang format to this dependency just to make it easy to run auto-format.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1968,7 +1998,7 @@ namespace Babylon
 
         private:
             Napi::ObjectReference m_jsSession{};
-            std::array<Napi::ObjectReference, 2> m_jsRenderTargetTextures;
+            std::array<Napi::ObjectReference, 3> m_jsRenderTargetTextures;
```
syntheticmagus, line 2001: nit: If this no longer has a guaranteed two elements (which I guess probably hasn't been true for a while now), should it be a vector now instead of an array?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -178,12 +211,11 @@ namespace xr
...
-            std::vector<XrViewConfigurationView> ConfigViews{};
-            std::vector<Swapchain> ColorSwapchains{};
-            std::vector<Swapchain> DepthSwapchains{};
-            std::vector<XrCompositionLayerProjectionView> ProjectionLayerViews{};
-            std::vector<XrCompositionLayerDepthInfoKHR> DepthInfoViews{};
+            std::unordered_map<XrViewConfigurationType, ViewConfigurationState> ViewStates{};
+            std::unordered_map<XrViewConfigurationType, std::vector<Swapchain>> ColorSwapchains{};
+            std::unordered_map<XrViewConfigurationType, std::vector<Swapchain>> DepthSwapchains{};
+            std::unordered_map<XrViewConfigurationType, std::vector<XrCompositionLayerProjectionView>> ProjectionLayerViews{};
+            std::unordered_map<XrViewConfigurationType, std::vector<XrCompositionLayerDepthInfoKHR>> DepthInfoViews{};
```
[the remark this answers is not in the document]
chrisfromwork: sounds good

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: Do iOS and Android use more than one render target texture? Right now, it looks like we pre-construct these textures when creating the NativeRenderTargetProvider. We may be creating two more textures than needed when executing the iOS and Android scenarios.

It seems a bit unclear whether GetRenderTargetForEye is called based on information provided in the NativeRenderTargetProvider callback. This makes me hesitant in looking into dynamically creating render targets textures as needed. It feels a bit beyond the scope of this review.

Dependencies/xr/Source/OpenXR/XR.cpp
chrisfromwork: this should now be fixed.

Dependencies/xr/Source/OpenXR/XR.cpp
chrisfromwork: this is no longer needed

Dependencies/xr/Source/OpenXR/XR.cpp
chrisfromwork: I have updated declarations for secondary view types

Dependencies/xr/Source/OpenXR/XR.cpp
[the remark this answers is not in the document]
chrisfromwork: this should now be fixed

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: this should now be addressed

PR #370, opened 2020-08-15
Shader traversers comments

syntheticmagus:
Added LOTS of comments and clarifications about how shader traversers work, hopefully making that code a little more approachable.


Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -237,9 +294,17 @@ namespace Babylon::ShaderCompilerTraversers
...
+            /// Because no accumulation or cross-correlation is necessary for this change
+            /// (i.e. each operation acts only on a single symbol), we do all the work inside
+            /// the traverser itself.
+            /// </summary>
+            /// <param name="symbol"></param>
             void visitSymbol(TIntermSymbol* symbol) override
             {
                 auto& type = symbol->getType();
+
+                // We only care about uniforms that are neither samplers nor matrices.
```
CedricGuillemet, line 307: bgfx only supports 4x4 matrices. If we have some 3x3 rotation matrices, it has to be replaced as well. the c++ code already does the NxM to 4x4 matrix : https://github.com/BabylonJS/BabylonNative/blob/85e0731e2eab931f2e500996d75d3fa7a0d36bce/Plugins/NativeEngine/Source/NativeEngine.cpp#L1177

CedricGuillemet:
Great documentation! It makes everything more clear. Thank you!
I'm curious about the performance. The amount of virtual calls/checks and allocation looks like it's a lot of work per shader.
Could you run some profiling on shader with medium complexity found in marbletower?
If the time for conversion magnitude is around 1ms, no need for further work. If it's in the 10-100ms, it might be interesting to push that work on the thread pool.

bghgary:
Using a thread pool is required as it will certainly cause hitching on slower devices. Shader compilation should be async.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
syntheticmagus: Created #401 to track the work of proving and solving this issue.

syntheticmagus:
Good points all, thanks for bringing those up! Created #402 to track this, as this definitely seems like a bigger bite than should be chewed in a documentation PR. 😃

CedricGuillemet (approved):

PR #374, opened 2020-08-19
Use correct buffer size in SetViewport

rgerd:
We were seeing a bug on some WMR HMD's as well as on HoloLens 2 where a region of the left eye view was cleared to the background color, and the region size corresponded to the size of the window created during initialization.

This was caused by two things: framebuffers being assigned view id's multiple times, and a framebuffer sharing the default framebuffer's view id after the default framebuffer is no longer used.

This fix keeps the default frame buffer distinct, as its size is tied to `bgfx::getStats()->[Width|Height]()`, so that the other bound framebuffers can be cleared properly and separately at their true sizes.


bghgary:
What's the status of this?

rgerd:
Thanks for pinging this. Almost forgot I made it. I will actually have a totally different PR which is related to this but provides an actual fix for (#387). Gonna close this one.

PR #376, opened 2020-08-21
Documentation

syntheticmagus:
Added six new documentation pages about a plethora of topics that may be informative/interesting to newer developers looking at the Babylon Native repository.

These pages are, at present _very_ top-of-my-head and neither the topics nor the content have been run by anyone else. I have also not proofread these as thoroughly as I would have liked, so I apologize in advance for typos and omitted words. Some of those I will catch on my own as I reread and revise in the coming days, but please don't hesitate to point out any errors you find; and, of course, all content- or topic-related questions/comments/concerns are very welcome and greatly appreciated. Thanks!


Documentation/Components.md
```diff
@@ -0,0 +1,246 @@
...
+This document gives a high-level overview of the components provided by 
+default in the Babylon Native repository. For information about the 
+definition of _component_ and the different categories of components, 
+please read the page on the 
+[Babylon Native build system](BuildSystem.md). For information about how to 
+create new components and add or reference them in Babylon Native, please
+read the page on 
+[extending Babylon Native](Extending.md).
+
+## Dependencies
```
bghgary, line 12: This seems like it should be merged with https://github.com/BabylonJS/BabylonNative#development-notes?

Documentation/Components.md
```diff
@@ -0,0 +1,246 @@
...
+
+Components in this category encapsulate functionality brought in from 
+dependencies which are not logically a part of Babylon Native. Most of these
+are external libraries, and many of them are included as submodules.
+
+### arcana
+
+[Arcana](https://github.com/microsoft/arcana.cpp) is a collection of C++ 
+utilities. Most prominently, it includes an open-source and cross-platform 
+task and asynchrony system modelled after the
```
bghgary, line 22: nit: modeled (for American english)
```suggestion
task and asynchrony system modeled after the 
```

Documentation/Components.md
```diff
@@ -0,0 +1,246 @@
...
+C++ utility library supplying several useful types and capabilities that
+are not supplied by the STL but are conceptually similar. It is used 
+internally throughout Babylon Native, but no GSL types are exposed by any
+Babylon Native API.
+
+### napi
+
+N-API is an integration layer that allows C and C++ native libraries to 
+interface with JavaScript engines in an engine-agnostic manner. Babylon
+Native's consumption of N-API is a heavily-modified spinoff of the
```
bghgary, line 61: > heavily-modified spinoff

I'm not sure this is accurate. Maybe merge with the wording we have from the README.

Documentation/Components.md
```diff
@@ -0,0 +1,246 @@
...
+[dedicated documentation page for the Playground app](Playground.md).
+
+### ValidationTests
+
+The ValidationTests app is exactly what its name implies: it is a mechanism
+for easily running validation tests to ensure Babylon Native continues to 
+run and render correctly under expected circumstances. The ValidationTests
+are designed to be run by the Babylon Native team as part of common 
+workflows and CI and are not intended to be highly customizable. The only
+supported platform for the validation tests is Win32.
```
bghgary, line 246: > The only supported platform for the validation tests is Win32.

I think you should omit this. We don't know if this will be true later.

Documentation/Playground.md
```diff
@@ -0,0 +1,143 @@
...
+being developed, sometimes checking in the change with an enabling flag if
+they feel that other developers may also benefit from the same test. 
+There are, however, a few additional features and use cases that are less 
+obvious.
+
+Though designed to execute experience.js by default, the Playground app is 
+also capable of loading and executing arbitrary JavaScript files. How this 
+is done varies by platform -- Win32 allows JS files to be passed as 
+command line arguments while UWP allows them to be executed via protocal 
+invocation, for example -- but the behavior is the same on all platforms
```
bghgary, line 26: typically, it is called protocol activation
```suggestion
activation, for example -- but the behavior is the same on all platforms
```

Documentation/Playground.md
```diff
@@ -0,0 +1,143 @@
...
+requested external JS file, then execute a special "playground runner" 
+intended to call behaviors that may have been defined in the requested JS
+file. Essentially, this allows for externally-defined JS files to control 
+the behavior of the Playground app as an alternative to experience.js --
+at runtime and without the need to recompile or redeploy.
+
+The Playground app also features a "hot reload" capability on certain
+platforms (Win32 and UWP, for now). The current implementation is that 
+pressing a button (`R` at present, likely to be remapped to `F5` in the 
+future) while the app is in focus causes the Playground to tear down
```
bghgary, line 38: > (`R` at present, likely to be remapped to `F5` in the future)

I recommend removing this and things like this to avoid having to update this documentation when things change.

bghgary (approved):
This is great and captures in much greater detail what is written in the README. I have some small comments.

Documentation/Components.md
syntheticmagus: We might could do so to some degree, but in my mind those two pieces of documentation serve meaningfully different purposes. The section in the README is, from what I recall, something along the lines of an "external credits" list intended to call out 3rd party dependencies we felt should be given particular notice. (It may be woefully out-of-date for this purpose.) `Components.md`, on the other hand, I specifically wrote to be a touchpoint for new Babylon Native developers who just built their first BabylonNative.sln and want some high-level information information about what all these components they're seeing are, regardless of the origin and nature of those components. Consequently, things like base-n show up in the README but not here since it doesn't show up as a component (it's a header-only implementation detail of UrlLib, I think) while things like UrlLib show up here but nowhere else because it's not external. `Components.md` also goes out of its way to clarify several cases where things are and aren't part of externally-exposed contracts. For these reasons, I'm inclined to keep these lists distinct so that they can serve separate purposes. I don't think it's hugely important, though, so I'm open to merging them if you feel strongly that that's a better thing to do.

Documentation/Playground.md
syntheticmagus: Unless you feel strongly about this one, I'd prefer to leave this and have us take on the responsibility of keeping these docs up-to-date because I feel we need a place we can send new Babylon Native developers and say, "Read this," and have it really tell them the sorts of things they can do. This documentation page in particular will, I suspect, need major updates in the near future anyway as the vision/implementation for the Playground app develops. It is a bit awkward to have documentation specifically calling out things that are a certain way now but will likely change in known ways soon. The whole `Playground.md` is like that, though, and my intention for it was to serve as a place we could direct the people who are looking to jump into Babylon Native right now so that they could get context on the current conversation about the Playground app (incomplete as that conversation is) and, if they want to, get involved. Thoughts?

Documentation/Components.md
bghgary: > "external credits" list intended to call out 3rd party dependencies we felt should be given particular notice

I didn't write it to be that. I wrote it as a call-out to certain things that were happening in the code. I think if you move some of the info into your section, the README section would be unnecessary.

> base-n

This is used by the Window polyfill for `atob`. I don't think we need to call this out at all.

Documentation/Playground.md
bghgary: Either way is fine.

Issue #380, opened 2020-08-26
XR dependency component should expose OpenXR components for usage by extension plugins


chrisfromwork:
The xr component for OpenXR currently contains things like the OpenXR session and OpenXR supported extensions. When writing extension components for preview OpenXR extensions, it's necessary to access these components to make the appropriate OpenXR calls. In order to build preview OpenXR functionality in extensions, we will need to expose a way to access the OpenXR session from places outside the xr component.

PR #381, opened 2020-08-26
Add XrRegistry for accessing OpenXR components in plugins outside of xr component

chrisfromwork:
This review addresses issue: https://github.com/BabylonJS/BabylonNative/issues/380

It contains the following changes:
1. Headers in xr component for windows OpenXR builds have been moved to the include directory and can now be referenced by plugins referencing the xr component. SupportedExtensions was also renamed to XrSupportedExtensions for consistency in naming.
2. The xr component now consumes openxr headers provided by the OpenXR-MixedReality repo compared to those in the OpenXR-SDK repo. This gives access to MSFT preview extensions that aren't distributed with the OpenXR-SDK repo.
3. XrRegistry, XrSessionContext and XrSessionContext::Impl have been defined to enable access to OpenXR components outside the xr plugin. XR.cpp logic now instantiates OpenXR components in a static XrSessionContext::Impl. This allows a singleton location for obtaining the OpenXR components while restricting the ability to manipulate/update content.

It would be great if reviewers could specifically provide feedback around the changes described in 3 before we convert this into a non-draft pull request. The current mechanism for accessing XrSession and XrSpace isn't thread safe.


Dependencies/xr/CMakeLists.txt
```diff
@@ -49,11 +47,12 @@ elseif (IOS)
...
 else()
     if(WIN32)
-        target_include_directories(xr PRIVATE "Source/OpenXR/Windows")
+        target_include_directories(xr PRIVATE "Include/OpenXR/Windows")
     endif()
 
-    target_include_directories(xr PRIVATE "Dependencies/OpenXR-SDK/include")
-    target_include_directories(xr PRIVATE "Source/OpenXR")
+    target_include_directories(xr PUBLIC "Dependencies/OpenXR-MixedReality/openxr_preview/include")
+    target_include_directories(xr PUBLIC "Dependencies/OpenXR-MixedReality/shared/XrUtility")
```
syntheticmagus, line 54: Publicly exposing an explicit OpenXR dependency is a pretty significant paradigm shift for XR, which to date has represented an opaque integration point allowing consuming logic to be completely decoupled from the underlying platform. An explicit OpenXR dependency isn't necessarily a bad thing, certainly in the long term; but this might be premature since it will open the door for comingled XR- and OpenXR-dependent code at a time when OpenXR isn't yet available on all the platforms we need to support.

I believe the intent here is to allow specific dependent consumers to access internal implementation details that aren't part of the broadly-supported core contract. If this is the case, it might be better to expose these as a separate secondary contract, similar to how we create [`Internal` contracts](https://github.com/BabylonJS/BabylonNative/blob/master/Plugins/NativeWindow/CMakeLists.txt#L20-L24) for certain Plugins that need to provide a deeper level of visibility to some, but not all, consumers. This would allow the core implementation-agnostic to remain intact and platform independent while still allowing extensions that need them to have more invasive access to the implementation details of the platform.

chrisfromwork: Although an example extension isn't provided here, we want to write extensions that specifically take an openxr dependency. We also don't know that these openxr specific extensions will ever end up in the babylonnative repo. Some of the functionality in these extensions doesn't have webxr counterparts. There are also a wide variety of openxr types (XrInstance, XrSystemId, XrSession, frame time, etc) that need to be exposed to these OpenXR-specific extensions. We could refactor the xr component to somehow expose these component types, but I have the following concerns in doing this:
1. XR components won't map 1:1 across different platforms. Accessing a session on OpenXR may not have a parallel on ARCore/ARKit
1. Exposing generic components through the existing xr contract requires some sort of mechanism for obscuring types. This leaves more room for error around memory management/component lifecycles if we start handing around void pointers.
1. It will be difficult to maintain an xr component querying contract that is actually implemented by all platforms. We've seen a variety of instances where arcore/arkit work has already tacked in non-operable functionality in the OpenXR xr component. I would prefer to avoid doing this moving forward.

I agree with your observation that the intent is to access internal implementation details that aren't part of the broadly-supported core contract. Is there an example for where NativeWindowInternal is used today? It may be that Internal contracts would fulfill our needs but I can't say I fully understand the intention/implementation details of an internal contract based on that cmake definition.

chrisfromwork: chatted offline, it made since to turn this into an internal contract. cmake files have been updated to reflect this ask.

tymlipari (approved):

rgerd (approved):

Dependencies/xr/Source/OpenXR/XrRegistry.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include "XrSupportedExtensions.h"
+
+#ifndef XR_REGISTRY
+#define XR_REGISTRY
```
syntheticmagus, line 6: nit: Is this idiom not redundant on `#pragma once`?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -75,14 +26,61 @@ namespace xr
...
+
+            extensions->PopulateDispatchTable(instance.Get());
+        }
+
+        XrTime GetDisplayTime()
+        {
+            return displayTime;
+        }
+
+        InstanceHandle instance;
```
syntheticmagus, line 51: nit: Stylistically, the convention in the broader Babylon Native system is to capitalize public struct members (and use `m_` for private members). I'm fine with this being left for a later clang format to pick up on, though; it's also fine for the XR dependency to establish and use different conventions.

syntheticmagus (approved):

Dependencies/xr/Source/OpenXR/XrRegistry.h
chrisfromwork: this should now be fixed

Dependencies/xr/Source/OpenXR/XR.cpp
chrisfromwork: this should now be fixed

PR #384, opened 2020-08-31
Implement OpenXR anchoring polyfill

rgerd:
This change provides OpenXR implementations for the functions required for creating, updating, and deleting anchors through an XRFrame. Essentially an OpenXR analog to (#289).

### DIFF
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -387,8 +387,75 @@ namespace xr
             renderResource.DepthSwapchains.clear();
         }
 
+        Anchor CreateAnchor(Pose pose, NativeTrackablePtr trackable, XrTime time)
+        {
+            auto& apiExtensions = HmdImpl.Extensions;
+
+            if (!apiExtensions->SpatialAnchorSupported)
+                throw std::runtime_error("Spatial anchors are not supported for this device.");
+
+            if (trackable != nullptr)
+                throw std::runtime_error("Anchors created from hit test results not yet implemented for OpenXR.");
+
+            auto anchorSpace = std::make_shared<OpenXRAnchorSpace>();
+
+            XrSpatialAnchorCreateInfoMSFT anchorCreateInfo{XR_TYPE_SPATIAL_ANCHOR_CREATE_INFO_MSFT};
+            anchorCreateInfo.space = SceneSpace;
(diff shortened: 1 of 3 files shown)


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -387,9 +387,76 @@ namespace xr
             renderResource.DepthSwapchains.clear();
         }
 
+        Anchor CreateAnchor(Pose pose, NativeTrackablePtr trackable, XrTime time)
+        {
+            auto& apiExtensions = HmdImpl.Extensions;
+
```
chrisfromwork, line 393: nit: consider const auto& to avoid accidentally setting the extensions unique_ptr

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -387,9 +387,76 @@ namespace xr
...
+            if (!apiExtensions->SpatialAnchorSupported)
+                throw std::runtime_error("Spatial anchors are not supported for this device.");
+
```
chrisfromwork, line 396: nit: for consistency please put this and the below throw in their own brackets compared to not having a bracket

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -387,9 +387,76 @@ namespace xr
...
+            const auto nativeAnchorPtr = reinterpret_cast<NativeAnchorPtr>(anchorSpace.get());
+            openXRAnchors[nativeAnchorPtr] = anchorSpace;
+            return { pose, nativeAnchorPtr };
+        }
+
+        void UpdateAnchor(Anchor& anchor, XrTime time) const
+        {
+            auto anchorSpace = openXRAnchors.at(anchor.NativeAnchor);
+
```
chrisfromwork, line 426: Check if this key exists

Dependencies/xr/Source/OpenXR/XR.cpp
syntheticmagus: Could also dereference the pointer if you wanted to avoid arrow syntax.

```
const auto& apiExtensions = *HmdImpl.Extensions;
```

Dependencies/xr/Source/OpenXR/XR.cpp
tymlipari: `.at` will throw an exception if the key isn't present in the map. Would we want to throw a different exception, perhaps with a string that explains the issue? Or would we want to do something like `if (itr == openXRAnchors.end()) return;` and skip the function entirely?

PR #389, opened 2020-09-02
[iOS + XR] Fix camera rotation for portrait display orientations

ryantrem:
This bug manifested as the transformation of screen coordinates to picking rays resulting in rays with the x/y reversed from what was expected. Ultimately the problem was that when we rotate the camera pose around the z axis for portrait and portrait upside down display orientations, the signs were swapped. Therefore, in regular portrait mode, the camera was actually upside down. There was other code that tried to account for this by flipping the UVs for both the Babylon texture and the camera texture. So the fix here is to:
1. Rotate the camera pose around the z-axis the correct direction for portrait/portrait upside down display orientations.
1. Don't reverse the UVs for the Babylon texture (note that the V is still swapped, but this is expected based on UV conventions in Metal - see https://developer.apple.com/documentation/metal/creating_and_sampling_textures).
1. Don't reverse the UVs for the camera texture.


bghgary (approved):

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -318,10 +310,10 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
         displayOrientationQuat = simd_quaternion(0.0f, 0.0f, 0.0f, 1.0f);
     }
```
Alex-MSFT, line 311: Did you double check the landscape orientation modes as well?  I would think these would also need to be flipped.

Alex-MSFT (approved):

Dependencies/xr/Source/ARKit/XR.mm
ryantrem: You're right, good catch.

PR #395, opened 2020-09-11
Workaround for crash on startup on devices without XR runtimes.

syntheticmagus:
A recent change to XR introduced the XrRegistry, which included a static member whose constructor depended upon the existence of an OpenXR runtime. This was causing the app to crash on initialization when running on a device without an OpenXR runtime because the constructor of the XrSessionContext would fail.

To avoid this, this PR switches from a static member to a global singleton, thereby avoiding static-ness as well. This isn't my favorite pattern, but it's the simplest adaptation I've thought of so far that behaves similarly to the original intent while not crashing if OpenXR isn't available. Definitely open to other designs, though, if a better one comes to mind.

### DIFF
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -24,9 +24,9 @@ namespace xr
 
             return swapchainImageIndex;
         };
-    }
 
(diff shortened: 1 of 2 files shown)


chrisfromwork (approved):
:shipit:

PR #399, opened 2020-09-14
CI and artifacts Documentation

CedricGuillemet:
I'm not sure we are all on the same page with the current state of the test infrastructure.
Perfect time for some documentation on CI/ValidationTests and artifacts then :)


Documentation/ContinuousIntegration.md
```diff
@@ -0,0 +1,145 @@
...
+- The test datas
+- The ValidationTests app
+- The platform integration
+
+# Test datas
+They are shared with babylonjs for the most part.
+Validation test scenes are listed in ```Apps\ValidationTests\Scripts\config.json```. It lists playground identifier that will be opened sequentialy. Per PG is may have additional parameters for validation pixels difference threshold.
+This json is a subset of babylonjs' one. The goal is to use the exact same file once the engine is in a sufficiently stable state.
+
+```
```
ryantrem, line 18: Specify the content type to improve syntax highlighting in some markdown renderers.
```suggestion
```json
```

Documentation/ContinuousIntegration.md
```diff
@@ -0,0 +1,145 @@
...
+CI will also take care of code quality: Warnings are treated as error and will break the build. A PR can't be merged if the build fails.
+
+CI validation is composed of 3 main parts:
+- The test datas
+- The ValidationTests app
+- The platform integration
+
+# Test datas
+They are shared with babylonjs for the most part.
+Validation test scenes are listed in ```Apps\ValidationTests\Scripts\config.json```. It lists playground identifier that will be opened sequentialy. Per PG is may have additional parameters for validation pixels difference threshold.
```
ryantrem, line 15: Single back ticks should produce the same result for inline code snippets.
```suggestion
Validation test scenes are listed in `Apps\ValidationTests\Scripts\config.json`. It lists playground identifier that will be opened sequentialy. Per PG is may have additional parameters for validation pixels difference threshold.
```

Documentation/ContinuousIntegration.md
```diff
@@ -0,0 +1,145 @@
...
+            "title": "setParent",
+            "playgroundId": "#JD49CT#2",
+            "referenceImage": "setParent.png"
+        },
+        ...
+}
+```
+
+Once the PG is loaded and a screenshot is saved on disk, it's compared to a reference image.
+Those images are accessible here : https://github.com/BabylonJS/Babylon.js/tree/master/tests/validation/ReferenceImages
```
ryantrem, line 32: Not all markdown renderers will automatically create links for urls.
```suggestion
Those images are accessible here : [https://github.com/BabylonJS/Babylon.js/tree/master/tests/validation/ReferenceImages](https://github.com/BabylonJS/Babylon.js/tree/master/tests/validation/ReferenceImages)
```

Documentation/ContinuousIntegration.md
```diff
@@ -0,0 +1,145 @@
...
+
+If a new scene to test is added in the json file, a new reference must be added to babylonjs repo.
+If a new feature or a bug fix makes the reference image different, it must also be updated.
+
+# ValidationTests app
+
+To make it simple, the validationTests app initialize the engine and runs ```Apps\ValidationTests\Scripts\validation_native.js```.
```
ryantrem, line 39:
```suggestion
To make it simple, the validationTests app initialize the engine and runs `Apps\ValidationTests\Scripts\validation_native.js`.
```

Documentation/ContinuousIntegration.md
```diff
@@ -0,0 +1,145 @@
...
+That js script is responsible for loading the json, running each test PG, make the screen capture and compare images.
+When completed, the executable returns a value (0 for success, any other value for failure) that is used by the pipeline to determine the build success.
+
+## The app executable
+
+It's very similar to the playground app: 
+- creates a window
+- initialize engine and JS standard components
+- initialize validation JS components
+- loads and ```validation_native.js```
```
ryantrem, line 49:
```suggestion
- loads and `validation_native.js`
```

And same with the rest, I'm going to stop commenting on them :)

Documentation/ContinuousIntegration.md
```diff
@@ -0,0 +1,145 @@
...
+- returns a value depending on success
+
+The validation JS components are function needed for the execution of the validation script.
+It adds features like loading and saving .png from a byte array, asks the renderer to do a screen shot, get working directories and exiting the app.
```
ryantrem, line 53: > asks the renderer to do a screen shot

How does this work? I thought we didn't currently have a way to get a screen shot in Babylon Native?

ryantrem (approved):

Documentation/ContinuousIntegration.md
CedricGuillemet: We do have a screencapture function:
https://github.com/BabylonJS/BabylonNative/blob/adf2b5af40d8a8b066f93ab4a5245825980a86ac/Plugins/NativeEngine/Source/NativeEngine.cpp#L1657
It's async so you can provide a JS function to handle the uint8array parameter:
https://github.com/BabylonJS/BabylonNative/blob/adf2b5af40d8a8b066f93ab4a5245825980a86ac/Apps/ValidationTests/Scripts/validation_native.js#L49
I use it to do the comparison and there is a function to save that bytearray as a .png
https://github.com/BabylonJS/BabylonNative/blob/adf2b5af40d8a8b066f93ab4a5245825980a86ac/Apps/ValidationTests/Scripts/validation_native.js#L43
This only works with the back buffer. If the rendering is done in a rendertarget, we need another solution. For XR, the mix between 3D and camera is done per platform with a specific rendering code.
If we want to capture AR with the camera + the 3D, that could work if the mixed rendering was done with bjs/bgfx.

PR #400, opened 2020-09-14
Implementing hand tracking via OpenXR for BabylonNative

### LINKED ISSUE #394: Enable articulated hand tracking in babylonNative for OpenXR
Enable hand tracking for use with babylonNative through openXR, now that babylonJS has support for it.

rickfromwork:
This change introduces the hand tracking backbone in XR.cpp and NativeXR.cpp to implement hand tracking when using openXR in BabylonNative. It works with the existing hand tracking code for BabylonJS, found [here](https://github.com/BabylonJS/Babylon.js/pull/8753).

This resolves #394

### DIFF
--- a/Dependencies/xr/Include/XR.h
+++ b/Dependencies/xr/Include/XR.h
@@ -140,6 +140,12 @@ namespace xr
                     Pose Pose;
                 };
 
+                struct JointSpace : Space
+                {
+                    bool PoseTracked{ false };
+                    float PoseRadius{};
+                };
+
                 struct View
                 {
                     Space Space{};
@@ -178,9 +184,11 @@ namespace xr
 
                     const Identifier ID{ NEXT_ID++ };
                     bool TrackedThisFrame{};
+                    bool JointsTrackedThisFrame{};
                     Space GripSpace{};
                     Space AimSpace{};
                     HandednessEnum Handedness{};
+                    std::vector<JointSpace> HandJoints{};
 
                 private:
                     static inline Identifier NEXT_ID{ 0 };
--- a/Dependencies/xr/Include/XR.h
+++ b/Dependencies/xr/Include/XR.h
@@ -142,8 +142,8 @@ namespace xr
 
                 struct JointSpace : Space
                 {
-                    bool PoseTracked{ false };
                     float PoseRadius{};
+                    bool PoseTracked{ false };
                 };
 
                 struct View
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -259,6 +259,33 @@ namespace xr
             std::vector<Frame::Plane> Planes{};
             std::vector<FeaturePoint> FeaturePointCloud{};
         } ActionResources{};
+            
+        struct HandJointData {
+            void InitializeXRHandJoints()
+            {
+                locations.jointCount = XR_HAND_JOINT_COUNT_EXT;
+                locations.jointLocations = jointLocations;
+            };
+
+            XrHandJointLocationEXT jointLocations[XR_HAND_JOINT_COUNT_EXT];
+            XrHandJointLocationsEXT locations{XR_TYPE_HAND_JOINT_LOCATIONS_EXT};
+        };
+
+        struct HandInfo {
+            XrHandEXT hand{};
+            XrHandTrackerEXT handTracker{};
+            XrSpace handSpace{};
+
+            HandJointData handJointData{};
+        };
+        
+        struct
+        {
+            XrBool32 supportsHandTracking{ false };
+            XrBool32 HandsInitialized{ false };
+
+            std::array<HandInfo, 2> handsInfo;
+        } HandData;
 
         float DepthNearZ{ DEFAULT_DEPTH_NEAR_Z };
         float DepthFarZ{ DEFAULT_DEPTH_FAR_Z };
@@ -317,7 +344,9 @@ namespace xr
         void RequestEndSession()
         {
             const auto& session = HmdImpl.Context.Session();
-            xrRequestExitSession(session);
+
+            UninitializeHandResources();
+            xrRequestExitSession(Session);
         }
 
         Size GetWidthAndHeightForViewIndex(size_t viewIndex) const
@@ -485,9 +514,14 @@ namespace xr
 
         void InitializeRenderResources(XrInstance instance, XrSystemId systemId)
         {
-            // Read graphics properties for preferred swapchain length and logging.
-            XrSystemProperties systemProperties{ XR_TYPE_SYSTEM_PROPERTIES };
+            // Read graphics properties for preferred swapchain length and logging, and hand tracking availability.
+            XrSystemHandTrackingPropertiesEXT handTrackingSystemProperties{ XR_TYPE_SYSTEM_HAND_TRACKING_PROPERTIES_EXT };
+            XrSystemProperties systemProperties{ XR_TYPE_SYSTEM_PROPERTIES, &handTrackingSystemProperties };
             XrCheck(xrGetSystemProperties(instance, systemId, &systemProperties));
+
+            // Initialize the hand resources
+            HandData.supportsHandTracking = handTrackingSystemProperties.supportsHandTracking && HmdImpl.Extensions->HandTrackingSupported;
+            InitializeHandResources();
 
             const XrViewConfigurationType primaryType = HmdImpl.PrimaryViewConfigurationType;
             auto& primaryRenderResource = RenderResources.ResourceMap[primaryType];
@@ -499,6 +533,44 @@ namespace xr
 
             // Create the swapchains for the primary view configuration type. Secondary view configuration type swapchains will be populated once activated.
             PopulateSwapchains(primaryRenderResource.ViewState);
+        }
+
+        void InitializeHandResources()
+        {
+            if (!HandData.supportsHandTracking)
+            {
+                return;
+            }
+
+            constexpr std::array<XrHandEXT, 2> HANDEDNESS_EXT{XR_HAND_LEFT_EXT, XR_HAND_RIGHT_EXT};
+            XrHandTrackerCreateInfoEXT trackerCreateInfo{XR_TYPE_HAND_TRACKER_CREATE_INFO_EXT};
+            trackerCreateInfo.handJointSet = XR_HAND_JOINT_SET_DEFAULT_EXT;
+
+            for (int i = 0; i < HandData.handsInfo.size(); i++)
+            {
+                // Create the hand trackers
+                HandData.handsInfo[i].hand = HANDEDNESS_EXT[i];
+                trackerCreateInfo.hand = HANDEDNESS_EXT[i];
+                XrCheck(HmdImpl.Extensions->xrCreateHandTrackerEXT(Session, &trackerCreateInfo, &HandData.handsInfo[i].handTracker));
+                    
+                // Preallocate joint buffers
+                HandData.handsInfo[i].handJointData.InitializeXRHandJoints();
+            }
+
+            HandData.HandsInitialized = true;
+        }
+
+        void UninitializeHandResources()
+        {
+            if (HandData.HandsInitialized)
+            {
+                for (HandInfo handInfo : HandData.handsInfo)
+                {
+                    HmdImpl.Extensions->xrDestroyHandTrackerEXT(handInfo.handTracker);
+                }
+
+                HandData.HandsInitialized = false;
+            }
         }
 
         void InitializeActionResources(XrInstance instance)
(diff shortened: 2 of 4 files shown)


rickfromwork:
Waiting on a change in babylon.js to be completed before I can update the native code for js, and make this a proper PR.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2215,6 +2293,9 @@ namespace Babylon
             NativeWebXRRenderTarget::Initialize(env);
             NativeRenderTargetProvider::Initialize(env);
             XR::Initialize(env);
+
+            // XRHand needs to be defined in order to use hand tracking, but we don't actually need the interface
+            env.Global().Set("XRHand", true);
         }
```
chrisfromwork, line 2299: nit: its probably worth having an XRHand interface for consistency. long term i am pretty sure XRHand is supposed to be iteratable/may need a constructro compared to us being able to get by just setting XRHand to a boolean

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1505,6 +1565,24 @@ namespace Babylon
                 }
             }
 
+            Napi::Value GetJointPose(const Napi::CallbackInfo& info)
+            {
+                assert(info[0].IsExternal());
+
```
chrisfromwork, line 1571: this will only throw in debug builds. Do we want to have this function return undefined if this check fails in release builds?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -408,6 +444,44 @@ namespace xr
...
+                HandData.handsInfo[i].hand = HANDEDNESS_EXT[i];
+                trackerCreateInfo.hand = HANDEDNESS_EXT[i];
+                XrCheck(HmdImpl.Extensions->xrCreateHandTrackerEXT(Session, &trackerCreateInfo, &HandData.handsInfo[i].handTracker));
+                    
+                // Preallocate joint buffers
+                HandData.handsInfo[i].handJointData.InitializeXRHandJoints();
+            }
+
+            HandData.HandsInitialized = true;
+        }
```
chrisfromwork, line 470: is there any concern around reinitializing hand data?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -991,6 +1065,60 @@ namespace xr
...
+                    auto& inputSource = InputSources[idx];
+                    inputSource.JointsTrackedThisFrame = handInfo.handJointData.locations.isActive;
+                        
+                    // Set up the handJoints vector, skipping the palm joint as webXR doesn't support it
+                    // xrLocateHandJointsEXT will always return the full joint set (or an error), so jointCount should never change
+                    if (inputSource.HandJoints.size() != handInfo.handJointData.locations.jointCount - 1)
+                    {
+                        inputSource.HandJoints = std::vector<JointSpace>(handInfo.handJointData.locations.jointCount - 1);
+                        assert(inputSource.HandJoints.size() == handInfo.handJointData.locations.jointCount - 1);
+                    }
```
chrisfromwork, line 1088: would this ever be false?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -991,6 +1065,60 @@ namespace xr
...
+
+                    auto handInfo = sessionImpl.HandData.handsInfo[idx];
+                    XrCheck(sessionImpl.HmdImpl.Extensions->xrLocateHandJointsEXT(handInfo.handTracker, &jointLocateInfo, &handInfo.handJointData.locations));
+                    
+                    auto& inputSource = InputSources[idx];
+                    inputSource.JointsTrackedThisFrame = handInfo.handJointData.locations.isActive;
+                        
+                    // Set up the handJoints vector, skipping the palm joint as webXR doesn't support it
+                    // xrLocateHandJointsEXT will always return the full joint set (or an error), so jointCount should never change
+                    if (inputSource.HandJoints.size() != handInfo.handJointData.locations.jointCount - 1)
```
chrisfromwork, line 1084: >handInfo.handJointData.locations.jointCount - 1 [](start = 57, length = 47)

nit: given this value is reused multiple locations, it may be worth caching in a variable for legibility

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -991,6 +1065,60 @@ namespace xr
...
+                    {
+                        constexpr XrSpaceLocationFlags RequiredFlags =
+                            XR_SPACE_LOCATION_POSITION_VALID_BIT |
+                            XR_SPACE_LOCATION_ORIENTATION_VALID_BIT |
+                            XR_SPACE_LOCATION_POSITION_TRACKED_BIT |
+                            XR_SPACE_LOCATION_ORIENTATION_TRACKED_BIT;
+
+                        for (uint32_t i = 0; i < handInfo.handJointData.locations.jointCount - 1; i++)
+                        {
+                            auto joint = handInfo.handJointData.jointLocations[i + 1];
```
chrisfromwork, line 1101: > + 1 [](start = 80, length = 4)

nit: for legibility it may help to declare this as a constant JOINT_OFFSET = 1 somewhere compared to just having an arbitrary increment here

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -252,6 +254,33 @@ namespace xr
...
+            HandJointData handJointData{};
+        };
+        
+        struct
+        {
+            XrBool32 supportsHandTracking{ false };
+            XrBool32 HandsInitialized{ false };
+
+            std::array<HandInfo, 2> handsInfo;
+        } HandData;
```
chrisfromwork, line 283: >HandData [](start = 10, length = 8)

nit: would anyone ever want to access this hand data in a custom extension? Are there different hand related openxr extensions that may never map to webxr? It may be worth considering whether to add this to the XrRegistry/XrSessionContext compared to having it inaccessible to other components here.

If there are no known scenarios where this information would be needed outside of the hand tracking implementation in this component, theres no need to move this.

chrisfromwork (approved):
:shipit:

Dependencies/xr/Include/XR.h
```diff
@@ -140,6 +140,12 @@ namespace xr
                     Pose Pose;
                 };
 
+                struct JointSpace : Space
```
tymlipari, line 143: nit: put the `bool` after the `float`. The compiler is likely padding out to 32 bits right now.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -252,6 +254,33 @@ namespace xr
             std::vector<Frame::Plane> Planes{};
             std::vector<FeaturePoint> FeaturePointCloud{};
         } ActionResources{};
+            
+        struct HandJointData {
```
tymlipari, line 258: nit: inconsistent formatting. Some of the opening braces are on same line, some are on next line.

Dependencies/xr/Source/OpenXR/XR.cpp
tymlipari, line 283: nit: formatting consistency, put the struct name at the beginning instead of the end

tymlipari: It sounds like there are some HL2-specific pieces at the moment (e.g. pointing/arm ray) that aren't part of the WebXR spec and that aren't currently known to be supported on other devices (e.g. Quest). Because of that we may want to consider making whatever pieces necessary to light that up published outside of this module.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -408,6 +444,44 @@ namespace xr
...
+            HandData.HandsInitialized = true;
+        }
+
+        void UninitializeHandResources()
+        {
+            if (HandData.HandsInitialized)
+            {
+                for (HandInfo handInfo : HandData.handsInfo)
+                {
+                    HmdImpl.Extensions->xrDestroyHandTrackerEXT(handInfo.handTracker);
```
tymlipari, line 478: This looks awfully like something that could live in an object destructor and go for an RAII pattern.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -408,6 +444,44 @@ namespace xr
...
+            }
+
+            HandData.HandsInitialized = true;
+        }
+
+        void UninitializeHandResources()
+        {
+            if (HandData.HandsInitialized)
+            {
+                for (HandInfo handInfo : HandData.handsInfo)
```
tymlipari, line 476: nit: `const&`

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: Right now, the only data that's ever passed in is the jointSpace created above, on line 183. I talked with syntheticMagus about it, and decided to leave the assert in there in case a dev ever decides to implement JointSpace internally.

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: I was tempted to make an XRHand component, but after talking with @syntheticmagus we decided that, for now at least, we can get by with a boolean, since implementing XRHand here as a wrapped class (and an iterable one, at that) was a lot more work. More so since we can achieve the same result for now by creating an Napi::Array, and do so on line 179. We can always replace it later if needed.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -134,6 +134,61 @@ namespace
...
+            "RING_PHALANX_PROXIMAL",
+            "RING_PHALANX_INTERMEDIATE",
+            "RING_PHALANX_DISTAL",
+            "RING_PHALANX_TIP",
+
+            "LITTLE_METACARPAL",
+            "LITTLE_PHALANX_PROXIMAL",
+            "LITTLE_PHALANX_INTERMEDIATE",
+            "LITTLE_PHALANX_DISTAL",
+            "LITTLE_PHALANX_TIP"};
```
syntheticmagus, line 176: nit: Style. I believe this file is governed by the Babylon Native clang format; if so, you should just be able to run auto-formatting in VS.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -134,6 +134,61 @@ namespace
...
 
+    void SetXRHandInputSource(Napi::Object& jsInputSource, xr::System::Session::Frame::InputSource& inputSource)
+    {
+        // Don't set hands up if hand data isn't supported/available
+        if (!inputSource.JointsTrackedThisFrame)
+        {
+            return;
+        }
+
+        constexpr const size_t HAND_JOINT_COUNT{25};
```
syntheticmagus, line 145: nit: This pattern is fine, but another option might be to just hard-code this number into the declaration of `HAND_JOINT_NAMES`, then use `HAND_JOINT_NAMES.size()` instead of the second `constexpr` variable.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -134,6 +134,61 @@ namespace
...
+            "LITTLE_PHALANX_DISTAL",
+            "LITTLE_PHALANX_TIP"};
+
+        auto env = jsInputSource.Env();
+        auto handJointCollection = Napi::Array::New(env, HAND_JOINT_COUNT);
+
+        for (size_t i = 0; i < HAND_JOINT_COUNT; i++)
+        {
+            auto napiJoint = Napi::External<std::decay_t<decltype(*inputSource.HandJoints.begin())>>::New(env, &inputSource.HandJoints[i]);
+            handJointCollection.Set((int)i, napiJoint);
```
syntheticmagus, line 184:
```suggestion
            handJointCollection.Set(static_cast<int>(i), napiJoint);
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -134,6 +134,61 @@ namespace
...
+            handJointCollection.Set(Napi::String::New(env, HAND_JOINT_NAMES[i]), (int)i);
```
syntheticmagus, line 185:
```suggestion
            handJointCollection.Set(Napi::String::New(env, HAND_JOINT_NAMES[i]), static_cast<int>(i));
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -134,6 +134,61 @@ namespace
...
+        }
+
+        handJointCollection.Set("length", (int)HAND_JOINT_COUNT);
```
syntheticmagus, line 188:
```suggestion
        handJointCollection.Set("length", static_cast<int>(HAND_JOINT_COUNT));
```

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I cleaned it up a little, but this is an anonymous struct used for a single variable, HandData. HandData is a named struct since I don't want to try embedding the struct definition inside the std::array...

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -145,6 +200,7 @@ namespace
         jsInputSource.Set("handedness", Napi::String::New(env, HANDEDNESS_STRINGS[static_cast<size_t>(inputSource.Handedness)]));
         jsInputSource.Set("targetRayMode", TARGET_RAY_MODE);
         SetXRInputSourceSpaces(jsInputSource, inputSource);
+        SetXRHandInputSource(jsInputSource, inputSource);
```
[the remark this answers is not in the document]
rickfromwork: SetXRInputSourceSpaces is called in ProcessInputSources every frame on existing input, to "Ensure the correct spaces are associated with the existing input source"

I guess it'd be good to do the same for the hand data there as well?

Issue #368, opened 2020-08-12
Setting AbstractMesh.showBoundingBox to true crashes BabylonNative


Alex-MSFT:
Repro steps:
1.) Add this line to the experience.js playground:
`scene.meshes[0].showBoundingBox = true;`
2.) Run the app it will crash with on the below line:
![image](https://user-images.githubusercontent.com/27031140/90074865-ec989980-dcb0-11ea-8f5f-36c0a0b0ca1d.png)

Observations:
Callstack seems to be related to the fact that the _gl (WebRenderingContext) is undefined.

bghgary:
Yup, that function (and probably related functions) needs to be overridden in the nativeEngine.ts

CedricGuillemet:
I'm taking a quick look. It should be pretty easy to fix.

PR #404, opened 2020-09-15
depth tests parameters

CedricGuillemet:
fixes https://github.com/BabylonJS/BabylonNative/issues/368
bjs update with commit https://github.com/BabylonJS/Babylon.js/pull/8967/commits/cab5feeee5a04ff74fb9d017502aa344f282bcf5

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1580,16 +1580,16 @@ namespace Babylon
         switch (fillMode)
         {
         case 2: // MATERIAL_PointFillMode
-            fillModeState |= BGFX_STATE_PT_POINTS;
+            fillModeState = BGFX_STATE_PT_POINTS;
             break;
         case 4: // MATERIAL_LineListDrawMode
-            fillModeState |= BGFX_STATE_PT_LINES;
+            fillModeState = BGFX_STATE_PT_LINES;
             break;
         case 6: // MATERIAL_LineStripDrawMode
-            fillModeState |= BGFX_STATE_PT_LINESTRIP;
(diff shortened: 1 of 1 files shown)


syntheticmagus (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1577,9 +1577,20 @@ namespace Babylon
 
         // TODO: support other fill modes
         uint64_t fillModeState = 0; //indexed tri list
-        if (fillMode == 2)
+        switch (fillMode)
```
bghgary, line 1580: I think we should try to be more consistent in passing enums. Sometimes we pass bgfx enums. Sometimes we pass Babylon enums. Should we just pick one and stick with it?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1577,9 +1577,20 @@ namespace Babylon
...
         {
+        case 2: // MATERIAL_PointFillMode
             fillModeState |= BGFX_STATE_PT_POINTS;
```
bghgary, line 1583: Why are these `|=` and not just `=`?

bghgary (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Agree. Fixed.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Agree. Opening an issue for that. https://github.com/BabylonJS/BabylonNative/issues/406

Issue #385, opened 2020-09-01
Rendering differences of PBRMaterial in iOS v/s Android


lokiiarora:
Asset used: https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/DamagedHelmet/glTF-Binary/DamagedHelmet.glb?raw=true

iOS Screenshot:
![Screenshot 2020-09-01 at 3 15 56 PM](https://user-images.githubusercontent.com/20043762/91836458-6c8aa180-ec68-11ea-8532-8fa6771c838e.png)
Android Screenshot:
![Screenshot_20200901-152556](https://user-images.githubusercontent.com/20043762/91836213-1a498080-ec68-11ea-92bf-8129c2b84885.jpg)
![Screenshot_20200901-152530](https://user-images.githubusercontent.com/20043762/91836226-1e759e00-ec68-11ea-91bc-c9578088e717.jpg)

Devices Used:
Android: GM1901 Oneplus 7
iOS: iPhone 11 iOS 13.3 Simulator.

Issue #372, opened 2020-08-18
GLBs using ORM map in BabylonReactNative demo app will not render properly


RonnieAshlock:
Possibly related to #371

It looks like the OcclusionRoughnessMetallic map is not being rendered when loaded in the BabylonReactNative demo app.

In Sandbox:
![image.png](https://images.zenhubusercontent.com/5efb7d8ce397006536736e73/54d6166a-f355-4410-ac2e-2486c2f8877d)

In the demo app:
![Screenshot_20200818-095403.jpg](https://images.zenhubusercontent.com/5efb7d8ce397006536736e73/29281685-774b-4b26-bc45-ae8e2675faca)

![Screenshot_20200818-095411](https://user-images.githubusercontent.com/60451995/90544349-e173bc80-e13b-11ea-8a45-9e25c26adcef.jpg)

Here's the asset used in the test:
https://raw.githubusercontent.com/RonnieAshlock/TestAssets/master/GLB/air_compressor_512_ORM.glb

bghgary:
Caused by the same underlying issue:
https://github.com/BabylonJS/BabylonReactNative/issues/61
https://github.com/BabylonJS/BabylonReactNative/issues/70

Issue #300, opened 2020-06-18
Remove dependency on SPIRV-Cross for OpenGL scenarios


syntheticmagus:
When the target rendering path is OpenGL, SPIRV-Cross should not actually be necessary because glslang should be sufficient to do all our required conversions on its own. Note that to make this change, all of the logic in `NativeEngine::CreateProgram` that depends on SPIRV-Cross types will need to be rewritten to use glslang instead; that logic should probably also be pulled out of `NativeEngine`, too. This is therefore expected to be a fairly large change that will affect behavior in every rendering path because it will change how the bgfx shader representations are created.

bghgary:
Reopen if we have a need to make this work.

PR #411, opened 2020-09-18
 Fixes for Android shader compile

### LINKED ISSUE #371: Normals map not rendering correctly with BabylonReactNative demo app
Rendering like this in the demo app:

Same asset loaded in Sandbox:

To repro, use this asset:
https://raw.githubusercontent.com/RonnieAshlock/TestAssets/master/GLB/Head/Head_512Tex.glb

bghgary:
Fixes #371, #372, #385.

Also fixes BabylonJS/BabylonReactNative#61 once BabylonReactNative picks up this change.

The main issue with the Android shader code path is that the sampler locations for textures are wrong. This change include a couple of fixes to address this:

1. The conversion to SPIRV is now as minimal as possible. The splitting of samplers into texture/sampler and then recombining them is now gone. [Removing SPIRV for OpenGL situations](https://github.com/BabylonJS/BabylonNative/issues/300) should be easier after this change. This is a problem because when SPIRV recombines the texture/sampler pair, it mangles the name and thus bgfx can't bind the locations properly. I couldn't find a way for SPIRV not to do this, but removing this splitting/recombining fixes this issue. Lowering to ES 300 version also then removes the binding attributes completely.
2. Because the binding has been removed completely, the "stage" as bgfx calls them, needs to be generated explicitly. As far as I can tell, for WebGL/OpenGL, this is simply the order they appear in the shader. The change to AppendSamplers reflects this fix.

Finally, I purposefully did not fix Linux in this change. I expect Linux will have exactly the same issue as Android, but I don't want to handle that in this PR. Will address it separately.

### DIFF
--- a/Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
+++ b/Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
@@ -36,12 +36,14 @@ namespace Babylon
 
             auto compiler = std::make_unique<spirv_cross::CompilerGLSL>(parser->get_parsed_ir());
 
+#if !ANDROID
             compiler->build_combined_image_samplers();
+#endif
 
             spirv_cross::CompilerGLSL::Options options = compiler->get_common_options();
 
-#ifdef ANDROID
-            options.version = 310;
+#if ANDROID
+            options.version = 300;
             options.es = true;
 #else
             options.version = 330;
@@ -88,7 +90,10 @@ namespace Babylon
         ShaderCompilerTraversers::IdGenerator ids{};
         auto cutScope = ShaderCompilerTraversers::ChangeUniformTypes(program, ids);
         ShaderCompilerTraversers::AssignLocationsAndNamesToVertexVaryings(program, ids);
+
+#if !ANDROID
         ShaderCompilerTraversers::SplitSamplersIntoSamplersAndTextures(program, ids);
+#endif
 
         std::string vertexGLSL(vertexSource.data(), vertexSource.size());
         auto [vertexParser, vertexCompiler] = CompileShader(program, EShLangVertex, vertexGLSL);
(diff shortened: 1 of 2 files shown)


Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -36,12 +36,14 @@ namespace Babylon
...
+#if !ANDROID
             compiler->build_combined_image_samplers();
+#endif
 
             spirv_cross::CompilerGLSL::Options options = compiler->get_common_options();
 
-#ifdef ANDROID
-            options.version = 310;
+#if ANDROID
+            options.version = 300;
```
syntheticmagus, line 46: Does this work? I seem to remember there was some issue with 300, like SPIRV-Cross had some sort of problem with that version.

bghgary: What do you mean by "in isolation"? It certainly works for ANDROID builds AFAICT.

bghgary: BTW, this change is what removes the binding attributes. Bindings are not supported until 310, but using 310 causes the bindings to output with locations of 0 for each one. WebGL2 conforms closely to OpenGL ES 3.0. We should use 3.0 unless there is a specific reason not to.

syntheticmagus: By "in isolation," I meant were there other changes (for example, a different SPIRV-Cross submodule version) that weren't reflected here that caused this to work (realizing it's a draft and might not be complete). I don't remember exactly what it was, but _something_ led us to decide to use GLSL 310 instead of 300. If that was the bindings thing and it no longer applies, that's awesome!

bghgary: I don't remember/know for sure. But this change is tied with the removal of the the split and recombine in this file.

syntheticmagus (approved):

PR #414, opened 2020-09-18
index start/count for drawcalls

### LINKED ISSUE #408: Mesh.MergeMeshes does not work correctly with multiple materials
If you create two meshes (e.g. `Mesh.CreateBox`) and they each have a different material, and then you merge them with `Mesh.MergeMeshes([box1, box2], true, true, undefined, undefined, true)` the geometry merges as expected, but you end up with a single material which looks kind of like one of the original materials but without the alpha (or something like that). Same thing works fine in the Babylon.js Playground.

CedricGuillemet:
fixes #408

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -442,14 +442,14 @@ namespace Babylon
             };
             DoForHandleTypes(nonDynamic, dynamic);
         }
-
-        void SetBgfxIndexBuffer() const
+        
+        void SetBgfxIndexBuffer(uint32_t firstIndex, uint32_t numIndices) const
         {
-            constexpr auto nonDynamic = [](auto handle) {
-                bgfx::setIndexBuffer(handle);
+            const auto nonDynamic = [firstIndex, numIndices](auto handle) {
+                bgfx::setIndexBuffer(handle, firstIndex, numIndices);
             };
-            constexpr auto dynamic = [](auto handle) {
-                bgfx::setIndexBuffer(handle);
+            const auto dynamic = [firstIndex, numIndices](auto handle) {
+                bgfx::setIndexBuffer(handle, firstIndex, numIndices);
             };
             DoForHandleTypes(nonDynamic, dynamic);
         }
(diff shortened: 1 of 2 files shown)


ryantrem (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -768,10 +768,7 @@ namespace Babylon
         const auto& vertexArray = *(info[0].As<Napi::External<VertexArray>>().Data());
 
         // a vertex array might not have an index buffer associated with
```
bghgary, line 770: this comment doesn't make sense now

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -768,10 +768,7 @@ namespace Babylon
...
-        if (vertexArray.indexBuffer.data)
-        {
-            vertexArray.indexBuffer.data->SetBgfxIndexBuffer();
-        }
+        m_currentBoundIndexBuffer = vertexArray.indexBuffer.data;
```
bghgary, line 771: Given the way things are going, I wonder if it would be cleaner to store the vertexArray instead of just the index buffer. Then do the bgfx work when draw is called.

bghgary (changes requested):

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: It does! the m_currentBoundIndexBuffer is updated with the index buffer associated with it. It can be nullptr.

bghgary: Oh, right, sorry, I misinterpreted the comment.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Not sure if a vertex/indexbuffer is bound once with multiple drawcalls depending on it. But maybe in the future. I see a trend to store a lot more states in nativeengine class (clear/rendering states/index buffer).

bghgary: We should probably take care of this sooner than later.

bghgary (approved):

Issue #398, opened 2020-09-12
Add support for ShadowGenerator


ryantrem:
Attempting to construct a `ShadowGenerator` results in the following error:
```
TypeError: undefined is not an object (evaluating 'gl.TEXTURE_2D')
```
I'm guessing there is just something that hasn't been implemented in Babylon Native that is needed to support this.

CedricGuillemet:
I'am taking a look. Might not be a lot of work...

CedricGuillemet:
Actually, it's more work than anticipated.
Half is done with PR #415 Need to finish creating the depth texture support.

CedricGuillemet:
I've found an issue in bgfx shader patching code. Fixing it before moving on this issue.

CedricGuillemet:
Current status for shadows:
- [x] Basic shadow implemented
- [x] Filtering and order or rendering fixed with this PR https://github.com/BabylonJS/BabylonNative/pull/739
- [ ] Cubemap shadows not supported yet

brian-hay:
> Need to finish creating the depth texture support.

@CedricGuillemetis in relation to your comment above, is the depth texture / depth renderer supported in Babylon Native at this point in time?

brian-hay:
Hi @CedricGuillemet yes that post was from me :-)

I tried your proposed temporary workaround and it worked!

bghgary:
Closing for now. If someone needs this, we should open a new issue.

PR #415, opened 2020-09-18
More validation tests

### LINKED ISSUE #87: Support render target textures
- [ ] RTT textures are flipped verticaly.
   This is more complex to fix than the loaded textures.

CedricGuillemet:
Because I like to make your life more difficult by adding validation tests ;)

- 26 more tests mostly gltf related
- the validation test errorcode return was wrong. because of a wrong WM_
- error tolerance boosted to 8% or 5% because of missing text in some scene.
- success/failure task condition to push the appropriate artifact (results/errors)
- Y flipped for RTT fixes #87 
- Post process!

bjs max with commit https://github.com/BabylonJS/Babylon.js/pull/8999/commits/07ee198b57debcf7adad689de6c9433e1198514f


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -700,6 +700,8 @@ namespace Babylon
             bgfx::setViewRect(0, 0, 0, w, h);
 #ifdef __APPLE__
             bgfx::frame();
+#elif __linux__
+            bgfx::discard();
```
bghgary, line 704: This is looking to be a bit concerning. Why is the code different for different platforms here?

Apps/ValidationTests/Scripts/validation_native.js
```diff
@@ -3,7 +3,7 @@ var currentScene;
 var config;
 var justOnce;
 var threshold = 25;
-var errorRatio = 2.5;
+var errorRatio = 8.0;
```
bghgary, line 6: Is this higher than what Babylon.js uses now?

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Still trying to understand what happens.

Apps/ValidationTests/Scripts/validation_native.js
CedricGuillemet: Yes! 2.5 bjs 8 here. Mostly because text is rendered in some PG.

bghgary: I'm assuming you mean text is not being rendered because we don't have text rendering yet?

CedricGuillemet: Yes! That's correct.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Reverted change.

Apps/ValidationTests/Scripts/validation_native.js
bghgary: We discussed this offline and the plan is to update this value only for the tests that have text and put a comment somewhere that explains why (or, at a minimum, make sure the commit message says why).

CedricGuillemet:
Adding a bunch more test scenes this week. Back to draft.

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -270,7 +270,7 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
             }
             break;
         }
-        case WM_DESTROY:
+        case WM_CLOSE:
```
bghgary, line 273: Not sure about this one. I'm assuming you are seeing some shutdown issues. Usually the sequence is WM_CLOSE calls DestroyWindow which then causes WM_DESTROY which calls PostQuitMessage.

Apps/ValidationTests/Win32/App.cpp
```diff
@@ -270,7 +270,7 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
...
         {
             short exitCode = LOWORD(wParam);
```
bghgary, line 275: I'm a bit confused by this. Documentation says that the lParam and wParam of both WM_DESTROY or WM_CLOSE are not used. What does this do exactly?

Plugins/NativeEngine/CMakeLists.txt
```diff
@@ -52,6 +52,8 @@ endif()
 
 target_compile_definitions(NativeEngine
     PRIVATE NOMINMAX)
+target_compile_definitions(NativeEngine
+    PRIVATE API${GRAPHICS_API}) # OpenGL is defined in bgfx.h. Using APIXXX instead
```
bghgary, line 56: I think this should probably be in all CAPS to match typical macros which may avoid the conflict. Use https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#string-transformations to convert to uppercase. But ideally, this shouldn't be done at all. The specific code should be in the specific files instead of using macros.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1029,6 +1038,25 @@ namespace Babylon
...
     }
 
+    Napi::Value NativeEngine::CreateDepthTexture(const Napi::CallbackInfo& info)
+    {
+        const auto texture = info[0].As<Napi::External<TextureData>>().Data();
+        uint16_t width = static_cast<uint16_t>(info[1].As<Napi::Number>().Uint32Value());
+        uint16_t height = static_cast<uint16_t>(info[2].As<Napi::Number>().Uint32Value());
+        bgfx::FrameBufferHandle frameBufferHandle{};
+
+        // This is WIP
```
bghgary, line 1048: Is there an issue to track?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1317,13 +1344,56 @@ namespace Babylon
...
+                const ProgramData::UniformValue& value = it.second;
+                if (value.YFlip)
+                {
+                    float tmpMatrix[16];
+                    static const float flipMatrix[16] = { 1.f, 0.f, 0.f, 0.f,
+                        0.f,-1.f, 0.f, 0.f,
+                        0.f, 0.f, 1.f, 0.f,
+                        0.f, 0.f, 0.f, 1.f };
+                    bx::mtxMul(tmpMatrix, value.Data.data(), flipMatrix);
+                    bgfx::setUniform({ it.first }, tmpMatrix, value.ElementLength);
```
bghgary, line 1372: nit: clang format this code

bghgary (approved):
Many of these changes are not ideal and should have issues associated with them. The API macros thing should be removed completely, but maybe after my changes.

Plugins/NativeEngine/CMakeLists.txt
CedricGuillemet: I agree we should not see any `#ifdef platform` anywhere in the code but in specific .cpp files. It's mostly in shaderCompilerCommon.cpp so I'll let you do the change when you are done with your shader compilation work.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Yes https://github.com/BabylonJS/BabylonNative/issues/398

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Done

Apps/ValidationTests/Win32/App.cpp
CedricGuillemet: Replacing with direct errorCode for postquitmessage instead of using message param.

PR #418, opened 2020-09-23
Move bgfx shader construction into ShaderCompiler class

bghgary:
This is the first step towards fixing #300. This change only move codes around and does not change functionality.

Tested on Win32, Android, and macOS.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -889,123 +679,35 @@ namespace Babylon
 
     Napi::Value NativeEngine::CreateProgram(const Napi::CallbackInfo& info)
     {
-        const auto vertexSource = info[0].As<Napi::String>().Utf8Value();
-        const auto fragmentSource = info[1].As<Napi::String>().Utf8Value();
+        const std::string vertexSource{info[0].As<Napi::String>().Utf8Value()};
+        const std::string fragmentSource = info[1].As<Napi::String>().Utf8Value();
```
syntheticmagus, line 683: nit: Initializer list instead of assignment for this one too?

syntheticmagus (approved):

Issue #343, opened 2020-07-28
Logic in NativeXr.cpp and XR.cpp should be split into multiple files


chrisfromwork:
Looking into NativeXr.cpp and XR.cpp, a lot of classes are currently defined in a single file. This does not make for great legibility. For example, function definitions in XR.cpp (OpenXR) for System::Sesssion, System and System::Session::Frame are mixed in order throughout. I believe it would improve legibility if we started splitting this logic into separate files.

bghgary:
Reopen if we find this to be annoying and want to split.

PR #421, opened 2020-09-25
Converting XRHand in NativeXR into a proper class

rickfromwork:
Changes XRHand from a bool to a proper class, in order to implement the joint name indices used in WebXRHand::_PopulateHandPartsDefinition, found in WebXRHandTracking.ts.

XRHand was set as a bool before, since it looked like it just needed to be defined to work. Later, I noticed the _PopulateHandPartsDefinition call required XRHand to have entries for each joint, which this PR implements.

What is being implemented from WebXR:
- Static joint index mapping (HAND_JOINT_NAMES)
- joint count (length)

What isn't being implemented from WebXR:
- Instance-specific data, specifically the jointSpace list
  - XRHand instances are not being made or used at this point, so only the static data is necessary
  - 'XRHand' instances are actually xrController.inputSource.hand, a Napi array with both static length/joint labels, and joint instance members


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1459,6 +1459,49 @@ namespace Babylon
             XRFrame* m_frame{};
         };
 
+        class XRHand : public Napi::ObjectWrap<XRHand>
```
tymlipari, line 1462: Can we move `XRHand` into another file? This one's already getting kind of long.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1459,6 +1459,49 @@ namespace Babylon
...
+        {
+            static constexpr auto JS_CLASS_NAME = "XRHand";
+
+        public:
+            static void Initialize(Napi::Env env)
+            {
+                Napi::HandleScope scope{env};
+
+                std::vector<Napi::ClassPropertyDescriptor<XRHand>> initList = {};
```
tymlipari, line 1471: Since you know the size up front that you'll need for the `std::vector`, use the constructor that takes a `size_t` parameter so it'll pre-allocate its internal buffer to match.

Also nit: you don't need to do `= {}`. You can just do `std::vector<...> initList{}` if you're using the default constructor.

tymlipari (approved):

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: It is, but all the other XR* classes are defined in here as well so I would probably move them all out at that point...

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1459,6 +1459,51 @@ namespace Babylon
...
+
+        public:
+            static void Initialize(Napi::Env env)
+            {
+                Napi::HandleScope scope{env};
+
+                std::vector<Napi::ClassPropertyDescriptor<XRHand>> initList{};
+                initList.reserve(HAND_JOINT_NAMES.size() + 1);
+
+                for (int i = 0; i < HAND_JOINT_NAMES.size(); i++)
```
rickfromwork, line 1474:
```suggestion
                for (size_t idx = 0; idx < HAND_JOINT_NAMES.size(); idx++)
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1459,6 +1459,51 @@ namespace Babylon
...
+                {
+                    initList.push_back(StaticValue(HAND_JOINT_NAMES[i], Napi::Value::From(env, i)));
```
rickfromwork, line 1476:
```suggestion
                    initList.push_back(StaticValue(HAND_JOINT_NAMES[idx], Napi::Value::From(env, idx)));
```

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: I have an issue tracking this here: https://github.com/BabylonJS/BabylonNative/issues/343

If this can be moved into another file we should go ahead and do it to set precedent. I don't know when anyone will get to the above issue.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1459,6 +1459,51 @@ namespace Babylon
...
+                {
+                    initList.push_back(StaticValue(HAND_JOINT_NAMES[idx], Napi::Value::From(env, idx)));
+                }
+
+                initList.push_back(StaticAccessor("length", &XRHand::GetLength, nullptr));
+
+                Napi::Function func = DefineClass(
+                    env,
+                    JS_CLASS_NAME,
+                    initList
```
chrisfromwork, line 1484: If i remember correctly, this class was previously declared in the webxr types as an array/iteratable item. It looks like we only define the class to have two static fields:

1. N static variables with hand join names that then map to an index value
2. A static variable to get N

Do we need any instance methods on this class for accessing joint data? Or does this class only exist to specify hand joint indices?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -127,6 +127,40 @@ namespace
         }
     }
 
+    constexpr std::array<const char*, 25> HAND_JOINT_NAMES
+    {
+        "WRIST",
+
+        "THUMB_METACARPAL",
```
chrisfromwork, line 134: should these values be all capital with underscores now that they are exposed as XRHand variables? I have typically seen WebXR string definitions as all lower case with dashes.

chrisfromwork: nevermind, it looks like this is correct relative of the spec: [https://immersive-web.github.io/webxr-hand-input/](https://immersive-web.github.io/webxr-hand-input/)

---
In reply to: [496088316](https://github.com/BabylonJS/BabylonNative/pull/421#discussion_r496088316) [](ancestors = 496088316)

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: We would need to specify joint data if this class was being used to pass joint data around as well, but that is being done by the JointSpace vector on InputSource. If the data gets populated elsewhere in the future, like Pose, this would have to be fleshed out, but since all joint setting/getting is done in NativeXR, we can bypass using the class for that, and instead do what we do above, in SetXRInputSourceData(), which essentially creates a NativeXR version of XRHand.

(It's done this way since making a class indexable through N-API is, from what I can tell, impossible. I've played with alternatives, but they aren't nearly as clean)

chrisfromwork: It looks like we may still be missing setting this class up as iterable<XRJointSpace> and an instance accessor that allows querying XRJointSpaces: https://immersive-web.github.io/webxr-hand-input/#xrhand-interface

Is there a reason that this additional functionality isn't implemented?

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: I'm naming them to match https://immersive-web.github.io/webxr-hand-input/#xrhand-interface

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: TL:DR the class is only used for static access, and instance access is done through inputSource.HandJoints

chrisfromwork: So we don't have the ability to define the iterable with N-API, is there a reason we aren't defining the getter for a joint?

chrisfromwork: looking at this more it seems like XRHand isn't a type that babylon.js developers will directly interact with/have access to. This seems reasonable for now

chrisfromwork (approved):
:shipit:

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: I might have made it a little harder to do this by placing HAND_JOINT_NAMES in an anonymous namespace in NativeXR, but I'll give it a shot

Issue #409, opened 2020-09-16
Javascript debugging should not be enabled for chakra backend on HoloLens


chrisfromwork:
We've determined that chakra script debugging does not work on hololens. This causes AppRuntimeChakra.cpp to throw an exception when kicking off a debug babylon native build on a HoloLens device due to a call to JsStartDebugging.

Workarounds include commenting out this line of code or building for release. But, in delivering a high quality developer experience, we should prevent this crash for debug builds on HoloLens. Third party developers shouldn't need to comment out code in BabylonNative releases to get HoloLens debug images to work.

tymlipari:
Worth noting that the reason this crashes is due to the fact that VS isn't correctly deploying the JS debugging support along with the remote debugger package. We've pinged them about this, but we should figure out if we can gracefully fall back when it's not present while we wait for a fix.

tymlipari:
Since it looks like we're unlikely to get Chakra debugging support for ARM targets, the remaining work to enable a Babylon Native debugging story on HoloLens 2 is tracked by #463 and #464.

PR #422, opened 2020-09-26
Disable Chakra debugging on ARM platforms

tymlipari:
Visual Studio currently doesn't deploy the required binaries needed for JS debugging on ARM-based platforms such as HoloLens 2. While we wait for the required VS fixes, this change disables the `JsStartDebugging` call on those architectures.

Related to #409

### DIFF
--- a/Core/AppRuntime/Source/AppRuntimeChakra.cpp
+++ b/Core/AppRuntime/Source/AppRuntimeChakra.cpp
@@ -46,7 +46,7 @@ namespace Babylon
             &dispatchFunction));
         ThrowIfFailed(JsProjectWinRTNamespace(L"Windows"));
 
-#ifdef _DEBUG
+#if defined(_DEBUG) && !(defined(_M_ARM64) || defined(_M_ARM))
         // Put Chakra in debug mode.
         ThrowIfFailed(JsStartDebugging());
 #endif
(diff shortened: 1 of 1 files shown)


chrisfromwork:
Do we anticipate anyone to be developing BabylonNative uwps for arm laptops?

chrisfromwork (approved):

Core/AppRuntime/Source/AppRuntimeChakra.cpp
```diff
@@ -46,7 +46,7 @@ namespace Babylon
             &dispatchFunction));
         ThrowIfFailed(JsProjectWinRTNamespace(L"Windows"));
 
-#ifdef _DEBUG
+#if defined(_DEBUG) && !(defined(_M_ARM64) || defined(_M_ARM))
```
bghgary, line 49: I think it would be better to output a warning if it fails instead of throw. Does that work for you?

tymlipari: Switched to no longer throw on failure but instead print a debug warning

Core/AppRuntime/Source/AppRuntimeChakra.cpp
```diff
@@ -46,9 +46,15 @@ namespace Babylon
...
 
-#ifdef _DEBUG
+#if defined(_DEBUG)
         // Put Chakra in debug mode.
-        ThrowIfFailed(JsStartDebugging());
+        {
+            auto result = JsStartDebugging();
+            if (result != JsErrorCode::JsNoError)
+            {
+                OutputDebugStringW(L"Failed to initialize debugging support.\n");
```
bghgary, line 55:
```suggestion
                OutputDebugStringW(L"Failed to initialize JavaScript debugging support.\n");
```

bghgary (approved):

PR #429, opened 2020-09-30
fix XR renderTarget regression

CedricGuillemet:
Because of a vertical flip needed for render targets on Direct3D and Metal, XR rendertarget was also flipped.
But as it is used for direct use with particular UV on Metal and as back buffer on Windows, the 3D displayed was flipped.
For XR, the render target must behave like a back buffer and not like a render target that would be reused by a pixel shader.


bghgary (approved):
This is fine for now, but we should figure out if there is a better fix for the flipping issue.

PR #430, opened 2020-09-30
Fix some regressions in XR mode on Android and iOS

ryantrem:
Several changes were recently made that broke XR on Android and iOS, and this PR fixes them.
- NAPI JSI needs a `StaticAccessor` implementation that doesn't throw and a `Reference::Reset` that is functional. I think there is more we need to do to actually make `StaticAccessor` work rather than just not crash. If that's the case, I can add a comment or something.
- NativeEngine needs to clear out the `RequestAnimationFrame` callback after calling it so that when we transition to XR mode we don't end up rendering twice (e.g. calling the stale callback for the regular `RequestAnimationFrame` in addition to the current callback for the XR `RequestAnimationFrame`).
- `DeviceInputSystem` needs to access the JS native object via the new `NativeObject` class.
- NativeXR should use the `Set` overload that takes a `const char*` (there should be no need to wrap it in a napi string, and we don't have that implemented for NAPI JSI currently).
- NativeXR should use `XRHand::PropertyDescriptor` rather than `ClassPropertyDescriptor<XRHand>`. In NAPI direct, `ObjectWrap<T>::PropertyDescriptor` is just type defed as `ClassPropertyDescriptor<T>`, so this is really more correct in that case. In NAPI JSI, `ClassPropertyDescriptor<T>` doesn't even exist, so we must use `ObjectWrap<T>::PropertyDescriptor` at this point.
- Removed the line from the README that says we don't have React Native integration. :)

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -512,12 +512,12 @@ namespace Babylon
 
             try
             {
-                if (!m_requestAnimationFrameCalback.IsEmpty())
+                if (!m_requestAnimationFrameCallback.IsEmpty())
                 {
                     // We can get here from either the normal RequestAnimationFrame or the XR RequestAnimationFrame,
                     // so we need to clear out the regular RequestAnimationFrame callback to make sure we don't incorrectly
                     // call it when we have transitioned to the XR RequestAnimationFrame.
-                    auto callback{std::move(m_requestAnimationFrameCalback)};
+                    auto callback{std::move(m_requestAnimationFrameCallback)};
                     callback({});
                 }
                 GetFrameBufferManager().Reset();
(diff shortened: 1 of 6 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -512,7 +512,15 @@ namespace Babylon
...
             try
             {
-                m_requestAnimationFrameCalback.Call({});
+                if (!m_requestAnimationFrameCalback.IsEmpty())
+                {
+                    // We can get here from either the normal RequestAnimationFrame or the XR RequestAnimationFrame,
+                    // so we need to clear out the regular RequestAnimationFrame callback to make sure we don't incorrectly
+                    // call it when we have transitioned to the XR RequestAnimationFrame.
+                    auto callback = std::move(m_requestAnimationFrameCalback);
+                    m_requestAnimationFrameCalback.Reset();
```
bghgary, line 521: do you need this? it's already moved to a local variable?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -512,7 +512,15 @@ namespace Babylon
 
             try
             {
-                m_requestAnimationFrameCalback.Call({});
+                if (!m_requestAnimationFrameCalback.IsEmpty())
+                {
+                    // We can get here from either the normal RequestAnimationFrame or the XR RequestAnimationFrame,
+                    // so we need to clear out the regular RequestAnimationFrame callback to make sure we don't incorrectly
+                    // call it when we have transitioned to the XR RequestAnimationFrame.
+                    auto callback = std::move(m_requestAnimationFrameCalback);
```
bghgary, line 520:
```suggestion
                    auto callback{std::move(m_requestAnimationFrameCalback)};
```

bghgary (approved):

PR #431, opened 2020-10-01
vsync Android

### LINKED ISSUE #390: Android frame rate is unlimited
On Android, the render loop goes as fast as it can. With a simple scene, it might render at 300+ FPS. This is rather wasteful, and we should probably initially just hard code it to be limited by vsync or something.

CedricGuillemet:
fixes #390 
bgfx reset flags was not sent to bgfx::reset

### DIFF
--- a/Core/Graphics/Source/Graphics.cpp
+++ b/Core/Graphics/Source/Graphics.cpp
@@ -137,7 +137,7 @@ namespace Babylon
         pd.backBuffer = nullptr;
         pd.backBufferDS = nullptr;
         bgfx::setPlatformData(pd);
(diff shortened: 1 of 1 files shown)


ryantrem (commented):
So we don’t have to tie into SurfaceView callback/delegate after all?

CedricGuillemet:
Yep! runs at 60Hz out of the box.

ryantrem:
Nice. Is it a no-op on iOS? And a no-op in XR mode?

CedricGuillemet:
I'm just starting work on iOS and sync rendering.
VSync in XR as well. Got 30FPS with AR.

ryantrem (approved):

PR #434, opened 2020-10-05
XR animation frame request scheduling bug fix

syntheticmagus:
Attempt to fix scheduling problems with animation frame requesting for scenarios where the creation of XR predates the creation of the render loop. Without this change, if there was no pre-existing render-loop, there was a one-frame gap before XR's render loop would become self-sustaining, and consequently it would never get started in earnest. I believe this may be the fix, or a meaningful part of it, for the recent soft-lock issues encountered by several teams consuming Babylon Native for XR on multiple platforms. I don't have access to all the required platforms, though, so further testing is needed to ensure this actually fixes the problem.

### DIFF
--- a/Plugins/NativeXr/Source/NativeXr.cpp
+++ b/Plugins/NativeXr/Source/NativeXr.cpp
@@ -243,6 +243,10 @@ namespace Babylon
             }
             m_isFrameScheduled = true;
 
+            m_graphicsImpl.GetAfterRenderTask().then(arcana::inline_scheduler, arcana::cancellation::none(), [this] {
+                m_engineImpl->ScheduleRender();
+            });
+
             m_graphicsImpl.GetBeforeRenderTask().then(arcana::inline_scheduler, arcana::cancellation::none(), [this, callback = std::move(callback)] {
                 // Note: we are guaranteed to be on the render thread.
                 m_isFrameScheduled = false;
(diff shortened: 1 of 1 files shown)


rickfromwork:
This change works for both our desktop and HoloLens testing. 👍

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -243,6 +243,10 @@ namespace Babylon
             }
             m_isFrameScheduled = true;
 
+            m_graphicsImpl.GetAfterRenderTask().then(arcana::inline_scheduler, arcana::cancellation::none(), [this] {
+                m_engineImpl->ScheduleRender();
+            });
+
```
bghgary, line 249: nit: maybe move this after GetBeforeRenderTask so that Before is before and After is after?

bghgary (approved):

ryantrem (approved):

Plugins/NativeXr/Source/NativeXr.cpp
syntheticmagus: Could do, but this ordering was intended to communicate the logical progression, specifically how the after render even precipitates the before render event on the next frame. I'm not sure it gets the message across, though.

Issue #387, opened 2020-09-01
Need a call to bgfx::touch to properly clear updated views


rgerd:
When views are created or updated through calls like [FrameBufferManager::Bind](https://github.com/BabylonJS/BabylonNative/blob/1b5333f06c9a4714ceb3c8ffb2049388209b4cdc/Plugins/NativeEngine/Source/NativeEngine.h#L248) and [NativeEngine::SetViewPort](https://github.com/BabylonJS/BabylonNative/blob/1b5333f06c9a4714ceb3c8ffb2049388209b4cdc/Plugins/NativeEngine/Source/NativeEngine.cpp#L1647), the [views are cleared](https://github.com/BabylonJS/BabylonNative/blob/1b5333f06c9a4714ceb3c8ffb2049388209b4cdc/Plugins/NativeEngine/Source/NativeEngine.h#L159) in order to draw the current frame on a clean slate. ([Original context for the bgfx::touch call](https://github.com/BabylonJS/BabylonNative/pull/41/commits/a90fc7c96d9191acdb238e7e602661b0ebd6fae0#r320950899)).

There was a bug we were seeing (#362) where in stereo rendering a view seemingly corresponding to the default back buffer was getting cleared on top of the render target texture for the left eye (after rendering), clipping content for just that eye. When using a desktop HMD (such as HP Reverb), resizing the window to resize the default back buffer was enough to resize the area of the view causing the extra clear.

(#386) provides a preliminary fix for (#362), but the `bgfx::touch` call is still important for clearing a view even when no other draw calls contribute to the current frame. It seems the most likely scenario is that functionality on the Babylon.js side has changed and some assumptions about behavior at this level are no longer valid. We need to work through the current logic and find a proper home for the `bgfx::touch` call.

Issue #344, opened 2020-07-29
[NativeXR] RenderTargets lifecycle


CedricGuillemet:
When a color texture is destroyed (rotation on iOS for example), the rendertarget linked to that texture is also destroyed:
https://github.com/BabylonJS/BabylonNative/blob/38aa6f455357932b31c4ef56860393b68a1d1453/Plugins/NativeXr/Source/NativeXr.cpp#L305
The issue happens when that rendertarget is still bound in NativeEngine. We end up with a dangling pointer https://github.com/BabylonJS/BabylonNative/blob/38aa6f455357932b31c4ef56860393b68a1d1453/Plugins/NativeEngine/Source/NativeEngine.h#L279

To mitigate this issue, I unbind the rendertarget to be destroyed if it's bound. And do not assert here https://github.com/BabylonJS/BabylonNative/blob/38aa6f455357932b31c4ef56860393b68a1d1453/Plugins/NativeEngine/Source/NativeEngine.h#L261

My first thought was to use a smartpointer so the rendertarget will only be destroyed when it's unbound. This *might* not be the good approach because the textures bound to that rendertarget are already destroyed and bgfx will assert at some point...or overriding bgfx render target texture with a dummy one until it's released.

syntheticmagus:
This seems odd. As I understand it, framebuffer binding and unbinding should always happen in pairs, so a non-default framebuffer state should never happen except when the JS thread is actively rendering, and frame transitions should only occur between rendering cycles. Have I gotten myself confused somewhere? If not, then any encounter with a bound framebuffer during frame transition (which is the only time, I think, that textures should be getting deleted) may be indicative a deeper/weirder problem.

ryantrem:
If sounds like one of the symptoms of this was leaked memory when the display is rotated. I will check this scenario with the latest code as part of my overall current work looking for and trying to fix memory leaks.

rgerd:
Here's the callback that should be run to unbind the framebuffer (and clear the pointer reference in the frame buffer manager) when the texture is destroyed:
https://github.com/BabylonJS/BabylonNative/blob/b766c35f2bc6a42c2b26d672fbcebe9c4c0eaaad/Plugins/NativeXr/Source/NativeXr.cpp#L390-L402

The PR that introduced it is #342.

It seems like this callback essentially fixes this issue, although as you mentioned @CedricGuillemet there's a definite possibility for improvement with smart pointers.

ryantrem:
[...]

PR #436, opened 2020-10-05
Reassign framebuffer view id's per-frame instead of per-bind

rgerd:
This PR:

1) Keeps us from unnecessarily creating multiple views for a single framebuffer per-frame.

2) Fixes bug #387, re-adding the call to `bgfx::touch` so that bgfx properly clears the view even when no geometry is submitted to the view for rendering.

3) Addresses bug #344, unbinding the currently-bound frame buffer at the end of the frame.

4) Fixes an issue where we were clearing the view as a side-effect of setting the viewport. Because Babylon.js does not unbind the frame buffer to which the scene was last rendered, the left eye was getting clipped by a call to setViewport at the beginning of the right eye being rendered. On top of this, the size of the view rect created by setViewport always corresponded to the size of the default back buffer (defined by `bgfx::getStats()->[width/height]()`), and not to the currently bound back buffer. [See here for a discussion of this issue on the Babylon.js forum.](https://forum.babylonjs.com/t/bound-active-camera-frame-buffer-not-unbound-in-scene-render/14559)

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1397,7 +1397,7 @@ namespace Babylon
 
             uint64_t m_engineStateYFlipped = m_engineState;
             m_engineStateYFlipped &= ~BGFX_STATE_CULL_MASK;
-            m_engineStateYFlipped |= (cullCCW | cullCCW) << BGFX_STATE_CULL_SHIFT;
+            m_engineStateYFlipped |= (cullCW | cullCCW) << BGFX_STATE_CULL_SHIFT;
 
             bgfx::setState(m_engineStateYFlipped | fillModeState);
         }
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1462,20 +1462,9 @@ namespace Babylon
         const auto y = info[1].As<Napi::Number>().FloatValue();
         const auto width = info[2].As<Napi::Number>().FloatValue();
         const auto height = info[3].As<Napi::Number>().FloatValue();
-
-        auto& backBuffer = m_frameBufferManager.GetBound();
-        const auto isDefault = m_frameBufferManager.IsRenderingToDefault();
-
-        const auto backbufferWidth = isDefault ? bgfx::getStats()->width : backBuffer.Width;
-        const auto backbufferHeight = isDefault ? bgfx::getStats()->height : backBuffer.Height;
         const float yOrigin = bgfx::getCaps()->originBottomLeft ? y : (1.f - y - height);
 
-        const bgfx::ViewId viewId = m_frameBufferManager.GetBound().ViewId;
-        bgfx::setViewRect(viewId,
-            static_cast<uint16_t>(x * backbufferWidth),
-            static_cast<uint16_t>(yOrigin * backbufferHeight),
-            static_cast<uint16_t>(width * backbufferWidth),
-            static_cast<uint16_t>(height * backbufferHeight));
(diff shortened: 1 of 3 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -520,7 +520,6 @@ namespace Babylon
                     auto callback{std::move(m_requestAnimationFrameCallback)};
                     callback({});
                 }
-                GetFrameBufferManager().Reset();
```
syntheticmagus, line 523: Without a reset or a recycling mechanism, the view IDs will increase without bound over time, eventually exceeding `bgfx::getCaps()->limits.maxViews`, right? Or is there another mechanism preventing this from happening?

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -234,7 +235,7 @@ namespace Babylon
     {
         FrameBufferManager()
         {
-            m_boundFrameBuffer = m_backBuffer = new FrameBufferData(BGFX_INVALID_HANDLE, GetNewViewId(), bgfx::getStats()->width, bgfx::getStats()->height);
+            m_boundFrameBuffer = m_defaultBackBuffer = new FrameBufferData(BGFX_INVALID_HANDLE, 0, bgfx::getStats()->width, bgfx::getStats()->height);
```
syntheticmagus, line 238: Is this a behavior change from the way it worked before? If so, do we know why it used to be different?

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -250,10 +251,13 @@ namespace Babylon
         void Bind(FrameBufferData* data)
         {
             m_boundFrameBuffer = data;
+            
+            const bool isNewFrameBuffer = m_boundFrameBuffer->ViewId == 0 && m_boundFrameBuffer != m_defaultBackBuffer;
+            const auto viewId = isNewFrameBuffer ? GetNewViewId() : m_boundFrameBuffer->ViewId;
```
syntheticmagus, line 256: Why do this here and not at `FrameBufferData`'s creation time? The check on 255 is a bit odd.

syntheticmagus (changes requested):
I think there's more going on here than meets the eye. The main thing that jumps out at me is that we probably need to be able to "recycle" view IDs one way or another so that long-running apps that spawn many views (among other scenarios) won't run out. That was what the old `RecycleSet` was for, but we specifically [got rid of the `RecycleSet`](https://github.com/BabylonJS/BabylonNative/pull/86/files) during a fix for some broken behavior that had to do with the validation tests. (I wonder if this has something to do with the current validation test failures, too?) I don't fully remember what exactly was broken that was fixed by #86, but it's probably worth checking into since this change seems to be targeting a shift back toward the sort of state-preserving architecture that we consciously made "a clean break from" when we adopted the current approach. Perhaps @CedricGuillemet remembers more about the context for this?

rgerd:
@syntheticmagus The PR you linked cleared some stuff up for me. I led myself into a false assumption after our earlier conversation that we wanted view id's to persist permanently, but now I see that my original read was more correct-- that we are trying to have view id's persist per-frame only. 

Please correct me if I'm wrong, but reassigning the view id's on bind and setViewport seems to be our method for reassigning view id's per-frame, but since Babylon.js could call setViewport multiple times on the same bound framebuffer per-frame, it seems like we need to actually just reassign per-frame.

I'm currently rolling back most of my changes to see if I can drive reassignment off of calls to `frameBufferManager->Reset()` (which is called at the end of each frame) and preserve most of the original core logic, which was correct the first time around.

I'll get this right eventually 👀

rgerd:
Going to rebase on the new threading changes and put up a fresh PR.

PR #437, opened 2020-10-06
Checking the added anchor at its new location

### LINKED ISSUE #435: Invalid argument error when calling XRFrame::CreateNativeAnchor
napiAnchor is being moved to the end of m_trackedAnchors, but is later referenced during the call to deferred.Resolve(). As it was previously moved, its value is now null.

rickfromwork:
Also replacing the push_back call with an emplace_back call, for a slight efficiency bonus.

Fixes #435

### DIFF
--- a/Plugins/NativeXr/Source/NativeXr.cpp
+++ b/Plugins/NativeXr/Source/NativeXr.cpp
@@ -1574,11 +1574,11 @@ namespace Babylon
                 xrAnchor->SetAnchor(nativeAnchor);
 
                 // Add the anchor to the list of tracked anchors.
-                m_trackedAnchors.push_back(std::move(napiAnchor));
+                m_trackedAnchors.emplace_back(std::move(napiAnchor));
 
                 // Resolve the promise with the newly created anchor.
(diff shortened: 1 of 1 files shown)


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1574,11 +1574,11 @@ namespace Babylon
                 xrAnchor->SetAnchor(nativeAnchor);
 
                 // Add the anchor to the list of tracked anchors.
-                m_trackedAnchors.push_back(std::move(napiAnchor));
+                m_trackedAnchors.emplace_back(std::move(napiAnchor));
 
                 // Resolve the promise with the newly created anchor.
                 auto deferred = Napi::Promise::Deferred::New(info.Env());
-                deferred.Resolve(napiAnchor.Value());
+                deferred.Resolve(m_trackedAnchors.back().Value());
```
Alex-MSFT, line 1581: D'oh thanks for the fix!  This was due to a hasty compiler warning fix when pulling it into ReactNative.

Alex-MSFT (approved):

satyapoojasama (approved):

PR #439, opened 2020-10-06
throw runtime_error on bad image data or missing app bundle resource

### LINKED ISSUE #292: [Apple] Crash when trying to load missing bundle resource
in UrlLib, on iOS (must be the same with MacOS), when trying to load a bundle asset (app: schema) that is not present, pathForResource returns an empty NSString that will trigger an exception with [NSURL fileURLWithPath]

CedricGuillemet:
fixes #292 

 - Added a comment for building MacOS with Xcode 12
 - Correctly thow an exception that is catched by JS when unable to decode image
 - throw when a missing resource in app bundle 
 
engine.createTexture("https://url_to_an_invalid_image"); is correctly trapped with an error in the JS console (by the thow line 1087 n NativeEngine.cpp)

but

engine.createTexture("app:///invalidResource.png"); is not trapped and the program stops despite the throw line 49 in UrlRequest.mm

I tried to see what's the difference but I was not able to put a finger on it. Any clue @bghgary ?

### DIFF
--- a/Dependencies/UrlLib/Source/Apple/UrlRequest.mm
+++ b/Dependencies/UrlLib/Source/Apple/UrlRequest.mm
@@ -44,6 +44,10 @@ namespace UrlLib
             if ([scheme isEqual:@"app"])
             {
                 NSString* path{[[NSBundle mainBundle] pathForResource:url.path ofType:nil]};
+                if (path == nil)
+                {
+                    throw std::runtime_error{"Invalid resource path in app bundle."};
+                }
                 url = [NSURL fileURLWithPath:path];
             }
             
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1082,8 +1082,10 @@ namespace Babylon
         arcana::make_task(arcana::threadpool_scheduler, m_cancelSource,
             [this, dataSpan, generateMips, invertY]() {
                 bimg::ImageContainer* image = bimg::imageParse(&m_allocator, dataSpan.data(), static_cast<uint32_t>(dataSpan.size()));
-                // todo: bimg::imageParse will return nullptr when trying to load a texture with an url that is not a valid texture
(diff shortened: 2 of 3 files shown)


syntheticmagus (approved):
Don't know if the non-trapping problem is still a work-in-progress, but as far as it goes the current change looks good to me.

Does the untrapped error throw a different kind of error? If the "app:///" URL leads down a different code path (which seems likely) and ends up generating something uncatchable like a segfault, that could explain the difference. If it's coming from inside our loading code, we might be able to preempt whatever's causing the problem and throw a normal exception instead, which should be catchable.

CedricGuillemet:
@syntheticmagus yes, http and app schemas have different code path. What I'm missing here is for both LoadTexture and ResponseType I check for null pointer and throw if necessary. the exception is trapped in one case and not the other.

Dependencies/UrlLib/Source/Apple/UrlRequest.mm
```diff
@@ -44,6 +44,10 @@ void ResponseType(UrlResponseType value)
             if ([scheme isEqual:@"app"])
             {
                 NSString* path{[[NSBundle mainBundle] pathForResource:url.path ofType:nil]};
+                if (path == nil)
+                {
+                    throw std::runtime_error{"Invalid resource path in app bundle."};
```
bghgary, line 49: Does this string end up on the JS side? If so, can we put the path in the error?

bghgary (approved):

bghgary:
I'll have to debug it to know what is going on, but we can probably merge regardless for now.

Dependencies/UrlLib/Source/Apple/UrlRequest.mm
CedricGuillemet: Not this one but the other throw is. When throwing, the reason is not forwarded to JS. JS handles the exception and display its own error message. But you get the JS callstack so it's pretty easy to figure out what's causing the exception.

PR #441, opened 2020-10-09
Canvas polyfill

CedricGuillemet:
Canvas polyfill plugin.
Can render text to texture and support some features of GUI rendering.
Need this babylon js PR: https://github.com/BabylonJS/Babylon.js/pull/10043


rickfromwork:
I noticed when rendering meshes to be used for the MeshButton3D, the meshes appeared inside-out.
The code run in the babylonNative app is equivalent to the provided playground scene - this does not repro on the playground.
https://www.babylonjs-playground.com/#4IN15I
[20201120_175553_HoloLens.zip](https://github.com/BabylonJS/BabylonNative/files/5576855/20201120_175553_HoloLens.zip)

CedricGuillemet:
@rickfromwork It looks like a depth test issue. I'll take a closer look.

rickfromwork:
> @CedricGuillemet
> I noticed when rendering meshes to be used for the MeshButton3D, the meshes appeared inside-out.
> The code run in the babylonNative app is equivalent to the provided playground scene - this does not repro on the playground.
> https://www.babylonjs-playground.com/#4IN15I
> 
> 5fb87a5b462b3239175721
> [20201120_175553_HoloLens.zip](https://github.com/BabylonJS/BabylonNative/files/5576855/20201120_175553_HoloLens.zip)

Another visual thing I noticed was this in VR:
image
Visually, the button rotates as I move my head around it, but in the opposite direction. It doesn't happen all the time, though.

CedricGuillemet:
I've changed your PG to scale only the root as the other meshes are parented

https://www.babylonjs-playground.com/#4IN15I#1

And it fixes both PG and babylon native. I pushed the change on this branch.

SeniorPlayer:
@CedricGuillemet Hi! Look at this [https://github.com/jdryg/vg-renderer](url) Will it help ?

CedricGuillemet:
> https://github.com/jdryg/vg-renderer

I'm using nanovg for 2 main reasons: older and more supported and bgfx already has a renderer.

SeniorPlayer:
> > https://github.com/jdryg/vg-renderer
> 
> I'm using nanovg for 2 main reasons: older and more supported and bgfx already has a renderer.

That renderer is just an example. It has bugs. should be careful.
[https://github.com/bkaradzic/bgfx/issues/1912](https://github.com/bkaradzic/bgfx/issues/1912)

rickfromwork:
@CedricGuillemet 
Hey, I ran a side-by-side comparison of the same scene running in babylon.js and babylonNative, and noticed that the text was not rendering, alongside the issue mentioned above. 
Babylon.js in-browser:
image

BabylonNative standalone:
image

I had to disable controller tracking for babylon.js, so there's no controller ray to go with it. The text panel is a copy of the experience.js code mentioned in the description.

CedricGuillemet:
I could not repro in the branch. current state looks like this:
image

rgerd:
@CedricGuillemet is it possible that @rickfromwork's issue isn't reproing in normal windowed mode because of an issue with swapped culling flags when rendering to a render target? In the second photo of his post, he's running in the Windows Mixed Reality simulator, so he's rendering to the eye render targets.

CedricGuillemet:
@rgerd Good point! I don't think render target are properly supported in this branch. I'm taking a look! Thanks!

arpu:
Hey,

not sure this is the right place to ask this, but with this Canvas Polyfill could i use threejs WebglRenderer?

CedricGuillemet:
> Hey,
> 
> not sure this is the right place to ask this, but with this Canvas Polyfill could i use threejs WebglRenderer?

For questions on Babylonjs and BabylonNative, you can create a post on the forum : https://forum.babylonjs.com/

On a general note, any PR on BabylonNative repo is meant to be for BabylonNative only without regards to any other rendering engine.

chrisfromwork:
@CedricGuillemet is there an update on the timeline for this work? It seems like 3d gui support in babylonnative will require this change?

CraigFeldspar:
Hey @CedricGuillemet, I'm trying to work on your fork, but didn't manage to build because of `nanovg_babylon.cpp` and `nanovg_babylon.h` missing. Any chance they weren't added to the git index ?
Thanks for this polyfill btw !

CedricGuillemet:
@CraigFeldspar This PR is under heavy work. I don't think it's usable right now. I'm working on making it work with latest master so I can do iterations. Keep an eye on this PR!

CedricGuillemet:
I've found a fix that will unblock me on this PR : https://github.com/BabylonJS/BabylonNative/pull/720
Let me finish updating to the latest master and then we can work in parallel.

CedricGuillemet:
@CraigFeldspar @chrisfromwork I've just pushed some changes that will build and render a quad with some text on it. I need to do some more changes to this PR : https://github.com/BabylonJS/Babylon.js/pull/10043 to fix 3d panel.
Then, we can split the work. 
The list in this comment is still relevant : https://github.com/BabylonJS/BabylonNative/issues/616#issuecomment-789693687
"A proper abstraction for the GUI " is the work on babylonjs mentioned earlier.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -108,6 +109,7 @@ extern "C"
                 Babylon::Polyfills::Window::Initialize(env);
 
                 Babylon::Polyfills::XMLHttpRequest::Initialize(env);
+                Babylon::Polyfills::Canvas::Initialize(env);
```
chrisfromwork, line 112: it might be worth having a default font specified during this initialization. Having to do a static call to the native engine to load a font by default feels kind of clunky.

Do we know whether we could distribute any default ttfs with the babylonnative codebase compared to relying on a http request to load one? BabylonNative will need to support offline usage

Apps/Playground/Scripts/experience.js
```diff
@@ -90,6 +90,111 @@ CreateBoxAsync().then(function () {
         scene.createDefaultLight(true);
     }
 
+    engine.updateDynamicTexture = function (texture, canvas, invertY, premulAlpha, format) {
```
chrisfromwork, line 93: what are we doing so that folks don't have to add this javascript code for canvasess to function correctly?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1021,6 +1022,13 @@ namespace Babylon
             });
     }
 
+    void NativeEngine::CopyTexture(const Napi::CallbackInfo& info)
+    {
+        const auto textureDestination = info[0].As<Napi::External<TextureData>>().Data();
```
chrisfromwork, line 1027: is there any clean up required for the original texture this data's handle was referencing?

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
CedricGuillemet: You can bundle ttf fonts in your app (iOS or Android and other) and then load it using app:// scheme

I agree loading a font async when starting the app is far from great.

Apps/Playground/Scripts/experience.js
CedricGuillemet: This code will end up being in nativeEngine.ts. I didn't open a PR for that yet because it may change a little during develpment and because I want the canvasAbstraction PR to be merged first.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: that's part of the 'copy texture' task on this list : https://github.com/BabylonJS/BabylonNative/issues/616#issuecomment-831162396

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
ryantrem: How does this compare to the browser? Do browsers have “built in” fonts, or do you just have access to system fonts, or what? Web pages don’t have to “bundle” a font to be able to display text, including displaying text in Babylon.js correct?

CedricGuillemet: Only guess but I think browsers use system font and (maybe) system API to render text. I believe it's possible to render text inside a CPU memory buffer and copy that buffer to the GPU if needed (in the case of Canvas to texture copy). That's something we might evaluate in the future (and compare quality and performance). That said, we can use ttf from the system instead of providing our own on Windows. But it's not possible to have access to fonts on Apple devices.

CedricGuillemet: I think I can add a method to do ttf loading when initializing the canvas plugin.

Apps/Playground/Win32/App.cpp
```diff
@@ -143,6 +145,7 @@ namespace
         });
 
         Babylon::ScriptLoader loader{*runtime};
+
```
bghgary, line 148: nit: why newline here?

Polyfills/Canvas/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+    "Source/Context.h"
+    "Source/MeasureText.cpp"
+    "Source/MeasureText.h"
+    "Source/nanovg_babylon.cpp"
+    "Source/nanovg_babylon.h"
+    )
+
+file(GLOB FONT_SOURCES ${BGFX_DIR}/examples/common/font/*.cpp)
+file(GLOB NANOVG_SOURCES ${BGFX_DIR}/examples/common/nanovg/nanovg.cpp)
+set(ATLAS_SOURCES ${BGFX_DIR}/examples/common/cube_atlas.cpp)
```
bghgary, line 15: Oy, using code from the examples. Is there a better way to handle this?

Polyfills/Canvas/Include/Babylon/Polyfills/Canvas.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <napi/env.h>
+
+namespace Babylon
+{
+    class Graphics;
+}
```
bghgary, line 8: What's this for?

Polyfills/Canvas/Source/Canvas.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <Babylon/Polyfills/Canvas.h>
+#include <GraphicsImpl.h>
+
+namespace Babylon::Polyfills::Internal
+{
+    class NativeCanvas final : public Napi::ObjectWrap<NativeCanvas>
+    {
+    public:
+        using ParentT = Napi::ObjectWrap<NativeCanvas>;
```
bghgary, line 11: Why is this public?

bghgary (commented):
This needs an virtual in-person review also.

Polyfills/Canvas/CMakeLists.txt
CedricGuillemet: Another solution is to find each github subrepo et add them to github.

Polyfills/Canvas/Include/Babylon/Polyfills/Canvas.h
CedricGuillemet: Left over. Removed.

Polyfills/Canvas/Source/Canvas.h
CedricGuillemet: Moved.

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+    void NativeCanvas::CreateInstance(Napi::Env env)
+    {
+        Napi::HandleScope scope{env};
+
+        Napi::Function func = ParentT::DefineClass(
+            env,
+            JS_CONSTRUCTOR_NAME,
+            {
+                ParentT::StaticMethod("loadTTF", &NativeCanvas::LoadTTF),
```
chrisfromwork, line 21: nit: "ParentT::" isn't needed.

Polyfills/Canvas/Source/Canvas.h
chrisfromwork: why do we define this ParentT type?

chrisfromwork: it feels like we don't do this elsewhere/having a custom type for a commonly used type may generate more confusion than worth

.github/workflows/nightly.yml
```diff
@@ -62,4 +62,4 @@ jobs:
       with:
           name: ResultImages
           path: |
-            buildWin32_x64/Apps/ValidationTests/Results
+            buildWin32_x64/Apps/ValidationTests/Results
```
chrisfromwork, line 65: nit: if this didn't change, please revert

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
chrisfromwork: Is there an example for ttf loading in a pure typescript context? iirc, providing the ttf data to babylonnative required accessing private/inaccessible babylon.js functions.

Polyfills/Canvas/CMakeLists.txt
chrisfromwork: @bghgary is using example code here a reason to block this review? or will it suffice for the time being if we file a follow up issue?

bghgary: We don't need to block the review.

chrisfromwork (approved):
In general i think this is in a good state for a first check in. @CedricGuillemet its still somewhat unclear what we need to do to load the first font in a pure typescript context since some of the methods for providing the font data to the engine isn't accessible through public types/methods. If you have suggestions for how to do this please let us know

Apps/ValidationTests/Scripts/text_rendering_test.js
```diff
@@ -0,0 +1,139 @@
...
+    var webGLTexture = texture._hardwareTexture.underlyingResource;
+    this._native.copyTexture(webGLTexture, canvas.getCanvasTexture());
+    texture.isReady = true;
+};
+
+console.log("Reading font");
+var textureGround;
+var font = "bold 44px monospace";
+BABYLON.Tools.LoadFile("https://raw.githubusercontent.com/CedricGuillemet/dump/master/droidsans.ttf", (data) => {
+    _native.NativeCanvas.loadTTF("droidsans", data);
```
chrisfromwork, line 43: @CedricGuillemet This is the specific function I'm not 100% sure how to access with react-native. Is the _native global something we can easily access in a pure typescript environment? @ryantrem do you know what this code would look like if we added something to the brn .ts files or the app.tsx?

Core/Graphics/Include/Shared/Babylon/Graphics.h
```diff
@@ -25,6 +25,7 @@ namespace Babylon
         void UpdateSize(size_t width, size_t height);
 
         void AddToJavaScript(Napi::Env);
+        static Graphics& GetFromJavaScript(Napi::Env);
```
bghgary, line 28: Is this necessary?

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -108,6 +108,8 @@ namespace Babylon
         // if no other draw calls are submitted to the view.
         if (doTouch)
         {
+            // discard everything or touch call will use what is bound.
+            encoder->discard();
```
bghgary, line 112: I don't think this is necessary anymore because we are discarding everything except for bindings now.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: It would be good to put a comment here so that people know this code is not correct.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -0,0 +1,694 @@
...
+        m_cancellationSource->cancel();
+    }
+
+    NVGcolor StringToColor(const std::string& colorString)
+    {
+        std::string str = colorString;
+        std::transform(str.begin(), str.end(), str.begin(),
+            [](unsigned char c) { return std::tolower(c); });
+
+        static const std::map<std::string, uint32_t> webColors = {
```
bghgary, line 104: consider putting this in a separate header file

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -0,0 +1,694 @@
...
+            {"wheat", 0xf5deb3},
+            {"white", 0xffffff},
+            {"whitesmoke", 0xf5f5f5},
+            {"yellow", 0xffff00},
+            {"yellowgreen", 0x9acd32}};
+        
+        if (str[0] == '#' && str.length() == 4)
+        {
+            unsigned int components[4];
+            int count = sscanf(str.c_str(), "#%1x%1x%1x", &components[0], &components[1], &components[2]);
```
bghgary, line 256: Try using sscanf_s (https://en.cppreference.com/w/c/io/fscanf)?

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -0,0 +1,694 @@
...
+    }
+
+    void Context::FillRect(const Napi::CallbackInfo& info)
+    {
+        auto left = info[0].As<Napi::Number>().FloatValue();
+        auto top = info[1].As<Napi::Number>().FloatValue();
+        auto width = info[2].As<Napi::Number>().FloatValue();
+        auto height = info[3].As<Napi::Number>().FloatValue();
+
+        NVGpaint paint = nvgLinearGradient(m_nvg, 0, 5, 0, 10, nvgRGBA(0, 160, 192, 255), nvgRGBA(0, 160, 192, 255));
```
bghgary, line 300: Should put comments on functions that are not complete yet.

bghgary: Or a comment at the top saying that these functions are not all done and some are done but not fully tested.

Polyfills/Canvas/Source/nanovg_babylon.cpp
```diff
@@ -0,0 +1,1175 @@
...
+#include "nanovg.h"
+
+#include <bgfx/bgfx.h>
+#include <bgfx/embedded_shader.h>
+
+#include <bx/bx.h>
+#include <bx/allocator.h>
+#include <bx/uint32_t.h>
+
+#include "GraphicsImpl.h"
```
bghgary, line 36: is this necessary?

bghgary (approved):

Apps/ValidationTests/Scripts/text_rendering_test.js
ryantrem, line 43: I'm not sure if this is on the radar, but should we be exposing this through a [`FontFace`](https://developer.mozilla.org/en-US/docs/Web/API/FontFace) polyfill? Then the calling code would just do:
```ts
const fontFace = new FontFace('droidsans', 'url(https://raw.githubusercontent.com/CedricGuillemet/dump/master/droidsans.ttf)');
fontFace.load().then(() => {
    ...
    const textBlock = new BABYLON.GUI.TextBlock();
    textBlock.fontFamily = 'droidsans';
    ...
});
```

Core/Graphics/Include/Shared/Babylon/Graphics.h
CedricGuillemet: Fixed

Polyfills/Canvas/Source/nanovg_babylon.cpp
CedricGuillemet: uint32_t.h is not. Others are.

Polyfills/Canvas/Source/Context.cpp
CedricGuillemet: yes, next PR.

Polyfills/Canvas/Source/Context.cpp
CedricGuillemet: Will do tests and push it in next PR if interesting results.

Issue #272, opened 2020-05-22
Readme Update - HoloLens 2


thomlucc:
Add info and steps in the Readme/Quick start guide on HoloLens 2 support

bghgary:
Be good to add a few lines to explain status for the 4.2 release.

PR #449, opened 2020-10-14
Add HoloLens 2 support to documentation and Playground app

chrisfromwork:
This review does the following:
1. We now have a flag that can be enabled in the Playground app to run through HoloLens 2 hand tracking scenarios
1. We now have documentation that calls out how to setup a HoloLens 2 Visual Studio solution

This change is related to: https://github.com/BabylonJS/BabylonNative/issues/272


Apps/Playground/Scripts/experience.js
```diff
@@ -219,7 +220,33 @@ CreateBoxAsync().then(function () {
                     });
                 }
 
-                xr.baseExperience.enterXRAsync(vr ? "immersive-vr" : "immersive-ar", "unbounded", xr.renderTarget);
+                let sessionMode = vr ? "immersive-vr" : "immersive-ar"
+                if (hololens) {
+                    // Because HoloLens 2 is a head mounted display, its Babylon.js immersive experience more closely aligns to vr
+                    sessionMode = "immersive-vr";
```
rickfromwork, line 226: We could inline this above as
`// Because HoloLens 2 is a head mounted display, its Babylon.js immersive experience more closely aligns to vr
let sessionMode = (vr || hololens) ? "immersive-vr" : "immersive-ar"`

Apps/Playground/Scripts/experience.js
```diff
@@ -219,7 +220,33 @@ CreateBoxAsync().then(function () {
...
+
+                    // Below is an example for enabling hand tracking. The code is not unique to HoloLens 2, and may be reused for other WebXR hand tracking enabled devices.
+                    // Create a mesh to display for each hand joint
+                    const jointMesh = BABYLON.SphereBuilder.CreateSphere("jointParent", { diameter: 1, }, scene);
```
rickfromwork, line 230: It's alright to leave this in, but joints will still be displayed even without passing in jointMesh. What passing in a mesh allows us to do is to customize the joints used (default joint is a grey sphere).

rickfromwork: If we want to show a bare minimum approach, we can get rid of the jointMesh part, and only pass xr.input to the features manager.

Apps/Playground/Scripts/experience.js
chrisfromwork: I don't have a strong attachment to how this is currently written, but my personal preference would be to have things related to hololens clearly called out in their own code stub for legibility. What do others think?

Apps/Playground/Scripts/experience.js
chrisfromwork: sounds good

syntheticmagus (approved):

rickfromwork (approved):

Issue #339, opened 2020-07-24
Animated GLB not loading in BabylonReactNative


RonnieAshlock:
Scenes will not load on device or emulator:

https://raw.githubusercontent.com/BabylonJS/Assets/master/meshes/alien.glb
https://raw.githubusercontent.com/BabylonJS/Assets/master/meshes/fish.glb
https://raw.githubusercontent.com/RonnieAshlock/TestAssets/master/GLB/ZeroGravity.glb

CedricGuillemet:
At least 2 different issues:
- shader compilation (glslang/c++ side) for alien.glb
- missing sampling constants for fish.glb

CedricGuillemet:
Fix for fish.glb in babylonjs : https://github.com/BabylonJS/BabylonNative/issues/339
@RonnieAshlock alien.glb is listed twice. Do you have a 3rd .glb to check?

CedricGuillemet:
@syntheticmagus It asserts in glslang due to the traversers (it seems). Can you please take a quick look at trying to load alien.glb ?

RonnieAshlock:
@CedricGuillemet Thank you for catching that. I updated the list with the third GLB I tested that would not load. Apologies for the error.

CedricGuillemet:
https://raw.githubusercontent.com/BabylonJS/Assets/master/meshes/fish.glb
https://raw.githubusercontent.com/RonnieAshlock/TestAssets/master/GLB/ZeroGravity.glb

Related to the same fix in babylonjs regarding sampling modes.

syntheticmagus:
Observations on this so far:
- The assert was caused by an unhandled fall-through case in one of the traversers for types that are not considered either vectors or matrices; I have a fix for this in a branch.
- Once this assert is resolved, it reveals another problem with the collision of attribute positions due to there being non-generic attributes which can be placed at positions after `FIRST_GENERIC_ATTRIBUTE_LOCATION`. This is not a complicated problem, but it's not completely clear to me what the right solution to it is; and since I'm pretty sure @bghgary already had this model working in an earlier version of Babylon Native, I want to ask him about it before trying naive things like simply increasing the `FIRST_GENERIC_ATTRIBUTE_LOCATION` variable.
[...]

PR #450, opened 2020-10-14
Change vertex attribute slot assignment.

Drigax:
This PR proposes a fix for #339.

Due to limitations with OpenGL's GL_MAX_VERTEX_ATTRIBS constraint, and SPIR-V adding location decorators to our vertex attribute declaractions, we run out of locations using our current attribute assignment schema:

staticAttributeLocationSchema

This proposal changes the current traverser behavior for android to use arbitrary slot locations, and our vertex buffer construction logic to be able to gracefully re-arrange how our attributes are passed to match the order that we declare them in our shader.

1) Traversers now use incrementing attribute locations.
2) Change BabylonJS::NativeEngine.RecordVertexBuffer to record vertex buffers in ascending native attribute location order. (via https://github.com/BabylonJS/Babylon.js/pull/9183)
2) Change BabylonNative::NativeEngine::RecordVertexBuffer to use the parsed attribute location order for vertex buffer layout indexing.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -26,23 +24,17 @@ namespace Babylon
...
-            constexpr uint32_t NEAREST_LINEAR =                 BGFX_SAMPLER_MAG_POINT  |   BGFX_SAMPLER_DEFAULT    |   BGFX_SAMPLER_DEFAULT    ;
-            constexpr uint32_t NEAREST_NEAREST_MIPLINEAR =      BGFX_SAMPLER_MAG_POINT  |   BGFX_SAMPLER_MIN_POINT  |   BGFX_SAMPLER_DEFAULT    ;
-            constexpr uint32_t LINEAR_NEAREST_MIPNEAREST =      BGFX_SAMPLER_DEFAULT    |   BGFX_SAMPLER_MIN_POINT  |   BGFX_SAMPLER_MIP_POINT  ;
-            constexpr uint32_t LINEAR_NEAREST_MIPLINEAR =       BGFX_SAMPLER_DEFAULT    |   BGFX_SAMPLER_MIN_POINT  |   BGFX_SAMPLER_DEFAULT    ;
-            constexpr uint32_t LINEAR_LINEAR_MIPNEAREST =       BGFX_SAMPLER_DEFAULT    |   BGFX_SAMPLER_DEFAULT    |   BGFX_SAMPLER_MIP_POINT  ;
-            constexpr uint32_t LINEAR_NEAREST =                 BGFX_SAMPLER_DEFAULT    |   BGFX_SAMPLER_MIN_POINT  |   BGFX_SAMPLER_DEFAULT    ;
-            // clang-format on
+            constexpr uint32_t NEAREST = BGFX_SAMPLER_MAG_POINT | BGFX_SAMPLER_MIN_POINT;                               // nearest is mag = nearest and min = nearest and mip = linear
```
Drigax, line 27: Not sure where all these formatting changes came from, going to give this another pass.

bghgary:
Nice picture! :-)

> SPIR-V adding location decorators to our vertex attribute declaractions

Ideally, we shouldn't be using SPIR-V at all. Hopefully that doesn't change things.

> Change BabylonJS::NativeEngine.RecordVertexBuffer

Can you put your changes in a branch and send a link to compare against master? Diffing babylon.max.js doesn't work on GitHub since it's too big.

Drigax:
Sure, the matching PR in Babylon.js is: https://github.com/BabylonJS/Babylon.js/pull/9183

Currently we are passing an attribute index which corresponds to the attribute ordering in Babylon.js, however after we recompile our shader, we aren't guaranteed the same attribute ordering using this new scheme. 

Also, since we have a mapping of attribute->actual index via `m_currentProgram->VertexAttributeLocations` it might make more sense to use the mapping instead of passing an index?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -680,15 +680,17 @@ namespace Babylon
 
         bgfx::VertexLayout vertexLayout{};
         vertexLayout.begin();
-        const bgfx::Attrib::Enum attrib = static_cast<bgfx::Attrib::Enum>(location);
+
+        const auto locationIndex = m_currentProgram->VertexAttributeLocations[kind];
```
bghgary, line 684: I don't understand this change. Why is this necessary? The js side gets the location ultimately via `NativeEngine::GetAttributes` which is all on the native side of things. What does using strings for kind do?

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -451,8 +451,21 @@ namespace Babylon::ShaderCompilerTraversers
 #define IF_NAME_RETURN_ATTRIB(varyingName, attrib, newName)  \
     if (std::strcmp(name, varyingName) == 0)                 \
     {                                                        \
-        return {static_cast<unsigned int>(attrib), newName}; \
+        return {static_cast<unsigned int>(attrib), name}; \
     }
+#if ANDROID
```
CedricGuillemet, line 456: Is this change only for Android or for OpenGL?

Drigax: This should be for OpenGL, not just android. I just used this preprocessor flag to test. How should we branch based off the current renderer?

Plugins/NativeEngine/Source/NativeEngine.cpp
Drigax: I agree, I don't think this (or my other js changes) are necessary either, since we are already given the proper vertex location. I think currently the order in which we put elements in `vertexArray.vertexBuffers` is causing my current issues with data ordering

Drigax: I removed this additional lookup and reverted my changes to the babylon.js contract, instead I believe the real issue is how we're passing vertex arrays from our babylon representation to bgfx's, i opted to change our datastructure for representing which array maps to a given attribute location so that we can insert these elements in an arbitrary order.

Drigax:
OK, I think this is ready for proper review.

The current approach is to modify our input attribute names during traversal such that the name now matches the attribute string defined via:

https://github.com/bkaradzic/bgfx/blob/master/src/renderer_gl.cpp#L45-L65

The upside of this approach is that 1) we don't need to modify bgfx's internal workings to support this, instead we're mangling our shader to fit the expected attribute - index schema.

The downside of this is that we're mangling our shader, any inspection into the shader contents will not really make sense, since we're now changing what the actual variables mean (eg: in the shader the symbol `a_position0` may now be used as the vertex color)

An alternative that @bghgary proposed is to instead modify this attribute table according to our attribute order that we encounter during traversal. 

The upside of this approach is that the shader would retain its original composition. only the symbol-bgfx::Attrib association will be modified (ex: a_texcoord0 may now be associated with bgfx::Attrib::Position internally)

The downside of this approach is that now we need to manage this table as we change our program, each shader program will require this table to be modified before we perform any vertex binding. This may have unintended side effects.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
Drigax: Changed `ANDROID` preprocessor to `APIOpenGL`, that should be appropriate, right?

CedricGuillemet: yep!

Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
```diff
@@ -7,11 +7,17 @@
 #include <SPIRV/GlslangToSpv.h>
 #include <spirv_parser.hpp>
 #include <spirv_glsl.hpp>
+#include <bgfx/bgfx.h>
+
+namespace bgfx::gl
+{
+    void setAttribName(bgfx::Attrib::Enum, const char*);
+}
```
bghgary, line 15: Are you still using this?

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -446,13 +449,49 @@ namespace Babylon::ShaderCompilerTraversers
                 }
             }
 
+#if __APPLE__ | APIOpenGL
```
bghgary, line 452: Probably mean this?
```suggestion
#if __APPLE__ || APIOpenGL
```

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -446,13 +449,49 @@ namespace Babylon::ShaderCompilerTraversers
...
+            BX_STATIC_ASSERT(bgfx::Attrib::Count == BX_COUNTOF(s_attribName) );
+#endif
+
             std::pair<unsigned int, const char*> GetVaryingLocationAndNewNameForName(const char* name)
             {
 #define IF_NAME_RETURN_ATTRIB(varyingName, attrib, newName)  \
     if (std::strcmp(name, varyingName) == 0)                 \
     {                                                        \
         return {static_cast<unsigned int>(attrib), newName}; \
     }
```
bghgary, line 485: This should be in the else case?

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -446,13 +449,49 @@ namespace Babylon::ShaderCompilerTraversers
...
+                "a_texcoord0",
+                "a_texcoord1",
+                "a_texcoord2",
+                "a_texcoord3",
+                "a_texcoord4",
+                "a_texcoord5",
+                "a_texcoord6",
+                "a_texcoord7",
+            };
+            BX_STATIC_ASSERT(bgfx::Attrib::Count == BX_COUNTOF(s_attribName) );
```
bghgary, line 476: nit:
```suggestion
            BX_STATIC_ASSERT(bgfx::Attrib::Count == BX_COUNTOF(s_attribName));
```

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -446,13 +449,49 @@ namespace Babylon::ShaderCompilerTraversers
                 }
             }
 
+#if __APPLE__ | APIOpenGL
+            // This table is a copy of the table bgfx uses for vertex attribute -> shader symbol association.
+            // copied from renderer_gl.cpp.
+            constexpr static const char * s_attribName[] =
```
bghgary, line 455: nit:
```suggestion
            constexpr static const char* s_attribName[] =
```

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -5,12 +5,15 @@
 #include <glslang/MachineIndependent/RemoveTree.h>
 
 #include <bgfx/bgfx.h>
+#include <bx/bx.h>
```
bghgary, line 8: is this necessary?

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -5,12 +5,15 @@
...
 
 #include <arcana/experimental/array.h>
 
 #include <gsl/gsl>
 
 #include <stdexcept>
+#include <iostream>
+#include <arcana/macros.h>
```
bghgary, line 16: What are these used for?

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -446,13 +449,49 @@ namespace Babylon::ShaderCompilerTraversers
...
         return {static_cast<unsigned int>(attrib), newName}; \
     }
+#if __APPLE__ | APIOpenGL
+                // For OpenGL and Metal platforms, we have an issue where we have a hard limit on the number shader attributes supported.
+                // To work around this issue, instead of mapping our attributes to the most similar bgfx::attribute, instead replace
+                // the first attribute encountered with the symbol bgfx uses for attribute 0 and increment for each subsequent attribute encountered.
+                // This will cause our shader to have nonsensical naming, but will allow us to efficiently "pack" the attributes.
+                UNUSED(name);
+                m_genericAttributesRunningCount++;
+                return {static_cast<unsigned int>(m_genericAttributesRunningCount-1), s_attribName[std::min(static_cast<unsigned int>(m_genericAttributesRunningCount-1), static_cast<unsigned int>(bgfx::Attrib::Count)-1)]};
```
bghgary, line 493: if m_genericAttributesRunningCount is bigger than bgfx::Attrib::Count, we should probably throw or something. Silently using the last index doesn't seem right.

Plugins/NativeEngine/Source/ShaderCompilerCommon.h
```diff
@@ -5,6 +5,7 @@
 #include <gsl/gsl>
 #include <spirv_cross.hpp>
 #include <spirv_parser.hpp>
+#include <map>
```
bghgary, line 8:
```suggestion
#include <unordered_map>
```

bghgary: Should probably include <string> as well.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
Drigax: iostream - unused, I'll remove this.

arcana/macros.h - I'm currently using their UNUSED(variable) macro. (L488)

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
Drigax: good point, moving it down.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
Drigax: fixing.

Plugins/NativeEngine/Source/ShaderCompilerOpenGL.cpp
Drigax: No, I'll remove this as well.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
Drigax: Currently using it for their static assert and array count implementation (L476), This may not be necessary, however.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
Drigax: Ok, that seems more responsible than reusing the last index for multiple attributes.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
bghgary: We haven't been consistently doing this, but using this seems like a good idea.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
bghgary: Ok, it's fine so that you can copy and paste. Is it a copy and paste?

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
bghgary: If you don't mind, also fix the UV one.

Drigax: I changed the other attribute location choosing code path to do this check as well.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
Drigax: yes, it is a copy paste

bghgary (approved):
Looks good. Make sure it works on all the relevant platforms.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -674,7 +675,7 @@ namespace Babylon
 
         vertexBufferData->EnsureFinalized(info.Env(), vertexLayout);
 
-        vertexArray.vertexBuffers.push_back({vertexBufferData, byteOffset / byteStride, bgfx::createVertexLayout(vertexLayout)});
+        vertexArray.vertexBuffers[location] = {vertexBufferData, byteOffset / byteStride, bgfx::createVertexLayout(vertexLayout)};
```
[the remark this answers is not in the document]
Drigax: the only time I can see this is if we have a shader with duplicate attributes with identical names, which will probably fail compilation, otherwise we seem to ensure that we assign unique locations to each attribute, would it be a good idea to add an assertion to ensure this?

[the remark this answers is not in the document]
bghgary: use `.insert` instead of `[]` and it will return if the entry already exists.

PR #453, opened 2020-10-19
min RT size of 8x8

CedricGuillemet:


bghgary (commented):
Looks okay to me, but probably better for someone else to sign off.

syntheticmagus (commented):
+1 to what Gary said. Looks good to me, but I think @ryantrem or @Alex-MSFT are the most knowledgeable about the implications this will have.

ryantrem:
Under what circumstances do we get a 0 width or height?

ryantrem (approved):

CedricGuillemet:
We get 0 sized window when the engine view is toggled on. I guess the view is created but layout didn't happen yet so  eglQuerySurface(display, surface, EGL_WIDTH/EGL_HEIGHT) returns 0;

PR #456, opened 2020-10-21
Fix a few memory management issues in xr.mm

ryantrem:
The main issue being fixed here is that after
`id<MTLTexture> texture = [metalDevice newTextureWithDescriptor:textureDescriptor];`
We need an explicit call to `retain`. In Cocoa, these static factory functions always call `autorelease` for the object being constructed, which means `release` will automatically be called when the containing `@autoreleasepool` scope exits. However, we previously didn't have an `@autoreleasepool` scope because we were managing the JS thread scheduling ourselves (and I didn't understand memory management patterns with Objective C and Cocoa). React Native's `CallInvoker::invokeAsync` does this correctly by wrapping the call to the function passed to it in an `@autoreleasepool`. What this means is that the `@autoreleasepool` is actually auto releasing the `MTLTexture`, so the textures are actually being released too early (I guess it usually works because the memory hasn't been overwritten yet) and then our code also explicitly releases it, so it is also double released (or rather we try to release an object that has already been released). So the proper way to deal with this is to explicitly call `retain` to match our explicit call to `release`.

I fixed up some additional things in this file now that I have a better understanding of Objective C + Cocoa memory management. 
- Use a factory function to create `MTLTextureDescriptor`s (which calls `autorelease`) and then just let the `CallInvoker::invokeAsync` scoped `@autoreleasepool` clean them up.
- Remove all direct usages of `dealloc`. This is equivalent to calling `delete` on a ref counted object, which is bad.
- Fix the fallout from the above (e.g. `release` the child `MTKView` after inserting it into the parent, since the parent will `retain` it and `release` it when it is removed from the children collection).

### DIFF
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -346,7 +346,7 @@ namespace {
         , displayOrientedTransform.columns[3][2] };
 }
 
-- (void)session:(ARSession*)__unused session didAddAnchors:(nonnull NSArray<__kindof ARAnchor *> *)anchors {
+- (void)session:(ARSession *)__unused session didAddAnchors:(nonnull NSArray<__kindof ARAnchor *> *)anchors {
     if (!planeDetectionEnabled) {
         return;
     }
@@ -366,7 +366,7 @@ namespace {
     [self UnlockPlanes];
 }
 
-- (void)session:(ARSession*)__unused session didUpdateAnchors:(nonnull NSArray<__kindof ARAnchor *> *)anchors {
+- (void)session:(ARSession *)__unused session didUpdateAnchors:(nonnull NSArray<__kindof ARAnchor *> *)anchors {
     if (!planeDetectionEnabled) {
         return;
     }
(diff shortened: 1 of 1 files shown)


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -250,7 +250,7 @@ - (CVMetalTextureRef)getCameraTexture:(CVPixelBufferRef)pixelBuffer plane:(int)p
  Checks whether the camera UVs need to be updated based on the orientation and size of the view port, and updates them if necessary.
  @return True if the camera UVs were updated, false otherwise.
 */
-- (Boolean)checkAndUpdateCameraUVs:(ARFrame *)frame {
```
bghgary, line 253: I'm not sure you should change this convention in ObjC files. It seems to be the default for Xcode. That said, I don't have much experience here.

bghgary (approved):

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Yeah I don't mind either way, but the standard convention I've seen is to have the space.

Alex-MSFT (approved):

CedricGuillemet:
LGTM! Did you see a difference in allocated memory reported by Xcode when switching on/off XR? I saw a 10mb memory footprint increase each time I exit XR.

ryantrem:
> LGTM! Did you see a difference in allocated memory reported by Xcode when switching on/off XR? I saw a 10mb memory footprint increase each time I exit XR.

I still see the 10mb footprint increase each XR cycle :/

PR #457, opened 2020-10-22
Enable plane detection on HoloLens 2/OpenXR devices and expose logic for richer scene understanding

chrisfromwork:
In this pull request we have the following changes:
1. We move to the 104 OpenXR release where Scene Understanding is available as an OpenXR preview extension.
1. A new SceneUnderstanding struct has been created and added to the OpenXR session context for windows uwp OpenXR builds.
1. SceneUnderstanding now populates plane information on XRFrames based on the existing xr plane detection API. These planes will now be available through the Babylon.js WebXRPlaneDetector feature (similar to existing support for ARCore and ARKit).
1. SceneUnderstanding exposes extension points for obtaining richer, world geometry information such as meshes and object categorization. This functionality can be accessed by external libs referencing the xrInternal cmake definition. It allows developers to start using scene understanding now with BabylonNative while we better define a non-plane world geometry story in Babylon.js


Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -0,0 +1,497 @@
...
+                }
+
+                auto& babylonPlane = *m_babylonPlanes[key];
+                babylonPlane.Center = XrPoseToBabylonPose(object->Pose);
+                babylonPlane.PolygonFormat = xr::PolygonFormat::XYZ;
+
+                constexpr uint8_t VALUES_IN_POINT = 3;
+                constexpr uint8_t VALUES_IN_XYZ_QUAD = 12;
+                babylonPlane.Polygon.resize(VALUES_IN_XYZ_QUAD);
+                babylonPlane.Polygon[0] = -1.f * xrPlane.extent.width / 2.f;
```
rgerd, line 373: Worth noting here that the order of these vertices implicitly defines a front and back of the plane, and since we don't have any normal information we can't properly identify which is which. This means we can't use the simple (but important) performance optimization of backface culling when visualizing the planes, and I assume it could be confusing to see both sides of a plane, for example if we have one aligned to a wall on one side of a corner and we're standing on the other side.

Not requesting action, maybe a comment, otherwise just pointing out.

chrisfromwork: yeah im not sure what the best practice will be here. The OpenXR spec around scene object detection is also in flux. I'll definitely add this comment for the time being.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -0,0 +1,497 @@
+#define NOMINMAX // avoid including macro max definitions in favor of limit max definitions
```
tymlipari, line 1: nit: Wrap this in an `#ifndef NOMINMAX` to avoid problems if someone ever adds this to a precompiled header or something

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -0,0 +1,497 @@
...
+#include "XrRegistry.h"
+
+using namespace xr;
+
+constexpr SceneUnderstanding::DetectionBoundaryType DEFAULT_BOUNDARY_TYPE = SceneUnderstanding::DetectionBoundaryType::Sphere;
+constexpr float DEFAULT_SPHERE_RADIUS = 5.f;
+constexpr double DEFAULT_UPDATE_INTERVAL_IN_SECONDS = 2;
+constexpr double NANOSECONDS_IN_SECOND = 1000000000;
+
+inline xr::Pose XrPoseToBabylonPose(XrPosef pose)
```
tymlipari, line 13: If this function is only used in this file (I'm assuming it is) add the `static` qualifier to keep it local to this TU

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -0,0 +1,497 @@
...
+        },
+        {
+            pose.orientation.x,
+            pose.orientation.y,
+            pose.orientation.z,
+            pose.orientation.w
+        }};
+}
+
+SceneUnderstanding::SceneUnderstanding::InitOptions::InitOptions(
```
tymlipari, line 30: Should this class be called `XrSceneUnderstanding` for consistency?

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -0,0 +1,497 @@
...
+    , SceneObjects(sceneObjects)
+    , UpdatedObjects(updatedObjects)
+    , RemovedObjects(removedObjects)
+{
+}
+
+struct SceneUnderstanding::SceneUnderstanding::Impl
+{
+public:
+    Impl() {}
```
tymlipari, line 89: nit: If this is empty, can we omit it? Not sure what the convention in this project is.

Also add a break between this and the next function.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+            m_sceneObjectLocations.resize(sceneObjects.size());
+            XrSceneObjectLocationsMSFT locations{ XR_TYPE_SCENE_OBJECT_LOCATIONS_MSFT };
+            locations.sceneObjectCount = static_cast<uint32_t>(m_sceneObjectLocations.size());
+            locations.sceneObjectLocations = m_sceneObjectLocations.data();
+
+            CHECK_XRCMD(extensions.xrLocateSceneObjectsMSFT(m_sceneHandle.Get(), &locateInfo, &locations));
+
+            // Get a set of the known objects to understand if any objects were removed
+            std::set<XrSceneObjectKeyMSFT> knownObjects{};
```
tymlipari, line 247: nit: consider `std::unordered_set`. Might be somewhat more efficient for this use case.

tymlipari (approved):

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1415,14 +1424,18 @@ namespace xr
...
     {
-        throw std::runtime_error("Planes not yet implemented for OpenXR.");
+        const auto& su = m_impl->sessionImpl.HmdImpl.Context.SceneUnderstanding();
+        return su.TryGetPlaneByID(id);
     }
 
     void System::Session::SetPlaneDetectionEnabled(bool) const
     {
-        throw std::runtime_error("Planes not yet implemented for OpenXR");
+        const auto& session = m_impl->HmdImpl.Context.Session();
```
Alex-MSFT, line 1435: Probably best to put this in an if statement for when the past in bool is true.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -0,0 +1,497 @@
...
+            m_updatedPlanes.clear();
+            m_removedObjects.clear();
+            m_removedMeshes.clear();
+            m_removedPlanes.clear();
+
+            switch (m_state)
+            {
+            case State::Ready:
+            {
+                HandleReady(time, space, extensions);
```
Alex-MSFT, line 154: Nit: The formatting on this makes the switch case kind of hard to read.  I would personally tab over everything under the switch once so the outer scopes are not inline with the inner scopes.

 also any reason to have the breaks outside of the scopes?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -283,6 +286,8 @@ namespace xr
         float DepthNearZ{ DEFAULT_DEPTH_NEAR_Z };
         float DepthFarZ{ DEFAULT_DEPTH_FAR_Z };
 
+
```
Alex-MSFT, line 289: Nit: remove

Alex-MSFT (approved):
The fulfillment of the contract looks good to me, but I am not too familiar with the APIs being used in SceneUnderstanding.cpp.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: I'm not sure if long term that will generate more confusion/lead folks to believe that this type is an OpenXR type. But, I also don't mind making this change/will go a head and update things.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: no preference on the breaks/brackets. I'll go ahead and make these changes :)

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: sounds good, I didn't know you could iterate over unordered_sets

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: Actually, I think we should keep this as SceneUnderstanding for now. We don't want to rename the header to XrSceneUnderstanding, because thats already used by the OpenXR-MixedReality dependency.

chrisfromwork:
going to try closing and reopening to retry ci

PR #461, opened 2020-10-26
Various error handling fixes

### LINKED ISSUE #224: Correctly propagate and handle JavaScript errors
- [x] Use Napi::Error for exceptions in native code
- [x] Make sure there is a good experience when JS throws an exception

### LINKED ISSUE #444: Error handling with NAPI/JSI is incomplete
From @ryantrem:
- [x] 1. An exception thrown in JS code where the bottom of the stack is from native code that is part of React Native - this behaves normally, in that we get a "red box" error in the app and an error on the console, but it doesn't cause the app to crash.
- [x] 2. An exception thrown in JS code where the bottom of the stack is from native code that is part of Babylon Native - this either causes the app to hang or crash, and the only place you can see the error is in logcat on Android or the debug output on iOS when the debugger is attached.
- [x] 3. An exception is thrown in native code where the stack contains JS code (e.g. called into native from JS) - this causes the app to crash and you can maybe see some native exception information in logcat.

### LINKED ISSUE #455: JavaScript exceptions from Scene.beforeRender cause bgfx to get into a bad state
If an exception is thrown from within `Scene.beforeRender`, it leaves the host app (seemingly something related to bgfx) in a bad and unrecoverable state. The easiest way to see this is actually to test it out in the Babylon React Native app. You can simply add `throw new Error("test");` to the existing `scene.beforeRender` callback and run the app.

bghgary:
Fixes #224
Fixes #444
Fixes #455

- Add default error handling behavior to AppRuntime that logs to platform-specific debug output.
- Add custom error handling callback to AppRuntime to enable error handling customization.
- Update WorkQueue to trap arcana errors. When error occurs, forward to error handler and stop the work queue.
- Update Graphics component to handle errors properly.
- Add a couple of helper methods to NAPI C++ layer for converting std::exception to Napi::Error.
- Implement Napi::Error in NAPI/JSI and a bunch of cleanup.
- Fix a problem in napi v8 where error messages were being lost.
- Update a bunch of code to use Napi::Error when errors are supposed to propagate to JS and std::runtime_error when they are not.


Core/AppRuntime/Source/AppRuntimeAndroid.cpp
```diff
@@ -1,9 +1,26 @@
...
+    {
+        try
+        {
+            std::rethrow_exception(ptr);
+        }
+        catch (const Napi::Error& error)
+        {
+            std::stringstream ss{};
+            ss << "Uncaught Error: " << error.Message() << std::endl;
+            __android_log_write(ANDROID_LOG_ERROR, "BabylonNative", ss.str().data());
```
ryantrem, line 23: Should this really be the default behavior, or should we defer to `console.error`? It seems like worst case, `console.error` already does this, and best case, you are in an environment where `console.error` does something better (obviously this doesn't apply to React Native specifically, but using it as an example, it already provides a nice `console.error` experience (e.g. surfaced in the app UI, not just in debug output).

Core/AppRuntime/Source/WorkQueue.cpp
```diff
@@ -42,6 +43,15 @@ namespace Babylon
         while (!m_cancelSource.cancelled())
         {
             m_dispatcher.blocking_tick(m_cancelSource);
+
+            // Call the unhandled exception handler if an exception is thrown in one of the tasks.
+            m_task = m_task.then(arcana::inline_scheduler, arcana::cancellation::none(), [this](const arcana::expected<void, std::exception_ptr>& result) {
+                if (result.has_error())
+                {
+                    m_unhandledExceptionHandler(result.error());
+                    m_cancelSource.cancel();
```
ryantrem, line 52: Are you sure we want to stop the loop? It seems like it is a lot more typical in JS environments to continue execution.

Core/AppRuntime/Source/AppRuntime.cpp
```diff
@@ -1,11 +1,15 @@
...
 
 namespace Babylon
 {
     AppRuntime::AppRuntime()
-        : m_workQueue{std::make_unique<WorkQueue>([this] { RunPlatformTier(); })}
+        : AppRuntime{DefaultUnhandledExceptionHandler}
+    {
+    }
+
+    AppRuntime::AppRuntime(std::function<void(std::exception_ptr)> unhandledExceptionHandler)
```
ryantrem, line 11: I kind of think this function should return a bool indicating whether the unhandled exception is being handled, and if not, then let it fall back to the default.

Core/AppRuntime/Source/WorkQueue.cpp
```diff
@@ -42,6 +43,15 @@ namespace Babylon
         while (!m_cancelSource.cancelled())
         {
             m_dispatcher.blocking_tick(m_cancelSource);
+
+            // Call the unhandled exception handler if an exception is thrown in one of the tasks.
+            m_task = m_task.then(arcana::inline_scheduler, arcana::cancellation::none(), [this](const arcana::expected<void, std::exception_ptr>& result) {
+                if (result.has_error())
+                {
+                    m_unhandledExceptionHandler(result.error());
```
ryantrem, line 51: It looks like providing this as an argument to the `WorkQueue` constructor is optional? Do we need to check it before calling it?

Core/Graphics/Source/Graphics.cpp
```diff
@@ -88,8 +94,15 @@ namespace Babylon
         }
         else
         {
-            return whenAllTask.then(Dispatcher, arcana::cancellation::none(), [this, &finished, &workDone]() mutable {
-                return RenderCurrentFrameAsync(finished, workDone);
+            return whenAllTask.then(Dispatcher, arcana::cancellation::none(), [this, &finished, &workDone, &error](const arcana::expected<void, std::exception_ptr>& result) mutable {
+                if (result.has_error())
+                {
+                    finished = true;
+                    error = result.error();
```
ryantrem, line 101: Why is this needed? Can't the caller just get it in a task continuation?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -168,22 +168,48 @@ namespace Babylon::Polyfills::Internal
...
+                {
+                    throw Napi::Error::New(env, result.error());
+                }
+
+                SetReadyState(ReadyState::Done);
+            });
+        }
+        catch (const std::exception& ex)
+        {
+            throw Napi::Error::New(info.Env(), ex.what());
```
ryantrem, line 211: I'm not quite following the distinction between when you pass in the actual `exception_ptr` vs the `what()` string. Can you explain?

Core/AppRuntime/CMakeLists.txt
```diff
@@ -4,10 +4,15 @@ if(NOT NAPI_JAVASCRIPT_ENGINE STREQUAL "JSI")
...
         "Source/AppRuntime.cpp"
         "Source/AppRuntime${NAPI_JAVASCRIPT_ENGINE}.cpp"
-        "Source/AppRuntime${BABYLON_NATIVE_PLATFORM}.cpp"
         "Source/WorkQueue.cpp"
         "Source/WorkQueue.h")
 
+    if(APPLE)
+        set(SOURCES ${SOURCES} "Source/AppRuntimeApple.mm")
+    else()
+        set(SOURCES ${SOURCES} "Source/AppRuntime${BABYLON_NATIVE_PLATFORM}.cpp")
```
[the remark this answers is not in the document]
bghgary: Right. I could just keep `${BABYLON_NATIVE_PLATFORM}`.

Core/AppRuntime/Source/WorkQueue.cpp
bghgary: Huh? JS shouldn't continue if there is an unhandled exception?

Core/AppRuntime/Source/AppRuntimeAndroid.cpp
bghgary: `console.error` may not be available if the polyfill is not included. I can perhaps check, but I'm not sure that's a good idea.

Core/AppRuntime/Source/AppRuntime.cpp
bghgary: I like it. I'll make the change.

Core/AppRuntime/Source/WorkQueue.cpp
bghgary: The default version points to a default function. `m_unhandledExceptionHandler` is always set to something.

Core/Graphics/Source/Graphics.cpp
bghgary: > Can't the caller just get it in a task continuation?

No. I thought the same thing, but this code doesn't return the last task of the chain. It returns the first task of the chain.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
bghgary: I'm not sure I understand the question. There are two separate things that can throw exceptions in this code.
1. The synchronous part of SendAsync may throw an exception.
2. The continuation may be in an exceptional state.

Does that answer the question? 🤔

syntheticmagus (approved):
Might take a little cajoling to merge this with master, but looks good to me.

ryantrem (approved):

Core/AppRuntime/Source/AppRuntime.cpp
bghgary: I ended up not doing this since it made the code a bit more difficult to understand.

Core/AppRuntime/Source/WorkQueue.cpp
bghgary: Removed.

bghgary:
@syntheticmagus Can you re-review the changes to WorkQueue? @ryantrem and I had a discussion about it and we need to be able to catch each continuation separately or some of them won't run when an exception happens.

PR #467, opened 2020-10-28
Update deps+ bgfx trace outputs

### LINKED ISSUE #494: bgfx update for vertex layout leak fix
fix is in bgfx.
An update is in progress in babylon native.

CedricGuillemet:
- update bgfx
~~- update open xr~~
~~- update arcore-android~~
- update ios cmake toolchain
~~- update glslang~~
- update babylon js with framebuffer destruction leak fixed
- optional lambda for bgfx trace output
- enable debug/leak checks in bgfx
- dispose engine in validation tests
- work on ValidationTests MacOS (does not work yet might be an issue in bgfx)

Change on babylonjs is visible here : https://github.com/BabylonJS/Babylon.js/pull/9410

with d3d11, there is still a leak with annotation `BGFX WARN 0x5f190380 RefCount is 4 (expected 0).`
Not an issue for React. I don't know how it affects memory footprint.

fixes #494


Core/Graphics/Include/Babylon/Graphics.h
```diff
@@ -31,6 +31,8 @@ namespace Babylon
             FinishRenderingCurrentFrame();
         }
 
+        void SetDiagnosticOutput(std::function<void(const char* output)> outputFunction);
```
syntheticmagus, line 34: I don't think it's a huge deal, but contract-wise this feels a little weird. While it's not exposing anything substantially new through this contract *per se*, using this method (unless I misunderstand it's purpose) will sign the user up to receive diagnostic messages from bgfx, which from the perspective of this contract is an implementation detail. A little odd, but maybe not *too* odd, so I can think of three ways we could address this.
1. Leave it here as-is and just document that its behavior exposes diagnostic information that is beyond the scope of our own contract.
2. Have this function only on the `Impl` such that it is part of the internal contract, where the exposure of implementation details and the expectation that users know what they're doing are not new.
3. Leave this here but find some way to designate it as something tricky, perhaps by something as simple as renaming it or by putting it on an internal object or something.

I personally kind of like option 2, but I think any of them could work. @bghgary, thoughts or preferences?

syntheticmagus (approved):
Other than the one contract consideration with the Graphics component, looks good to me.

Core/Graphics/Include/Babylon/Graphics.h
CedricGuillemet: If I move this function to the impl, how can I call it to set the output in the validation tests app?

CedricGuillemet:
I've found 2 issues while testing more in depth:
- MSAA is broken in bgfx. enabling it will crash on Android
- Texture mismatch on Metal (looks like normals are used instead of diffuse for texturing). Can it be an issue on our side @Drigax ?

ryantrem:
This is a big change! Will you test it out in Babylon React Native too?

CedricGuillemet:
@ryantrem yes! the previous bgfx update went smoothly but this one brings a lot of issues. I already did 1 PR on bgfx to fix DX shader compilation. I think it will need another one for Metal impl and maybe openGL.

Apps/Playground/Win32/App.cpp
```diff
@@ -127,7 +127,7 @@ namespace
         std::string scriptsRootUrl = GetUrlFromPath(GetModulePath().parent_path().parent_path() / "Scripts");
 
         Babylon::ScriptLoader loader{*runtime};
-        loader.Eval("document = {}", "");
+        loader.Eval("document = {removeEventListener:function(){}}", "");
```
bghgary, line 130: I think we need to try to avoid making this hack worse. Can we file an issue?

Apps/ValidationTests/CMakeLists.txt
```diff
@@ -76,6 +76,7 @@ target_link_to_dependencies(ValidationTests
     PRIVATE XMLHttpRequest)
 
 if(APPLE AND NOT IOS)
+    target_link_libraries(ValidationTests PUBLIC "-framework MetalKit")
```
bghgary, line 79: Does this need to be PUBLIC? Since it's the final target, there should be no one depending on this target right? Does PRIVATE work?

Apps/ValidationTests/Shared/TestUtils.h
```diff
@@ -29,6 +29,12 @@ namespace
     int errorCode{};
 }
 
+// can't externalize variable with ObjC++. Using a function instead.
+int GetExitCode()
+{
+    return errorCode;
+}
```
bghgary, line 36: where is this used?

Core/Graphics/Include/Babylon/Graphics.h
bghgary: Option 2 doesn't seem possible. This is exposed all the way to the app level. I'm not sure we should expose this on the graphics level though. I would think if we want some diagnostic output, then it should be global to all the modules we write and not just graphics. Hooking up bgfx debug output is just part of that story which probably should be configurable.

Apps/Playground/Win32/App.cpp
CedricGuillemet: Yes! And I will remove this change in the app.

Apps/ValidationTests/CMakeLists.txt
CedricGuillemet: I will try but I think it will work.

Apps/ValidationTests/Shared/TestUtils.h
CedricGuillemet: in MacOS appdelegate. I forgot to push that change.

PR #468, opened 2020-10-28
Improve framebuffer view logic

rgerd:
This PR improves the logic around managing view id assignments for framebuffers: 
* Assign new view id's per-frame instead of per-bind, and notify framebuffers of dirty view id's by maintaining a table of self-destructible weak references.
* Set the view rect to the size of the framebuffer by default, and conditionally set the size to match the window, instead of always just sizing to the window, since in XR we aren't always rendering to a window-sized backbuffer.
* Keep the default backbuffer assigned to view id 0 in order to stay consistent with bgfx's convention. This is strictly better than what we were doing before (assigning `BGFX_INVALID_HANDLE` to view id 1) because we were effectively assigning two view id's to the default backbuffer, one through default bgfx behavior and one explicitly in the framebuffer manager constructor.
* Fix a bug in the culling swap logic for render targets (discovered when a post-processing validation test failed) by manually swapping the bits instead of using XOR.
* Re-add bgfx::touch call so that we clear our views even when no geometry is rendered inside the clipping volume.

### DIFF
--- a/Apps/ValidationTests/Scripts/config.json
+++ b/Apps/ValidationTests/Scripts/config.json
@@ -43,6 +43,7 @@
         {
             "title": "Black and White post-process",
             "playgroundId": "#N55Q2M#0",
+            "renderCount": 20,
             "referenceImage": "bwpp.png"
         },
         {
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1386,12 +1386,16 @@ namespace Babylon
                     bgfx::setUniform({it.first}, value.Data.data(), value.ElementLength);
                 }
             }
-            // change culling
+
+            // We need to explicitly swap the culling state flags (instead of XOR)
+            // because we would like to preserve the no culling configuration, which is 00.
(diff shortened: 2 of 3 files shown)


Apps/ValidationTests/Scripts/config.json
```diff
@@ -43,6 +43,7 @@
         {
             "title": "Black and White post-process",
             "playgroundId": "#N55Q2M#0",
+            "renderCount": 20,
```
syntheticmagus, line 46: Is this an inherent part of the change, or was this motivated by the validation test strangeness we've been seeing recently?

syntheticmagus (approved):
Looks good to me. In light of recent changes and validation test behavior, we should make sure to manually test this thoroughly; but if manual testing doesn't throw up any red flags, I'm on board!

Apps/ValidationTests/Scripts/config.json
rgerd: I did this to pass the test locally. It seems we need a couple of frames in both post-processing scenarios for the rendering to "settle." I have a feeling it's because we're calling `bgfx::touch` again-- we have the renderCount on chromatic aberration from when we had the touch call, but we didn't have it on this test which was added after the touch call was removed.

PR #476, opened 2020-10-29
Enable WebXRMeshDetector for OpenXR/UWP in BabylonNative

chrisfromwork:
This review is the native implementation for the following Babylon.js review:
https://github.com/BabylonJS/Babylon.js/pull/9307

When testing this change locally, one will need the 104 OpenXR release installed on a HoloLens 2 device. They will also need to enable preview OpenXR extensions


Dependencies/xr/Include/XR.h
```diff
@@ -120,6 +111,46 @@ namespace xr
...
+    };
+
+    struct GeometryDetectorOptions
+    {
+        xr::DetectionBoundary DetectionBoundary{};
+        double UpdateInterval{ 10 };
+    };
+
+    typedef int64_t GeometryId;
+    constexpr int32_t INVALID_GEOMETRY_ID = -1;
```
chrisfromwork, line 152: this should be int64_t

chrisfromwork: we could also use size_t

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -34,32 +34,27 @@
 using namespace android;
 using namespace android::global;
 
-namespace xr
-{
+namespace xr {
```
syntheticmagus, line 37: It looks like you got attacked by a rogue clang formatter. :smiley: If you didn't intend to change this file (or, at least, not this substantially), can you revert the unintended changes to make it easier to identify the intended ones?

Dependencies/xr/Include/XR.h
```diff
@@ -223,6 +298,7 @@ namespace xr
                 void UpdateAnchor(Anchor&) const;
                 void DeleteAnchor(Anchor&) const;
                 Plane& GetPlaneByID(Plane::Identifier) const;
+                Mesh& GetMeshByID(Mesh::Identifier) const;
```
syntheticmagus, line 301: nit: I think we might want to move away from the pattern of having behavior on the frame at some point (having the frame be pure state makes it easier to reason about stuff like threading). This one's of a piece with the ones that are currently here, though, so I think it's not a big deal either way.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -5,11 +5,32 @@
 #include "SceneUnderstanding.h"
 #include "XrRegistry.h"
 
+#pragma region DirectX
```
syntheticmagus, line 8: Typically in other parts of Babylon Native we've split these sorts of things out into platform-specific headers/implementations that are included as appropriate by CMake. Could that approach be used for this?

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -375,6 +441,17 @@ struct SceneUnderstanding::SceneUnderstanding::Impl
...
 
+                babylonPlane.ParentGeometryId = static_cast<xr::GeometryId>(objectKey);
+
+                if (s_sceneObjectTypeMap.count(object->Kind) > 0)
+                {
+                    babylonPlane.ParentGeometryType = s_sceneObjectTypeMap.at(object->Kind);
+                }
+                else
+                {
+                    babylonPlane.ParentGeometryType = s_sceneObjectTypeMap.at(XrSceneObjectKindTypeMSFT::XR_SCENE_OBJECT_KIND_TYPE_UNKNOWN_MSFT);
```
syntheticmagus, line 452: nit: I think this file might be exclusively intended for exploring capabilities which are currently Microsoft-specific, in which case this is already in an area where it won't clash. But just in case, if I'm incorrect and this is in a code path that other vendors/platforms may encounter, we should be proactive about preparing to have other use cases coexist alongside our vendor-specific implementations.

Dependencies/xr/Include/XR.h
chrisfromwork: I'm not sure what the preference will be here. Currently collections for planes, meshes, etc are declared on the XRFrame for WebXR. Is the call out here that we would prefer to have mesh/plane information maintained elsewhere in native code? Or is the suggestion to try and change the WebXR spec?

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: scene understanding is currently a microsoft specific preview extension. I think there's hope that long term it will become a khronos group extension that other vendors implement. I would guess that at this time, the XrSceneObjectKindTypeMSFT will be renamed.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: this seems reasonable. The directx specific region was a weaker attempt at this :)

Dependencies/xr/Source/ARCore/XR.cpp
chrisfromwork: this should hopefully now be fixed

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: moved to XrPlatform.h, which is windows specific

Dependencies/xr/Include/XR.h
```diff
@@ -120,6 +111,63 @@ namespace xr
...
+    struct Frustum
+    {
+        Pose Pose{};
+        FieldOfView FOV{};
+        float FarDistance;
+    };
+
+    struct DetectionBoundary
+    {
+        DetectionBoundaryType Type{ DetectionBoundaryType::Sphere };
```
satyapoojasama, line 138: nit: if only sphere is used, do we want to name it accordingly?

chrisfromwork: sphere is the default value used here. The detection boundary type is specified through the Babylon.js api

ryantrem:
What are all the submodule updates about?

chrisfromwork:
> What are all the submodule updates about?

These aren't intentional. I'll look to revert them now.

Dependencies/xr/Include/XR.h
```diff
@@ -120,6 +111,63 @@ namespace xr
...
+        Unknown,
+        Background,
+        Wall,
+        Floor,
+        Ceiling,
+        Platform,
+        Undefined
+    };
+
+    const std::map<xr::SceneObjectType, std::string> SceneObjectTypeNames
```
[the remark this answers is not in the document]
chrisfromwork: yes in UpdateSceneObjects in NativeXr.cpp

Dependencies/xr/Include/XR.h
```diff
@@ -212,18 +269,42 @@ namespace xr
...
                 private:
-                    static inline Identifier NEXT_ID{0};
+                    static inline Identifier NEXT_ID{ 0 };
+                };
+
+                struct Mesh
+                {
+                    typedef float PositionType;
+                    typedef uint32_t IndexType;
+                    typedef float NormalType;
```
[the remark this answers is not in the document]
chrisfromwork: no specific reason, looks like using is used throughout this file. I will update this

Dependencies/xr/Include/XR.h
```diff
@@ -212,18 +269,42 @@ namespace xr
...
+
+                    using Identifier = size_t;
+                    const Identifier ID{ NEXT_ID++ };
+                    std::vector<PositionType> Positions{};
```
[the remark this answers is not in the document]
chrisfromwork: right now this vector is similar to the float array used in Babylon.js vertex data. it is a vector containing 3n floats where n is the number of vertices. That going said, we could instead do a vector3f type here that then is just copied to the Babylon.js vertex data array in NativeXr

Dependencies/xr/Include/XR.h
```diff
@@ -120,6 +111,63 @@ namespace xr
...
+        Box,
+        Frustum,
+        Sphere
+    };
+
+    struct Frustum
+    {
+        Pose Pose{};
+        FieldOfView FOV{};
+        float FarDistance;
```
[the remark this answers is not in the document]
chrisfromwork: I don't have a strong preference on this one. Do we have an example for deconstructing a projection matrix into these values?

Dependencies/xr/Include/XR.h
```diff
@@ -59,21 +59,23 @@ namespace xr
         size_t Height{};
     };
 
+    struct Vector3f
```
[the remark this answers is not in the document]
chrisfromwork: I personally prefer having x/y/z accessors compared to using array indices. I'll introduce a vector4f datatype as well.

---
In reply to: [533654130](https://github.com/BabylonJS/BabylonNative/pull/476#discussion_r533654130) [](ancestors = 533654130)

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
chrisfromwork: FWIW, the babylon.js native webxr spec has a frustum definition mapping 1:1 to these values [https://github.com/BabylonJS/Babylon.js/blob/66883a9689a94440dd060dffe702a7636879155d/src/LibDeclarations/webxr.nativeextensions.d.ts#L18](https://github.com/BabylonJS/Babylon.js/blob/66883a9689a94440dd060dffe702a7636879155d/src/LibDeclarations/webxr.nativeextensions.d.ts#L18)

---
In reply to: [533716211](https://github.com/BabylonJS/BabylonNative/pull/476#discussion_r533716211) [](ancestors = 533716211)

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
ryantrem: I think that would make it a lot easier to understand the code, and prevent the possibility of missing values (e.g. an vector with a length that is not a multiple of 3).

Dependencies/xr/Include/XR.h
chrisfromwork: my suggestion would be to keep things this way for now. we can update things here as the WebXR spec changes

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
chrisfromwork: Vector4f has been added and used throughout

Dependencies/xr/Include/XR.h
```diff
@@ -120,6 +111,63 @@ namespace xr
...
+        FieldOfView FOV{};
+        float FarDistance;
+    };
+
+    struct DetectionBoundary
+    {
+        DetectionBoundaryType Type{ DetectionBoundaryType::Sphere };
+        Frustum Frustum{};
+        float SphereRadius{ 5.f };
+        Vector3f BoxDimensions;
```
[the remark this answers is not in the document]
chrisfromwork: this should now be fixed throughout

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
chrisfromwork: these have been updated to usings

chrisfromwork:
I believe the majority of feedback is addressed in this review. We should avoid checking this in until we have an understanding around exposing mesh winding order: https://github.com/BabylonJS/Babylon.js/pull/9575/files

Dependencies/xr/Source/OpenXR/SceneUnderstanding.h
```diff
@@ -66,39 +53,20 @@ namespace xr
...
-
         SceneUnderstanding();
         ~SceneUnderstanding();
-        void Initialize(const InitOptions options) const;
+        void Initialize(InitOptions options) const;
         void UpdateFrame(UpdateFrameArgs args) const;
-        void UpdateSceneObjects(UpdateSceneObjectsArgs args) const;
+        System::Session::Frame::SceneObject& TryGetSceneObjectByID(const System::Session::Frame::SceneObject::Identifier id) const;
         System::Session::Frame::Plane& TryGetPlaneByID(const System::Session::Frame::Plane::Identifier id) const;
+        System::Session::Frame::Mesh& TryGetMeshByID(const System::Session::Frame::Mesh::Identifier id) const;
```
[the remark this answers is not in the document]
syntheticmagus: nit: Looking at the implementation, it looks like they throw if they fail. I don't know if they need `Try` in the name if they throw; I think `Try` only makes sense if it returns a `bool` and takes a non-`const` ref to update.

syntheticmagus (approved):
I have no blocking comments on this, other than the CI checks. However, I'm not working in this code on a day-to-day basis, so I think my comments should probably carry less weight than those of people who are directly involved in the current ongoing work.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1185,6 +1189,10 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
         return m_impl->sessionImpl.GetPlaneByID(planeID);
     }
```
Alex-MSFT, line 1191: Does iOS not need a stub implementation for System::Session::Frame::GetSceneObjectByID?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -203,6 +203,90 @@ namespace
         return Napi::Persistent(jsInputSource);
     }
 
+    xr::DetectionBoundary CreateDetectionBoundary(const Napi::Object& object)
```
Alex-MSFT, line 206: Would it be better for this to take in an xr::DetectionBoundary&, and just fill the values on the object that way rather than creating a new DetectionBoundary object then copying the values over after returning?

Alex-MSFT (approved):

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: this seems like a good suggestion

---
In reply to: [540498049](https://github.com/BabylonJS/BabylonNative/pull/476#discussion_r540498049) [](ancestors = 540498049)

Dependencies/xr/Source/OpenXR/SceneUnderstanding.h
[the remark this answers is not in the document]
chrisfromwork: IMO, long term we should avoid failing fast in favor of returning undefineds/booleans for this platform code. That going said the current naming was chosen to remain consistent with TryGetPlaneByID

---
In reply to: [539762959](https://github.com/BabylonJS/BabylonNative/pull/476#discussion_r539762959) [](ancestors = 539762959)

Dependencies/xr/Source/ARKit/XR.mm
chrisfromwork: yes it does

Dependencies/xr/Source/OpenXR/SceneUnderstanding.h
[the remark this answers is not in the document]
chrisfromwork: actually this is specifically scoped to scene understanding, i misunderstood what was going on here. I'll rename things for now to drop the Try*

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -34,61 +55,23 @@ SceneUnderstanding::SceneUnderstanding::InitOptions::InitOptions(
...
-    , UpdatedObjects(updatedObjects)
-    , RemovedObjects(removedObjects)
-{
-}
-
 struct SceneUnderstanding::SceneUnderstanding::Impl
 {
 public:
-    void Initialize(const InitOptions& options)
+    void Initialize(InitOptions options)
```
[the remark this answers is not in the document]
chrisfromwork: this should be fixed/changed back

Dependencies/xr/Source/OpenXR/SceneUnderstanding.h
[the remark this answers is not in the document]
chrisfromwork: this should now be fixed

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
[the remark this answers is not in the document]
chrisfromwork: this should now be fixed

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1181,10 +1187,18 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
         m_impl->sessionImpl.DeleteAnchor(anchor);
     }
 
+System::Session::Frame::SceneObject& System::Session::Frame::GetSceneObjectByID(System::Session::Frame::SceneObject::Identifier) const {
```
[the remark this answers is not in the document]
chrisfromwork: This should now be fixed

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
chrisfromwork: This should now be fixed

Dependencies/xr/Include/XR.h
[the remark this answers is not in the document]
chrisfromwork: this should now be fixed.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1195,10 +1199,18 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
         m_impl->sessionImpl.DeleteAnchor(anchor);
     }
 
+    System::Session::Frame::SceneObject& System::Session::Frame::GetSceneObjectByID(System::Session::Frame::SceneObject::Identifier) const {
+        throw std::runtime_error("not implemented");
```
satyapoojasama, line 1203: are these to be implemented later as well?

Dependencies/xr/Include/XR.h
```diff
@@ -120,6 +112,61 @@ namespace xr
...
+    struct DetectionBoundary
+    {
+        DetectionBoundaryType Type{ DetectionBoundaryType::Sphere };
+        std::variant<float, Frustum, Vector3f> Data{ 5.f };
+    };
+
+    struct GeometryDetectorOptions
+    {
+        xr::DetectionBoundary DetectionBoundary{};
+        double UpdateInterval{10};
```
rgerd, line 146: Do we need extra precision here?

Dependencies/xr/Source/ARKit/XR.mm
chrisfromwork: yes, this logic exists for ARKit, but does not currently exist for ARCore

Dependencies/xr/Include/XR.h
rgerd, line 146: nit: Do we need extra precision here?

Dependencies/xr/Include/XR.h
```diff
@@ -120,6 +112,61 @@ namespace xr
...
+
+    enum class SceneObjectType
+    {
+        Unknown,
+        Background,
+        Wall,
+        Floor,
+        Ceiling,
+        Platform,
+        Undefined
```
rgerd, line 157: nit: Is `Undefined` meant to silently communicate an error in recognition? Would it suffice to just have `Unknown` as the default and check if `ID == INVALID_ID` to see if the object has been initialized?

Dependencies/xr/Source/ARKit/XR.mm
satyapoojasama: nit: may be add a TODO comment if that helps

rgerd (approved):

Dependencies/xr/Source/ARKit/XR.mm
chrisfromwork: I've added https://github.com/BabylonJS/BabylonNative/issues/542

I'll add a TODO calling out this issue if i need to submit additional changes/rerun ci.

Dependencies/xr/Include/XR.h
chrisfromwork: iirc, javascript defaults to doubles for number values and this value comes from typescript/javascript. I don't know that anyone will specify this amount of precision, but its currently possible and we can support it.

Dependencies/xr/Include/XR.h
chrisfromwork: Unknown is a type specific to OpenXR. ARKit has its own enumeration. I think we should relook this when consolidating other platform's scene object/compound mesh types. I had been using Unknown with the expectation that a platform attempted enumerating the object but did not know what it was. vs Undefined where the platform did not attempt any enumeration/does not support enumeration. But, I see how this can be confusing/improved.

satyapoojasama (approved):

PR #478, opened 2020-10-30
Bring back XHR fix

ryantrem:
- Ensure validation tests are able to log even when the app crashes.
- Switch to v8 as it doesn't crash.
- Add build steps to produce crash dumps and upload them as an artifact. This also includes the exe, but I haven't figured out how to create the pdb yet. My vote is we try to figure that out, and once we have that, we can disable v8, repro the crash, and then debug. I think this is less urgent than some of the other remaining work.


CedricGuillemet (approved):
LGTM! Can you please update the github action for the nightlty as well?

ryantrem:
> LGTM! Can you please update the github action for the nightlty as well?

I made the changes here: https://github.com/ryantrem/BabylonNative/commit/220c4576d9e56d5e7e33457a526a5f13bf5bc5fe
But haven't figured out how to test them, so I'll save that for another PR.

Issue #465, opened 2020-10-27
UrlRequest does not handle errors


ryantrem:
`UrlRequest` does not currently handle errors, which can easily lead to app crashes. For example, simply trying to load a glb from a URL without an internet connection crashes the host app (on Android, not sure yet what the behavior on iOS is).

ryantrem:
For Android, functions like `JNIEnv::CallVoidMethod` don't throw a C++ exception when a Java exception occurs. Rather, after calling functions like `JNIEnv::CallVoidMethod` we need to call `JNIEnv::ExceptionCheck()` to determine whether there was a Java exception, and if so call `JNIEnv::ExceptionOccurred()` to get the throwable (which itself we will need a jni wrapper for) and pull out info like the exception message, then call `JNIEnv::ExceptionClear()` to clear out the current Java exception. Maybe wrap all this up into a macro or something that we can wrap all our other JNIEnv calls in?

PR #479, opened 2020-11-03
UrlLib/XMLHttpRequest error handling

ryantrem:
On both Android and iOS, if a client side error (such as no internet connection) occurred, the app would crash. In the case of iOS, it was because we were throwing exceptions inside a Cocoa callback rather than propagating them throw the arcana task returned from `SendAsync`. In the case of Android, it was because we were not checking for Java exceptions and converting them to C++ exceptions.

I also snuck one other small change in here, which is to not copy data out of `NSData` into a `std::vector`, but rather just take a strong reference to the `NSData` and return a `gsl::span` that directly references the `NSData`. This just eliminates an extra data copy, which can be rather large depending on the file being downloaded.

This mostly addresses https://github.com/BabylonJS/BabylonNative/issues/465, though we also need https://github.com/BabylonJS/Babylon.js/pull/9335 (included in @babylonjs/core@4.2.0-rc.2) for everything to be fully working.

### DIFF
--- a/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
+++ b/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
@@ -108,10 +108,14 @@ namespace java::lang
     {
     public:
         Throwable(jthrowable throwable);
+        ~Throwable();
 
         String GetMessage() const;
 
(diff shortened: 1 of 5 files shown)


bghgary (approved):
Chatted offline..

PR #481, opened 2020-11-04
Set cmake overrides for OpenXR-SDK's jsoncpp dependency to unblock ci build

chrisfromwork:
CI machines now resolve jsoncpp based on C:/msys64/mingw64/lib/cmake/jsoncpp, which did not exist before. This change updates our cmake to never use system resolve jsoncpp

### DIFF
--- a/Dependencies/xr/CMakeLists.txt
+++ b/Dependencies/xr/CMakeLists.txt
@@ -12,6 +12,9 @@ elseif (IOS)
     set(SOURCES ${SOURCES}
         "Source/ARKit/XR.mm")
 else()
+    # Avoid picking up system installed jsoncpp in favor of source distributed with openxr_loader project
+    set(JSONCPP_FOUND OFF CACHE BOOL "disable using system installed jsoncpp" FORCE)
+    set(BUILD_WITH_SYSTEM_JSONCPP OFF CACHE BOOL "disable using system installed jsoncpp" FORCE)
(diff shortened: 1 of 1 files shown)


Dependencies/xr/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ elseif (IOS)
     set(SOURCES ${SOURCES}
         "Source/ARKit/XR.mm")
 else()
+    # Avoid picking up system installed jsoncpp in favor of source distributed with openxr_loader project
+    set(BUILD_WITH_SYSTEM_JSONCPP OFF CACHE BOOL "disable using system installed jsoncpp" FORCE)
```
bghgary, line 16: You probably don't want to use FORCE here since that will prevent users from using a custom one if they really wanted to.

bghgary (approved):
Looks good. Thanks a bunch for fixing!

Dependencies/xr/CMakeLists.txt
chrisfromwork: We can try not using force, my concern is that removing force would cause this value to get overwritten in the jsoncpp setup

chrisfromwork: please avoid checking this in, ill push a commit to test whether force is needed

chrisfromwork: looks like ci succeeded without the force flag, ill go ahead and update this once the build completes

PR #484, opened 2020-11-06
small fix to setTimeout

chrisfromwork:
Right now setTimeout(0) executes synchronously. This breaks logic in Babylon.js. In this change we force all setTimeout callbacks to have to occur on the next frame.


bghgary (approved):
Might be good to do a few tests to make sure everything is fine.

chrisfromwork:
> Might be good to do a few tests to make sure everything is fine.

I edited experience.js locally and saw that the change indeed fixed the problem:

const observable = new BABYLON.Observable();

for (let i = 0; i < 10; i++) {
    const value = i;
    observable.addOnce(() => { console.log("attempt " + value); });
}

observable.notifyObservers();

Are there any other suggested tests to run?

bghgary:
Probably good enough.

PR #487, opened 2020-11-06
Support updating window in ARKit XR

ryantrem:
We were previously just passing a `void* windowPtr` into the XR layer (specifically for iOS/ARKit right now), but this isn't good enough since the window can change during an XR session. The change here is to instead pass in a `std::function<void*()> windowProvider` so the XR layer can query the current window. This happens inside `GetNextFramed`, which is similar to what the ARCore XR layer does (it polls the current gl surface on each GetNextFrame). This addresses https://github.com/BabylonJS/BabylonReactNative/issues/106. Note the reason it was *crashing* before is because we immediately released the `xrView` and allowed the superview to manage its lifetime, but the superview (the window) can be replaced, in which case the `xrView` is released and we continue to try to use it. Now we retain shared ownership and don't release the `xrView` until we end the session.

### DIFF
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -542,16 +542,17 @@ namespace xr {
         float DepthNearZ{ DEFAULT_DEPTH_NEAR_Z };
         float DepthFarZ{ DEFAULT_DEPTH_FAR_Z };
         
-        Impl(System::Impl& systemImpl, void* graphicsContext, void* window)
-            : SystemImpl{ systemImpl } {
-            metalDevice = id<MTLDevice>(graphicsContext);
-            UIView* MainView = (UIView*)window;
-            
+        Impl(System::Impl& systemImpl, void* graphicsContext, std::function<void*()> windowProvider)
+            : SystemImpl{ systemImpl }
+            , getMainView{ [windowProvider{ std::move(windowProvider) }] { return reinterpret_cast<UIView*>(windowProvider()); } }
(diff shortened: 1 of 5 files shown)


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -542,16 +542,17 @@ bool TryInitialize() {
...
-        Impl(System::Impl& systemImpl, void* graphicsContext, void* window)
-            : SystemImpl{ systemImpl } {
-            metalDevice = id<MTLDevice>(graphicsContext);
-            UIView* MainView = (UIView*)window;
-            
+        Impl(System::Impl& systemImpl, void* graphicsContext, std::function<void*()> windowProvider)
+            : SystemImpl{ systemImpl }
+            , getMainView{ [windowProvider{ std::move(windowProvider) }] { return reinterpret_cast<UIView*>(windowProvider()); } }
+            , metalDevice{ id<MTLDevice>(graphicsContext) } {
+
```
Alex-MSFT, line 549: Nit: No new line

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -542,16 +542,17 @@ bool TryInitialize() {
...
+            UIView* mainView = getMainView();
+
```
Alex-MSFT, line 551:
```suggestion
```

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT, line 549:
```suggestion
```

Alex-MSFT (approved):

PR #493, opened 2020-11-10
Fixing the UrlRequest on Android for local files

CoPrez:
With the current implementation of UrlRequest for Android it does not support loading local files. When using a URL in the format of `file://<local_path>` Babylon-Native crashes on Android (but does work on iOS).

The root cause of the crash is because of an incorrect cast from UrlConnection to HttpUrlConnection where depending on the URL the underlying class may be a FileUrlConnection. This change matches the iOS implementation where it checks if the UrlConnection is of the type HttpUrlConnection and if so it checks the response code. If the UrlConnection is of any other type it returns HTTP 200. I verified that if you give it a local file url that is invalid it gracefully handles the error and properly returns a status code of 0.

In order to accomadate the ability to check the java class type I refactored the wrapper class to expose the JNI function [IsAssignableFrom](https://docs.oracle.com/en/java/javase/13/docs/specs/jni/functions.html#isassignablefrom) on a new Class wapper object. The Class object can either be constructed from a string or from a jobject. I replaced the jclass property of the Object wrapper with the new Class wrapper class to make accessing it easier.

With the new changes you can do a typecheck like this:
`if (connection.Class().IsAssignableFrom({"java/net/HttpURLConnection"}))`

### DIFF
--- a/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
+++ b/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
@@ -82,6 +82,8 @@ namespace java::lang
         Class(const char* className);
         Class(const jclass classObj);
 
+        operator jclass() const;
+
         bool IsAssignableFrom(Class otherClass);
 
     protected:
@@ -99,9 +101,8 @@ namespace java::lang
         Object(const char* className, jobject object);
 
         JNIEnv* m_env;
-        const jclass m_class;
+        const class Class m_class;
         jobject m_object;
-        class Class m_classWrapper;
     };
 
     class String
--- a/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
+++ b/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
@@ -76,10 +76,24 @@ namespace java::lang
         jbyteArray m_byteArray;
     };
 
+    class Class
+    {
+    public:
+        Class(const char* className);
+        Class(const jclass classObj);
+
+        bool IsAssignableFrom(Class otherClass);
+
+    protected:
+        JNIEnv* m_env;
+        const jclass m_class;
+    };
+
     class Object
     {
     public:
         operator jobject() const;
+        Class Class();
 
     protected:
         Object(const char* className, jobject object);
@@ -87,6 +101,7 @@ namespace java::lang
         JNIEnv* m_env;
         const jclass m_class;
         jobject m_object;
+        class Class m_classWrapper;
     };
 
     class String
--- a/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
+++ b/Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
@@ -98,7 +98,8 @@ namespace java::lang
         Class Class();
 
     protected:
-        Object(const char* className, jobject object);
+        Object(const char* className);
+        Object(jobject object);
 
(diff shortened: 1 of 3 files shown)


Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
```diff
@@ -76,16 +76,33 @@ namespace java::lang
...
+        Class Class();
 
     protected:
-        Object(const char* className, jobject object);
+        Object(const char* className);
+        Object(jobject object);
 
         JNIEnv* m_env;
-        const jclass m_class;
+        const class Class m_class;
```
ryantrem, line 105: I don't think you need/want the `const class` part of this. I think it can just be `Class m_class`. @bghgary?

Dependencies/UrlLib/Source/Android/UrlRequest.cpp
```diff
@@ -98,7 +98,15 @@ namespace UrlLib
                         URLConnection connection{url.OpenConnection()};
                         connection.Connect();
 
-                        m_statusCode = static_cast<UrlStatusCode>(((HttpURLConnection)connection).GetResponseCode());
+                        if (connection.Class().IsAssignableFrom({"java/net/HttpURLConnection"}))
```
ryantrem, line 101: I didn't really think of this when we talked before, but it's not great to have to use this JNI string in `URLRequest` since the purpose of JavaWrappers is to hide these JNI details. I'm not sure how to make this cleaner without even more changes though, like perhaps having `Object` be constructed from a `Class` rather than a `const char*`, and then having pre-defined `Class` instances for classes we care about. Some thing like:

```c++
Class HttpURLConnectionClass{"java/net/HttpURLConnection"};
...
Object::Object(Class _class) { ... } // instead of taking a const char*
...
HttpUrlConnection::HttpUrlConnection() : Object(HttpURLConnectionClass) // hypothetical to illustrate a potential pattern (since we don't create new HttpURLConnections right now)
...
connection.Class().IsAssignableFrom(HttpURLConnectionClass)
```

@bghgary any thoughts?

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
bghgary: const seems okay to me. The `class` part is extraneous though.

```suggestion
        const Class m_class;
```

Dependencies/UrlLib/Source/Android/UrlRequest.cpp
bghgary: This would be nice, but maybe as a separate change?

bghgary (approved):

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
CoPrez: The class identifier was required in this case because I was also adding a method to the Object wrapper named `Class()`. I've renamed the method to GetClass to more accurately match the Java syntax which also means I can get rid of the class identifier here.

Dependencies/UrlLib/Source/Android/UrlRequest.cpp
CoPrez: Talked with @ryantrem about this one offline. I'm not going to make that change as part of this PR. To make things a little cleaner and a step towards that direction I'm adding a static method to HttpURLConnection to get the class so that this code no longer needs to know about the underlying java class name.

ryantrem (approved):

PR #495, opened 2020-11-10
Update README.md

Drigax:
update current project capabilities

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,12 @@ Then targeting a Ninja make file:
 cmake -GNinja -DJSCORE_LIBRARY=/usr/lib/x86_64-linux-gnu/libjavascriptcoregtk-4.0.so ..
 ```
 
-Ninja is not mandatory and make can be used instead.
+Ninja is not mandatory and make can be used instead:
+
+```
+cmake -DJSCORE_LIBRARY=/usr/lib/x86_64-linux-gnu/libjavascriptcoregtk-4.0.so ..
+```
+
 And finaly, run a build:
 
 ```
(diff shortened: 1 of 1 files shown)


Drigax:
We should also update this to ensure the Linux build instructions are accurate.

Drigax:
Linux instructions seem accurate from what I can tell, I was able to get started from fresh install by following the current guide. I'm not sure what we may be missing...

README.md
```diff
@@ -37,15 +37,13 @@ The following major features are partially implemented but not yet supported. No
...
-- Building and running the libraries and demo apps for the following platforms: 
-Android.
 - Mixed reality powered by OpenXR on the following platforms: Win32, UWP.
+- User Input.
 
 The following major features are not yet supported or implemented, even as previews, 
 but are expected to be supported in the future. Note that this list is not exhaustive.
 
 - Debugging JavaScript with V8.
-- User input.
```
bghgary, line 48: We still don't have this yet. Don't remove it.

README.md
```diff
@@ -350,8 +348,13 @@ Then targeting a Ninja make file:
 cmake -GNinja -DJSCORE_LIBRARY=/usr/lib/x86_64-linux-gnu/libjavascriptcoregtk-4.0.so ..
 ```
 
-Ninja is not mandatory and make can be used instead.
-And finaly, run a build:
+Ninja is not mandatory and make can be used instead:
```
bghgary, line 351: If `ninja-build` is a dependency, we should just tell people to use ninja to build. Having a fork at each step seems confusing to me.

Drigax: makes sense, I'll just remove the additional instruction.

bghgary (approved):

PR #497, opened 2020-11-12
Enable gamepad input for the Windows MR controller

satyapoojasama:
The following are introduced/updated in this PR:

- Introduced gamepad object to the native input source which contains trigger, squeeze, thumbstick and trackpad  
- created bindings corresponding to Windows MR controller and update the values for every frame
- made necessary changes in nativexr.cpp to make the gamepad values available in js layer

Now, with these changes, controller are its inputs are identified in a native scene. Tested the changes with a custom scene for all the controller inputs.


Dependencies/xr/Include/XR.h
```diff
@@ -146,6 +146,24 @@ namespace xr
...
+                    bool Pressed{ false };
+                    bool Touched{ false };
+                    double Value{0};
+                };
+
+                struct GamePad
+                {
+                    using Identifier = size_t;
+                    const Identifier ID{ NEXT_ID++ };
+                    std::vector<float> axes;
```
chrisfromwork, line 160: >axes [](start = 39, length = 4)

nit: Axes and Buttons should be capitalized based on other struct definitions preference in capitalizing public fields

Dependencies/xr/Include/XR.h
```diff
@@ -146,6 +146,24 @@ namespace xr
...
+                    std::vector<Button> buttons; 
+
+                    private:
+                    static inline Identifier NEXT_ID{0};
```
chrisfromwork, line 164: nit: this has an extra tab of spaces compared to other struct

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -637,12 +678,153 @@ namespace xr
...
             }
 
-            // Provide suggested bindings to instance
-            XrInteractionProfileSuggestedBinding suggestedBindings{ XR_TYPE_INTERACTION_PROFILE_SUGGESTED_BINDING };
-            XrCheck(xrStringToPath(instance, ActionResources.DEFAULT_XR_INTERACTION_PROFILE, &suggestedBindings.interactionProfile));
-            suggestedBindings.suggestedBindings = bindings.data();
-            suggestedBindings.countSuggestedBindings = (uint32_t)bindings.size();
-            XrCheck(xrSuggestInteractionProfileBindings(instance, &suggestedBindings));
+
+            // Create controller get trigger value action and suggested bindings
```
chrisfromwork, line 682: >  [](start = 0, length = 1)

nit: single space above between this line and the end parenthesis

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1168,7 +1350,122 @@ namespace xr
...
+                {
+                    // query input action state
+                    XrActionStateFloat triggerState{XR_TYPE_ACTION_STATE_FLOAT};
+                    XrActionStateGetInfo getInfo{XR_TYPE_ACTION_STATE_GET_INFO};
+                    getInfo.action = actionResources.ControllerGetTriggerValueAction;
+                    XrCheck(xrGetActionStateFloat(session, &getInfo, &triggerState));
+                    auto& inputSource = InputSources[idx];
+                    if (triggerState.changedSinceLastSync)
+                    {
+                        inputSource.GamepadObject.buttons[0].Value = triggerState.currentState;
```
chrisfromwork, line 1379: >       [](start = 97, length = 6)

nit: remove extra spaces

chrisfromwork, line 1379: >0 [](start = 58, length = 1)

nit: can this button index be a defined constexpr?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1168,7 +1350,122 @@ namespace xr
...
+
+                // Get trigger value data
+                {
+                    // query input action state
+                    XrActionStateFloat triggerState{XR_TYPE_ACTION_STATE_FLOAT};
+                    XrActionStateGetInfo getInfo{XR_TYPE_ACTION_STATE_GET_INFO};
+                    getInfo.action = actionResources.ControllerGetTriggerValueAction;
+                    XrCheck(xrGetActionStateFloat(session, &getInfo, &triggerState));
+                    auto& inputSource = InputSources[idx];
+                    if (triggerState.changedSinceLastSync)
```
chrisfromwork, line 1377: nit: it seems like throughout the inputSource could be declared inside the changedSinceLastSync checks

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1828,6 +1866,12 @@ namespace Babylon
             static constexpr auto JS_CLASS_NAME = "XRSession";
             static constexpr auto JS_EVENT_NAME_END = "end";
             static constexpr auto JS_EVENT_NAME_INPUT_SOURCES_CHANGE = "inputsourceschange";
+            static constexpr auto JS_EVENT_NAME_INPUT_SQUEEZE_START = "squeezestart";
```
chrisfromwork, line 1869: >JS_EVENT_NAME_INPUT_SQUEEZE_START [](start = 34, length = 33)

do we ever fire these new events?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2003,19 +2048,32 @@ namespace Babylon
 
                     current.insert(inputSource.ID);
 
-                    auto found = m_idToInputSource.find(inputSource.ID);
-                    if (found == m_idToInputSource.end())
+                    auto inputSourceFound = m_idToInputSource.find(inputSource.ID);
+                    if (inputSourceFound == m_idToInputSource.end())
                     {
                         // Create the new input source, which will have the correct spaces associated with it.
```
chrisfromwork, line 2054: @satyapoojasama @rickfromwork does this mean we are always populating both hand information and game pad information for all OpenXR UWP form factors? What does the input source declare for vr headsets in terms of hands currently? What does hololens declare in terms of gamepad?

IIRC, hololens shouldn't support a gamepad. I don't think vr should support hands. But, right now it looks like everything is always populated. What are folks thoughts here?

rickfromwork: @chrisfromwork why shouldn't HoloLens support gamepad? It's possible to connect a wireless one via bluetooth and use it in other apps and the shell.

tymlipari: Thinking longer-term, since this is technically OpenXR-based, it's conceivable to me that we'll end up in situations where we'll see novel combinations of input modalities and immersive vs. augmented rendering. Think Quest + Hands and HoloLens + controller (which, HoloLens 1 did ship with the clicker).

Ideally, we'd rely on the OpenXR runtime to tell us what is and isn't present and just expose the capabilities of that runtime.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -637,12 +678,153 @@ namespace xr
...
+                actionInfo.subactionPaths = ActionResources.ControllerSubactionPaths.data();
+                XrCheck(xrCreateAction(ActionResources.ActionSet, &actionInfo, &ActionResources.ControllerGetTriggerValueAction));
+                // For each controller subaction
+                for (size_t idx = 0; idx < ActionResources.CONTROLLER_SUBACTION_PATH_PREFIXES.size(); ++idx)
+                {
+                    // Create suggested binding
+                    std::string path{ ActionResources.CONTROLLER_SUBACTION_PATH_PREFIXES[idx] };
+                    path.append(ActionResources.CONTROLLER_GET_TRIGGER_VALUE_PATH_SUFFIX);
+                    bindings.push_back({ ActionResources.ControllerGetTriggerValueAction });
+                    XrCheck(xrStringToPath(instance, path.data(), &bindings.back().binding));
```
tymlipari, line 698: What happens if a controller input profile doesn't support this path? I'm assuming the `XrCheck` throws and we're left in a bad state?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1168,7 +1350,122 @@ namespace xr
...
                 }
-                
+
+                {
+                    // Initialize the gamepad object components
+                    auto& inputSource = InputSources[idx];
+                    auto controllerInfo = sessionImpl.ControllerInfo;
+                    if (inputSource.GamepadObject.buttons.size() != controllerInfo.CONTROLLER_BUTTONS)
+                    {
+                        inputSource.GamepadObject.buttons = std::vector<Button>(controllerInfo.CONTROLLER_BUTTONS);
```
tymlipari, line 1360: Just use the `.clear()` member function on `std::vector` rather than assigning an empty one. You'll avoid new allocations that way.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1527,10 +1564,10 @@ namespace Babylon
                     JS_CLASS_NAME,
                     {
                         InstanceMethod("getViewerPose", &XRFrame::GetViewerPose),
-                        InstanceMethod("getPose", &XRFrame::GetPose),
-                        InstanceMethod("getHitTestResults", &XRFrame::GetHitTestResults),
-                        InstanceMethod("createAnchor", &XRFrame::CreateAnchor),
-                        InstanceMethod("getJointPose", &XRFrame::GetJointPose),
+                            InstanceMethod("getPose", &XRFrame::GetPose),
```
tymlipari, line 1567: nit: Whitespace issues

Plugins/NativeXr/Source/NativeXr.cpp
satyapoojasama: good catch, they are to be removed, i forgot. Thanks!

Dependencies/xr/Source/OpenXR/XR.cpp
satyapoojasama: i thought all the Windows MR controllers have a trigger button. Also, one more question i have is - do we want to support other MR controllers? because right now, we are only supporting windows MR controllers

Dependencies/xr/Source/OpenXR/XR.cpp
satyapoojasama: I'm curious on what difference that would make?

Dependencies/xr/Source/OpenXR/XR.cpp
tymlipari: I think it's okay for us to only currently add in the WMR controllers, but since this is the OpenXR module it should be able to eventually support non-WMR controllers as well.

Dependencies/xr/Include/XR.h
```diff
@@ -146,6 +146,24 @@ namespace xr
                     bool PoseTracked{ false };
                 };
 
+                struct Button
```
rickfromwork, line 149: Does it make sense to have Button be a part of Frame as it is here? It could probably be embedded within the Gamepad struct, or alongside Ray and Anchor outside of System. Since only Gamepad uses it, I'd lean towards including it inside Gamepad.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -256,13 +256,62 @@ namespace xr
...
+            static constexpr char* MICROSOFT_XR_INTERACTION_PROFILE{ "/interaction_profiles/microsoft/motion_controller" };
 
             std::vector<Frame::InputSource> ActiveInputSources{};
             std::vector<Frame::Plane> Planes{};
             std::vector<FeaturePoint> FeaturePointCloud{};
         } ActionResources{};
 
+        struct
+        {
+            static constexpr uint32_t CONTROLLER_BUTTONS{ 4 };
```
rickfromwork, line 303: nit: personally I think it looks cleaner if related lines are grouped, i.e.:
```
static constexpr uint32_t CONTROLLER_BUTTONS{ 4 };
static constexpr uint32_t TRIGGER_BUTTON = 0;
static constexpr uint32_t SQUEEZE_BUTTON = 1;
static constexpr uint32_t TRACKPAD_BUTTON = 2;
static constexpr uint32_t THUMBSTICK_BUTTON = 3;

static constexpr uint32_t CONTROLLER_AXES{ 4 };
static constexpr uint32_t TRACKPAD_X_AXIS = 0;
static constexpr uint32_t TRACKPAD_Y_AXIS = 1;
static constexpr uint32_t THUMBSTICK_X_AXIS = 2;
static constexpr uint32_t THUMBSTICK_Y_AXIS = 3;
```

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I think its mostly to keep related code in one place. So since inputSource and controllerInfo aren't used outside of the conditional, moving their declaration inside the conditional would make it visually clearer where they are being used.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -184,25 +184,62 @@ namespace
...
+    }
+
     Napi::ObjectReference CreateXRInputSource(xr::System::Session::Frame::InputSource& inputSource, Napi::Env& env)
     {
         constexpr std::array<const char*, 2> HANDEDNESS_STRINGS{
             "left",
             "right"};
         constexpr const char* TARGET_RAY_MODE{"tracked-pointer"};
-
         auto jsInputSource = Napi::Object::New(env);
```
rickfromwork, line 222: nit: was this whitespace shift intentional?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1597,6 +1634,7 @@ namespace Babylon
         private:
             const xr::System::Session::Frame* m_frame{};
             Napi::ObjectReference m_jsXRViewerPose{};
+            Napi::ObjectReference m_jsTriggerValue{};
```
rickfromwork, line 1637: Where is this used?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2003,19 +2042,32 @@ namespace Babylon
...
-                    auto found = m_idToInputSource.find(inputSource.ID);
-                    if (found == m_idToInputSource.end())
+                    auto inputSourceFound = m_idToInputSource.find(inputSource.ID);
+                    if (inputSourceFound == m_idToInputSource.end())
                     {
                         // Create the new input source, which will have the correct spaces associated with it.
                         m_idToInputSource.insert({inputSource.ID, CreateXRInputSource(inputSource, env)});
+                        
+                        //Now that input Source is created, create a gamepad object for the input source
+                        inputSourceFound = m_idToInputSource.find(inputSource.ID);
```
rickfromwork, line 2052: Since this adds entries alongside m_idToInputSource, it should also remove entries when m_idToInputSource removes them (line 2120).

Dependencies/xr/Include/XR.h
satyapoojasama: fair enough! i will move it

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: Side note on my comment - I keep forgetting gamepad here means a 6dof controller. I'm used to gamepad meaning xbox controller. I don't think we currently support 6dof on HoloLens (though I do remember talks in the past about wanting to).

Plugins/NativeXr/Source/NativeXr.cpp
satyapoojasama: nope it is not, i can reverse that

Plugins/NativeXr/Source/NativeXr.cpp
satyapoojasama: good catch, its not.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -637,12 +686,152 @@ namespace xr
...
+                    XrCheck(xrStringToPath(instance, path.data(), &bindings.back().binding));
+                }
+            }
+
+            // Provide default Microsoft suggested binding to instance
+            XrInteractionProfileSuggestedBinding microsoftSuggestedBindings{ XR_TYPE_INTERACTION_PROFILE_SUGGESTED_BINDING };
+            XrCheck(xrStringToPath(instance, ActionResources.MICROSOFT_XR_INTERACTION_PROFILE, &microsoftSuggestedBindings.interactionProfile));
+            microsoftSuggestedBindings.suggestedBindings = bindings.data();
+            microsoftSuggestedBindings.countSuggestedBindings = (uint32_t)bindings.size();
+            XrCheck(xrSuggestInteractionProfileBindings(instance, &microsoftSuggestedBindings));
```
syntheticmagus, line 804: As above, unless we're consciously in a place that is designated (and named) as Microsoft-specific, we should make sure that we're not doing things that will clash with the broader vision of OpenXR. Especially with Oculus and Valve now getting OpenXR runtimes into the public, we should anticipate people wanting to add support for those in the near future and try to plan ways for the implementations to share what can be shared and clearly designate what must be separated.

Please correct me if I'm misreading, but it looks like parts of this PR are _replacing_ the KHR simple controller as default with a Microsoft-specific alternative. If that's what's intended, I recommend reconsidering. The KHR simple controller exists as a default partly because it encapsulates a minimum functionality set that many different vendor implementations can map to. I think it would be better for Microsoft's more custom and feature-rich implementation to live alongside the simplistic default as opposed to replacing it. This will both leave the simplistic default intact and help establish the patterns that will eventually allow multiple feature-rich controller implementations (Oculus, Vive, Index, etc.) to coexist in the code alongside what we're currently actively implementing.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -184,25 +184,62 @@ namespace
         }
     }
 
+    void SetXRGamepadObjectData(Napi::Object& jsInputSource, Napi::Object& jsGamepadObject, xr::System::Session::Frame::InputSource& inputSource)    
+    {
+        auto env = jsInputSource.Env();
+        //Set Gamepad Object
+
+        auto gamepadButtons = Napi::Array::New(env, inputSource.GamepadObject.Buttons.size());
```
syntheticmagus, line 192: nit: Just as a general rule, if this is being done per-frame, it might be better to create a "scratch" object that is used repeatedly because JavaScript allocations tend to be _much_ more computationally expensive than seems entirely reasonable. @bghgary can probably comment more informedly on this, but I believe this holds true when allocating objects through N-API. I know that NativeXr.cpp currently does a lot of per-frame JS allocation, so it might be preferable to just deal with all of those at once in a perf pass rather than trying to keep up with them here, so I don't think this needs to be a significant concern right now. It's just something to be aware of.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -184,25 +184,62 @@ namespace
...
-
         auto jsInputSource = Napi::Object::New(env);
+
         jsInputSource.Set("handedness", Napi::String::New(env, HANDEDNESS_STRINGS[static_cast<size_t>(inputSource.Handedness)]));
         jsInputSource.Set("targetRayMode", TARGET_RAY_MODE);
         SetXRInputSourceData(jsInputSource, inputSource);
 
         auto profiles = Napi::Array::New(env, 1);
-        Napi::Value string = Napi::String::New(env, "generic-trigger-squeeze-touchpad-thumbstick");
+        Napi::Value string = Napi::String::New(env, "windows-mixed-reality");
```
syntheticmagus, line 230: My concern here is in line with some of my other comments, but it's more emphatic here: this cannot be in this file. The `generic-trigger-squeeze-touchpad-thumbstick` string was already probably not as it should be, but this file definitely cannot be making decisions that target a specific platform. I think there might already be a way to provide this string, or something very much like it, to Babylon.js's XR abstraction; if you could leverage that to designate the preferred controller type and just use that string here, that would probably be ideal. If that's too much of a change to be a part of this PR and it's more convenient to have this hard-coded in the C++ for now, it's probably alright to have this in a header and use one of our CMake platform tricks to have it only take this new value on Windows (and have it be either generic or undefined on other platforms). I don't think it can go here, though. NativeXr.cpp exists to bridge the gap between Babylon.js and XR.h; it _cannot_ know about Windows.

Dependencies/xr/Source/OpenXR/XR.cpp
satyapoojasama: thanks for bringing this up Justin, initially I thought there will be a runtime way of determining what controller is being used in openXR  but did not find any specific details of that sort. Later, after chatting with Tym, I understood that we would like to support all windows MR controllers as a first step. But this makes sense, I can dig deeper to see if it is possible to support a fall back option incase something is not a windows controller. Also, I would like to get @tymlipari 's perspective on this.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -637,12 +712,96 @@ namespace xr
...
             }
 
-            // Provide suggested bindings to instance
-            XrInteractionProfileSuggestedBinding suggestedBindings{ XR_TYPE_INTERACTION_PROFILE_SUGGESTED_BINDING };
-            XrCheck(xrStringToPath(instance, ActionResources.DEFAULT_XR_INTERACTION_PROFILE, &suggestedBindings.interactionProfile));
-            suggestedBindings.suggestedBindings = bindings.data();
-            suggestedBindings.countSuggestedBindings = (uint32_t)bindings.size();
-            XrCheck(xrSuggestInteractionProfileBindings(instance, &suggestedBindings));
+            // Create controller get trigger value action and suggested bindings
+            {
```
rgerd, line 716: nit: Don't need these scope braces anymore

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1168,7 +1376,71 @@ namespace xr
...
-                
+
+                // Get gamepad data
+                {
+                    auto& inputSource = InputSources[idx];
+                    auto controllerInfo = sessionImpl.ControllerInfo;
+
+                    // Get trigger value data
+                    float currentTriggerValue;
+                    bool triggerChangedSinceLastSync = m_impl->QueryControllerFloatAction(actionResources.ControllerGetTriggerValueAction, session, currentTriggerValue);
```
rgerd, line 1387: uber nit: Could check the return value directly in the `if` instead of assigning the result to a variable (since the out reference will get assigned if the condition evaluates as true), but feel free to ignore this if you'd rather keep it as-is

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -256,13 +256,62 @@ namespace xr
             XrAction ControllerGetAimPoseAction{};
             std::array<XrSpace, CONTROLLER_SUBACTION_PATH_PREFIXES.size()> ControllerAimPoseSpaces{};
 
-            static constexpr char* DEFAULT_XR_INTERACTION_PROFILE{ "/interaction_profiles/khr/simple_controller" };
```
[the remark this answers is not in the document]
satyapoojasama: I added back the khr controller as a fallback option

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2066,6 +2127,15 @@ namespace Babylon
...
                         m_idToInputSource.erase(id);
+
+                        // remove the corresponding gampad if it exists
+                        for (const auto gamepadId : enabledGamepads)
+                        {
+                            if (id == gamepadId)
+                            {
+                                m_idToGamepadObject.erase(id);
+                            }
+                        }
```
syntheticmagus, line 2138: This block of logic is a bit peculiar. Without diving into the code very deeply, I'm not 100% sure what I think this should be, but this seems like a very odd way to remove an entry from a map. If the intention is just to remove the ID from `m_idToGamepadObject` if it exists, I don't think you need to check; [`std::map::erase`](https://en.cppreference.com/w/cpp/container/map/erase) can legally be given nonexistent keys to erase, in which case it will just do nothing and return 0. In that case, the only line you'd need of this block is the line that erases.

It's possible I'm missing a nuance, though. The presence of `enabledGamepads` here confuses me; is it indicating that we _only_ want to remove gamepads in this way if they were added this frame? How would they get removed if they were added on a different frame?

syntheticmagus (commented):
This should get approval from people who are more closely involved with this code on a day-to-day basis, but other than a few minor considerations, nothing blocking jumps out at me. Nice PR! 👍

Plugins/NativeXr/Source/NativeXr.cpp
satyapoojasama: you are right Justin, by the above logic i was trying to prevent removing a non-existent id. I will remove this then. Thank you!

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1168,7 +1380,67 @@ namespace xr
                         inputSource.AimSpace.Pose.Orientation.W = location.pose.orientation.w;
                     }
                 }
-                
+
+                // Get gamepad data 
+                auto controllerInfo = sessionImpl.ControllerInfo;
```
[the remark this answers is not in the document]
satyapoojasama: which "const auto&" are you talking about Justin?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2003,19 +2044,39 @@ namespace Babylon
...
-                    auto found = m_idToInputSource.find(inputSource.ID);
-                    if (found == m_idToInputSource.end())
+                    auto inputSourceFound = m_idToInputSource.find(inputSource.ID);
+                    if (inputSourceFound == m_idToInputSource.end())
                     {
                         // Create the new input source, which will have the correct spaces associated with it.
                         m_idToInputSource.insert({inputSource.ID, CreateXRInputSource(inputSource, env)});
+                        
+                        //Now that input Source is created, create a gamepad object if enabled for the input source
+                        inputSourceFound = m_idToInputSource.find(inputSource.ID);
```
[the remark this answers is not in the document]
satyapoojasama: I agree, the reason i did this is because when I tried to save the output of CreateXRInputSource as a Napi::ObjectReference and use it later for gamepad creation, the m_idToInputSource::insert is not allowing to insert the saved napi::objectReference for some overloading reasons, not sure why.

Dependencies/xr/Source/OpenXR/XR.cpp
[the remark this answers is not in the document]
syntheticmagus:
```
const auto& controllerInfo = sessionImpl.ControllerInfo;
```

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: +1 with Justin's suggestion here. It seems like we grab a reference to the javascript gamepad that we then use throughout the Babylon.js codebase. It seems like theres not an expectation for gamepad buttons/state to be immutable.

That going said, if WebXR declares any of these items as immutable (like immutable sets with plane detection), we may be forced to conform to recreating javascript objects every frame. I just wanted to call out that we may only be able to introduce these improvements case by case.

---
In reply to: [523280642](https://github.com/BabylonJS/BabylonNative/pull/497#discussion_r523280642) [](ancestors = 523280642)

chrisfromwork (approved):
:shipit:

rgerd (approved):

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1168,7 +1380,67 @@ namespace xr
                         inputSource.AimSpace.Pose.Orientation.W = location.pose.orientation.w;
                     }
                 }
-                
+
+                // Get gamepad data 
+                const auto& controllerInfo = sessionImpl.ControllerInfo;
+                if (!controllerInfo.DefaultControllerBinding)
```
rickfromwork, line 1386: nit (naming) - it makes more sense to me if you're checking to see if controllerInfo is using the MR controller binding, compared to checking if we aren't using the default bindings (since the code inside the conditional is specific to the MR bindings).

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: nit: there's still new whitespace

Dependencies/xr/Include/XR.h
```diff
@@ -185,6 +200,8 @@ namespace xr
                     const Identifier ID{ NEXT_ID++ };
                     bool TrackedThisFrame{};
                     bool JointsTrackedThisFrame{};
+                    bool GamepadEnabledThisFrame{};
```
rickfromwork, line 203: Would it make more sense to call this GamepadTrackedThisFrame, to match the others above it?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2003,19 +2042,38 @@ namespace Babylon
...
+                        if (inputSource.GamepadEnabledThisFrame)
+                        {
+                            auto inputSourceVal = inputSourceFound->second.Value();
+                            m_idToGamepadObject.insert({inputSource.ID, CreateXRGamepadObject(inputSourceVal, inputSource)});
```
rickfromwork, line 2056: After this, isn't the result of CreateXRGamepadObject stored in both `m_idToGamepadObject[inputSource.ID]` *and* `m_idToInputSource[inputSource.ID].gamepad`? Do we need it stored in two different places if it's added/updated/removed from both at the same time?

satyapoojasama: I don't think it is stored in both Rick, we are creating napi object corresponding to the gamepad object in CreateXRGamepadObject. And we do not create that in CreateXRInputSource. So, i dont think we need to worry.

satyapoojasama: you have a point, but I am not sure we can extract member values of a napi object. m_idToInputSource[inputSource.ID] is a napi object reference to input source. Can we extract the gamepadvalue from it?

rickfromwork: It looks like you can call `m_idToInputSource[inputSource.ID].Get("gamepad")` to get the gamepad object. You can either test the returned value with a call to `.IsObject()` to see if it exists, or `m_idToInputSource[inputSource.ID].Has("gamepad")`

rickfromwork: (I'm basing this off of what I see in the constructors for XRRay and XRHitTestSource)

satyapoojasama: I see, i will try and see :)

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2003,19 +2040,34 @@ namespace Babylon
...
                         // Ensure the correct spaces are associated with the existing input source.
-                        auto val = found->second.Value();
-                        SetXRInputSourceData(val, inputSource);
+                        auto inputSourceVal = inputSourceFound->second.Value();
+                        SetXRInputSourceData(inputSourceVal, inputSource);
+
+                        //inputSource already exists, find the corresponding gamepad object if enabled and set to correct values
+                        if (inputSourceVal.Has("gamepad"))
+                        {
+                            auto gamepadObject = inputSourceVal.Get("gamepad").As<Napi::Object>();
```
rickfromwork, line 2068: If you wanted to, you could now move this line inside of SetXRGamepadObject, or heck even put this whole if block inside of SetXRInputSourceData

satyapoojasama: So it seems like I can do that from the surface but its not possible. SetXRGamepadObject is also called from CreateXRGamepadObject before the gamepad is even set. So, here is where the gamepad property is set to the inputsource.  SetXRGamepadObject has to be called every other time with already created gamepad except for the first time  . If I move the gamepad extraction into the SetXRGamepadObject, it won't work the first time.

rickfromwork: Oh, I missed that part.

rickfromwork (approved):

PR #498, opened 2020-11-13
Small fixes in JavaWrappers

ryantrem:
- Release char buffers after we are done with them (e.g. copying them into an std::string).
- Don't return a pointer to a stack allocated char buffer from Throwable::what (instead store the message as an std::string in the Throwable).

### DIFF
--- a/Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
+++ b/Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
@@ -108,7 +108,10 @@ namespace java::lang
 
     String::operator std::string() const
     {
-        return m_env->GetStringUTFChars(m_string, nullptr);
+        const char* buffer{m_env->GetStringUTFChars(m_string, nullptr)};
+        std::string str{buffer};
+        m_env->ReleaseStringUTFChars(m_string, buffer);
+        return str;
     }
 
     Throwable::Throwable(jthrowable throwable)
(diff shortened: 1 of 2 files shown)


Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -108,12 +108,16 @@ namespace java::lang
 
     String::operator std::string() const
     {
-        return m_env->GetStringUTFChars(m_string, nullptr);
+        const char* buffer{m_env->GetStringUTFChars(m_string, nullptr)};
```
bghgary, line 111: the `nullptr` is for isCopy right? Didn't we say we need to get this result to know if we need to release or not?

ryantrem: Rereading the docs, I don't think so. They say this:

> Returns a pointer to an array of bytes representing the string in modified UTF-8 encoding. **This array is valid until it is released by ReleaseStringUTFChars().**

I think this means you must release the buffer regardless, and this StackOverflow suggests the same: https://stackoverflow.com/questions/5859673/should-you-call-releasestringutfchars-if-getstringutfchars-returned-a-copy

I think the JVM docs also agree with this, though they are confusing. For example, on one page it says:

> You must call Release<Type> once for every Get<Type> call, regardless of the value of the isCopy parameter. This step is necessary because calling Release<Type> deletes JNI local references that might otherwise prevent garbage collection.

This seems pretty clear. But then on another page it says:

> Ensure that you free copies made by a Get<Type> function by calling the corresponding Release<Type> function when the copy is no longer needed.

Which could suggest that a Release* call is exclusively needed for copies, or could just mean that a Release* call is inclusively needed for copies.

https://www.ibm.com/support/knowledgecenter/SSYKE2_7.0.0/com.ibm.java.win.70.doc/diag/understanding/jni_mode_flag.html

bghgary: hmm, okay.

bghgary (approved):

PR #500, opened 2020-11-17
AndroidStudio+Linux

CedricGuillemet:
- fix uppercase for V8 sources
- add build def for AndroidStudio on Linux


Dependencies/napi/napi-direct/source/env_V8.cc
```diff
@@ -1,6 +1,6 @@
 #include <napi/env.h>
 #include <napi/js_native_api_types.h>
-#include "js_native_api_v8.h"
+#include "js_native_api_V8.h"
```
bghgary, line 3: I don't know what happened here, but the file in the source control should be lowercase. They are in the node.js repo, so I'm not sure how they became uppercase in the source control.

bghgary (changes requested):

Dependencies/napi/napi-direct/source/env_V8.cc
CedricGuillemet: I did change the case for the filenames because NAPI_JAVASCRIPT_ENGINE is V8 and not v8. I can change the cmake files to replace by v8 is you prefer.

bghgary: Oh sorry. I missed the fact that you updated the file name casing in this change. I would recommend updating cmake to look for lowercase names and change the JavaScriptCore files to be lowercase in source control.

Apps/Playground/Android/app/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
 if (${ANDROID_JSENGINE_LIBNAME} STREQUAL "jsc")
     set(NAPI_JAVASCRIPT_ENGINE "JavaScriptCore" CACHE STRING "JavaScript engine for N-API.")
 elseif (${ANDROID_JSENGINE_LIBNAME} STREQUAL "v8android")
-    set(NAPI_JAVASCRIPT_ENGINE "V8" CACHE STRING "JavaScript engine for N-API.")
+    set(NAPI_JAVASCRIPT_ENGINE "v8" CACHE STRING "JavaScript engine for N-API.")
```
bghgary, line 30: Sorry, this isn't what I mean. We should continue to use V8 as the input, but convert it to lowercase when accessing the file.

Dependencies/napi/napi-direct/CMakeLists.txt
```diff
@@ -1,14 +1,20 @@
+if(NAPI_JAVASCRIPT_ENGINE STREQUAL "V8")
+    set(ENGINE_FILE_POSTFIX "v8")
+else()
+    set(ENGINE_FILE_POSTFIX ${NAPI_JAVASCRIPT_ENGINE})
+endif()
+
```
bghgary, line 6:
```suggestion
string(TOLOWER NAPI_JAVASCRIPT_ENGINE ENGINE_FILE_POSTFIX)
```

Then update the local file for JSC to lower case.

bghgary (approved):

Issue #375, opened 2020-08-21
OpenXR backend session state is not handled correctly


chrisfromwork:
Repro:
Open the start menu on a HoloLens 2 while running a BabylonNative application that contains a lot of meshes.

Issue:
OpenXR backend does not correctly tear down rendering when transitioning to an idle OpenXr state. When application perf causes OpenXr to transition to a XR_SESSION_STATE_STOPPING state, BabylonNative ends the OpenXR session. This does not currently ensure that there won't be additional attempts to submit frames, which can cause OpenXr exceptions.

PR #501, opened 2020-11-18
Add a check to stop frame submissions when OpenXR session state is idle

chrisfromwork:
This change is related to https://github.com/BabylonJS/BabylonNative/issues/375

OpenXR does not support wait/begin/end frame calls when in an idle state. This change updates frame logic to be a no-op when in an OpenXR idle state.

This change fixes scenarios where the start menu is opened/closed on HoloLens. Without this change, opening and closing the start menu on HoloLens crashes the rendering thread.


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -910,6 +909,12 @@ namespace xr
             depthInfoView.subImage.imageRect = imageRect;
             depthInfoView.subImage.imageArrayIndex = 0;
         }
+
+        bool IsFrameSubmissionSupported()
+        {
+            const auto& sessionState = sessionImpl.HmdImpl.Context.State();
+            return sessionState != XrSessionState::XR_SESSION_STATE_IDLE;
```
tymlipari, line 916: Should this be `return sessionState > XrSessionState::XR_SESSION_STATE_IDLE && sessionState < XrSessionState::XR_SESSION_STATE_STOPPING;`?

chrisfromwork: Chatted with OpenXR owners, sounds like they'd actually suggest creating a flag based on whether we have called xrBeginSession/xrEndSession. I'll update this now

satyapoojasama (approved):

Issue #473, opened 2020-10-29
Viewer rotation changes are overcompensated in XR mode


Alex-MSFT:
When in XR mode rotating the device causes an overcompensation in the Babylon XR Camera orientation.  Here are 3 pictures that show the orientation of an object shifting based on rotations along the X-Axis.

My first guess is we are overcompensating for the X rotation when applying the transformation from the view matrix of the device to the camera in NativeXR.  I see the same effect on both iOS and Android.

Here is a link to a sample playground app on android in the browser which does not see the same drift.
https://playground.babylonjs.com/#IFECJ1#3

Segment as measured:
![1603993145737](https://user-images.githubusercontent.com/27031140/97612045-25a0fa00-19d4-11eb-9593-0474e6b2b7f2.jpg)

Segment after rotating phone towards the floor:
![1603993152324_crop](https://user-images.githubusercontent.com/27031140/97612622-e2935680-19d4-11eb-9c96-e1471516baca.jpg)

Segment after rotating phone towards the wall:
![1603993155537](https://user-images.githubusercontent.com/27031140/97612629-e58e4700-19d4-11eb-9d82-0a068b527fdc.jpg)

PR #503, opened 2020-11-18
Push XR projection matrix generation down to platform implementations

Alex-MSFT:
This PR attempts to address #473, which effectively expanded the projection matrix to maintain the aspect ratio, but incorrectly calculated the real FOV.  I think this was occurring because we were applying both the angle FOV values with the aspect ratio from the color texture.

In this PR I'm pushing the generation of projection matrices down to the platform level implementations.  For ARCore and ARKit we are provided with valid projection matrices.  For OpenXR we need to generate them ourselves based off the FoV angles.

Here are some comparisons from my testing attempting to measure a 5' long measuring tape.  In both cases I measured by getting close to the tape and walking along its length then stepping back to a reasonable distance to take a picture.

old:

new:

### DIFF
--- a/Dependencies/xr/Source/ARCore/XR.cpp
+++ b/Dependencies/xr/Source/ARCore/XR.cpp
@@ -527,15 +527,7 @@ namespace xr
                 // Get the current projection matrix
                 glm::mat4 projectionMatrix{};
                 ArCamera_getProjectionMatrix(session, camera, DepthNearZ, DepthFarZ, glm::value_ptr(projectionMatrix));
-
-                // Copy the projection matrix values into the view.
-                for(int row = 0; row < 4; row++)
-                {
-                    for(int column = 0; column < 4; column++)
-                    {
-                        ActiveFrameViews[0].ProjectionMatrix[row * 4 + column] = projectionMatrix[row][column];
-                    }
-                }
+                memcpy(ActiveFrameViews[0].ProjectionMatrix.data(), &projectionMatrix, sizeof(float) * 16);
             }
 
             ActiveFrameViews[0].DepthNearZ = DepthNearZ;
--- a/Dependencies/xr/Source/ARCore/XR.cpp
+++ b/Dependencies/xr/Source/ARCore/XR.cpp
@@ -526,18 +526,16 @@ namespace xr
             {
                 // Get the current projection matrix
                 glm::mat4 projectionMatrix{};
-                ArCamera_getProjectionMatrix(session, camera, ActiveFrameViews[0].DepthNearZ, ActiveFrameViews[0].DepthFarZ, glm::value_ptr(projectionMatrix));
+                ArCamera_getProjectionMatrix(session, camera, DepthNearZ, DepthFarZ, glm::value_ptr(projectionMatrix));
 
-                // Calculate the aspect ratio and field of view
-                float a = projectionMatrix[0][0];
-                float b = projectionMatrix[1][1];
-
-                float aspectRatio = b / a;
-                float fieldOfView = std::atan(1.0f / b);
-
-                // Set the horizontal and vertical field of view
-                ActiveFrameViews[0].FieldOfView.AngleDown = -(ActiveFrameViews[0].FieldOfView.AngleUp = fieldOfView);
-                ActiveFrameViews[0].FieldOfView.AngleLeft = -(ActiveFrameViews[0].FieldOfView.AngleRight = fieldOfView * aspectRatio);
+                // Copy the projection matrix values into the view.
+                for(int row = 0; row < 4; row++)
+                {
+                    for(int column = 0; column < 4; column++)
+                    {
+                        ActiveFrameViews[0].ProjectionMatrix[row * 4 + column] = projectionMatrix[row][column];
+                    }
+                }
             }
 
             ActiveFrameViews[0].DepthNearZ = DepthNearZ;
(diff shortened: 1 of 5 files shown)


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -46,30 +46,19 @@ namespace
...
-        const float aspectRatio = static_cast<float>(view.ColorTextureSize.Width) / static_cast<float>(view.ColorTextureSize.Height);
-        const float deltax = (r - l);
-        const float deltay = (t - b);
-        const float afterTangentAspectRatio = deltax / deltay;
-        const float compensationRatio = afterTangentAspectRatio / aspectRatio;
-        const float tc{std::tanf(view.FieldOfView.AngleUp * compensationRatio) * n};
-        const float bc{std::tanf(view.FieldOfView.AngleDown * compensationRatio) * n};
+        // Calculate total vertical and horizontal FoV and the aspect ratio.
+        const float verticalFoV{abs(view.FieldOfView.AngleUp) + abs(view.FieldOfView.AngleDown)};
+        const float horizontalFoV{abs(view.FieldOfView.AngleRight) + abs(view.FieldOfView.AngleLeft)};
```
Alex-MSFT, line 51: This solution assumes symmetric FOVs.  Do we have cases in our use cases on OpenXR with asymmetric FOVs?

@syntheticmagus @rgerd @chrisfromwork

chrisfromwork: I'm not sure that we can assume FOVs are symmetrical. I can test this out on a HoloLens tomorrow

chrisfromwork: Was able to get to this today. HoloLens FOV's aren't symmetrical. we should sync on this offline tomorrow.

Alex-MSFT: Sounds good for now I am going to fork the behavior and keep the old implementation for asymmetric FoVs.  We should investigate if whether the old calculation was correct for these cases.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -74,6 +93,22 @@ namespace
         return bxResult;
     }
 
+    std::array<float, 16> CreateProjectionMatrix(const xr::System::Session::Frame::View& view)
+    {
+        const bool isCentered{
+            abs(view.FieldOfView.AngleUp + view.FieldOfView.AngleDown) < .01 &&
+            abs(view.FieldOfView.AngleRight + view.FieldOfView.AngleLeft) < .01};
```
chrisfromwork, line 100: nit: can we get 0.01 defined as a constexpr?

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: After a lengthy discussion, we're going to instead push projection matrix creation down to platform specific implementations.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -526,18 +526,16 @@ namespace xr
...
-                float b = projectionMatrix[1][1];
-
-                float aspectRatio = b / a;
-                float fieldOfView = std::atan(1.0f / b);
-
-                // Set the horizontal and vertical field of view
-                ActiveFrameViews[0].FieldOfView.AngleDown = -(ActiveFrameViews[0].FieldOfView.AngleUp = fieldOfView);
-                ActiveFrameViews[0].FieldOfView.AngleLeft = -(ActiveFrameViews[0].FieldOfView.AngleRight = fieldOfView * aspectRatio);
+                // Copy the projection matrix values into the view.
+                for(int row = 0; row < 4; row++)
```
CedricGuillemet, line 532: use memcpy?

Alex-MSFT: Good idea, done!

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -526,18 +526,8 @@ namespace xr
...
-                float b = projectionMatrix[1][1];
-
-                float aspectRatio = b / a;
-                float fieldOfView = std::atan(1.0f / b);
-
-                // Set the horizontal and vertical field of view
-                ActiveFrameViews[0].FieldOfView.AngleDown = -(ActiveFrameViews[0].FieldOfView.AngleUp = fieldOfView);
-                ActiveFrameViews[0].FieldOfView.AngleLeft = -(ActiveFrameViews[0].FieldOfView.AngleRight = fieldOfView * aspectRatio);
+                ArCamera_getProjectionMatrix(session, camera, DepthNearZ, DepthFarZ, glm::value_ptr(projectionMatrix));
+                memcpy(ActiveFrameViews[0].ProjectionMatrix.data(), &projectionMatrix, sizeof(float) * 16);
```
ryantrem, line 530: Can't we statically get the size of the whole array? Seems like we shouldn't need to repeat the `16` here.

ryantrem, line 530: If you are doing a `memcpy` doesn't that indicate the memory layout is exactly the same? And if that is the case, can't we just cast `ProjectionMatrix` and pass it directly to `ArCamera_getProjectionMatrix`?

Alex-MSFT: Good point I'll try that

PR #504, opened 2020-11-19
Expose tracking state for ARCore and ARKit

CoPrez:
BabylonJS has an existing paradigm in place to expose the tracking state of the WebXRCamera ([see here](https://github.com/BabylonJS/Babylon.js/blob/f64b0af5d1e97597f76b7dfe5d25bd122d478628/src/XR/webXRCamera.ts#L163)). However the implementation in BabylonNative never plumbed through the emulatedPosition property (and it always returned a defined pose even before tracking started). That resulted in BabylonJS declaring that it always had tracking.

This change addresses that issue by updating the BabylonNative ARCore and ARKit WebXR implementations to follow the same behavior as WebXR itself. Until tracking has been established it will return undefined when calling getViewerPose. Additionally I've added the [XRPose.emulatedPosition](https://developer.mozilla.org/en-US/docs/Web/API/XRPose/emulatedPosition) property to the BabylonNative XRViewerPose implementation so that BabylonJS can detect when tracking has been lost.

This PR does not address exposing tracking states for OpenXR. The OpenXR implementation will maintain it's current behavior where it is assumed that it always has tracking.


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1195,6 +1199,14 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
         return m_impl->sessionImpl.GetPlaneByID(planeID);
     }
 
+    bool System::Session::Frame::IsTracking() const {
+        // There are three different tracking states as defined in ARKit: https://developer.apple.com/documentation/arkit/artrackingstate
+        // From my testing even while obscuring the camera for a long duration the state still registers as ARTrackingStateLimited
+        // rather than ARTrackingStateNotAvailable. For that reason the only state that should be considered to be trully tracking is
+        // ARTrackingStateNormal.
+        return m_impl->sessionImpl.GetTrackingState() == ARTrackingState::ARTrackingStateNormal;
```
ryantrem, line 1207: I would do this inside the impl and call it `IsTracking`. The purpose of the impl is just to hide the private members so that different implementations can have different implementation details. But for the most part, the wrapper should just be a wrapper and not introduce any additional logic when possible.

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1428,6 +1428,13 @@ namespace xr
         return su.TryGetPlaneByID(id);
     }
 
+    bool System::Session::Frame::IsTracking() const
+    {
+        // TODO - Plumb tracking states from OpenXR. For now this will maintain the current behavior where
+        // BabylonJS assumes tracking is always available.
+        return true;
+    }
```
ryantrem, line 1436: @chrisfromwork @rgerd FYI

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1605,15 +1617,31 @@ namespace Babylon
...
+            bool m_hasBegunTracking = false;
 
             Napi::Value GetViewerPose(const Napi::CallbackInfo& info)
             {
+                // To match the WebXR implementation we should return undefined here until we have gotten
+                // initial tracking. After that point we can just continue returning the position
+                // as it will be marked as estimated if tracking is lost.
+                if (!m_hasBegunTracking && !m_frame->IsTracking())
+                {
+                    return info.Env().Undefined();
```
ryantrem, line 1630:
```suggestion
                    return info.Env().Null();
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -873,16 +875,21 @@ namespace Babylon
...
+                for (uint32_t idx = 0; idx < static_cast<uint32_t>(frame->Views.size()); ++idx)
                 {
-                    const auto& view = views[idx];
+                    const auto& view = frame->Views[idx];
                     m_views[idx]->Update(idx, CreateProjectionMatrix(view), view.Space, view.IsFirstPersonObserver);
                 }
+                
+                // Check the frame to see if it has valid tracking, if it does not then the position should
+                //  be flagged as being emulated.
+                m_jsIsEmulatedPosition = Napi::Persistent(Napi::Boolean::New(info.Env(), !frame->IsTracking()));
```
ryantrem, line 886: This is going to update every frame. It seems like we should either:
1. Store a regular C++ bool, and have `GetEmulatedPosition` just return a new `Napi::Boolean` each time it is called (thoughts @bghgary?).
1. Only replace `m_jsIsEmulatedPosition` if the value has changed (e.g. compare `m_jsIsEmulatedPosition.Value()` to `frame->IsTracking()` before replacing the value).

Dependencies/xr/Source/OpenXR/XR.cpp
chrisfromwork: can we get an issue filed for this work?

---
In reply to: [527098276](https://github.com/BabylonJS/BabylonNative/pull/504#discussion_r527098276) [](ancestors = 527098276)

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: why is the preference null compared to undefined?

---
In reply to: [527103137](https://github.com/BabylonJS/BabylonNative/pull/504#discussion_r527103137) [](ancestors = 527103137)

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -873,16 +875,21 @@ namespace Babylon
...
-                for (uint32_t idx = 0; idx < static_cast<uint32_t>(views.size()); ++idx)
+                for (uint32_t idx = 0; idx < static_cast<uint32_t>(frame->Views.size()); ++idx)
                 {
-                    const auto& view = views[idx];
+                    const auto& view = frame->Views[idx];
                     m_views[idx]->Update(idx, CreateProjectionMatrix(view), view.Space, view.IsFirstPersonObserver);
                 }
+                
+                // Check the frame to see if it has valid tracking, if it does not then the position should
+                //  be flagged as being emulated.
```
chrisfromwork, line 885: >  [](start = 19, length = 1)

nit remove extra space

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: My naive instinct is that the recreation of an Napi::boolean every frame is likely less performant than the second option. That going said, I'd be in favor of the second option. :)

---
In reply to: [527109083](https://github.com/BabylonJS/BabylonNative/pull/504#discussion_r527109083) [](ancestors = 527109083)

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1605,15 +1617,31 @@ namespace Babylon
             XRRigidTransform& m_transform;
             Napi::ObjectReference m_jsPose{};
             Napi::ObjectReference m_jsJointPose{};
+            
+            bool m_hasBegunTracking = false;
```
chrisfromwork, line 1622: nit:  i think the preference is initialization in brackets

chrisfromwork (approved):
:shipit:

Dependencies/xr/Source/ARKit/XR.mm
CoPrez: Updated to move the logic into SessionImpl

ryantrem (approved):

Dependencies/xr/Source/OpenXR/XR.cpp
CoPrez: Issue filed ([here](https://github.com/BabylonJS/BabylonNative/issues/505)). I've updated the comment in code with a link to the issue as well.

Plugins/NativeXr/Source/NativeXr.cpp
CoPrez: Removed.

Plugins/NativeXr/Source/NativeXr.cpp
bghgary: Creating a js bool isn't too expensive. The persistent part is more expensive. I'm not sure which is faster in this case.

Plugins/NativeXr/Source/NativeXr.cpp
bghgary: This completely depends on the contract of GetViewerPose. `null` and `undefined` are two different things in js. I'm assuming the contract is nullable?

bghgary: Actually, no, it's not. The contract is undefined or the pose.

https://github.com/BabylonJS/Babylon.js/blob/b8209e7d7c27abcf021d55a0cfc69aa6e0951a0d/src/LibDeclarations/webxr.d.ts#L130

Plugins/NativeXr/Source/NativeXr.cpp
CoPrez: Updated

Plugins/NativeXr/Source/NativeXr.cpp
CoPrez: Keeping it as undefined. Thanks for finding the source of truth on that one @bghgary

Alex-MSFT (approved):

Dependencies/xr/Include/XR.h
```diff
@@ -223,6 +223,7 @@ namespace xr
                 void UpdateAnchor(Anchor&) const;
                 void DeleteAnchor(Anchor&) const;
                 Plane& GetPlaneByID(Plane::Identifier) const;
+                bool IsTracking() const;
```
Alex-MSFT, line 226: One thing we might want to consider setting IsTracking at frame creation time, and not having this query the  AR Session everytime it is called.  In general the frame object should just be a representation of state at the time the frame was created.

Plugins/NativeXr/Source/NativeXr.cpp
ryantrem: Interesting. The docs I was looking at before indicated null, but I guess we should go by the d.ts.

Plugins/NativeXr/Source/NativeXr.cpp
CoPrez: I did a test where I created two different persistent bools (one for true and one for false) so that I could always return whichever persistent bool value was correct at the time. That should have the least possible overhead as you don't need to do a comparison across the Napi layer. I then compared that to code that always just returns a new Napi bool. There was no effective difference to the performance of the scene I was testing with. I used a scene that was sufficiently complex to be limited by the CPU/GPU.

Because of that I'm just going to go with the cleanest code which is to return a new Napi bool each time the method is called.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -835,21 +836,22 @@ namespace Babylon
                 , m_jsTransform{Napi::Persistent(XRRigidTransform::New(info))}
                 , m_jsViews{Napi::Persistent(Napi::Array::New(info.Env(), 0))}
                 , m_transform{*XRRigidTransform::Unwrap(m_jsTransform.Value())}
+                , m_isEmulatedPosition{true}
             {
             }
 
-            void Update(const Napi::CallbackInfo& info, gsl::span<const xr::System::Session::Frame::View> views)
+            void Update(const Napi::CallbackInfo& info, const xr::System::Session::Frame* frame)
```
syntheticmagus, line 843: I strongly recommend against taking a raw pointer as an argument unless it is intended that this argument can genuinely be null. Since I don't believe this function should ever be called on `nullptr`, I recommend changing this signature to `const xr::System::Session::Frame&` and dereferencing in the calling code.

CoPrez: You are correct that in the context the frame is used here it should never be null. I've updated it to pass it by reference now rather than pointer.

PR #506, opened 2020-11-20
Add support for bgfx re-initialization

ryantrem:
Bgfx holds on to a fair bit of memory. Some of it seems like memory that should be held on to (like pre-allocated temporary buffers), and some of it seems like memory that should not be held on to and is probably a bug (e.g. index buffers, vertex buffers, and textures allocated during model load) (though this memory seems to be released when a NativeEngine is re-instantiated, which doesn't really make sense, so we need to investigate that more). This change makes it possible to choose to disable/suspend/deinitialize Graphics to release all bgfx memory (via `bgfx::shutdown`), and then enable/resume/re-initialize (via `bgfx::init`).

Previously, we exposed a promise instance from the `_native` object that was completed once graphics were initialized. Since graphics can now be re-initialized, I've changed this to exposing a function from the `_native` object that you can call to get a promise to await. This is the only required contract change - otherwise calling the new `EnableRendering`/`DisableRendering` are optional (only if you have a scenario where you want to release Graphics/bgfx related memory).

I've tested this on iOS and Android in Babylon React Native Playground. On Android, there seems to be a lot of memory that is still not released, so I need to investigate this more, but maybe after #467 is completed. On iOS, most of the memory is released, and I think the rest will be released after #467 is complete.

I also tested this on Android in the regular Babylon Native Playground and it seems to work fine. **I haven't manually tested it on any other platforms for regular Babylon Native Playground, including Win32 multi-threading, so could use some help there**. Presumably the validation tests will give me coverage on most platforms though.


Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -53,6 +59,7 @@ namespace Babylon
             bool Dirty{};
```
ryantrem, line 59: I could replace `Initialized` with the thread affinity object, since `m_renderThreadAffinity.is_set()` always has the same value is the current `Initialized` flag. Any thoughts on this?

Core/Graphics/Source/Graphics.cpp
```diff
@@ -82,31 +84,62 @@ namespace Babylon
...
+        assert(m_renderThreadAffinity.check());
+
         if (m_rendering)
         {
             throw std::runtime_error{"Current frame cannot be started before prior frame has been finished."};
         }
         m_rendering = true;
 
+        // Ensure rendering is enabled
+        EnableRendering();
```
syntheticmagus, line 138: nit: As we chatted about offline, if this is exposed publicly such that you _can_ call it, I think I'd prefer that you _have_ to call it, and we assert/fail if a call to render happens while rendering isn't enabled. This applies to calling `DisableRendering` in the destructor, too: off the top of my head, it seems better to make calling that the owner's job with the destructor just asserting/failing if shutting down while rendering is enabled, especially since the lifetime of the Graphics object isn't guaranteed to be managed on the render thread. (This was a problem with the pre-existing call to `bgfx::shutdown`, too, come to think of it.) It's not a deal-breaker for me, but I would definitely prefer that these methods, if they must exist, always be required.

Core/Graphics/Source/GraphicsImpl.h
syntheticmagus: If you add an inline predicate `Initialized()` to `BgfxState` that just checks the affinity, I think that'd be fine. Just leaving it as a bool seems fine too. Having `m_renderThreadAffinity.is_set()` checks in the code without additional context or naming seems like it could be confusing, though.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -2,13 +2,15 @@
...
+#include <cassert>
+
 #define BGFX_RESET_FLAGS (BGFX_RESET_VSYNC | BGFX_RESET_MSAA_X4 | BGFX_RESET_MAXANISOTROPY)
 
 namespace Babylon
 {
     namespace
     {
-        constexpr auto JS_SENTINEL_NAME = "graphicsInitializationPromise";
+        constexpr auto JS_GRAPHICS_READY_NAME = "whenGraphicsReady";
```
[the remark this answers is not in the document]
ryantrem: I don't mind changing the name, though we already use the "when" prefix for task/promise returning functions in many places in both the native and JS code, so I don't think it is out of place. Do you have suggestions for a different name (that is still consistent with other parts of the code)?

As far as a promise returning function vs an event/callback, I strongly prefer a promise returning function. We need to await this state inside initialization code, and having to do it through state+event is both harder for the consumer and more error prone. For example, you need expose the current state and the caller needs to check it before registering for the event. If you forget to check the state, then your event handler will never trigger, and your code basically hangs. This could easily happen depending on the order you call the various initialization functions in Babylon Native. Additionally, even if you got this code right, you'd still have to "convert" the state+event into a promise that you can await in your init code. All of that seems way more complicated for the consumer than just awaiting a promise returning function.

Core/Graphics/Source/Graphics.cpp
ryantrem: Yea, I chatted with @bghgary offline as well and his recommendation was to still do it automatically to not break the current contract. I tend to agree with this, as it simplifies what I believe to be the common case (automatic rendering with no intention of releasing bgfx resources until the app dies). In general, my preference is to optimize for the common case, but allow additional complexity for the uncommon case. That said, I'm willing to defer to you and @bghgary. I guess you two can battle it out? :)
Fight!

bghgary: I don't have a strong preference either way.

Core/Graphics/Source/Graphics.cpp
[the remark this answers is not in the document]
bghgary: We need to agree on the contract for initialization in general. We can decide on the naming then.

Core/Graphics/Source/Graphics.cpp
syntheticmagus: I think it's perfectly fine to go in as-is. We have the discussion and rationale recorded here, so we can always revisit it if we decide we want to change something later.

Core/Graphics/Source/Graphics.cpp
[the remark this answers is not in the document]
syntheticmagus: Do we want to address that in this PR, or separately? I think the name here is perfectly workable, so if we want to check it in now and address the initialization contract in a separate issue/meeting/PR, that sounds reasonable to me.

syntheticmagus (approved):
As mentioned in the comment conversations, there are aspects to this change that I either don't prefer or that are simply a part of larger issues that require further conversations and PRs; but with those considerations in mind, I think this change is fine to go in as a part of moving forward.

PR #515, opened 2020-11-24
ValidationTests Android

CedricGuillemet:
Android ValidationTests.
Project + CI build
A 2nd PR is expected with VT running Emulator.


Apps/ValidationTests/Android/app/src/main/java/BabylonNative/BabylonView.java
```diff
@@ -0,0 +1,121 @@
+package BabylonNative;
```
syntheticmagus, line 1: nit: Is this file essentially duplicated from a BabylonReactNative predecessor? If so, should we consider putting this in as a utility, perhaps an Android-specific plugin?

syntheticmagus (approved):

Apps/ValidationTests/Android/app/src/main/java/BabylonNative/BabylonView.java
CedricGuillemet: This is a long lasting issue. Basically, JNI cpp + glue is the same as the playground testapp. we could factor that into an Android library and use it (as a package?) in both playground & validation. For now, we only provide BN as a source code subrepo. I hope this PR will also push forward to do something with proper packages and releases.

PR #519, opened 2020-11-26
fix leaks + removed dead code

### LINKED ISSUE #295: Memory leak in Console polyfill
https://github.com/BabylonJS/BabylonNative/blob/master/Polyfills/Console/Source/Console.cpp#L21

The allocation on this line is never released. Should also use the data parameter of the DefineClass instead of passing in an argument.

CedricGuillemet:
- fixes #295 
- removed dead code
- fix leak in testutils


Polyfills/Console/Source/Console.cpp
```diff
@@ -16,15 +16,15 @@ namespace Babylon::Polyfills::Internal
...
 
-        Napi::Object console = func.New({Napi::External<Babylon::Polyfills::Console::CallbackT>::New(env, new Babylon::Polyfills::Console::CallbackT(std::move(callback)))});
-        env.Global().Set(JS_INSTANCE_NAME, console);
+        env.Global().Set(JS_INSTANCE_NAME, func.New({}));
     }
 
     Console::Console(const Napi::CallbackInfo& info)
         : ParentT{info}
-        , m_callback{*info[0].As<Napi::External<Babylon::Polyfills::Console::CallbackT>>().Data()}
+        , m_callback{std::move(*(Babylon::Polyfills::Console::CallbackT*)info.Data())}
```
syntheticmagus, line 27: nit: This seems like an unusual pattern. I don't think it will cause problems as written, but taking ownership of the callback this way seems a bit confusing to me. Could we instead set the callback member directly from the static function? In other words, change the initializer here to 
```
, m_callback{}
```
and change the calling code to 
```
auto console = func.New({});
Console::Unwrap(console)->m_callback = std::move(callback);
env.Global().Set(JS_INSTANCE_NAME, console);
```
or something like that? Slightly less efficient, but very clear what it's doing.

syntheticmagus (approved):

PR #520, opened 2020-11-26
more validation tests

CedricGuillemet:
6 more tests. 1 with draco gltf.


Apps/ValidationTests/Scripts/validation_native.js
```diff
@@ -6,6 +6,13 @@ var saveResult = true;
 var testWidth = 600;
 var testHeight = 400;
 
+// Random replacement
```
syntheticmagus, line 9: nit: If this is a common algorithm, can you add a link in a comment about where it came from? Alternatively, there's a simple random function used with the Babylon.js procedural textures (it's from a `vec2`, but could be adapted easily), so we could just use that one and reference it there. Not a big deal, just a leftover sensibility I have about keeping track of random algorithms.

syntheticmagus (approved):

Apps/ValidationTests/Scripts/validation_native.js
CedricGuillemet: I've added a note in the validation tests infrastructure issue to refactor the validation_native.js with the bjs version. I'll take care of that rando, function at the same time.

Apps/BabylonScripts/draco_decoder_gltf.js
```diff
@@ -0,0 +1,48 @@
+
```
bghgary, line 1: We don't want the webassembly version?

CedricGuillemet: It doesn't work yet.

bghgary: Okay, we should make a note or file an issue or something. It would be good to avoid adding incomplete work if possible.

CedricGuillemet: The goal of VT is to avoid regression. Any test scene that can pass the test must be added. If we only test feature complete PG we wouldn't test anything as xmlhttprequest is still incomplete. Issue is still open for WASM btw.

PR #525, opened 2020-12-02
Fix swimminess on iOS

Alex-MSFT:
This PR addresses the swimminess on iOS detailed in [BabylonReactNative#74](https://github.com/BabylonJS/BabylonReactNative/issues/74).  After investigation it was determined there was a 1 frame delay between each render call and the content being presented on iOS.  This was due to use not setting the BGFX_RESET_FLIP_AFTER_RENDER flag on iOS.  By default bgfx will only "flip" (render) the content in its buffer when preparing the next frame, so the actual texture draw would only occur on the subsequent frame.

I'm also making a small change to XR.mm to keep a single reference to the current frame retained for the duration of the frame, and changing all other frame dependent calls to operate against that reference.  Since ARKit is working off the main thread it is possible that the frame will change during execution, so this ensures that we're always operating off the same data for the entirety of the frame.


Core/Graphics/Source/Graphics.cpp
```diff
@@ -9,6 +9,8 @@
 // See issue https://github.com/BabylonJS/BabylonNative/issues/494#issuecomment-731135918
 // for explanation
 #define BGFX_RESET_FLAGS (BGFX_RESET_VSYNC | BGFX_RESET_MAXANISOTROPY)
+#elif __APPLE__
+#define BGFX_RESET_FLAGS (BGFX_RESET_VSYNC | BGFX_RESET_MSAA_X4 | BGFX_RESET_MAXANISOTROPY | BGFX_RESET_FLIP_AFTER_RENDER)
```
ryantrem, line 13: Probably worth adding a comment?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -639,7 +640,7 @@ bool TryInitialize() {
         std::unique_ptr<System::Session::Frame> GetNextFrame(bool& shouldEndSession, bool& shouldRestartSession, std::function<void(void* texturePointer)> deletedTextureCallback) {
             shouldEndSession = sessionEnded;
             shouldRestartSession = false;
-
+
```
ryantrem, line 643:
```suggestion

```

ryantrem:
> I'm also making a small change to XR.mm to keep a single reference to the current frame retained for the duration of the frame, and changing all other frame dependent calls to operate against that reference. Since ARKit is working off the main thread it is possible that the frame will change during execution, so this ensures that we're always operating off the same data for the entirety of the frame.

Is this actually needed, or just a safeguard? I wouldn't expect anything to change under the hood on us. I would only expect it to change if we pull another frame (e.g. `[session currentFrame]`). If we retain the frame, then I believe `[session currentFrame]` will keep returning the same frame until we release it. But obviously if you are already holding on to it so you can release it later, then we may as well use that member var. So I guess what I'm saying is I think this change makes sense, but for a different reason than in your description, and so I want to make sure we have a shared understanding of how this works.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -650,7 +651,8 @@ bool TryInitialize() {
                     [desiredSuperview addSubview:xrView];
                 }
                 
-                ARFrame* currentFrame = [session currentFrame];
+                currentFrame = [session currentFrame];
+                [currentFrame retain];
```
ryantrem, line 655: It doesn't matter too much, but if you wanted to make this more symmetrical, you could pull these two lines out and put them before the `dispatch_sync` call. Then we are acquiring/retaining the `currentFrame` and releasing it from the same thread. I tested this out a couple weeks ago and as far as I could tell it worked correctly.

ryantrem (approved):

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: That is what I originally had, if you think this is cleaner that is good with me.  I had put them like this so that the lifetime of currentFrame would be tied to the lifetime of the XRFrame here, but I was waffling between the two places.

Alex-MSFT:
> > I'm also making a small change to XR.mm to keep a single reference to the current frame retained for the duration of the frame, and changing all other frame dependent calls to operate against that reference. Since ARKit is working off the main thread it is possible that the frame will change during execution, so this ensures that we're always operating off the same data for the entirety of the frame.
> 
> Is this actually needed, or just a safeguard? I wouldn't expect anything to change under the hood on us. I would only expect it to change if we pull another frame (e.g. `[session currentFrame]`). If we retain the frame, then I believe `[session currentFrame]` will keep returning the same frame until we release it. But obviously if you are already holding on to it so you can release it later, then we may as well use that member var. So I guess what I'm saying is I think this change makes sense, but for a different reason than in your description, and so I want to make sure we have a shared understanding of how this works.

Yes you're reading is right if we retain it we should not be served new frames until it is released, so using session.currentFrame should work just fine.  But what you are saying is the same logic I went through in my head.  If we are going ahead and retaining it we need to be able to release it later on anyways, so might as well make it a member variable.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Oh oops read this comment wrong the first time.  Yeah that makes sense to keep it all on the same thread, will do.

bghgary (approved):

ryantrem (approved):

PR #526, opened 2020-12-03
Add dom static lib to polyfill DOMPointReadOnly for WebXR APIs

chrisfromwork:
We have some WebXR definitions that require constructing and handing around DOMPointReadOnlys. This change introduces an incomplete polyfill for DOMPointReadOnly that allows us to unblock WebXR world geometry scenarios in BabylonNative

### DIFF
--- a/Apps/Playground/Android/app/CMakeLists.txt
+++ b/Apps/Playground/Android/app/CMakeLists.txt
@@ -63,7 +63,8 @@ target_link_libraries(BabylonNativeJNI
         Console
         Window
         ScriptLoader
-        XMLHttpRequest)
+        XMLHttpRequest
+        DOM)
 
 configure_file(
     "${ANDROID_NDK}/sources/cxx-stl/llvm-libc++/libs/${ANDROID_ABI}/libc++_shared.so"
--- a/Polyfills/DOM/Source/DOM.cpp
+++ b/Polyfills/DOM/Source/DOM.cpp
@@ -1,0 +1,69 @@
+#include <Babylon/Polyfills/DOM.h>
+
+namespace Babylon::Polyfills::DOM
+{
+    class DOMPointReadOnly final : public Napi::ObjectWrap<DOMPointReadOnly>
+    {
+        static constexpr auto JS_CLASS_NAME = "DOMPointReadOnly";
+
+    public:
+        static void Initialize(Napi::Env env) {
+            Napi::HandleScope scope{ env };
+
+            Napi::Function func = DefineClass(
+                env,
+                JS_CLASS_NAME,
+                {
+                    InstanceAccessor("x", &DOMPointReadOnly::GetX, nullptr),
+                    InstanceAccessor("y", &DOMPointReadOnly::GetY, nullptr),
+                    InstanceAccessor("z", &DOMPointReadOnly::GetZ, nullptr),
+                    InstanceAccessor("w", &DOMPointReadOnly::GetW, nullptr)
+                });
+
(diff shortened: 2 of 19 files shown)


Apps/Playground/Android/app/CMakeLists.txt
```diff
@@ -63,7 +63,8 @@ target_link_libraries(BabylonNativeJNI
         Console
         Window
         ScriptLoader
-        XMLHttpRequest)
+        XMLHttpRequest
+        DOM)
```
bghgary, line 67: I'm not sure about calling this polyfill `DOM` though I'm not sure I have a better name. I don't think we will ever implement all of the Web APIs that start with DOM.

Polyfills/DOM/Source/DOM.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Babylon::Polyfills::DOM
+{
+    class DOMPointReadOnly final : public Napi::ObjectWrap<DOMPointReadOnly>
+    {
+        static constexpr auto JS_CLASS_NAME = "DOMPointReadOnly";
+
+    public:
+        static void Initialize(Napi::Env env) {
+            Napi::HandleScope scope{ env };
```
bghgary, line 11: nit: format with clang-format before merging

Polyfills/DOM/Source/DOM.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+            if (info.Length() > 2)
+            {
+                _z = info[2].As<Napi::Number>();
+            }
+
+            if (info.Length() > 3)
+            {
+                _w = info[3].As<Napi::Number>();
+            }
```
bghgary, line 39: nit: this can be all done with member initializers to avoid double initialization

Polyfills/DOM/Source/DOM.cpp
```diff
@@ -0,0 +1,69 @@
...
+                    InstanceAccessor("w", &DOMPointReadOnly::GetW, nullptr)
+                });
+
+            env.Global().Set(JS_CLASS_NAME, func);
+        }
+
+        DOMPointReadOnly(const Napi::CallbackInfo& info) : Napi::ObjectWrap<DOMPointReadOnly>{info}
+        {
+            _x = info[0].As<Napi::Number>();
+            _y = info[1].As<Napi::Number>();
```
bghgary, line 29: according to spec, even the first two values are optional, so you probably need to check the length for >0 and >1 also.

bghgary (approved):

Polyfills/DOM/Source/DOM.cpp
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            if (info.Length() > 3)
+            {
+                _w = info[3].As<Napi::Number>();
+            }
+        }
+
+    private:
+        float _x{ 0 };
```
tymlipari, line 43: Should these be `float` or `double`? Not sure what precision level is expected from `DOMPointReadOnly`.

bghgary: Ooh, this is a good point. all floating point values in JS are doubles. I would recommend using doubles.

ryantrem:
I feel like I need to see this change in context of its usage to really understand it. Is `DOMPointReadOnly` an input to the XR APIs that code on the JS side needs to construct, or is it only an output of the XR APIs?

chrisfromwork:
we are going to avoid checking this in in favor of creating a class that implements the DOMPointReadOnly interface in our typescript library

Issue #512, opened 2020-11-23
[APPLE]Metal/bgfx concurrency with renderAutomatically off


CedricGuillemet:
I'm trying ValidationTests on Mac.
With renderAutomatically on, everything works as expected.
When it's off, I do the rendering in the MTKView delegate, it works fine for some scenes and then it crashes.
The callstacks I get for main and JS threads look like this:
![image](https://user-images.githubusercontent.com/1312968/99984820-d620d480-2dad-11eb-9c96-939a792a4261.png)
Basically, a shader is created on the JS thread while the UI thread responsible for rendering is decoding and uploading a texture. It looks like both threads are accessing Metal resources which leads to a crash.
My guess is to use a rendering encode in the JS thread to not have direct access to graphic API.
What's your thoughts @syntheticmagus ?

CedricGuillemet:
I think this issue can be closed. Do you confirm @bghgary @syntheticmagus @ryantrem  ?

bghgary:
This shouldn't be an issue anymore.

CedricGuillemet:
Closing.

PR #530, opened 2020-12-04
small fixes for macos

### LINKED ISSUE #510: [APPLE]Heap corruption when loading Cubemap w/mipmaps
with automaticRendering on on mac and validation tests, I got a recurring crash with napi here: https://github.com/BabylonJS/BabylonNative/blob/cb2f58449157f78c56109787d247c67f732e84ed/Plugins/NativeEngine/Source/NativeEngine.cpp#L1162
I recoded the lambda to something like this:
```
                auto* pointer = static_cast<uint8_t*>(typedArray.ArrayBuffer().Data()) + typedArray.ByteOffset();
                std::vector<uint8_t> memory(pointer, pointer + typedArray.ByteLength());
                tasks[(face * numMips) + mip] = arcana::make_task(arcana::threadpool_scheduler, m_cancelSource, [this, memory = std::move(memory)]() {
                    bimg::ImageContainer* image = bimg::imageParse(&m_allocator, memory.data(), memory.size());
                    FlipY(image);
                    return image;
                });
```
And crash disappeared.
I wanted to make a capture of the callstack of the crash for this issue but once my code discarded, I get random crash with heap corruption.
At this point, I don't know if the code is the culprit or moving/replacing code will hide the memory corruption.

I finally managed to get a callstack with the crash in napi:
image

CedricGuillemet:
- fix URL containing forbiden characters
- bool RENDER_ON_JS_THREAD on macOS validationtests

### DIFF
--- a/Apps/ValidationTests/macOS/ViewController.mm
+++ b/Apps/ValidationTests/macOS/ViewController.mm
@@ -13,6 +13,7 @@
 
 std::unique_ptr<Babylon::Graphics> graphics{};
 std::unique_ptr<Babylon::AppRuntime> runtime{};
+constexpr bool RENDER_ON_JS_THREAD{true};
 
 @interface EngineView : MTKView <MTKViewDelegate>
 
@@ -29,7 +30,7 @@ std::unique_ptr<Babylon::AppRuntime> runtime{};
 
 - (void)drawInMTKView:(MTKView *)__unused view
 {
-    if (graphics != nullptr) {
+    if (graphics != nullptr && !RENDER_ON_JS_THREAD) {
         graphics->RenderCurrentFrame();
     }
 }
@@ -75,7 +76,7 @@ std::unique_ptr<Babylon::AppRuntime> runtime{};
         Babylon::Polyfills::XMLHttpRequest::Initialize(env);
 
         graphics->AddToJavaScript(env);
-        Babylon::Plugins::NativeEngine::Initialize(env, false); // render on UI Thread
+        Babylon::Plugins::NativeEngine::Initialize(env, RENDER_ON_JS_THREAD);
 
         Babylon::TestUtils::CreateInstance(env, windowPtr);
     });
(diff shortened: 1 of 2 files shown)


Dependencies/UrlLib/Source/Apple/UrlRequest.mm
```diff
@@ -56,6 +58,11 @@ void ResponseType(UrlResponseType value)
 
             NSURLSession* session{[NSURLSession sharedSession]};
             NSURLRequest* request{[NSURLRequest requestWithURL:url]};
+            if (url == nil)
+            {
+                // Complete the task, but retain the default status code of 0 to indicate a client side error.
+                return arcana::task_from_result<std::exception_ptr>();
```
bghgary, line 64: We still need to work on getting the status text propagated back out for this to be fully correct.

Apps/ValidationTests/macOS/ViewController.mm
```diff
@@ -75,7 +76,7 @@ - (void)refreshBabylon {
         Babylon::Polyfills::XMLHttpRequest::Initialize(env);
 
         graphics->AddToJavaScript(env);
-        Babylon::Plugins::NativeEngine::Initialize(env, false); // render on UI Thread
+        Babylon::Plugins::NativeEngine::Initialize(env, RENDER_ON_JS_THREAD);
```
bghgary, line 79: We are now changing this from false to true. What is the reason why it doesn't work?

bghgary (approved):
Okay to merge, but I have some comments/questions.

Apps/ValidationTests/macOS/ViewController.mm
CedricGuillemet: Because of this : https://github.com/BabylonJS/BabylonNative/issues/512

bghgary: 👍

PR #531, opened 2020-12-07
Enabling default and microsoft controller bindings side-by-side

rickfromwork:
We were using the MS controller bindings and defaulting to default bindings when they weren't accessible, but this creates a problem. Aim/grip space is only populated when the input device is a MS controller, and other devices, such as hands, do not populate aim/grip space.

Also, if the default profile *was* used, xrSuggestInteractionProfileBindings would hit an error as the list of bindings included invalid bindings for the default profile.

### DIFF
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -733,7 +733,7 @@ namespace xr
                 ActionResources.CONTROLLER_GET_TRIGGER_VALUE_ACTION_LOCALIZED_NAME,
                 ActionResources.CONTROLLER_GET_TRIGGER_VALUE_PATH_SUFFIX,
                 &ActionResources.ControllerGetTriggerValueAction,
-               microsoftControllerBindings
+                microsoftControllerBindings
                 instance);
 
             // Create controller get squeeze click action and suggested bindings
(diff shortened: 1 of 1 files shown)


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -784,30 +793,22 @@ namespace xr
...
             // Provide Microsoft suggested binding to instance
             XrInteractionProfileSuggestedBinding microsoftSuggestedBindings{ XR_TYPE_INTERACTION_PROFILE_SUGGESTED_BINDING };
             XrCheck(xrStringToPath(instance, ActionResources.MICROSOFT_XR_INTERACTION_PROFILE, &microsoftSuggestedBindings.interactionProfile));
-            microsoftSuggestedBindings.suggestedBindings = bindings.data();
-            microsoftSuggestedBindings.countSuggestedBindings = (uint32_t)bindings.size();
-
-            // Fallback on the default bindings if the Microsoft bindings fail
-            if (XR_FAILED(xrSuggestInteractionProfileBindings(instance, &microsoftSuggestedBindings)))
-            {
-                ControllerInfo.ControllerBinding = false;
```
satyapoojasama, line 799: you also might want to remove the "ControllerBinding" parameter from ControllerInfo all together.

satyapoojasama (approved):

tymlipari (approved):

Issue #522, opened 2020-11-30
Rendering stops working if an XR session starts+stops without getting tracking


ryantrem:
On both Android and iOS, rendering gets in a bad state if you start then stop XR without XR ever actually getting tracking. This can easily be reproed by just starting XR with the device laying flat on a desk. If you rotate the device 90 degrees (and the view/window is reset), then it starts rendering correctly again. This behavior was introduced with https://github.com/BabylonJS/BabylonNative/commit/23f066b567d5eb7054c984a0726e6bc59fe74d06.

Drigax:
When in the repro state, I can see that we are now drawing the equivalent of multiple frames in a single frame:

![image](https://user-images.githubusercontent.com/3952729/101378707-49bfe880-3868-11eb-8006-5c903187bd1d.png)


Compared with a working frame, we may be drawing hundreds of times:
![image](https://user-images.githubusercontent.com/3952729/101379004-9efbfa00-3868-11eb-962d-9d3f4042d6f9.png)

PR #537, opened 2020-12-11
Fix rendering no longer working when exiting XR

Drigax:
In some scenarios we can create multiple framebuffers with conflicting ViewID. This can cause us to have a scenario where the currently bound framebuffer has a view id, but BGFX's underlying view is mapped to another framebuffer. This is currently observed in #522, where where we create a framebuffer via `NativeXR::::BeginFrame()` but it is never bound via `NativeEngine`.

Proposed fix is to guarantee `FrameBufferManager.m_boundFrameBuffer` has clean view id when referenced.


Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -280,8 +280,15 @@ namespace Babylon
...
-        FrameBufferData& GetBound() const
+        FrameBufferData& GetBound()
         {
+            // Babylon.js minimizes how frequently Bind() is called, causing there
+            // to be some scenarios where the currently bound frame buffer can retain
+            // an invalid view id when another FrameBuffer is created but not bound.
+            // To prevent this, make sure the bound frameBuffer has a clean ViewId.
+            if (m_boundFrameBuffer->IsViewIdDirty){
+                Bind(m_boundFrameBuffer);
+            }
```
ryantrem, line 291:
```suggestion
            if (m_boundFrameBuffer->IsViewIdDirty)
            {
                Bind(m_boundFrameBuffer);
            }
```

ryantrem (approved):

syntheticmagus (approved):

PR #539, opened 2020-12-11
Only creating gamepad data when gamepad information is available

rickfromwork:
Since the action profile's availability is separate from whether we use it, this change only creates a gamepad component on the input source if gamepad data is available.

### DIFF
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -1446,7 +1446,7 @@ namespace xr
                 // Get gamepad data 
                 {
                     const auto& controllerInfo = sessionImpl.ControllerInfo;
-                    auto gamepadObject = InputSources[idx].GamepadObject;
+                    auto& gamepadObject = InputSources[idx].GamepadObject;
 
                     // Update gamepad data
                     if ((m_impl->TryUpdateControllerFloatAction(actionResources.ControllerGetTriggerValueAction, session, gamepadObject.Buttons[controllerInfo.TRIGGER_BUTTON].Value)) &&
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -312,8 +312,6 @@ namespace xr
             static constexpr uint32_t TRACKPAD_Y_AXIS = 1;
             static constexpr uint32_t THUMBSTICK_X_AXIS = 2;
             static constexpr uint32_t THUMBSTICK_Y_AXIS = 3;
-
-            XrBool32 ControllerBinding{ true };
         } ControllerInfo;
 
         struct HandInfo
@@ -808,7 +806,7 @@ namespace xr
             XrCheck(xrStringToPath(instance, ActionResources.MICROSOFT_XR_INTERACTION_PROFILE, &microsoftSuggestedBindings.interactionProfile));
             microsoftSuggestedBindings.suggestedBindings = microsoftControllerBindings.data();
             microsoftSuggestedBindings.countSuggestedBindings = (uint32_t)microsoftControllerBindings.size();
-            ControllerInfo.ControllerBinding = XR_SUCCEEDED(xrSuggestInteractionProfileBindings(instance, &microsoftSuggestedBindings));
+            XrCheck(xrSuggestInteractionProfileBindings(instance, &microsoftSuggestedBindings));
 
             XrSessionActionSetsAttachInfo attachInfo{ XR_TYPE_SESSION_ACTION_SETS_ATTACH_INFO };
             attachInfo.countActionSets = 1;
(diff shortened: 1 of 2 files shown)


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1431,63 +1444,22 @@ namespace xr
...
-                    }
-
-                    // Get trackpad axes data
-                    float currentTrackpadXAxis, currentTrackpadYAxis;
-                    if (m_impl->QueryControllerVector2fAction(actionResources.ControllerGetTrackpadAxesAction, session, currentTrackpadXAxis, currentTrackpadYAxis))
+                    const auto& controllerInfo = sessionImpl.ControllerInfo;
+                    auto gamepadObject = InputSources[idx].GamepadObject;
+
+                    // Update gamepad data
+                    if ((m_impl->TryUpdateControllerFloatAction(actionResources.ControllerGetTriggerValueAction, session, gamepadObject.Buttons[controllerInfo.TRIGGER_BUTTON].Value)) &&
```
chrisfromwork, line 1452: >TryUpdateControllerFloatAction [](start = 33, length = 30)

nit: can any of these fail while others succeed?

chrisfromwork (approved):
:shipit:

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I don't think so, but if any do fail I'd rather not send incomplete data (since babylonJS consumes all or none of it)

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1431,63 +1444,22 @@ namespace xr
...
-                        inputSource.GamepadObject.Axes[controllerInfo.TRACKPAD_Y_AXIS] = currentTrackpadYAxis;
+                        // Only signal that gamepad data is available if the actions were available
+                        InputSources[idx].GamepadTrackedThisFrame = true;
                     }
-
-                    //Get trackpad click data
-                    bool currentTrackpadValue;
-                    if (m_impl->QueryControllerBooleanAction(actionResources.ControllerGetTrackpadClickAction, session, currentTrackpadValue))
-                    {
-                        inputSource.GamepadObject.Buttons[controllerInfo.TRACKPAD_BUTTON].Pressed = currentTrackpadValue;
```
satyapoojasama, line 1466: why did you remove these? how are you updating the gamepad current values?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1115,53 +1113,68 @@ namespace xr
...
-           // query input action state
-           XrActionStateBoolean booleanActionState{XR_TYPE_ACTION_STATE_BOOLEAN};
-           XrActionStateGetInfo getInfo{XR_TYPE_ACTION_STATE_GET_INFO};
-           getInfo.action = controllerAction;
-           XrCheck(xrGetActionStateBoolean(session, &getInfo, &booleanActionState));
-           auto controllerInfo = sessionImpl.ControllerInfo;
-           if (booleanActionState.changedSinceLastSync)
-           {
+        // Returns true if the action is supported on the current input
+        bool TryUpdateControllerBooleanAction(XrAction controllerAction, XrSession session, bool& currentBooleanState)
```
satyapoojasama, line 1138: I believe you mixed up two things here, action supported and changed since last frame are two different things. you might want probably a different function for action supported. It is a bit ambiguous

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I'm updating them inside of TryUpdateController*Action

satyapoojasama: okay, I just realized it :)

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1431,63 +1444,22 @@ namespace xr
...
+                        (m_impl->TryUpdateControllerBooleanAction(actionResources.ControllerGetTrackpadTouchAction, session, gamepadObject.Buttons[controllerInfo.TRACKPAD_BUTTON].Touched)) &&
+                        (m_impl->TryUpdateControllerVector2fAction(actionResources.ControllerGetThumbstickAxesAction, session, gamepadObject.Axes[controllerInfo.THUMBSTICK_X_AXIS], gamepadObject.Axes[controllerInfo.THUMBSTICK_Y_AXIS])) &&
+                        (m_impl->TryUpdateControllerBooleanAction(actionResources.ControllerGetThumbstickClickAction, session, gamepadObject.Buttons[controllerInfo.THUMBSTICK_BUTTON].Pressed)))
                     {
-                        inputSource.GamepadObject.Axes[controllerInfo.TRACKPAD_X_AXIS] = currentTrackpadXAxis;
-                        inputSource.GamepadObject.Axes[controllerInfo.TRACKPAD_Y_AXIS] = currentTrackpadYAxis;
+                        // Only signal that gamepad data is available if the actions were available
+                        InputSources[idx].GamepadTrackedThisFrame = true;
```
rgerd, line 1461: nit: Could set the `GamepadTrackedThisFrame` member directly to the result of the &'s instead of using an if, but I realize this may be a personal preference.

rgerd (approved):

satyapoojasama (approved):

PR #546, opened 2020-12-16
2D raw texture

### LINKED ISSUE #543: Support Raw textures in Babylon Native
Support raw textures in Babylon Native to enable RGBA buffers to be used as textures on shaders/materials. 

Should currently work in Babylon already, work is to bring support to Babylon Native.

CedricGuillemet:
fixes #543 
PG https://playground.babylonjs.com/#PS73E8


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1100,6 +1102,39 @@ namespace Babylon
             });
     }
 
+    void NativeEngine::LoadRawTexture(const Napi::CallbackInfo& info)
```
ryantrem, line 1105: Why don't we have `onSuccess`/`onError` callbacks for this one? It looks like all the other texture related loading functions have those.

CedricGuillemet: It's a JS synchronous operation https://github.com/BabylonJS/Babylon.js/blob/35ebb5c27d0cfbe42954a69a8a03843d1ca0a6e1/src/Engines/Extensions/engine.rawTexture.ts#L251

bghgary: If it is actually synchronous, we need to make the code actually synchronous. It's currently create a task that doesn't go anywhere. If an error occurs, it will go no where.

ryantrem (approved):

Issue #540, opened 2020-12-11
Setting mesh.renderingGroupId doesn't work


bghgary:
Use the code from the https://playground.babylonjs.com/#1J6IGD in Babylon Native PG.

Result:
![image](https://user-images.githubusercontent.com/19158740/101957656-b3b4f680-3bb6-11eb-8289-38aab1d9af17.png)

Expected:
![image](https://user-images.githubusercontent.com/19158740/101957665-b9aad780-3bb6-11eb-909e-2c0058256d90.png)

bghgary:
This is because `engine.clear` doesn't work properly in NativeEngine. NativeEngine currently just sets the clear states for the view which is only called once during bgfx::submit. I think we need to set the states and then call bgfx::touch.

bghgary:
Had some discussions with @syntheticmagus and @CedricGuillemet. Basically, bgfx handles clear state via views. We have to use bgfx views to do clears which we are currently not. We also want to ensure that the view clear state handling will work for XR scenarios. For example, make sure this [PR](https://github.com/BabylonJS/BabylonNative/pull/468) doesn't get regressed.

bghgary:
This is "working" no with PR #547, but there is more work to clean up this code and make it work properly for XR scenarios.

PR #547, opened 2020-12-16
Change clear to increment the bgfx view id

bghgary:
This is a quick fix for #540. There is more work to do to fix this properly.

The gist of the change is to increment the view id before setting the view clear state. This will ensure bgfx issues a clear call when clear is called by Babylon.js.

There is a corresponding change in Babylon.js here: https://github.com/BabylonJS/Babylon.js/pull/9619


bghgary:
@rgerd This isn't the final fix, but can you test it in your scenarios. I'd like to get this fix in first to unblock a partner.

syntheticmagus (approved):
This is marked as a draft, so I'm not sure if I'm supposed to approve it yet, but I have no qualms about getting this (and its partner in the other repo) checked in to unblock someone now since the more permanent solution will take a bit more work.

bghgary:
I'm going to merge this for now. @rgerd If you get a chance to test and find something, let me know.

PR #549, opened 2020-12-18
Fixes for react native UWP support

chrisfromwork:
These fixes are needed for BabylonReactNative UWP support.

1. Some libraries need silenced deprecation warnings
2. napi-jsi is missing a convert to boolean operator
3. comparisons of uint32_t and int32_t types generating warnings/compilation errors
4. We need to expose the ability to set a window type for graphics/bgfx in order to support xaml applications


Dependencies/CMakeLists.txt
```diff
@@ -32,6 +32,8 @@ elseif(ANDROID)
     add_compile_definitions(BGFX_CONFIG_RENDERER_OPENGLES=30)
 elseif(UNIX)
     add_compile_definitions(BGFX_CONFIG_RENDERER_OPENGL=33)
+elseif(WINDOWS_STORE)
+    set(WINDOWS_WINRT_NO_RUNTIME ON CACHE BOOL "Avoid consuming windows runtime library in bgfx static libs")
```
chrisfromwork, line 36: nit: this requires submodule changes. I should remove this for now

Core/Graphics/Source/Graphics.cpp
```diff
@@ -277,6 +286,11 @@ namespace Babylon
         m_impl->SetNativeWindow(windowPtr);
     }
 
+    void Graphics::UpdateWindowType(void* windowType)
```
bghgary, line 289: Would you ever update the window type after creating the graphics object? Seems like this can be part of UpdateWindow and CreateGraphics (possibly as overloads)?

Dependencies/napi/napi-jsi/CMakeLists.txt
```diff
@@ -16,4 +16,8 @@ target_include_directories(napi
 target_link_to_dependencies(napi
     PUBLIC jsi)
 
+if(WIN32)
+    target_compile_definitions(napi PRIVATE _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING)
```
bghgary, line 20: Probably should make this PUBLIC instead of PRIVATE and then you don't need to add it to all the remaining cmakelists.

Dependencies/napi/napi-jsi/include/napi/napi.h
```diff
@@ -959,6 +959,7 @@ namespace Napi {
 
     bool operator ==(const Reference<T> &other) const;
     bool operator !=(const Reference<T> &other) const;
+    operator bool() const;
```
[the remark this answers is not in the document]
chrisfromwork: missed this in first pass but broke in linker. I pushed an updated change that just returns !IsEmpty(). if there's a better file to add the implementation to let me know :)

Dependencies/napi/napi-jsi/CMakeLists.txt
chrisfromwork: ah that explains the headache :P

Core/Graphics/Source/Graphics.cpp
chrisfromwork: no, this seems like a good suggestion

Dependencies/napi/napi-jsi/include/napi/napi.h
```diff
@@ -959,6 +959,7 @@ namespace Napi {
 
     bool operator ==(const Reference<T> &other) const;
     bool operator !=(const Reference<T> &other) const;
+    operator bool() const { return !IsEmpty(); }
```
bghgary, line 962: Implementation should go in napi-inl.h

tymlipari, line 962: Should this be `explicit` (requiring a cast) or left as implicit? Not sure how this is intended to be used.

Core/Graphics/Include/Babylon/Graphics.h
```diff
@@ -18,6 +18,7 @@ namespace Babylon
 
         template<typename NativeWindowT>
         void UpdateWindow(NativeWindowT window);
+        void UpdateWindowType(void* type);
```
ryantrem, line 21: What is an example of what this type would be?

chrisfromwork: this type looks like its only supposed to be used on linux platforms but it has windows content hidden behind it as well:

https://github.com/bkaradzic/bgfx/blob/e2eb3d56e6c43db6c531194d72e344817e6c4fb0/src/dxgi.cpp#L457

Dependencies/napi/napi-jsi/include/napi/napi.h
ryantrem: I think the intention is for it to be implicit so you can do something like `if (myRef)`

bghgary: This mimics the behavior in napi-direct: https://github.com/BabylonJS/BabylonNative/blob/master/Dependencies/napi/napi-direct/include/napi/napi.h#L1112 so that you can do what Ryan said.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -262,19 +262,34 @@ namespace Babylon
...
+    {
+        m_impl->SetNativeWindow(windowPtr, nullptr);
+    }
+
+    template<>
+    void Graphics::UpdateWindow<void*, void*>(void* windowPtr, void* windowTypePtr)
+    {
+        m_impl->SetNativeWindow(windowPtr, windowTypePtr);
+    }
+
```
bghgary, line 276: Is the following still necessary? If so, does it need to change?

https://github.com/BabylonJS/BabylonNative/blob/03a5c2dafe96c0fe354fc3a462ffa86184762e4c/Core/Graphics/Source/Graphics.cpp#L26-L30

Core/Graphics/Source/Graphics.cpp
chrisfromwork: this should now be updated

Core/Graphics/Include/Babylon/Graphics.h
chrisfromwork: this has been updated to provide generics/require custom calls to specify window types

Core/Graphics/Source/Graphics.cpp
chrisfromwork: good catch, didn't see these before. We no longer need them

bghgary (approved):

ryantrem (approved):

PR #550, opened 2020-12-18
Fix potential crash in xr.mm

ryantrem:
We hold on to the main view (an `MTKView`) in xr.mm and use it, but we don't do anything to explicitly contribute to the lifetime of the object. This means it could be deleted at any time and we would still try to use it. With some refactoring I'm doing in Babylon React Native, the timing/ordering has changed slightly and I started seeing this issue. The fix here is to just explicitly retain/release the main view as long as we are intending to use it.


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -649,14 +651,18 @@ bool TryInitialize() {
...
             }
-            
+
             dispatch_sync(dispatch_get_main_queue(), ^{
                 // Check whether the main view has changed, and if so, reparent the xr view.
                 UIView* currentSuperview = [xrView superview];
                 UIView* desiredSuperview = getMainView();
                 if (currentSuperview != desiredSuperview) {
+                    [currentSuperview release];
                     [xrView removeFromSuperview];
```
Alex-MSFT, line 661: Does this ordering need to be flipped?

Alex-MSFT (approved):

Dependencies/xr/Source/ARKit/XR.mm
ryantrem: Yea that’d be safer.

ryantrem: Actually I don't think it matters. The issue I was facing was that the `superview` unexpectedly became `nil`. I think views only keep a weak reference to their superviews, and when a view is deleted, it automatically removes child views. Since calling `removeFromSuperview` is a no-op if there is no superview, then this shouldn't cause any problem.

PR #551, opened 2020-12-21
Pass frame buffer ownership to JS

Alex-MSFT:
This fix addresses https://github.com/BabylonJS/BabylonReactNative/issues/134

The issue was caused by a change in BabylonJS which properly implemented _releaseFrameBufferObjects to prevent us from leaking frame buffers in BabylonNative when cleaning up textures: https://github.com/BabylonJS/Babylon.js/pull/9410

For the XR Render texture we were keeping ownership in NativeXR, but passing the reference back to JavaScript, so on scene dispose we were double releasing the FrameBuffer causing a crash.

To fix this I'm adding a flag to mark whether the frame buffer is owned by JavaScript or the native engine, and checking that flag at deletion time to check where the clean up should occur.


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -488,7 +488,25 @@ namespace Babylon
...
+            if (!pair.second->OwnedByJS)
+            {
+                frameBuffersDestroyed = true;
+                delete pair.second;
+            }
+        }
+
+        // If any frame buffers were destroyed, we need to call bgfx::frame once to force clean up.
+        // This is likely broken for multi-threaded scenarios.
+        if (frameBuffersDestroyed)
```
bghgary, line 505: This should probably warrant a new issue or at least a TODO.

bghgary (approved):

Plugins/NativeXr/Source/NativeXr.cpp
Alex-MSFT: Created issue here: https://github.com/BabylonJS/BabylonNative/issues/555

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1258,8 +1258,12 @@ namespace Babylon
 
     void NativeEngine::DeleteTexture(const Napi::CallbackInfo& info)
     {
-        const auto texture = info[0].As<Napi::External<TextureData>>().Data();
-        delete texture;
+        // It is possible that we were called on an null/undefined texture if so no action required.
+        if (!info[0].IsNull() && !info[0].IsUndefined())
```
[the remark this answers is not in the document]
Alex-MSFT: Added a check in nativeEngine.h here: https://github.com/BabylonJS/Babylon.js/pull/9643/files

PR #557, opened 2021-01-04
converts R8 to RGB8 textures

### LINKED ISSUE #556: Grayscale JPEG files are shown in red when used as diffuse textures
`moon.js`:
```js
function createMoon(scene, texturePath, pos) {
    var moon = BABYLON.Mesh.CreateSphere("moon", 32, 0.3);
    moon.position.x = pos;
    var moonMat = new BABYLON.StandardMaterial("moonMaterial", scene);
    moonMat.diffuseTexture = new BABYLON.Texture(texturePath, scene);
    moonMat.specularColor = new BABYLON.Color3(0, 0, 0);
    moon.material = moonMat;
    return moon;
}

function createScene(engine) {
    var scene = new BABYLON.Scene(engine);
    scene.createDefaultCamera(true);
    scene.createDefaultLight(true);

    var moonGrayscale = createMoon(scene, "https://user-images.githubusercontent.com/1297574/103034701-2d14f780-456e-11eb-9485-c73581dd7c80.jpg", -0.2);
    var moonColor = createMoon(scene, "https://user-images.githubusercontent.com/1297574/103034707-30a87e80-456e-11eb-8644-88d12f0a6e8d.jpg", 0.2);

    BABYLON.Tools.Log("Scene created");
    return scene;
}
```
Grayscale texture:
moon

Same texture but saved as RGB:
moon-color

`./Playground ./moon.js`:
playground_screenshot

Reproduced on Linux and Android.

CedricGuillemet:
fixes #556 
Converts R8 textures to RGB8 using bimg conversion function

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1132,6 +1132,10 @@ namespace Babylon
             dataRefs[face] = Napi::Persistent(typedArray);
             tasks[face] = arcana::make_task(arcana::threadpool_scheduler, m_cancelSource, [this, dataSpan, generateMips]() {
                 bimg::ImageContainer* image = bimg::imageParse(&m_allocator, dataSpan.data(), static_cast<uint32_t>(dataSpan.size()));
+                // if texture is R8, it needs to be converted as luminance (r=g=b=luminance and alpha = 1)
+                // see what's done in loadTexture
+                // keeping an assert here until we find some assets to test.
+                assert(image->m_format != bimg::TextureFormat::R8);
                 if (generateMips)
                 {
                     GenerateMips(&m_allocator, &image);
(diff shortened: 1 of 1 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1071,6 +1071,16 @@ namespace Babylon
...
+                if (image->m_format == bimg::TextureFormat::R8)
+                {
+                    // Images with only 1 channel are interpreted as luminance texture with RGB containing the same value as R and alpha as 255
+                    // To emulate this behavior, the format is switched to alpha8 and converted to RGB8 when packing and unpacking take care of 
+                    // component swizzling.
+                    image->m_format = bimg::TextureFormat::A8;
+                    bimg::ImageContainer* rgba = bimg::imageConvert(&m_allocator, bimg::TextureFormat::RGB8, *image, false);
+                    bimg::imageFree(image);
+                    image = rgba;
+                }
```
bghgary, line 1083: Is this the only place this needs to happen? Can cube textures be grayscale?

CedricGuillemet: yes, I don't see anything preventing cube textures to use grayscale. With bjs, more checks are done as well on PoT sizes and consistent dimensions. I don't think our impl is as much robust. What's the tool used to generate .env?

bghgary: You don't need a `.env` file. Any cube texture will have the same issue right?

PR #558, opened 2021-01-04
Update BabylonNative to consume newer bgfx.cmake submodule commit

chrisfromwork:
We checked in some fixes to bgfx/bx to avoid consuming the windows runtime library in static libs. This change moves us to newer commits that support this fix.


CedricGuillemet:
bgfx and bgfx.cmake subrepos moved from @syntheticmagus 's repo to babylonjs ones:
https://github.com/BabylonJS/bgfx
https://github.com/BabylonJS/bgfx.cmake

chrisfromwork:
> bgfx and bgfx.cmake subrepos moved from @syntheticmagus 's repo to babylonjs ones:
> https://github.com/BabylonJS/bgfx
> https://github.com/BabylonJS/bgfx.cmake

@CedricGuillemet  does that mean that there are issues with this review?

bghgary:
I think @CedricGuillemet may be referring to the failed builds. GitHub should auto-redirect the repo due to the move, but we should update the links to the BabylonJS one when it's convenient. If you are not in a rush, this PR may be the right one to do it and perhaps it'll fix the build issue also?

chrisfromwork:
> I think @CedricGuillemet may be referring to the failed builds. GitHub should auto-redirect the repo due to the move, but we should update the links to the BabylonJS one when it's convenient. If you are not in a rush, this PR may be the right one to do it and perhaps it'll fix the build issue also?

Up for whatever, but what do these changes entail? Am I needing to update a .gitmodules file?

bghgary:
It requires an update to this file: https://github.com/BabylonJS/bgfx.cmake/blob/master/.gitmodules

chrisfromwork:
@CedricGuillemet  and @bghgary i've moved this to the newest bgfx.cmake commit that contains the .gitmodules fix. We'll see what happens with CI

chrisfromwork:
let's default to @Alex-MSFT's review open elsewhere. they were able to fix CI

PR #561, opened 2021-01-05
Update `app:///` scheme to be consistent across platforms

### LINKED ISSUE #319: Investigate using CMAKE_CFG_INTDIR to copy scripts for WIN32
See if https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/CMakeLists.txt#L175-L181 can be fixed using https://cmake.org/cmake/help/v3.17/variable/CMAKE_CFG_INTDIR.html

bghgary:
Fixes #319


bghgary:
Hmm, will have to update Linux also before merging.

CedricGuillemet (approved):
LGTM!

Issue #482, opened 2020-11-05
Support DPI


ryantrem:
Babylon.js has some support for scaling the display in different ways, based on the `adaptToDeviceRatio` `Engine` constructor parameter, `Engine.setHardwardScalingLevel`, and `window.devicePixelRatio`. I believe Babylon Native does not currently support any of this, and that currently we always specify view size in device pixels. I think that also means on mobile devices we are probably rendering at very high resolutions, perhaps higher than necessary with an associated resource cost. We should think about how to support this as it could have a huge impact on mobile performance.

Drigax:
Sounds good. I'll start looking into this

Drigax:
I added some rudimentary support for `NativeEngine.setHardwareScalingLevel` via this branch: https://github.com/Drigax/BabylonNative/tree/add-engine-hardware-scaling-level

I'll open a PR when I return

Drigax:
Looking at MSDN:
https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio

```
The devicePixelRatio of Window interface returns the ratio of the resolution in physical pixels to the resolution in CSS pixels for the current display device. This value could also be interpreted as the ratio of pixel sizes: the size of one CSS pixel to the size of one physical pixel. In simpler terms, this tells the browser how many of the screen's actual pixels should be used to draw a single CSS pixel.
```

This gives me the impression that window.devicePixelRatio can be implemented native-side as another scalar applied to calculate the resolution of the BGFX window. I'll most likely implment this similarily to how I'm currently implementing `Engine.hardwareScalingLevel`

ryantrem:
I think we can deduce `devicePixelRatio` on iOS/Android with:
iOS: [`UIScreen.mainScreen.nativeScale`](https://developer.apple.com/documentation/uikit/uiscreen/1617825-nativescale?language=objc)
Android: [`DisplayMetrics.density`](https://developer.android.com/reference/android/util/DisplayMetrics#density)
I'm not 100% sure this is right, so we'll need to test it out.

Drigax:
https://github.com/BabylonJS/BabylonNative/pull/582

PR #566, opened 2021-01-08
Native implementation of Engine.setHardwareScalingRatio

Drigax:
This PR implements and adds boilerplate to expose NativeEngine.getHardwareScalingRatio.

Render scaling is implemented via a scalar used to resize bgfx's rendering resolution.

Addresses #482


Core/Graphics/Source/Graphics.cpp
```diff
@@ -249,6 +252,25 @@ namespace Babylon
...
     }
 
+    float Graphics::Impl::GetHardwareScalingLevel()
+    {
+        return m_hardwareScalingLevel;
+    }
+
+    void Graphics::Impl::SetHardwareScalingLevel(float level)
+    {
+        if (level == 0)
```
Drigax, line 262: I currently don't like this solution, we should have some "approximately equal to" float comparitor instead.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -64,12 +64,15 @@ namespace Babylon
 
     void Graphics::Impl::Resize(size_t width, size_t height)
     {
-        std::scoped_lock lock{m_bgfxState.Mutex};
+        m_resolution.width = width;
```
Drigax, line 67: Using a lock here and changing this from JS via `OnBeforeRender` caused an exception. Though we should probably figure out why instead of removing the lock

Core/Graphics/Source/Graphics.cpp
```diff
@@ -64,12 +64,15 @@ namespace Babylon
...
+        m_resolution.height = height;
+
         m_bgfxState.Dirty = true;
 
         auto& res = m_bgfxState.InitState.resolution;
-        res.width = static_cast<uint32_t>(width);
-        res.height = static_cast<uint32_t>(height);
+        auto ratio = GetHardwareScalingLevel();
+        res.width = static_cast<uint32_t>(m_resolution.width / ratio);
+        res.height = static_cast<uint32_t>(m_resolution.height / ratio);
```
ryantrem, line 75: I would move all of this into a parameterless private `Resize` function and call that both here and from `SetHardwareScalingLevel`. Basically we are saying there are two different states that can affect final resolution: the view size, and the hardware scaling level. If either of those change, we need to do a `Resize` operation.

Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -61,6 +64,14 @@ namespace Babylon
             bool Dirty{};
         } m_bgfxState{};
 
+        struct
+        {
+            std::mutex Mutex{};
```
bghgary, line 69: where is this used?

Drigax: This struct reperesents our "target resolution" of the application. This mutex is used so we can retrieve and change this resolution in a thread-safe way.

Core/Graphics/Source/Graphics.cpp
Drigax: Re-added the lock.

Core/Graphics/Source/Graphics.cpp
Drigax: Good suggestion, that prevents the current awkward way we call resize using the current resolution when the scaling level is changed.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -249,6 +260,26 @@ namespace Babylon
...
 
+    float Graphics::Impl::GetHardwareScalingLevel()
+    {
+        std::scoped_lock lock(m_resolution.Mutex);
+        return m_resolution.hardwareScalingLevel;
+    }
+
+    void Graphics::Impl::SetHardwareScalingLevel(float level)
+    {
+        if (level == 0)
```
ryantrem, line 271: I guess we should check for <= 0 since it is an unsigned float?

Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -61,6 +65,15 @@ namespace Babylon
...
+            size_t width;
+            size_t height;
```
ryantrem, line 72:
```suggestion
            size_t width{};
            size_t height{};
```

Core/Graphics/Source/Graphics.cpp
CedricGuillemet: <= FLT_EPSILON . Never compare floats with literal values.

Drigax: Thanks, both these changes make sense.

bghgary: `std::numeric_limits<float>::epsilon()` if you want the C++ way.

ryantrem: Looks like `FLT_EPSILON` is not supported on some platforms anyway.

bghgary: It should be. Probably just need to include the header. I would recommend using the C++ one though.

Core/Graphics/Source/GraphicsImpl.h
bghgary: I don't see any references to this variable. Am I missing something?

bghgary: Oh, you added it just now.

Core/Graphics/Source/Graphics.cpp
Drigax: Ok, decided to go with the c++ style for this, thanks for the suggestions.

bghgary: Include <limits>

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1530,4 +1532,13 @@ namespace Babylon
             function();
         });
     }
+
+    void NativeEngine::SetHardwareScalingLevel(const Napi::CallbackInfo& info){
```
syntheticmagus, line 1536: nit: Style. I forget how to run it in VS, but I think the clang formatter should be able to format this for you.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -64,12 +64,23 @@ namespace Babylon
 
     void Graphics::Impl::Resize(size_t width, size_t height)
     {
-        std::scoped_lock lock{m_bgfxState.Mutex};
-        m_bgfxState.Dirty = true;
+        {
+            std::scoped_lock lock(m_resolution.Mutex);
```
syntheticmagus, line 68: Is the reason `m_resolution` has its own mutex because it doesn't seem like a natural part of `m_bgfxState`? If so, it might be better to rename `m_bgfxState` to just `m_internalState`, or
```
struct
{
    std::mutex Mutex{};

    struct
    {
        ...
    } Resolution{};

    struct
    {
        ...
    } Bgfx{};
} m_state{};
```
or something like that, to avoid having to unlock and lock two different mutexes that are both trying to guard two very closely related parts of the Graphics object's state.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -64,12 +64,23 @@ namespace Babylon
...
+            m_resolution.width = width;
+            m_resolution.height = height;
+        }
+        Resize();
+    }
 
+    void Graphics::Impl::Resize()
```
syntheticmagus, line 75: This name confuses me a bit, I think because it's a private helper method overloading a public contract method. I think this method might disappear/change substantially in refactors related to other comments, but if it doesn't, it might be clearer to name it `UpdateBgfxResolution`, or something like that, to clarify its distinction as just an internal helper.

Core/Graphics/Source/Graphics.cpp
Drigax: Added the include.

Core/Graphics/Source/Graphics.cpp
Drigax: That was my reasoning for the most part, this isn't quite appropriate to call this the state of BGFX, but some internal state that we keep and use to set BGFX's state. I think locking all of this behind a single mutex isn't a bad change, especially considering how a change to one affects the other.

Core/Graphics/Source/Graphics.cpp
Drigax: makes sense, functions shouldn't be overloaded without good reason.

Core/Graphics/Source/Graphics.cpp
Drigax: refactored the structs to a single struct following your suggestion, definitely makes the locking a bit more understandable.

Plugins/NativeEngine/Source/NativeEngine.cpp
Drigax: oopsie. beautify in vsCode seems to follow the clang formatting style :)

Core/Graphics/Source/Graphics.cpp
```diff
@@ -64,12 +65,22 @@ namespace Babylon
...
     void Graphics::Impl::Resize(size_t width, size_t height)
     {
-        std::scoped_lock lock{m_bgfxState.Mutex};
-        m_bgfxState.Dirty = true;
+        {
+            std::scoped_lock lock(m_state.Mutex);
+            m_state.Resolution.width = width;
+            m_state.Resolution.height = height;
+        }
+        UpdateBgfxResolution();
```
syntheticmagus, line 73: nit: In both places where this is called, the state mutex is locked immediately before calling it, then that lock is abandoned, then it's locked again inside the call. If the contract for this method became that it's unsafe (maybe by putting it on the struct and/or calling it `UnsafeUpdate...()`?) then we could just move the call inside the scope of the outer lock and not need to lock, unlock, and relock. Honestly, I'm not sure about this one; the two locks feel unnecessary, but making a helper method which exists exclusively to modify mutex-guarded state, then making that helper method explicitly unguarded, also feels kind of weird. Do we know if it's meaningfully worse perf-wise to just leave in the two locks?

Core/Graphics/Source/Graphics.cpp
```diff
@@ -354,4 +385,12 @@ namespace Babylon
     {
         m_impl->SetDiagnosticOutput(std::move(outputFunction));
     }
+
+    void Graphics::SetHardwareScalingLevel(float level){
```
syntheticmagus, line 389: nit: Style.

syntheticmagus (approved):
Other than two minor considerations, looks good to me.

Core/Graphics/Source/Graphics.cpp
ryantrem: I would just replace mutex with recursive_mutex which allows for re-entrancy.

ryantrem: You could also use lock_guard instead of scoped_lock here. scoped_lock is just a helper for locking on multiple mutexes in an raii fashion, but we’re only locking on one mutex.

Drigax: good suggestion, that definitely removes the need for awkward block scoping to prevent re-locking the same mutex in the caller.

I'm going to keep with `scoped_lock`, I'm not sure if there's anything to gain in using the older lock implementation, I was under the impression that this supplanted `lock_guard`

Issue #562, opened 2020-07-21
BabylonNative does not provide a mechanism to get access to its frame buffer


hendryems:
What's the mechanism for an React Native app to be able to get the frame/(back) buffer for the purposes of streaming out the contents of EngineView in real time?

dylanleatham:
Tech Assist has a need for this feature to support our calling and video recording scenarios. For our asset capture feature, we want to access the video frame buffer (with mixed reality annotations) and combine that with captured audio data to produce a video file of the scenario. In a call, we will be streaming the video buffer through IC3 to other call participants as well as capturing it to produce a video file.

bghgary:
Fixed in #180

PR #571, opened 2021-01-14
Draw AR camera to babylon render texture

ryantrem:
This change is the first part of the solution to #562.

Prior to this change, we would clear the render texture, have Babylon render to it, then in a custom shader compose the Babylon render texture on top of the camera texture. The problem with this is that BGFX never sees the camera texture, and so there is no way from the context of Babylon Native to get an rgba byte buffer of what is being displayed to the screen in XR mode.

With this change, for both ARCore and ARKit, we now draw the camera texture onto the render texture that Babylon draws to, and then we just present that final texture to the screen. This way the render texture that Babylon Native (and therefore BGFX) is dealing with has the camera texture.

I did a little bit of white space and uniform initialization cleanup as well, so the change looks bigger than it really is. Sorry!

### DIFF
--- a/Dependencies/xr/Source/ARCore/XR.cpp
+++ b/Dependencies/xr/Source/ARCore/XR.cpp
@@ -229,20 +229,13 @@ namespace xr
                 return gsl::finally([blendFuncSFactor, previousBlendFuncTFactor]() { glBlendFunc(blendFuncSFactor, static_cast<GLenum>(previousBlendFuncTFactor)); });
             }
 
-//            auto ClearColor(GLfloat red, GLfloat green, GLfloat blue, GLfloat alpha)
-//            {
-//                GLfloat previousClearColor[4];
-//                glGetFloatv(GL_COLOR_CLEAR_VALUE, previousClearColor);
-//                glClearColor(red, green, blue, alpha);
-//                return gsl::finally([red = previousClearColor[0], green = previousClearColor[1], blue = previousClearColor[2], alpha = previousClearColor[3]]() { glClearColor(red, green, blue, alpha); });
-//            }
-
-            auto Sampler(int unit)
+            auto BindSampler(GLenum unit, GLuint id)
             {
-                glActiveTexture(GL_TEXTURE0 + unit);
-                GLint previousSampler;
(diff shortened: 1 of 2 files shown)


dylanleatham (approved):

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -542,6 +536,43 @@ namespace xr
 
             ArCamera_release(camera);
 
+            // Suppress rendering if the camera did not produce the first frame yet.
+            // This is to avoid drawing possible leftover data from previous sessions if
+            // the texture is reused.
+            int64_t frameTimestamp{};
```
Alex-MSFT, line 542: I think maybe it makes sense to have this comment instead say something like: 
"Draw the camera texture to the Babylon render texture, suppress rendering if the camera did not produce the first frame yet."

That way you can easily tell at a glance what this chunk of code is responsible for before getting into the scope, since this chunk and the below chunk have the same comment it doesn't become apparent until you start reading the code inside the if statement.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -758,82 +798,51 @@ Size GetWidthAndHeightForViewIndex(size_t) const {
...
-                @try {
-                    if(renderPassDescriptor != nil) {
-                        renderPassDescriptor.colorAttachments[0].texture = drawable.texture;
-                        renderPassDescriptor.colorAttachments[0].loadAction = MTLLoadActionClear;
-                        renderPassDescriptor.colorAttachments[0].clearColor = MTLClearColorMake(0.0,0.0,0.0,1.0);
-                        
-                        // Create a render command encoder.
-                        id<MTLRenderCommandEncoder> renderEncoder = [commandBuffer renderCommandEncoderWithDescriptor:renderPassDescriptor];
-                        renderEncoder.label = @"XRDisplayEncoder";
+                if(renderPassDescriptor != nil) {
```
Alex-MSFT, line 811:
```suggestion
                if (renderPassDescriptor != nil) {
```

Alex-MSFT (approved):

Alex-MSFT (approved):

PR #572, opened 2021-01-15
depth 1 for subrepo clones

CedricGuillemet:


CedricGuillemet:
Not significant enough build time difference. Closing.

PR #578, opened 2021-01-20
NativeCapture first implementation

syntheticmagus:
The first implementation of an integration of bgfx's frame capture mechanism via a few new methods on the `Graphics` object and a new core component `NativeCapture`. This works, but I'd especially like feedback on a few specific points:

- Having an entirely separate component and type to handle captures feels a bit questionable, but at least in its current form it's doing something very native-specific.
- Having the callbacks just receive the raw data from bgfx as a `Napi::External` also feels a bit weird, but it's unclear to me that a component like this should have the authority to manipulate that data before passing it along.
- If you make two of these `NativeCapture`s at the same time and try to give them different lifespans, they will interfere with each other (the actual capture will stop as soon as the first one dies). I think they shouldn't be constructible at all, so that will likely be factored into the next iteration of how this is exposed (which I guess will emerge from discussion on this PR).

Thanks for taking a look!


Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,81 @@
...
+            auto listener = info[0].As<Napi::Function>();
+            m_callbacks.push_back(Napi::Persistent(listener));
+        }
+
+        void CaptureDataReceived(const BgfxCallback::CaptureData& data)
+        {
+            std::vector<uint8_t> bytes{};
+            bytes.resize(data.DataSize);
+            std::memcpy(bytes.data(), data.Data, data.DataSize);
+            m_runtime.Dispatch([this, data = data, bytes = std::move(bytes)](Napi::Env env) mutable {
```
bghgary, line 51:
```suggestion
            m_runtime.Dispatch([this, data, bytes{std::move(bytes)}](Napi::Env env) mutable {
```

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+                auto external = Napi::External<BgfxCallback::CaptureData>::New(env, &data);
+                for (const auto& callback : m_callbacks)
+                {
+                    callback.Call({external});
+                }
+            });
+        }
+
+        void Dispose(const Napi::CallbackInfo&)
```
bghgary, line 62: What happens if dispose isn't called? Should we call dispose in the destructor automatically?

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+        static void Initialize(Napi::Env env)
+        {
+            Napi::HandleScope scope{env};
+
+            Napi::Function func = NativeCapture::DefineClass(
+                env,
+                JS_CLASS_NAME,
+                {
+                    NativeCapture::InstanceMethod("addOnCaptureCallback", &NativeCapture::AddOnCaptureCallback),
```
bghgary, line 23: nit: AddOnCaptureCallback or AddCaptureCallback?

Core/Graphics/Source/BgfxCallback.cpp
```diff
@@ -123,15 +125,28 @@ namespace Babylon
...
+    void BgfxCallback::captureFrame(const void* data, uint32_t size)
     {
+        m_captureData.Data = data;
+        m_captureData.DataSize = size;
+
+        std::scoped_lock lock{m_captureCallbacksMutex};
+        for (const auto& callback : m_captureCallbacks)
+        {
+            callback(m_captureData);
+        }
```
bghgary, line 150: Might be good to reset m_captureData to defaults here to avoid confusion. Maybe do it in captureEnd?

bghgary: Maybe don't store m_captureData at all and just store each individual width, height, pitch, etc. instead.

bghgary (commented):
I'm not sure I like the contract. Also, is there an example usage somewhere?

Plugins/NativeCapture/Source/NativeCapture.cpp
ryantrem: Is your nit about casing? I'm confused because the current casing looks right for JS, so maybe I'm missing something.

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,81 @@
...
+            m_callbacks.push_back(Napi::Persistent(listener));
+        }
+
+        void CaptureDataReceived(const BgfxCallback::CaptureData& data)
+        {
+            std::vector<uint8_t> bytes{};
+            bytes.resize(data.DataSize);
+            std::memcpy(bytes.data(), data.Data, data.DataSize);
+            m_runtime.Dispatch([this, data = data, bytes = std::move(bytes)](Napi::Env env) mutable {
+                data.Data = bytes.data();
```
ryantrem, line 52: Why is this here? It looks like it is assigning the data source to the copy that was made.

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+                auto external = Napi::External<BgfxCallback::CaptureData>::New(env, &data);
+                for (const auto& callback : m_callbacks)
+                {
+                    callback.Call({external});
```
ryantrem, line 57: Isn't this basically passing an opaque object to the JS side? We need the JS side to be able to read the actual bytes (or pass it along to other native code that doesn't know anything about Napi or Babylon Native). I was expecting we'd pass something like a `Uint8Array` and a width/stride.

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,81 @@
...
+    private:
+        void AddOnCaptureCallback(const Napi::CallbackInfo& info)
+        {
+            auto listener = info[0].As<Napi::Function>();
+            m_callbacks.push_back(Napi::Persistent(listener));
+        }
+
+        void CaptureDataReceived(const BgfxCallback::CaptureData& data)
+        {
+            std::vector<uint8_t> bytes{};
```
ryantrem, line 48: Related to my comment below, but rather than allocating a vector each frame, can we just allocate a JS `Uint8Array` early on that we hold onto for the lifetime of this object, and just copy the source data into it each frame?

ryantrem: Maybe only recreate it if/when the data size changes, which in practice will probably be infrequent.

ryantrem (commented):
Overall, I actually like having a separate `NativeCapture` that can be instantiated from JS, and where the capture session has the lifetime of this object. This makes it easy for consumers to use this directly now, and later we find a good way to abstract this at the Babylon.js level. I think this is a better approach than what I had previously been imagining (mashing some new functions into `NativeEngine` and having people use the private `_native` directly like the validation tests do for screenshots).

My main feedback is about the `Napi::External` that is passed to the JS callback (see my comments in the code).

Plugins/NativeCapture/Source/NativeCapture.cpp
bghgary: No, it's whether the "On" part should be there or not.

bghgary:
Would it be good to also move NativeEngine::GetFrameBufferData to this class?

Plugins/NativeCapture/Source/NativeCapture.cpp
syntheticmagus: I guess we have to. This really shouldn't be allowed -- the purpose of having dispose is to ensure that a capture's lifetime is controlled by program logic as opposed to the whims of the garbage collector -- but we may not have a choice because the JS runtime can theoretically be torn down while the graphics object survives, so I suppose we need a fallback in the destructor to avoid potentially leaving graphics in a bad state.

Core/Graphics/Source/BgfxCallback.cpp
syntheticmagus: I'm not sure I understand. Can you elaborate on what confusion these actions might help to avoid?

Plugins/NativeCapture/Source/NativeCapture.cpp
syntheticmagus: Spoke offline. This is completing the deep copy begun in the lambda capture on line 51, which will have successfully copied most of the `data` object but will have kept the raw pointer to the original memory, which may not exist at the time this callback is executed, so we have to use the copy.

Plugins/NativeCapture/Source/NativeCapture.cpp
syntheticmagus: Spoke offline about this. It is passing an opaque object with the current expectation that that object would be interpreted by other native objects for maximum efficiency. We also discussed a mechanism to "un-opaque" this object if JS wants to read it directly; is that something we currently have a use case for such that we should put it in right now, or should we add that later if/as it becomes necessary?

Plugins/NativeCapture/Source/NativeCapture.cpp
syntheticmagus: Spoke offline. Unfortunately because the callback here is dispatched to the JS thread (starting from the graphics thread), we need to copy it to a memory location that will be long-lived enough to make it onto the JS thread even if the JS runtime is a bit behind; and to avoid accidentally paving over the last frame's data in case we receive two graphics thread callbacks before processing one on the JS side, we cannot assume it's safe to reuse the same buffer. If we want to avoid allocating these vectors every single frame, we could probably make a resource pool to create and recycle them as needed. That feels like a performance optimization, though, so it might be better to tackle that in a separate PR improving on this one once we start using it.

syntheticmagus:
Regarding moving `GetFrameBufferData` to this (or at least to be more like this), I think that might make sense, but I haven't looked at it too deeply. It feels like a separate PR, though.

Plugins/NativeCapture/Source/NativeCapture.cpp
ryantrem: The JS does not need a readable but array, but it needs a pointer that can be passed to C++ code (via JSI).

ryantrem:
> Regarding moving `GetFrameBufferData` to this (or at least to be more like this), I think that might make sense, but I haven't looked at it too deeply. It feels like a separate PR, though.

Yea it could be, it just might mean changing the contract, but that’s fine.

Plugins/NativeCapture/Source/NativeCapture.cpp
syntheticmagus: In the latest iteration, I modified the callback to take a JS object that automatically exposes the callback data, including an ArrayBuffer for the bytes. This ArrayBuffer is made using N-API's external ArrayBuffer mechanism, which does not take ownership of the memory. This implementation balances what I recall to be the results from our conversation about this a few weeks ago, but it definitely has some tradeoffs that I'll be happy to discuss and revise. (For example, I'm not a huge fan of `new`ing a N-API object for every callback, but since caching had pave-over issues I'm not sure if another pattern would be better. It could certainly be more efficient if I just pass up a raw pointer, but that consideration must be balanced against safety/clarity/ease of use for the relevant scenario, so I'm interested in perspectives on that.)

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,117 @@
...
+                {
+                    case bgfx::TextureFormat::BGRA8:
+                        jsData.Set(FORMAT_MEMBER_NAME, "BGRA8");
+                        break;
+                    default:
+                        jsData.Set(FORMAT_MEMBER_NAME, env.Undefined());
+                        break;
+                }
+                jsData.Set("yFlip", data.YFlip);
+                jsData.Set("data", Napi::ArrayBuffer::New(env, bytes.data(), data.DataSize));
```
ryantrem, line 80: Talked offline and I think we agreed to re-use a `Napi::ArrayBuffer` instead of allocating a new one when possible (maybe we need to create a new one if the `DataSize` changes?).

Core/Graphics/Source/BgfxCallback.h
```diff
@@ -1,16 +1,32 @@
...
+
+#include <arcana/containers/ticketed_collection.h>
+
 #include <bgfx/bgfx.h>
 #include <bgfx/platform.h>
-#include <napi/napi.h>
+
+#include <mutex>
 #include <queue>
+#include <vector>
```
ryantrem, line 12: It looks like we're not using this in the end? Can we remove it?

ryantrem (approved):

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -0,0 +1,126 @@
...
+                }
+                jsData.Set("yFlip", data.YFlip);
+
+                auto jsBytes = jsData.Get("data").As<Napi::ArrayBuffer>();
+                if (data.DataSize != jsBytes.ByteLength())
+                {
+                    jsBytes = Napi::ArrayBuffer::New(env, data.DataSize);
+                    jsData.Set("data", jsBytes);
+                }
+                std::memcpy(jsBytes.Data(), bytes.data(), data.DataSize);
```
ryantrem, line 90: Can release the `bytes` memory at this point, or do we need it for the external? If we need it for the external and it is adding extra memory, maybe we should remove it for now?

PR #580, opened 2021-01-22
discard texture bindings because of mismatch in shaders

CedricGuillemet:
Reported failing Nightly Build.
Texture binding was not discarded and subsequent drawcall shaders failed with d3d11.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1421,7 +1421,7 @@ namespace Babylon
             bgfx::setState(m_engineState | fillModeState);
         }
 
-        bgfx::submit(m_frameBufferManager.GetBound().ViewId, m_currentProgram->Program, 0, BGFX_DISCARD_INSTANCE_DATA | BGFX_DISCARD_STATE | BGFX_DISCARD_TRANSFORM);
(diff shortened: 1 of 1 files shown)


CedricGuillemet:
It looks like it will be harder to fix that anticipated. I exposed the issue to Alexis to see how babylonjs commits can cause the issue.
The crash reported by d3d11 when running UnindexedMesh PG is:
```
D3D11 ERROR: ID3D11DeviceContext::DrawIndexed: The Shader Resource View dimension declared in the shader code (TEXTURE2D) does not match the view type bound to slot 0 of the Pixel Shader unit (TEXTURECUBE).  This mismatch is invalid if the shader actually uses the view (e.g. it is not skipped due to shader code branching). [ EXECUTION ERROR #354: DEVICE_DRAW_VIEW_DIMENSION_MISMATCH]
D3D11: **BREAK** enabled for the previous message, which was: [ ERROR EXECUTION #354: DEVICE_DRAW_VIEW_DIMENSION_MISMATCH ]
```
The Fresnel PG must be run before UnindexedMesh.
It seems like Fresnel binds some TextureCubes that are not unbound or replaced by new 2D textures. When rendering UnindexedMesh, the shader is expecting a different texture type and crashes.

To repro, you have to use the same Preprocessor as the nightly when building solution: -DBGFX_CONFIG_MEMORY_TRACKING=ON -DBGFX_CONFIG_DEBUG=ON

I tried : 
- discard bindings after bgfx::submit but subsequent drawcalls that share the same state will fail
- bind invalidHandle to each texture slot used by the shader when setProgram is called
- call bgfx::discard when binding a new program

CedricGuillemet:
Closing this PR and opening an issue instead

PR #582, opened 2021-01-22
Add platform polyfills for retrieving window.deviceScalingRatio

### LINKED ISSUE #482: Support DPI

Drigax:
This PR modifies the Window polyfill contract to retain a reference to the native window implementation, and creates platform implementations of retreiving the DPI scaling of that window.

Dependent on Babylon.js pull request: https://github.com/BabylonJS/Babylon.js/pull/9892 in order to enable auto-scaling on engine creation.

Fixes #482 #620


Polyfills/Window/CMakeLists.txt
```diff
@@ -3,16 +3,31 @@ set(SOURCES
     "Source/Window.h"
     "Source/Window.cpp")
 
+if(WIN32)
```
CedricGuillemet, line 6: For platform specific files, the practice is to use the platform in the file name https://github.com/BabylonJS/BabylonNative/blob/ab4991d247ca7d74923c6924fe4258f67f8bb2df/Plugins/NativeEngine/CMakeLists.txt#L25

Core/Graphics/Source/Graphics.cpp
```diff
@@ -64,12 +65,20 @@ namespace Babylon
 
     void Graphics::Impl::Resize(size_t width, size_t height)
     {
-        std::scoped_lock lock{m_bgfxState.Mutex};
-        m_bgfxState.Dirty = true;
+        std::scoped_lock lock(m_state.Mutex);
```
ryantrem, line 68:
```suggestion
        std::scoped_lock lock{m_state.Mutex};
```

Core/Graphics/Source/Graphics.cpp
```diff
@@ -64,12 +65,20 @@ namespace Babylon
...
+        m_state.Resolution.height = height;
+        UpdateBgfxResolution();
+    }
 
-        auto& res = m_bgfxState.InitState.resolution;
-        res.width = static_cast<uint32_t>(width);
-        res.height = static_cast<uint32_t>(height);
+    void Graphics::Impl::UpdateBgfxResolution()
+    {
+        std::scoped_lock lock(m_state.Mutex);
```
ryantrem, line 76:
```suggestion
        std::scoped_lock lock{m_state.Mutex};
```

Core/Graphics/Source/Graphics.cpp
```diff
@@ -249,6 +258,26 @@ namespace Babylon
         Callback.SetDiagnosticOutput(std::move(outputFunction));
     }
 
+    float Graphics::Impl::GetHardwareScalingLevel()
+    {
+        std::scoped_lock lock(m_state.Mutex);
```
ryantrem, line 263:
```suggestion
        std::scoped_lock lock{m_state.Mutex};
```

Core/Graphics/Source/Graphics.cpp
```diff
@@ -249,6 +258,26 @@ namespace Babylon
...
+    }
+
+    void Graphics::Impl::SetHardwareScalingLevel(float level)
+    {
+        if (level <= std::numeric_limits<float>::epsilon())
+        {
+            throw std::runtime_error{"HardwareScalingValue cannot be less than or equal to 0."};
+        }
+        {
+            std::scoped_lock lock(m_state.Mutex);
```
ryantrem, line 274:
```suggestion
            std::scoped_lock lock{m_state.Mutex};
```

Polyfills/Window/Source/Window.cpp
```diff
@@ -10,18 +11,23 @@ namespace Babylon::Polyfills::Internal
...
         constexpr auto JS_SET_TIMEOUT_NAME = "setTimeout";
         constexpr auto JS_A_TO_B_NAME = "atob";
+
         constexpr auto JS_ADD_EVENT_LISTENER_NAME = "addEventListener";
         constexpr auto JS_REMOVE_EVENT_LISTENER_NAME = "removeEventListener";
+        auto JS_DEVICE_PIXEL_RATIO_NAME = "devicePixelRatio";
     }
 
-    void Window::Initialize(Napi::Env env)
+    void Window::Initialize(Napi::Env env, void* windowPtr)
```
ryantrem, line 20: We shouldn't need to pass the windowPtr to `Window::Intialize` because `Window` can just get it from `Graphics::Impl::GetNativeWindow()` right?

Polyfills/Window/Source/Windows/WindowPlatform.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <Windows.h>
+#include <Babylon/JsRuntime.h>
+
+#include "..\Window.h"
```
syntheticmagus, line 6: You should be able to do this just as `<Window.h>`; CMake should have set up the include directories so that you can reference header without having to depend directly on the directory relationship to it.

Apps/Playground/Scripts/experience.js
```diff
@@ -53,7 +53,7 @@ function CreateInputHandling(scene) {
 
 _native.whenGraphicsReady().then(function () {
 
-    var engine = new BABYLON.NativeEngine();
+    var engine = new BABYLON.NativeEngine(true);
```
ryantrem, line 56: What is this extra param? When I look at `NativeEngine` in Babylon.js I don't see a constructor that takes a parameter.

Polyfills/Window/Source/Window.cpp
```diff
@@ -10,18 +11,23 @@ namespace Babylon::Polyfills::Internal
         constexpr auto JS_CLASS_NAME = "Window";
         constexpr auto JS_SET_TIMEOUT_NAME = "setTimeout";
         constexpr auto JS_A_TO_B_NAME = "atob";
+
         constexpr auto JS_ADD_EVENT_LISTENER_NAME = "addEventListener";
         constexpr auto JS_REMOVE_EVENT_LISTENER_NAME = "removeEventListener";
+        auto JS_DEVICE_PIXEL_RATIO_NAME = "devicePixelRatio";
```
ryantrem, line 17:
```suggestion
        constexpr auto JS_DEVICE_PIXEL_RATIO_NAME = "devicePixelRatio";
```

Polyfills/Window/Source/Window.cpp
```diff
@@ -10,18 +11,23 @@ namespace Babylon::Polyfills::Internal
...
     {
         Napi::HandleScope scope{env};
 
-        Napi::Function constructor = DefineClass(
+        Napi::Function constructor = Window::DefineClass(
             env,
             JS_CLASS_NAME,
-            {});
+            {
+                StaticAccessor(JS_DEVICE_PIXEL_RATIO_NAME, &Window::GetDevicePixelRatio, &Window::SetDevicePixelRatio)
```
ryantrem, line 28: Seems like this should be an `InstanceAccessor`. `GetDevicePixelRatio` is on the global `window` instance, not on the prototype.

ryantrem, line 28: As far as I can tell, `SetDevicePixelRatio` is not an API on `window`. Seems like this should be removed.

Polyfills/Window/Source/Window.cpp
Drigax: Yeah, its legal for our polyfills to take dependencies on the core components, so I can make that change, and remove the member variable from the window class.

Polyfills/Window/Source/Window.cpp
ryantrem: And by removed, I just mean pass in `nullptr` for the mutator function (which is what 99% of our `InstanceAccessor` use cases do).

Apps/Playground/Scripts/experience.js
Drigax: I will be adding an additional parameter to the NativeEngine constructor for an `adaptToDeviceRatio` flag, similar to the parameter used for the ThinEngine.

Polyfills/Window/Source/Window.cpp
Drigax: Thanks,  I must have changed this trying to fix my JS wrapper.

Polyfills/Window/Source/Window.cpp
Drigax: I'll try that but I don't believe this will work either.

Drigax: Its not. I'm still addressing some issues with the JS wrapper not properly working, and all of this code is subject to change. Currently the PR is not marked as ready for review, but I appreciate the feedback.

Polyfills/Window/Source/Windows/WindowPlatform.h
```diff
@@ -0,0 +1,21 @@
...
+
+using WindowType = HWND;
+
+namespace Babylon::Polyfills::Internal
+{
+    Napi::Value Window::GetDevicePixelRatio(const Napi::CallbackInfo& info)
```
ryantrem, line 12: Why is this in a header file instead of a cpp file? I feel like what I would have expected is:

```
Window\Source\Window.h
Window\Source\Window.cpp
Window\Source\Windows\Window.cpp
Window\Source\iOS\Window.cpp
Window\Source\Android\Window.cpp
etc.
```

But I'm not sure if there are different existing patterns in Babylon Native for this.

Drigax: I originally modeled this platform specific implementation off of:
https://github.com/BabylonJS/BabylonNative/blob/master/Dependencies/xr/Source/OpenXR/Windows/XrPlatform.h

But if that's not a great pattern, I'm definitely open to change it to match your pattern as well.

Drigax: Also according to @CedricGuillemet's comment above:https://github.com/BabylonJS/BabylonNative/pull/582#discussion_r562749856


We have 3 different patterns, It may be a good idea to align to a single pattern here.

Polyfills/Window/Source/Window.cpp
ryantrem: Ah ok, I assumed if a PR was published (even in draft) you were looking for feedback. Happy to wait until it is out of draft if that's the process folks tend to use!

Polyfills/Window/CMakeLists.txt
Drigax: Hey @CedricGuillemet, what are your thoughts on using something along the lines of:

"Polyfills/Window/Window${PLATFORM}.cpp"
vs
"Polyfills/Window/${PLATFORM}/Window.cpp" for the convention?

Polyfills/Window/CMakeLists.txt
```diff
@@ -3,16 +3,31 @@ set(SOURCES
...
+endif()
+
 target_link_to_dependencies(Window 
     PUBLIC napi
     PRIVATE base-n
     PRIVATE JsRuntime)
 
+if(WIN32)
+    target_link_to_dependencies(Window
+        PRIVATE "shlwapi.lib")
```
chrisfromwork, line 28: >shlwapi [](start = 17, length = 7)

Are we testing any uwp flavors? It looks currently like this has only been validated against win32

Apps/Playground/Win32/App.cpp
```diff
@@ -109,7 +109,7 @@ namespace
                 OutputDebugStringA(message);
             });
 
-            Babylon::Polyfills::Window::Initialize(env);
+            Babylon::Polyfills::Window::Initialize(env, (void*)&hWnd);
```
chrisfromwork, line 112: >hWnd [](start = 64, length = 4)

how does this work when hand in a pointer to a swapchainpanel comapred to a window?

[https://github.com/BabylonJS/BabylonReactNative/blob/95ecb8305edbe58f73f3a49e743d8b7ea428f8c5/Modules/%40babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp#L35](https://github.com/BabylonJS/BabylonReactNative/blob/95ecb8305edbe58f73f3a49e743d8b7ea428f8c5/Modules/%40babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp#L35)

Polyfills/Window/Source/Windows/WindowPlatform.h
chrisfromwork: id personally prefer to cpp and header files for this one

---
In reply to: [563957901](https://github.com/BabylonJS/BabylonNative/pull/582#discussion_r563957901) [](ancestors = 563957901)

Polyfills/Window/Source/Windows/WindowPlatform.h
```diff
@@ -0,0 +1,21 @@
...
+    {
+        auto& window = *static_cast<Window*>(info.Data());
+
+        UINT dpi = GetDpiForWindow(*((HWND*)(window.m_windowPtr)));
+        return Napi::Value::From(info.Env(), (float)dpi / 96.0f);
```
chrisfromwork, line 17: For BabylonReactNative windowPtr is the IUknown/abi pointer for a SwapChainPanel not a hwnd. I believe this cast will generate bad memory/an eventual access violation

Polyfills/Window/CMakeLists.txt
```diff
@@ -3,16 +3,31 @@ set(SOURCES
...
+    target_include_directories(Window 
+        PRIVATE "Source/Windows")
+endif()
+
 target_link_to_dependencies(Window 
     PUBLIC napi
     PRIVATE base-n
     PRIVATE JsRuntime)
 
+if(WIN32)
```
chrisfromwork, line 26: >WIN32 [](start = 3, length = 5)

iirc win32 evaluates to defined for cmake generated uwp solutions. I am not familiar with the below lib that is introduced by this change, but if it causes us to consume the windows runtime library it will break BabylonReactNative compilation. Do we know if shlwapi.lib is uwp friendly?

Polyfills/Window/CMakeLists.txt
chrisfromwork: What would ${Platform} evaluate to here?

---
In reply to: [565474829](https://github.com/BabylonJS/BabylonNative/pull/582#discussion_r565474829) [](ancestors = 565474829)

ryantrem: I think the second one is the convention we already have elsewhere. For example: https://github.com/BabylonJS/BabylonNative/tree/master/Dependencies/UrlLib/Source

Polyfills/Window/Source/Windows/WindowPlatform.h
Drigax: Hmm, that definitely complicates things for supporting Win32 and UWP. I suppose i should have separate implementations for both, enforced with some preprocessor directive? are there any preprocessor variables for UWP?

tymlipari: I'm not aware of a specific C++ macro for that, but you could have different implementation files for Win32 and UWP include them separately in the CMake configuration.

That said, in the case @chrisfromwork mentioned, we probably need the ability to specify the scale factor externally from the BabylonReactNative component. For UWP where we're backed by a `SwapChainPanel`, we'd presumably want to use `SwapChainPanel.CompositionScaleX` and `SwapChainPanel.CompositionScaleY` as the scaling ratio to correctly integrate with XAML's rendering.

Apps/Playground/Scripts/experience.js
tymlipari: A bool is kind of confusing here. Would it make sense to take a configuration object instead so you could do something like:

`var engine = new BABYLON.NativeEngine({ adaptToDeviceRatio: true });`

Polyfills/Window/Source/Windows/WindowPlatform.h
chrisfromwork: this isn't strictly a uwp vs win32 distinction. BabylonNative UWP flavors use a core window pointer for the window ptr:
https://github.com/BabylonJS/BabylonNative/blob/93005ea71a30f489eac3ad1d98878ac50532b51a/Apps/Playground/UWP/App.cpp#L146

BabylonReactNative UWP flavors use a swap chain panel for the window ptr:
https://github.com/BabylonJS/BabylonReactNative/blob/95ecb8305edbe58f73f3a49e743d8b7ea428f8c5/Modules/%40babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp#L35

it might be more practical to hand in a std::function that returns the scale factor needed compared to trying to determine what to do behind the scenes with the void* window ptr.

ryantrem: If I recall correctly, the idea behind having the `windowPtr` be templated was that in the future we'd have template specializations per platform that would make it more type safe (rather than just being `void*`). If we actually implemented that now, then it seems like the template specializations for each type (including `ICoreWindow` vs `SwapChainPanel`) can take care of ensuring we are doing the right thing to get the pixel scaling value. @syntheticmagus, @bghgary - any comments on the window type template specialization?

Polyfills/Window/Source/Window.cpp
Drigax: I'm more interested in large, architectural considerations (Is this even the right problem to solve? are we missing an essential usecase) than small nits (you missed spacing/you don't need this function, etc...) since a lot of this may be rewritten, but thanks

Polyfills/Window/CMakeLists.txt
Drigax: Sounds good, I can change to use that convention.

Polyfills/Window/CMakeLists.txt
Drigax: Not yet, I noticed that your UWP support is in now, so I should support that as well now.

Polyfills/Window/CMakeLists.txt
Drigax: I have no idea, I never worked with UWP prior to this. @syntheticmagus @bghgary do you have any idea?

Polyfills/Window/Source/Windows/WindowPlatform.h
bghgary: What @ryantrem said sound right to me. It should depend on the specializations that are allowed for the platform. For Win32, this is probably an HWND only. For UWP, this can either be a CoreWindow to take over the whole window or a SwapChainPanel. At some point, we may add support for exclusive fullscreen if bgfx adds support for it which may be a completely different parameter.

The switching between platforms should end up in the CMake step as much as possible. We should try to avoid using macros to switch.

Polyfills/Window/CMakeLists.txt
bghgary: CMake defines WIN32 for all windows platforms. WINDOWS_STORE is additionally defined for UWP. `shlwapi` is desktop only. There are hacks to make it work in UWP, but it probably won't pass [WACK](https://developer.microsoft.com/en-us/windows/downloads/app-certification-kit/).

Polyfills/Window/Source/Windows/WindowPlatform.h
bghgary: > We have 3 different patterns, It may be a good idea to align to a single pattern here.

I only see two patterns. The difference between switching the header vs. the cpp just depends on where the code needs to change. This decision needs to be made regardless of the platform switching.

I am personally fine with both patterns. If it's a simple switch there is no need to create a directory to hold it. If there are lots of files, then a directory makes more sense.

Polyfills/Window/Source/Window.cpp
Drigax: @bghgary helped me find a solution, since the javascript `Window` class we create here isn't actually the `window` reference we use in our environment. the following `Set()` declarations adds the other polyfill methods to the global object [which we set to also be global\["window"\]](https://github.com/BabylonJS/BabylonNative/blob/93005ea71a30f489eac3ad1d98878ac50532b51a/Core/JsRuntime/Source/JsRuntime.cpp#L20)

Manually adding an accessor for window .polyfill seems to give us comparable behavior to the window object in a browser.

Apps/Playground/Win32/App.cpp
Drigax: Good question, I'll have to look into how the BabylonNative module interacts with the rest of BabylonNative @syntheticmagus, do you know off hand how we initialize the graphics with a swapchainpanel?

I plan to remove this parameter, @ryantrem raises a good point that we can use `Graphics::Impl::GetNativeWindow()` instead of keeping a pointer to the window, we also persist a pointer to the window type as well.

Polyfills/Window/Source/Windows/WindowPlatform.h
ryantrem: I think since `Graphics` is where we will have the template specializations for window type, and since that is where we have hardware scaling level, and since using the `window` polyfill is optional, and finally since ideally this value would be accessible to native code as well as JavaScript, the core implementation of `GetDevicePixelRatio` should be implemented on and publicly exposed from `Graphics`, and then the `window` polyfill should just expose that value to JavaScript.

Polyfills/Window/CMakeLists.txt
Drigax: I have to keep the convention of each platform header sharing the same name here, since we want the same shared source to include it. But I was able to use ${BABYLON_NATIVE_PLATFORM} to handle the build flavor to minimize platform-specific ${SOURCE} declarations.

Polyfills/Window/Source/Window.cpp
Drigax: We can either a) Link against Graphics and manage a reference to the external class, or b) link against GraphicsInternal and all its dependencies, and use Javascript to get a reference to the Graphics::Impl. I currently updated this PR to use a), as I've had some strange compilation issues with b), but I can take another look at this

Polyfills/Window/Source/Windows/WindowPlatform.h
Drigax: Yes, that's definitely a better suggestion.

Polyfills/Window/Source/Window.h
```diff
@@ -9,19 +9,21 @@ namespace Babylon::Polyfills::Internal
...
     public:
-        static void Initialize(Napi::Env env);
+        static void Initialize(Napi::Env env, void* windowPtr);
         static Window& GetFromJavaScript(Napi::Env);
 
         Window(const Napi::CallbackInfo& info);
-
     private:
         JsRuntime& m_runtime;
+        void* m_windowPtr;
```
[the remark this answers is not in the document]
Drigax: Yes, changed templates to use WindowType instead of void*

Apps/Playground/Scripts/experience.js
Drigax: On the one hand, we might want to match the ThinEngine constructor as closely as possible, on the other hand, the constructor already uses things like HTMLCanvasElement that may not be all that useful for us...

Sure, a configuration object might be the most maintainable solution, since we can add more recognized members of this object as we add more options.

Polyfills/Window/Source/Windows/WindowPlatform.h
Drigax: Ok, moved the implementation to a .cpp, following the (Source or Include)/${BABYLON_NATIVE_PLATFORM}/File convention

Polyfills/Window/CMakeLists.txt
Drigax: We can switch off of `BABYLON_NATIVE_PLATFORM` thankfully somebody has already had the foresight to make sure that we aren't overlapping our build configuration for Win32 and UWP:

https://github.com/BabylonJS/BabylonNative/blob/b4f81771abb2722c3c18731a59b31c18298fcd40/CMakeLists.txt#L17

Apps/Playground/Scripts/experience.js
Drigax: Opened https://github.com/BabylonJS/Babylon.js/pull/9892 to add this to Babylon.js's constructor

Apps/Playground/Win32/App.cpp
Drigax: Changed in the latest implementation to no longer manage a reference to a `Graphics` object, we now statically retrieve a Graphics::Impl via Javascript to retrieve any Graphics related info.

Polyfills/Window/Source/Windows/WindowPlatform.h
Drigax: This implementation makes more sense, now the polyfill is just how we expose this functionality to Javascript, while we keep our Graphics::Impl object as our source of truth and where we should handle these system calls. The latest revision should reflect this change.

Polyfills/Window/Source/Window.cpp
Drigax: Had to tweak the Graphics cmake file, we can link against GraphicsInternal, and it cleans up our creation and makes it so we don't necessarily have to manage a reference to a Graphics object in the Window polyfill object. Definitely the better solution.

Core/Graphics/CMakeLists.txt
```diff
@@ -1,15 +1,24 @@
...
 add_library(Graphics ${SOURCES})
 warnings_as_errors(Graphics)
 
 target_include_directories(Graphics
-    PUBLIC "Include")
+    PUBLIC "Include"
+    PUBLIC "Include/Babylon/${BABYLON_NATIVE_PLATFORM}")
+
+if(WINDOWS_STORE)
+elseif(WIN32)
```
bghgary, line 18:
```suggestion
if(WIN32 AND NOT WINDOWS_STORE)
```

Core/Graphics/CMakeLists.txt
```diff
@@ -33,3 +42,6 @@ target_link_to_dependencies(GraphicsInternal
     INTERFACE bgfx
     INTERFACE bimg
     INTERFACE bx)
+
+    target_compile_definitions(GraphicsInternal
+    INTERFACE NOMINMAX)
```
bghgary, line 47:
```suggestion
target_compile_definitions(GraphicsInternal
    INTERFACE NOMINMAX)
```

Core/Graphics/Source/Graphics.cpp
```diff
@@ -44,25 +44,26 @@ namespace Babylon
...
-    void Graphics::Impl::SetNativeWindow(void* nativeWindowPtr, void* windowTypePtr)
+    void Graphics::Impl::SetNativeWindow(WindowType nativeWindowPtr, void* windowTypePtr)
     {
         std::scoped_lock lock{m_state.Mutex};
         m_state.Bgfx.Dirty = true;
 
         auto& pd = m_state.Bgfx.InitState.platformData;
         pd.ndt = windowTypePtr;
-        pd.nwh = nativeWindowPtr;
+        pd.nwh = (void*)nativeWindowPtr;
```
bghgary, line 60:
```suggestion
        pd.nwh = static_cast<void*>(nativeWindowPtr);
```

Core/Graphics/Source/Win32/Graphics.cpp
```diff
@@ -0,0 +1,12 @@
...
+#include <GraphicsPlatform.h>
+
+namespace Babylon
+{
+    float Graphics::Impl::GetDevicePixelRatio()
+    {
+        UINT dpi = GetDpiForWindow(GetNativeWindow());
+        return (float)dpi / 96.0f;
+    }
+}
```
bghgary, line 12:
```suggestion
        UINT dpi{GetDpiForWindow(GetNativeWindow())};
        return static_cast<float>(dpi) / 96.0f;
    }
}
```

Polyfills/Window/CMakeLists.txt
```diff
@@ -12,7 +12,9 @@ target_include_directories(Window
 target_link_to_dependencies(Window 
     PUBLIC napi
     PRIVATE base-n
-    PRIVATE JsRuntime)
+    PRIVATE JsRuntime
+    PRIVATE GraphicsInternal)
 
 set_property(TARGET Window PROPERTY FOLDER Polyfills)
 source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${SOURCES})
+
```
bghgary, line 20: nit: I don't know what is going on here. Wrong line ending maybe?

Polyfills/Window/Source/Window.cpp
```diff
@@ -1,7 +1,10 @@
-#include "Window.h"
 #include <basen.hpp>
 #include <chrono>
 #include <iterator>
+#include "Window.h"
+
+#include <Babylon/Graphics.h>
+#include "GraphicsImpl.h"
```
bghgary, line 7:
```suggestion
#include <GraphicsImpl.h>
```

Polyfills/Window/Source/Window.cpp
```diff
@@ -10,15 +13,17 @@ namespace Babylon::Polyfills::Internal
         constexpr auto JS_CLASS_NAME = "Window";
         constexpr auto JS_SET_TIMEOUT_NAME = "setTimeout";
         constexpr auto JS_A_TO_B_NAME = "atob";
+
```
bghgary, line 16: nit: extra newline?

Polyfills/Window/Source/Window.cpp
```diff
@@ -10,15 +13,17 @@ namespace Babylon::Polyfills::Internal
...
         constexpr auto JS_REMOVE_EVENT_LISTENER_NAME = "removeEventListener";
+        constexpr auto JS_DEVICE_PIXEL_RATIO_NAME = "devicePixelRatio";
     }
 
     void Window::Initialize(Napi::Env env)
     {
         Napi::HandleScope scope{env};
 
-        Napi::Function constructor = DefineClass(
+        Napi::Function constructor = Window::DefineClass(
```
bghgary, line 26: We don't do this anywhere else. Why just here?

Polyfills/Window/Source/Window.cpp
```diff
@@ -53,6 +58,18 @@ namespace Babylon::Polyfills::Internal
         {
             global.Set(JS_REMOVE_EVENT_LISTENER_NAME, Napi::Function::New(env, &Window::RemoveEventListener, JS_REMOVE_EVENT_LISTENER_NAME));
         }
+        
+        //global.Set(JS_DEVICE_PIXEL_RATIO_NAME,Napi::PropertyDescriptor::Accessor(env, global, JS_DEVICE_PIXEL_RATIO_NAME, &Window::GetDevicePixelRatio, napi_default, nullptr));
+        //auto prop = Napi::PropertyDescriptor::Accessor(env, global, JS_DEVICE_PIXEL_RATIO_NAME, &Window::GetDevicePixelRatio, napi_default, nullptr);
+        //global.DefineProperty(prop);
```
bghgary, line 64: Remove old code.

Polyfills/Window/Source/Window.cpp
```diff
@@ -53,6 +58,18 @@ namespace Babylon::Polyfills::Internal
...
+
+        // Create an accessor to add to the window object to define window.devicePixelRatio
+        Napi::Object descriptor{Napi::Object::New(env)};
+        descriptor.Set("enumerable", Napi::Value::From(env, true));
+        descriptor.Set("get", Napi::Function::New(env, &Window::GetDevicePixelRatio, "devicePixelRatio", jsWindow));
+        Napi::Object object{global.Get("Object").As<Napi::Object>()};
+        Napi::Function defineProperty{object.Get("defineProperty").As<Napi::Function>()};
+        defineProperty.Call(object, {global, Napi::String::New(env, "devicePixelRatio"), descriptor});
```
bghgary, line 72: It would probably be good to wrap this with an if check like the other global members in case this property already exists.

Polyfills/Window/Source/Window.cpp
```diff
@@ -109,6 +126,12 @@ namespace Babylon::Polyfills::Internal
             }
         });
     }
+
+    Napi::Value Window::GetDevicePixelRatio(const Napi::CallbackInfo& info) 
+    {
+        auto env = info.Env();
```
bghgary, line 132:
```suggestion
        auto env{info.Env()};
```

Core/Graphics/Source/Android/Graphics.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include <AndroidExtensions/Globals.h>
+#include <AndroidExtensions/JavaWrappers.h>
+
+namespace Babylon
+{
+    float Graphics::Impl::GetDevicePixelRatio()
+    {
+        auto dpi = android::global::GetAppContext().getResources().getConfiguration().getDensityDpi();
+        return (float)dpi/160.0f;
```
ryantrem, line 13: What is this division by 160? Will you add a comment and a link to associated Android documentation?

Core/Graphics/Include/Babylon/Android/GraphicsPlatform.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#include <android/native_window.h>
+
+using WindowType = ANativeWindow*;
```
ryantrem, line 5: If we do it this way, I think there are a couple implications:
1. There can only be one window type per platform that must be known at compile time. E.g. the same compilation can't use both `HWND` and `SwapChainPanel*`. I'm not sure if this is a problem or not.
2. I don't think there is any reason to even have Graphics be templated if we do it this way.

If we think that each compilation might need to support multiple window types, then I would think we would want to keep the template, and provide one or more template specializations per platform (for each window type supported by that platform). Thoughts @bghgary?

Core/Graphics/Source/Graphics.cpp
```diff
@@ -44,25 +44,26 @@ namespace Babylon
...
-    void* Graphics::Impl::GetNativeWindow()
+    WindowType Graphics::Impl::GetNativeWindow()
     {
         std::scoped_lock lock{m_state.Mutex};
-        return m_state.Bgfx.InitState.platformData.nwh;
+        return (WindowType)m_state.Bgfx.InitState.platformData.nwh;
     }
 
-    void Graphics::Impl::SetNativeWindow(void* nativeWindowPtr, void* windowTypePtr)
+    void Graphics::Impl::SetNativeWindow(WindowType nativeWindowPtr, void* windowTypePtr)
```
ryantrem, line 53: I think the way we really want this to work is for `SetNativeWindow` to take just one parameter, and only for certain specializations of the type of that parameter have `pd.ndt` get set. Like if the type of the `nativeWindowPtr` is known at compile time to be a `SwapChainPanel*`, then the type of `windowTypePtr` is also known to be 2. Therefore from a public API standpoint, `windowTypePtr` should not be part of the API, but rather an implementation detail. We couldn't do that before because we didn't know anything about the type of `nativeWindowPtr` at compile time, but if we are fixing that in this PR, then we should probably figure out how to get `windowTypePtr` out of the contract as well. @bghgary thoughts on the best way to do this?

ryantrem: I put this comment on `Graphics::Impl` but it should have been on `Graphics` itself (the public API). The same concept applies though.

ryantrem: Maybe what would make the most sense would be for `Graphics.h` to `#include GraphicsPlatform.h` and have non-templated members (such as `CreateGraphics` and `UpdateWindow`), but for the implementation to always defer to `GraphicsImpl`, which does have templated members, and then have template specializations per platform. Then we could have (for example) a platform agnostic `void*` version of `CreateGraphics`, and then each platform could provide a specialization for their window types (which would match the types in `GraphicsPlatform.h`) that could call the standard `void*` version and then optionally do additional stuff (like in the case of Windows, it could additionally set `pd.ndt`).

Core/Graphics/Source/Android/Graphics.cpp
Drigax: 160 dpi is the "baseline" android pixel density:
https://developer.android.com/training/multiscreen/screendensities

I'm currently looking to see if there are any constants to reference instead of using a magic number

Polyfills/Window/Source/Window.cpp
Drigax: Ok, makes sense. I'll clean up this portion.

Core/Graphics/Include/Babylon/Android/GraphicsPlatform.h
Drigax: Regarding 1) i was planning on handing cases where we may want multiple window types via defining our generic type with a variant, and using reflection to change behavior based off the current type at runtime, which may or may not be optimal.

I'm also a bit wary of whether or not we need our templated Graphics class as well, I'm interested in @syntheticmagus's opinion regarding what we're hoping to accomplish using it.

Core/Graphics/CMakeLists.txt
```diff
@@ -44,3 +101,12 @@ target_link_to_dependencies(GraphicsInternal
     INTERFACE bgfx
     INTERFACE bimg
     INTERFACE bx)
+
+target_compile_definitions(GraphicsInternal
+    INTERFACE NOMINMAX)
+
+# compile as objc++
+if (APPLE)
```
Drigax, line 109: One thing of note is now that Graphics includes the Apple platform headers, we need to compile all the modules that consume Graphics as objc++. so far this has caused no functional or compilation related issues, but this seems significant.

@ryantrem, do you have any concerns about this change?

Core/Graphics/Source/GraphicsImpl.cpp
```diff
@@ -37,23 +38,25 @@ namespace Babylon
...
+    void Graphics::Impl::SetNativeWindowInternal(GraphicsConfiguration config)
     {
         std::scoped_lock lock{m_state.Mutex};
         m_state.Bgfx.Dirty = true;
 
         auto& pd = m_state.Bgfx.InitState.platformData;
-        pd.ndt = windowTypePtr;
-        pd.nwh = nativeWindowPtr;
+        pd.ndt = nullptr;
+        pd.nwh = config.windowPtr;
```
Drigax, line 54: UWP is the only platform that makes use of the platformData.ndt to specify the type of window being drawn to, so it was added as a field to the configuration object used to create the Graphics object now. 

This should be easily extensible to include any other datatypes needed for UWP as seen in #606.

Polyfills/Window/CMakeLists.txt
Drigax: weird, removed.

Drigax:
Fixes #482 #620

Core/Graphics/CMakeLists.txt
ryantrem: Why do we need this flag? We use objective c++ in other places and didn't need to add any new flags, right?

Drigax: We have a few places where we will include objectivec++ headers, but as far as I see, we do this all in .mm source files, such as ViewController.mm for our MacOS playground, but my understanding is that when clang compiles a .mm file, it will compile it as objectivec++. If we want to include objectivec++ in other source, we'll have to add the flag.

ryantrem: Is this because a header with objective c++ types is being included in a shared .cpp file or something?

Drigax: Yes, the MetalKit header is included in Graphics

Core/Graphics/Source/UWP/GraphicsPlatform.cpp
```diff
@@ -0,0 +1,36 @@
...
+    {
+        SetNativeWindowInternal(config);
+
+        std::scoped_lock lock{m_state.Mutex};
+        auto& pd = m_state.Bgfx.InitState.platformData;
+        pd.ndt = config.windowTypePtr;
+        pd.nwh = config.windowPtr;
+        pd.context = nullptr;
+        pd.backBuffer = nullptr;
+        pd.backBufferDS = nullptr;
```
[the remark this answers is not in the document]
Drigax: I suppose, I went with originally to combine shared + platform specific behavior for whatever we need to do to setup our window. Right now the only difference is how we populate the `platformData` struct, so I can make that change.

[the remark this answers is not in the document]
Drigax: done

Core/Graphics/Source/GraphicsImpl.cpp
Drigax: Following @ryantrem, suggestion I removed `Graphics::Impl::SetNativeWindowInternal` in favor of having platform implementations of `Graphics::Impl::ConfigureBgfxPlatformData` called by a shared `Graphics::Impl::SetNativeWindow`

Core/Graphics/CMakeLists.txt
Drigax: I was able to modify Graphics such that components that are including Graphics and GraphicsImpl.h are no longer consuming the platform headers via some clever use of templates thanks to @syntheticmagus. However GetNativeWindow now has an awkward way to invoke it where we use the window type as a template specialization. Alternatives were to remove this from the class outright and create some utility function to handle this, but this seems a bit awkward as well...

Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -39,7 +40,11 @@ - (void)init:(void*)view width:(int)inWidth height:(int)inHeight xrView:(void*)x
     float height = inHeight;
     void* windowPtr = view;
     
-    graphics = Babylon::Graphics::CreateGraphics(windowPtr, static_cast<size_t>(width), static_cast<size_t>(height));
+    GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
+    graphicsConfig.windowPtr = (__bridge WindowType)windowPtr;
```
ryantrem, line 44: Can't the `(void*)view` parameter now be changed to `(MTKView*)view`?

Apps/ValidationTests/iOS/LibNativeBridge.mm
```diff
@@ -35,7 +36,11 @@ - (void)init:(void*)view width:(int)inWidth height:(int)inHeight
     float height = inHeight;
     void* windowPtr = view;
     
-    graphics = Babylon::Graphics::CreateGraphics(windowPtr, static_cast<size_t>(width), static_cast<size_t>(height));
+    GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
+    graphicsConfig.windowPtr = (__bridge WindowType)windowPtr;
```
ryantrem, line 40: Same comment about `MTKView`.

Core/Graphics/Include/Android/Babylon/GraphicsPlatform.h
```diff
@@ -0,0 +1,13 @@
+ #pragma once
+
+#include <android/native_window.h>
+
+using WindowType = ANativeWindow*;
+
+struct GraphicsConfiguration
+{
+    WindowType windowPtr;
+    void* windowTypePtr;
```
ryantrem, line 10: Can't this be removed now since there is a platform specific function for translating from `GraphicsConfiguration` to bgfx platform data?

Apps/ValidationTests/macOS/ViewController.mm
```diff
@@ -69,10 +70,14 @@ - (void)initialize {
     engineView.autoresizingMask = NSViewWidthSizable | NSViewHeightSizable;
     [[self view] addSubview:engineView];
     engineView.delegate = engineView;
-
+    
     void* windowPtr = (__bridge void*)engineView;
 
-    graphics = Babylon::Graphics::CreateGraphics(windowPtr, static_cast<size_t>(600), static_cast<size_t>(400));
+    GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
+    graphicsConfig.windowPtr = engineView;
```
ryantrem, line 77: How does this compile? Isn't `graphicsConfig.windowPtr` of type `MTKView*`?

Core/Graphics/Include/UWP/CoreWindow/Babylon/GraphicsPlatform.h
```diff
@@ -0,0 +1,15 @@
...
+#include <windows.ui.core.h>
+#include <winrt/windows.ui.core.h>
+#include <winrt/Windows.ApplicationModel.h>
+
+using WindowType = winrt::Windows::UI::Core::ICoreWindow*;
+
+struct GraphicsConfiguration
+{
+    WindowType windowPtr;
+    void* windowTypePtr;
```
ryantrem, line 12: You should be able to remove this as it is just supposed to indicate whether you are targeting an `ICoreWindow*` or an `ISwapChainPanel*`, which you already know at compile time.

Core/Graphics/Source/Apple/macOS/GraphicsPlatform.mm
```diff
@@ -0,0 +1,27 @@
...
+        pd.nwh = config.windowPtr;
+        pd.context = nullptr;
+        pd.backBuffer = nullptr;
+        pd.backBufferDS = nullptr;
+    }
+
+    float GraphicsImpl::UpdateDevicePixelRatio()
+    {
+        std::scoped_lock lock{m_state.Mutex};
+        MTKView* view = (MTKView*)GetNativeWindow<WindowType>();
```
ryantrem, line 23: If this has to be cast, then haven't we lost the benefit of the new approach?

Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -49,20 +51,20 @@ namespace Babylon
...
 
-        Impl();
-        ~Impl();
-
-        void* GetNativeWindow();
-        void SetNativeWindow(void* nativeWindowPtr, void* windowTypePtr);
+        GraphicsImpl();
+        virtual ~GraphicsImpl();
+        template<typename WindowT>
+        WindowT GetNativeWindow();
```
ryantrem, line 62: If I understand correctly, this is the change that tries to eliminate the need to compile with the objective C++ flag, is that right? It seems very awkward though, and like we have a weird mix of known and sort of unknown types in the API.

Core/Graphics/Source/UWP/GraphicsPlatform.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "../GraphicsImpl.h"
+
+#include <winrt/Windows.Graphics.Display.h>
+
+using namespace winrt;
+using namespace Windows::Graphics::Display;
+
+namespace Babylon
+{
+    void GraphicsImpl::ConfigureBgfxPlatformData(const GraphicsConfiguration& config, bgfx::PlatformData& pd)
```
ryantrem, line 12: I think I was imagining we'd have two versions of this, just like you do for the headers (`ICoreWindow*` vs. `ISwapChainPanel*`), and then they could just set the `pd.nwh` to the correct type.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -1,42 +1,29 @@
 #include <Babylon/Graphics.h>
+#include <Babylon/GraphicsPlatform.h>
 #include "GraphicsImpl.h"
+#include <Babylon/Graphics.h>
 
 namespace Babylon
 {
+    class Graphics::Impl : public GraphicsImpl { };
```
ryantrem, line 8: What is the purpose of this?

Apps/Playground/iOS/LibNativeBridge.mm
Drigax: Yeah, it probably should be.

Core/Graphics/Source/Graphics.cpp
Drigax: GraphicsImpl.h is now no longer included by Graphics.h,
So we declare Graphics::Impl in Graphics.h, then define it as our internal type in the .cpp where we use it.

The idea was to remove it from the public contract.

Core/Graphics/Source/UWP/GraphicsPlatform.cpp
Drigax: That's fine, I originally opted to share it as the implementations seemed identical, but this makes sense.

Core/Graphics/Source/Apple/macOS/GraphicsPlatform.mm
Drigax: Good point, I think that was an oversight, I don;t believe I've updated this code in a while.

Core/Graphics/Source/GraphicsImpl.h
Drigax: My understanding is that if we want this function as a member of `GraphicsImpl` we either A) go back to using void*, B) use the platform type in this header, causing compilation issues with our consumers if the headers arent c++, or c) use some templating to obscure the type in the header.

The alternative is to make it some odd utility function that uses GraphicsImpl? C++ makes these types of situations difficult.

Core/Graphics/Include/Android/Babylon/GraphicsPlatform.h
Drigax: Yeah we can define GraphicsConfiguration differently in the different platform headers now, I'll change that.

Core/Graphics/Source/Graphics.cpp
Drigax: Until I realized that we also need to change this for consumers of the private contract as well, so the intent of this change was wrong. I believe reverting this would be a reasonable option.

Core/Graphics/Source/GraphicsImpl.h
syntheticmagus: I agree that it's a little awkward, but (in my opinion) much less so than having a bunch of C++ libraries that have to be compiled as Objective C++ just so that they can find a type that they don't actually use in C++. The reason for templating this instead of forward-declaring `MTKView` is because Apple defines `MTKView` as an interface, which didn't seem to play nicely with our attempts to forward-declare it in C++. Because this is an internal contract, not a proper outward-facing API, I don't think it's too big a deal for it to have this small degree of awkwardness.

If it's bothersome, though, I can think of two other option to do this cleanly, though the first is (in my opinion) even more awkward: have a special platform-specific struct defined just like the config struct, the only purpose of which is to contain an `MTKView*`, then forward-declare _that_ and use it as the return type. This is awkward because it entails having an entire type which has to be implemented per-platform just for the purpose of avoiding a templated function.

The second approach would be to simply move this function definition to a platform-specific header, using our CMake practices to deliver the right header to the right consumer. This is probably my favorite of the three, but it's still a little awkward because you'd need to take it out of the class, making it `GetNativeWindowFromGraphicsImpl(GraphicsImpl&)` or something, which could get odd. Bottom line, I think all three of these approaches are equally viable and equally awkward, so I don't really have a strong personal preference among them.

Apps/ValidationTests/macOS/ViewController.mm
Drigax: we're declaring the interface `EngineView` to be an `MTKView` via L18

Apps/Playground/iOS/LibNativeBridge.mm
Drigax: changed.

Apps/ValidationTests/iOS/LibNativeBridge.mm
Drigax: changed.

Core/Graphics/Include/Android/Babylon/GraphicsPlatform.h
Drigax: Changed.

Core/Graphics/Include/UWP/CoreWindow/Babylon/GraphicsPlatform.h
Drigax: Yeah, I moved the logic to ConfigureBgfxPlatformData for the different platforms.

Core/Graphics/Source/Apple/macOS/GraphicsPlatform.mm
Drigax: removed the cast

Core/Graphics/Source/UWP/GraphicsPlatform.cpp
Drigax: Added a separate implementations for the different window types.

Plugins/NativeCapture/CMakeLists.txt
```diff
@@ -13,4 +13,4 @@ target_link_to_dependencies(NativeCapture
     PRIVATE GraphicsInternal)
 
 set_property(TARGET NativeCapture PROPERTY FOLDER Plugins)
-source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${SOURCES})
+source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${SOURCES})
```
syntheticmagus, line 16: nit: Undo changes to this file.

Plugins/NativeEngine/CMakeLists.txt
```diff
@@ -68,4 +68,4 @@ target_link_to_dependencies(NativeEngineInternal
     INTERFACE glslang
     INTERFACE SPIRV
     INTERFACE spirv-cross-hlsl
-    INTERFACE GraphicsInternal)
+    INTERFACE GraphicsInternal)
```
syntheticmagus, line 71: nit: Undo changes to this file.

Polyfills/Window/Include/Babylon/Polyfills/Window.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <napi/env.h>
+#include <Babylon/Graphics.h>
```
syntheticmagus, line 4: Undo?

Polyfills/Window/Source/Window.cpp
```diff
@@ -1,7 +1,10 @@
-#include "Window.h"
 #include <basen.hpp>
 #include <chrono>
 #include <iterator>
+#include "Window.h"
+
+#include <Babylon/Graphics.h>
```
syntheticmagus, line 6: Does this file need `Graphics`, or just `GraphicsImpl`?

Polyfills/Window/Source/Window.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <Babylon/JsRuntime.h>
+#include <Babylon/Graphics.h>
```
syntheticmagus, line 4: Undo?

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -70,7 +71,11 @@ extern "C"
             int32_t width  = ANativeWindow_getWidth(window);
             int32_t height = ANativeWindow_getHeight(window);
 
-            g_graphics = Babylon::Graphics::CreateGraphics<void*>(window, static_cast<size_t>(width), static_cast<size_t>(height));
+            GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
```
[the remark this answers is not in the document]
Drigax: Definitely appreciate the explanation! I think I'm too used to .NET convention where lvalues without assignment are null references.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -126,7 +131,11 @@ extern "C"
         {
             ANativeWindow *window = ANativeWindow_fromSurface(env, surface);
             g_runtime->Dispatch([window, width = static_cast<size_t>(width), height = static_cast<size_t>(height)](auto env) {
-                g_graphics->UpdateWindow<void*>(window);
+                GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
```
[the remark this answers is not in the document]
Drigax: I figured using the default constructor and  populating the fields independently in our sample playground gives us the flexibility to expand the `GraphicsConfiguration` object without needing to update existing code that uses this struct. I considered using a list initializer as well, however we're still constrained by the ordering of the members in the struct.

Apps/Playground/UWP/App.cpp
```diff
@@ -259,9 +260,13 @@ void App::RestartRuntime(Windows::Foundation::Rect bounds)
     m_displayScale = static_cast<float>(displayInformation->RawPixelsPerViewPixel);
     size_t width = static_cast<size_t>(bounds.Width * m_displayScale);
     size_t height = static_cast<size_t>(bounds.Height * m_displayScale);
-    auto* windowPtr = reinterpret_cast<ABI::Windows::UI::Core::ICoreWindow*>(CoreWindow::GetForCurrentThread());
+    auto* windowPtr = reinterpret_cast<winrt::Windows::UI::Core::ICoreWindow*>(CoreWindow::GetForCurrentThread());
 
-    m_graphics = Babylon::Graphics::CreateGraphics<void*>(windowPtr, width, height);
+    GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
```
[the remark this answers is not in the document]
Drigax: see Android/app/src/main/cpp/BabylonNativeJNI, going to resolve the like comments other than that one.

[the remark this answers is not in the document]
syntheticmagus: The main thing these comments were intended to call out was the assignment operator/double-constructor thing, not the constructor-with-arguments suggestion, so I'm not sure these are actually resolved based on that comment.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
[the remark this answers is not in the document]
syntheticmagus: The suggestion to have a constructor take arguments isn't really more than a nit, so it's not that important. That said, you can still choose to not use the constructor if it doesn't fit your use case, and you can have multiple constructors if you have several typical use cases. In this case, since you're writing the same four or five lines of code multiple times in multiple places, this initialization scenario seems common enough to warrant a constructor (which, if the struct is later expanded with other members, could just leave those blank if it wants to since they by definition won't pertain to this usage). As I said, though, it's a nit, so just a judgement call. The main thing this comment was to document was the same thing as the long one: the initialization code.

Apps/ValidationTests/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -116,9 +122,15 @@ extern "C"
...
-                g_graphics->UpdateSize(width, height);
+
+            GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
+            graphicsConfig.windowPtr = window;
+            graphicsConfig.width = static_cast<size_t>(width);
+            graphicsConfig.height = static_cast<size_t>(height);
+
+            g_runtime->Dispatch([graphicsConfig](auto env) {
+                g_graphics->UpdateWindow(graphicsConfig);
+                g_graphics->UpdateSize(graphicsConfig.width, graphicsConfig.height);
```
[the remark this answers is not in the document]
Drigax: I don't know why we're dispatching here. If its not necessary I can remove it.

@CedricGuillemet, do you know why we're doing this?

Core/Graphics/Include/Shared/Babylon/Graphics.h
```diff
@@ -34,12 +33,18 @@ namespace Babylon
...
+
     private:
         Graphics();
 
         Graphics(const Graphics&) = delete;
         Graphics(Graphics&&) = delete;
 
+        void UpdateWindowInternal(GraphicsConfiguration config);
+
+        float UpdateDevicePixelRatio();
```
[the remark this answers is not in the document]
Drigax: UpdateDevicePixelRatio is implemented in the platform .cpp files.

I don't see UpdateWindowInternal used anywhere, I'll remove that, surprised it didn't cause any compilation errors.

Core/Graphics/Source/Android/GraphicsPlatform.cpp
```diff
@@ -0,0 +1,28 @@
+#include <Babylon/Graphics.h>
+#include <Babylon/GraphicsPlatform.h>
+#include "../GraphicsImpl.h"
+
+#include <AndroidExtensions/Globals.h>
+#include <AndroidExtensions/JavaWrappers.h>
+
+namespace Babylon
+{
+    void GraphicsImpl::ConfigureBgfxPlatformData(const GraphicsConfiguration& config, bgfx::PlatformData& pd)
```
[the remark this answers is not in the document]
Drigax: the fields in the configuration differs between platforms, and the fields in the `PlatformData` that we want to populate differs between platforms as well. I'm not sure if we should use a shared implementation for this.

Core/Graphics/Source/Graphics.cpp
```diff
@@ -1,42 +1,29 @@
...
+    
     Graphics::Graphics()
-        : m_impl{std::make_unique<Impl>()}
+        : m_impl{std::make_unique<Graphics::Impl>()}
```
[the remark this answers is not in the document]
Drigax: No, we're already in Graphics, let me remove that.

Polyfills/Window/Source/Window.cpp
Drigax: Just GraphicsImpl

Core/Graphics/Source/Android/GraphicsPlatform.cpp
```diff
@@ -0,0 +1,28 @@
+#include <Babylon/Graphics.h>
+#include <Babylon/GraphicsPlatform.h>
+#include "../GraphicsImpl.h"
```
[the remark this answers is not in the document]
Drigax: So far both of those suggestions cause build failure as we cannot find the header. Not too sure why. I'm going to leave as-is for now.

Core/Graphics/CMakeLists.txt
```diff
@@ -11,14 +11,71 @@ set(SOURCES
...
 target_include_directories(Graphics
-    PUBLIC "Include"
-    PRIVATE "Source/${BABYLON_NATIVE_PLATFORM}")
+            PUBLIC "Include/Shared"
+            PRIVATE "Source/${BABYLON_NATIVE_PLATFORM}")
+
+if(WINDOWS_STORE)
+    if(BABYLON_NATIVE_WINDOWS_STORE_USE_SWAPCHAINPANEL)
+        target_include_directories(Graphics
+            PUBLIC "Include/${BABYLON_NATIVE_PLATFORM}/SwapChainPanel")
```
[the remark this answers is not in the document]
Drigax: Im sure it would condense some of the logic here, but at the cost of readibility. I'm hesitant to make this change, as here we're pretty clear what we're doing based on build flavor and option.

Core/Graphics/CMakeLists.txt
```diff
@@ -11,14 +11,71 @@ set(SOURCES
...
+
+if(WIN32 AND NOT WINDOWS_STORE)
+    target_link_to_dependencies(Graphics
+        PRIVATE "shlwapi.lib")
+elseif(ANDROID)
+    target_link_to_dependencies(Graphics
+        PRIVATE android
+        PRIVATE AndroidExtensions)
+elseif(UNIX AND NOT APPLE)
+    # link to linux libraries
```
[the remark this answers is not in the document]
Drigax: not needed

Core/Graphics/Include/Shared/Babylon/Graphics.h
```diff
@@ -4,6 +4,8 @@
 
 #include <memory>
 
+struct GraphicsConfiguration;
```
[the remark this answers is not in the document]
Drigax: I tried moving this forward declaration to Babylon::Graphics, however clang considers these expressions illegal:
```
/Users/runner/work/1/s/Core/Graphics/Source/Apple/../GraphicsImpl.h:23:12: error: forward declaration of struct cannot have a nested name specifier
    struct Graphics::Configuration;
           ^~~~~~~~~~
1 error generated.

```
I think our hands are a bit tied, this struct may have to live in `Babylon` namespace

Apps/Playground/X11/App.cpp
```diff
@@ -54,18 +55,23 @@ namespace
         inputBuffer.reset();
         graphics.reset();
     }
-    
-    void InitBabylon(int32_t window, int width, int height, int argc, const char* const* argv)
+
+    void InitBabylon(Window window, int width, int height, int argc, const char* const* argv)
```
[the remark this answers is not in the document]
Drigax: its a primitive. an `unsigned long` that stores the address where the Window struct data resides. See X.h.

Apps/Playground/X11/App.cpp
```diff
@@ -54,18 +55,23 @@ namespace
...
     {
         std::vector<std::string> scripts(argv + 1, argv + argc);
         std::string moduleRootUrl = GetUrlFromPath(GetModulePath().parent_path());
 
         Uninitialize();
 
         // Separately call reset and make_unique to ensure prior state is destroyed before new one is created.
-        graphics = Babylon::Graphics::CreateGraphics((void*)(uintptr_t)window, static_cast<size_t>(width), static_cast<size_t>(height));
+        GraphicsConfiguration graphicsConfig = GraphicsConfiguration();
+        graphicsConfig.windowPtr = (void*)(uintptr_t)window;
```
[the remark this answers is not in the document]
Drigax: Linux is a bit of a special case here. `Window` is defined as a `long unsigned int` rather than a pointer type, which makes our shared casting code not work for this case (clang does not allow us to static cast to a non pointer type)

Rather than

Core/Graphics/Include/Unix/Babylon/GraphicsPlatform.h
```diff
@@ -0,0 +1,12 @@
+ #pragma once
+
+#include <X11/Xlib.h>
+
+using WindowType = void*;
```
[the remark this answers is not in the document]
Drigax: This actually is intended. see my other response regarding the Linux typing.

ryantrem:
Hey @Drigax - since this has Babylon Native API breaking changes, are you planning to update Babylon React Native? There are some other PRs out with updates to both Babylon Native and Babylon React Native, and I think these are blocked until Babylon React Native is updated to deal with the API changes.

Drigax:
> Hey @Drigax - since this has Babylon Native API breaking changes, are you planning to update Babylon React Native? There are some other PRs out with updates to both Babylon Native and Babylon React Native, and I think these are blocked until Babylon React Native is updated to deal with the API changes.

I see, I don't believe the PRs are blocked, since they're updating the submodule to a commit prior to mine, however I have the changes staging here, and will have a PR open shortly:

https://github.com/Drigax/BabylonReactNative/tree/update-graphics-contract

PR #583, opened 2021-01-26
Update NativeInput to propagate mouse device type

chrisfromwork:
We need some sort of change like this to correctly send desktop mouse pointer events to BabylonNative from BabylonReactNative logic.


tymlipari (approved):

Plugins/NativeInput/Source/NativeInput.cpp
```diff
@@ -61,31 +61,33 @@ namespace Babylon::Plugins
         NativeInput::Impl::DeviceInputSystem::Initialize(env);
     }
 
-    void NativeInput::Impl::PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y)
+    void NativeInput::Impl::PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y, bool isMouse)
     {
-        m_runtimeScheduler([pointerId, buttonIndex, x, y, this]() {
-            const uint32_t inputIndex{GetPointerButtonInputIndex(buttonIndex)};
-            std::vector<int32_t>& deviceInputs{GetOrCreateInputMap(DeviceType::Touch, pointerId, { inputIndex, POINTER_X_INPUT_INDEX, POINTER_Y_INPUT_INDEX })};
+        const auto deviceType = isMouse ? DeviceType::Mouse : DeviceType::Touch;
```
ryantrem, line 66:
```suggestion
        const auto deviceType{ isMouse ? DeviceType::Mouse : DeviceType::Touch };
```

Plugins/NativeInput/Include/Babylon/Plugins/NativeInput.h
```diff
@@ -7,14 +7,25 @@ namespace Babylon::Plugins
...
+
         // TODO: Ideally instances of these should be scoped to individual views within an env, but we don't yet support multi-view.
         // See https://github.com/BabylonJS/BabylonNative/issues/147
         static NativeInput& CreateForJavaScript(Napi::Env);
         static NativeInput& GetFromJavaScript(Napi::Env);
 
-        void PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
-        void PointerUp(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
-        void PointerMove(uint32_t pointerId, uint32_t x, uint32_t y);
+        void PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y, DeviceType deviceType);
```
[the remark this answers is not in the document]
chrisfromwork: this should now be updated

Plugins/NativeInput/Source/NativeInput.cpp
```diff
@@ -61,31 +61,31 @@ namespace Babylon::Plugins
...
 
-    void NativeInput::Impl::PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y)
+    void NativeInput::Impl::PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y, DeviceType deviceType)
     {
-        m_runtimeScheduler([pointerId, buttonIndex, x, y, this]() {
-            const uint32_t inputIndex{GetPointerButtonInputIndex(buttonIndex)};
-            std::vector<int32_t>& deviceInputs{GetOrCreateInputMap(DeviceType::Touch, pointerId, { inputIndex, POINTER_X_INPUT_INDEX, POINTER_Y_INPUT_INDEX })};
+        m_runtimeScheduler([pointerId, buttonIndex, x, y, deviceType, this]() {
+            const uint32_t inputIndex{ GetPointerButtonInputIndex(buttonIndex) };
+            std::vector<int32_t>& deviceInputs{ GetOrCreateInputMap(deviceType, pointerId, { inputIndex, POINTER_X_INPUT_INDEX, POINTER_Y_INPUT_INDEX }) };
```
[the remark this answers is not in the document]
chrisfromwork: I believe this should now be fixed

Plugins/NativeInput/Source/NativeInput.cpp
chrisfromwork: this is no longer applicable

Plugins/NativeInput/Source/NativeInput.cpp
```diff
@@ -98,26 +162,31 @@ namespace Babylon::Plugins
...
         });
     }
 
     NativeInput::Impl::DeviceStatusChangedCallbackTicket NativeInput::Impl::AddDeviceConnectedCallback(NativeInput::Impl::DeviceStatusChangedCallback&& callback)
     {
-        return m_deviceConnectedCallbacks.insert(std::move(callback));
+        // Firing this callback on registration to mimic web behavior
+        for (auto [key, state] : m_inputs)
+        {
+            callback(key.first, key.second);
```
chrisfromwork, line 186: @ryantrem synchronous calls seem to work for this callback, but do we need to actually queue things with the event dispatcher?

Plugins/NativeInput/Source/NativeInput.h
```diff
@@ -32,9 +32,14 @@ namespace Babylon::Plugins
 
         Impl(Napi::Env);
 
-        void PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
-        void PointerUp(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
-        void PointerMove(uint32_t pointerId, uint32_t x, uint32_t y);
+        bool HasMouse();
```
chrisfromwork, line 35: TODO: i still need this to be platform specific

Plugins/NativeInput/CMakeLists.txt
```diff
@@ -1,9 +1,20 @@
...
-    "Source/DeviceInputSystem.cpp"
-    "Source/DeviceInputSystem.h")
+    "Source/Shared/NativeInput.cpp"
+    "Source/Shared/NativeInput.h"
+    "Source/Shared/DeviceInputSystem.cpp"
+    "Source/Shared/DeviceInputSystem.h")
+
+if (ANDROID)
+    set(SOURCES ${SOURCES}
+        "Source/Android/NativeInputAndroid.cpp")
```
ryantrem, line 10: Can we just name all these files `NativeInput.cpp`? I think that is more similar to how we do it in other places, like UrlLib.

Plugins/NativeInput/Source/Shared/NativeInput.cpp
```diff
@@ -40,55 +42,112 @@ namespace Babylon::Plugins
...
 
     NativeInput::Impl::Impl(Napi::Env env)
         : m_runtimeScheduler{JsRuntime::GetFromJavaScript(env)}
     {
         NativeInput::Impl::DeviceInputSystem::Initialize(env);
+
+        if (HasMouse())
+        {
+            const uint32_t inputIndex{ GetPointerButtonInputIndex(0) };
+            GetOrCreateInputMap(DeviceType::Mouse, MOUSE_POINTER_ID, { inputIndex, POINTER_X_INPUT_INDEX, POINTER_Y_INPUT_INDEX });
```
ryantrem, line 83: Is this just so that we can query these inputs before there are any mouse events?

Plugins/NativeInput/Include/Babylon/Plugins/NativeInput.h
```diff
@@ -12,9 +12,12 @@ namespace Babylon::Plugins
...
         static NativeInput& GetFromJavaScript(Napi::Env);
 
-        void PointerDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
-        void PointerUp(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
-        void PointerMove(uint32_t pointerId, uint32_t x, uint32_t y);
+        void MouseDown(uint32_t buttonIndex, uint32_t x, uint32_t y);
+        void MouseUp(uint32_t buttonIndex, uint32_t x, uint32_t y);
+        void MouseMove(uint32_t x, uint32_t y);
+        void TouchDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
+        void TouchUp(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y);
```
ryantrem, line 19: I don't think we need/want `buttonIndex` for these since there is only one "button" for touch. Seems like it should be:
```suggestion
        void TouchDown(uint32_t pointerId, uint32_t x, uint32_t y);
        void TouchUp(uint32_t pointerId, uint32_t x, uint32_t y);
```

Plugins/NativeInput/Source/Shared/NativeInput.cpp
```diff
@@ -40,55 +42,112 @@ namespace Babylon::Plugins
...
+    }
+
+    void NativeInput::Impl::MouseMove(uint32_t x, uint32_t y)
+    {
+        PointerMove(MOUSE_POINTER_ID, x, y, DeviceType::Mouse);
+    }
+
+    void NativeInput::Impl::TouchDown(uint32_t pointerId, uint32_t buttonIndex, uint32_t x, uint32_t y)
+    {
+        PointerDown(pointerId, buttonIndex, x, y, DeviceType::Touch);
```
ryantrem, line 104: Based on my previous comment, it seems like the `buttonIndex` should be hard coded to 0 (e.g. `TOUCH_BUTTON_INDEX`, like what you did for `MOUSE_POINTER_ID`).

Plugins/NativeInput/Source/Windows/NativeInputWindows.cpp
```diff
@@ -0,0 +1,9 @@
+#include "..\Shared\NativeInput.h"
+
+namespace Babylon::Plugins
+{
+    bool NativeInput::Impl::HasMouse()
+    {
+        return true;
```
ryantrem, line 7: I was thinking we would actually use platform specific APIs to determine whether a mouse is present (similar to what Babylon.js does in the browser). For UWP I think this could be determined from `Windows::Devices::Input::MouseCapabilities::MousePresent`.

Plugins/NativeInput/CMakeLists.txt
chrisfromwork: Can you include multiple cpp files with the same name in the same lib project?

Plugins/NativeInput/Source/Windows/NativeInputWindows.cpp
tymlipari: I had a similar thought while looking at this code, but wonder how we'd handle transient devices? For example on HoloLens we support bluetooth mice, but they might not be present at app launch. We'd presumably want to be responsive to runtime changes (though that might also be better to tackle with a separate PR).

Plugins/NativeInput/CMakeLists.txt
ryantrem: If they are in different folders I think so.

Plugins/NativeInput/Source/NativeInput.cpp
ryantrem: No we shouldn't need to dispatch. This is called from the JS thread so we are already on the correct thread, and the equivalent code that runs in the browser does this synchronously as well: https://github.com/BabylonJS/Babylon.js/blob/master/src/DeviceInput/deviceInputSystem.ts#L29-L40

Plugins/NativeInput/Source/Windows/NativeInputWindows.cpp
ryantrem: Yea we should handle that properly but I’m ok with it being in a later change personally.

Plugins/NativeInput/Source/Android/NativeInputAndroid.cpp
```diff
@@ -0,0 +1,9 @@
+#include "..\Shared\NativeInput.h"
+
+namespace Babylon::Plugins
+{
+    bool NativeInput::Impl::HasMouse()
+    {
+        return false;
```
syntheticmagus, line 7: nit: This is probably fine for now, but I don't think this is a safe assumption going forward, is it? Android devices do support mice, I think.

Plugins/NativeInput/CMakeLists.txt
chrisfromwork: yeah looks like this suggestion works. FWIW, I prefer this naming as well :)

Plugins/NativeInput/Include/Babylon/Plugins/NativeInput.h
chrisfromwork: sounds good

Plugins/NativeInput/Source/Shared/NativeInput.cpp
chrisfromwork: sounds good

Plugins/NativeInput/Source/Android/NativeInputAndroid.cpp
chrisfromwork: I'll add todos to these currently returning false

Plugins/NativeInput/Source/Windows/NativeInputWindows.cpp
chrisfromwork: sounds good

Plugins/NativeInput/Source/Shared/NativeInput.cpp
chrisfromwork: yes, that was the thought behind it

ryantrem (approved):

PR #589, opened 2021-01-27
Expose AR Feature Points on iOS

jomcclan:
This PR contains logic that surfaces feature points from ARKit on iOS. This replicates the logic that @Alex-MSFT added for ARCore back in August: https://github.com/BabylonJS/BabylonNative/pull/353.

This'll allow the applications consuming this logic to get the same experience with feature points on iOS as they currently have on Android, and furthers the feature parity across platforms.

### DIFF
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -538,6 +538,7 @@ namespace xr {
         ARFrame* currentFrame{};
         float DepthNearZ{ DEFAULT_DEPTH_NEAR_Z };
         float DepthFarZ{ DEFAULT_DEPTH_FAR_Z };
+        bool FeaturePointCloudEnabled{ false };
 
         Impl(System::Impl& systemImpl, void* graphicsContext, std::function<void*()> windowProvider)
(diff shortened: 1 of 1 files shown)


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -977,6 +978,40 @@ void UpdatePlanes(std::vector<Frame::Plane::Identifier>& updatedPlanes, std::vec
             }
         }
 
+        void UpdateFeaturePointCloud() {
+            if (!FeaturePointCloudEnabled) {
+                return;
+            }
+
+            ARPointCloud *pointCloud = currentFrame.rawFeaturePoints;
```
Alex-MSFT, line 986:
```suggestion
            ARPointCloud* pointCloud = currentFrame.rawFeaturePoints;
```

Alex-MSFT (approved):

Dependencies/xr/Source/ARKit/XR.mm
ryantrem: I think we are about 50/50 on pointer style in this doc, heh.

PR #592, opened 2021-02-02
Fix slash direction

chrisfromwork:
BabylonNative PR ci doesn't build NativeInput, but BabylonReactNative is seeing android and ios build failures due to slash direction


ryantrem (approved):

chrisfromwork:
this change fixed android and iOS BabylonReactNative builds here: https://github.com/BabylonJS/BabylonReactNative/pull/145/checks?check_run_id=1816460415

PR #593, opened 2021-02-03
Homogeneous Depth

CedricGuillemet:
changes made to be in sync with https://github.com/BabylonJS/Babylon.js/pull/9885
babylonjs with commit https://github.com/BabylonJS/Babylon.js/pull/9885/commits/4067b79f2f09c9d4d9369a3467c228f58194e757


ryantrem:
Does this change require a new Babylon.js to not be broken, or is a new Babylon.js only needed to call these new functions? In general, if we have breaking changes, I think we need to not check them in to Babylon Native until we have a new @babylonjs/core NPM package published, otherwise Babylon React Native cannot update Babylon Native and people get blocked in other work.

CedricGuillemet:
@ryantrem on the native side, there is only a new InstanceAccessor, so it can be used with any previous bjs version.
but the bjs PR https://github.com/BabylonJS/Babylon.js/pull/9885 makes this one mandatory

CedricGuillemet:
@ryantrem my fix PR hase been merged in bjs, so be careful when you specify a new bjs module version until this fix has been merged.

bghgary (approved):

syntheticmagus (approved):

PR #603, opened 2021-02-13
Disable hardware upscaling

### LINKED ISSUE #601: Setting engine.hardwareScalingLevel to a small value causes some android devices to reboot
Setting a very large rendering resolution causes Pixel 3 devices to reboot.

This can be reproduced by adding `engine.setHardwareScalingLevel(0.1)` to the playground experience.js.

Drigax:
Disable upscaling by disallowing setting engine.hardwareScalingLevel < 1.0
fixes #601. (kinda)

### DIFF
--- a/Core/Graphics/Source/Graphics.cpp
+++ b/Core/Graphics/Source/Graphics.cpp
@@ -270,13 +270,13 @@ namespace Babylon
 
     void Graphics::Impl::SetHardwareScalingLevel(float level)
     {
-        if (level <= std::numeric_limits<float>::epsilon())
+        if (level < 1) // Disallow upscaling.
         {
-            throw std::runtime_error{"HardwareScalingValue cannot be less than or equal to 0."};
+            throw std::runtime_error{"HardwareScalingValue cannot be less than 1."};
         }
         {
             std::scoped_lock lock{m_state.Mutex};
-            m_state.Resolution.HardwareScalingLevel = std::max(level, 1.0f); // Disallow upscaling.
+            m_state.Resolution.HardwareScalingLevel = level;
         }
 
         UpdateBgfxResolution();
--- a/Core/Graphics/Source/Graphics.cpp
+++ b/Core/Graphics/Source/Graphics.cpp
@@ -276,7 +276,7 @@ namespace Babylon
         }
         {
             std::scoped_lock lock{m_state.Mutex};
-            m_state.Resolution.HardwareScalingLevel = level;
+            m_state.Resolution.HardwareScalingLevel = std::max(level, 1.0f); // Disallow upscaling.
         }
 
         UpdateBgfxResolution();


Core/Graphics/Source/Graphics.cpp
```diff
@@ -276,7 +276,7 @@ namespace Babylon
         }
         {
             std::scoped_lock lock{m_state.Mutex};
-            m_state.Resolution.HardwareScalingLevel = level;
+            m_state.Resolution.HardwareScalingLevel = std::max(level, 1.0f); // Disallow upscaling.
```
ryantrem, line 279: It seems weird to check that `level` is greater than epsilon above, but then if it is between epsilon and 1 to clamp it to 1 here. To me it would make more sense to just change the check on line 275 to throw if the value is less than 1 with an error message that says upscaling is not supported. I think this will make it more immediately obvious why things are not behaving as expected if you set a level less than 1.

Drigax: Good idea, made the change.

ryantrem (approved):

Core/Graphics/Source/Graphics.cpp
```diff
@@ -270,9 +270,9 @@ namespace Babylon
 
     void Graphics::Impl::SetHardwareScalingLevel(float level)
     {
-        if (level <= std::numeric_limits<float>::epsilon())
+        if (level < 1) // Disallow upscaling.
```
syntheticmagus, line 273: Are we sure we want to categorically disable hardware upscaling? And, if so, is this the right place to do it?

The reason I ask is because the `Graphics` object is the very lowest level of rendering abstraction we have above bgfx, which we essentially treat as, "the platform." Disallowing hardware upscaling here is tantamount to a declaration that nobody anywhere ever, on any platform, should be using hardware upscaling with Babylon Native. This seems like quite a strong statement, particularly because [the Babylon.js docs imply this should be possible](https://doc.babylonjs.com/typedoc/classes/babylon.engine#sethardwarescalinglevel), and it makes me wonder if this shouldn't perhaps be handled at a higher level, such as a platform-specific N-API error coming from `NativeEngine` or something.

But if we do decide that hardware upscaling should be disabled from here, we should probably change this method's name to `SetHardwareDownscalingLevel` or something, just to clarify from the outside what the contract is.

Drigax: Good point, I'm in an awkward position here due to this device specific issue. My current options are either: 

1) Disable any codepaths that cause this on any device for every device, which is admitedly the nuclear option here.
2) Don't do anything
3) add logic to disable this codepath on the platform that this issue has been observed on
4) root cause why this particular device crashes

syntheticmagus: I like option 2. 😀 

Jokes aside, I don't know the context that led to this bug and change, so please don't let my comment stop you if this is blocking somebody; we can always circle back to this later if needs be. My inclination is to think that the described behavior (the rebooting) constitutes a usage-based platform bug on a narrow band of platforms, and that the best option might just be for the developer to not do the thing that's going to trigger their platform's bug. It's not clear to me that preventing errant usage is part of our business.

That said, this is a pretty hefty consequence for errant usage, especially since the "errant" usage is based on a platform bug. One option might be to put a `console.warn(...)` about this in NativeEngine.ts; this will avoid having us lock down capabilities based on a rare platform bug while still allowing any developer encountering the issue a better chance to understand what's going on and (hopefully) decide to not do what they were doing that led them to that point.

Again, though, I don't know the context behind this investigation; so if there's a usage-informed concern being addressed here, I'm happy for us to just solve the problem at hand and revisit the philosophical quandaries later.

Drigax: Hey, I definitely think we should strongly consider our options before disabling a feature across all our platforms before moving forward.

Looking at the bug linked to this PR (and the AOSP issue linked to the bug):
https://github.com/BabylonJS/BabylonNative/issues/601
https://issuetracker.google.com/issues/180116488

I'd reckon that option 1) is a bit overkill given that this does seem to be recognized as an issue with Android's OpenGL implementation.

@ryantrem originally reported this issue, but that was due to accidentally upscaling instead of downscaling on Android, I don't believe that we have any dependents that absolutely need 10x upscaling on android.

I think adding a warning is an appropriate amount of diligence, since we're atleast telling users "be careful doing this"

PR #606, opened 2021-02-18
Assign IDXGISwapChain1 to the ISwapChainPanelNative in Babylon React Native

nkari82:
This pull request solves several problems on the Windows version.

1. Problem passing the wrong window handle to the renderer.
2. Problems not being able to redraw when changing JavaScript.
3. Crash occurs when changing the window size in debug mode.


bghgary:
This is a contract change. @syntheticmagus @ryantrem @chrisfromwork @Drigax Any thoughts on this?

Core/Graphics/Source/GraphicsImpl.cpp
```diff
@@ -86,11 +86,14 @@ namespace Babylon
 
         if (!m_state.Bgfx.Initialized)
         {
+            auto& init{m_state.Bgfx.InitState};
+
             // Set the thread affinity (all other rendering operations must happen on this thread).
-            m_renderThreadAffinity = std::this_thread::get_id();
+            // If context is specified, it is checked in an external thread.
+            if (init.platformData.context == nullptr)
+                m_renderThreadAffinity = std::this_thread::get_id();
```
syntheticmagus, line 94: Can you clarify what intended usage is being described here?

syntheticmagus (commented):
Overall, this looks strongly aligned with things I think we want to do. I have one question about the thread affinity change. I also think that, since this seems to be a Windows-focused change, it may fit better in a platform-specific implementation file where all the `void*` variables can be replaced with real platform-native types. (This is the motivation behind the template specialization approach, which I believe @Drigax is building on in his PR right now.) So, I think this is very likely the way to go, but we should coordinate and time this with other PRs that are already in flight to ensure the best fit and least contention.

Core/Graphics/Source/GraphicsImpl.cpp
nkari82: If the device was created externally, I thought that thread affinity should be managed where it was created. Ultimately, it will disappear when multiview is supported...

ryantrem:
@nkari82 thanks for contributing this change! Regarding @syntheticmagus's comment:

> which I believe @Drigax is building on in his PR right now

We had a quick discussion on this yesterday and the current plan is to replace all the templated overrides with a single non-templated function that takes something like a `PlatformData` arg, which is defined per platform, along with a per platform function that takes a (platform specific) `PlatformData` and a `bgfx::PlatformData` and updates the `bgfx::PlatformData` from the Babylon platform-sepcific `PlatformData`. I expect that will make this change cleaner, so we don't have to introduce yet another overload.

Core/Graphics/Source/GraphicsImpl.cpp
syntheticmagus: I see. I think that makes sense, though I'm not sure I fully grasp all the implications; and I think it's aligned with the approach we want, though perhaps in a slightly different form. The changes coming in #591 will, I think, likely have significant implications for this because we're pretty dramatically refining the threading model. Under the new paradigm, the "render thread" *is* something that the Graphics object needs to know about in order to enforce that certain calls are executed correctly, but it should be possible for the "render thread" to change identities by calling `DisableRendering` on the old "render thread" and `EnableRendering` on the new one. We're hoping #591 will be able to go in quite soon, likely within a week. Since your change is also concerned with threading, at least tangentially, it might be best to wait for that PR to go in, after which we can operate on the improved and clarified threading paradigm it will introduce. Thoughts?

chrisfromwork:
@nkari82 I think we have fixed most of the bgfx bugs that we were working around with this code. Please reopen this pr if there are things in here we still need for improved react-native support.

PR #607, opened 2021-02-18
V8 inspector support for Win32

rgerd:
Get ready for BabylonNative debugging to reach a whole new level-- source-mapped everything, breakpoints, profiler, heap snapshots, and more!

Just start the app, go to `edge://inspect` and click inspect.

This is a pretty big PR, but these are the main things going on:
* @tymlipari Pulled in and dusted off a V8 inspector server implementation that provides a websocket server (built on [asio](https://github.com/chriskohlhoff/asio) + [llhttp](https://github.com/nodejs/llhttp)) which facilitates communication between the engine and DevTools.
* Added some CMake logic to generate the llhttp library since the native source is generated with TypeScript.
* Updated V8 because there are a lot of bug fixes and subtle features that have been introduced since we first pulled in the engine. This turned out to be crucial in getting the current inspector protocol properly supported.
* Checked in the babylon.max.js source map so that we can finally step through core code in the original typescript.
* Passing in the script URL so that V8 can correctly identify the origins of the scripts its running.

(To-do: write a documentation page in follow-up PR to guide people through the steps and features)

breakpoint
cpu-profiler


chrisfromwork:
> 	if(typeof exports === 'object' && typeof module === 'object')

Is there a published npm package for this babylon.max.js update?

---
Refers to: Apps/BabylonScripts/babylon.max.js:2 in 7eb19d3. [](commit_id = 7eb19d3a59e553231d0274939077d58202ef4a0d, deletion_comment = False)

CMakeLists.txt
```diff
@@ -24,6 +24,18 @@ else()
     message(FATAL_ERROR "Unrecognized platform: ${CMAKE_SYSTEM_NAME}")
 endif()
 
+if(${CMAKE_CXX_COMPILER} MATCHES "x86/cl.exe$")
+    set(ARCH "x86")
```
chrisfromwork, line 28: is this change specific to new dependencies? Can we get a comment here calling out what its for?

Dependencies/napi/napi-direct/source/env.cc
```diff
@@ -1,11 +1,18 @@
...
 {
-    Napi::Value Eval(Napi::Env env, const char* string, const char* sourceUrl)
+    Napi::Value Eval(Napi::Env env, const char* source, const char* sourceUrl)
     {
         napi_value result;
-        NAPI_THROW_IF_FAILED(env, napi_run_script(env, Napi::String::New(env, string), sourceUrl, &result));
+
+        // Append the source URL so V8 can locate the file.
+        std::ostringstream sourceWithURL;
+        sourceWithURL << source << "\n";
```
chrisfromwork, line 12: should this always be done or should it only be used for v8 scenarios behind a macro?

Core/AppRuntime/Source/v8/LICENSE_NODE
```diff
@@ -0,0 +1,19 @@
+Copyright Node.js contributors. All rights reserved.
+
```
chrisfromwork, line 2: is this logic distributed through a submodule/do we have to check this in?

Dependencies/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+# cmake 3.18+ to have the ARCHIVE_EXTRACT sub-command for files
+cmake_minimum_required(VERSION 3.18)
```
bghgary, line 2: Should we be consistent in what version of CMake we require? I.e. should we change this everywhere instead of just this file?

Dependencies/napi/CMakeLists.txt
```diff
@@ -3,8 +3,11 @@ cmake_minimum_required(VERSION 3.12)
 project(napi)
 
 # Set per-platform defaults if no JS engine is set.
-if (WIN32)
+if (WINDOWS_STORE)
     set(NAPI_JAVASCRIPT_ENGINE "Chakra" CACHE STRING "JavaScript engine for N-API.")
+elseif (WIN32)
+    set(NAPI_JAVASCRIPT_ENGINE "V8" CACHE STRING "JavaScript engine for N-API.")
+    set(V8_USE_INSPECTOR ON CACHE BOOL "Enable the V8 inspector.")
```
bghgary, line 10: This will make the default V8. Is that what we want?

CMakeLists.txt
bghgary: Looks like this was moved from the napi project? Do we need this logic in more than one place? Also, I'm not sure this is generic enough to be at the top.

Dependencies/napi/napi-direct/source/env.cc
```diff
@@ -1,11 +1,18 @@
...
+        sourceWithURL << "//# sourceURL=" << sourceUrl << "\n";
```
bghgary, line 13:
```suggestion
        sourceWithURL << source << std::endl;
        sourceWithURL << "//# sourceURL=" << sourceUrl << std::endl;
```

Dependencies/napi/napi-direct/source/env.cc
bghgary: This code should probably be inside the napi_run_script function for V8.

bghgary: Actually, does V8 have a way to specify this without changing the string?

rgerd: I scoured the v8 docs looking for a way to pass the source url to the engine and there doesn't seem to be a way. Fortunately at least the inspector seems to identify and hide the string from the source so it's basically invisible to the user.

Core/AppRuntime/Source/v8/LICENSE_NODE
syntheticmagus: If we do have to check this in, we probably also need to call it out in [NOTICE.md](https://github.com/BabylonJS/BabylonNative/blob/master/NOTICE.md).

Dependencies/CMakeLists.txt
rgerd: My reasoning here was that this is the only file that uses features up to this version, and we already have a handful of different cmake_minimum_required versions across BabylonNative (and then differences between our submodules), so we have a better idea of what parts of the cmake code need which version. That being said it looks like the root CMakeLists has the maximum minimum_required version across all of these files, so it makes sense to me to need to update that version as well.

CMakeLists.txt
rgerd: I moved it up so that I could have the ARCH variable to download node for llhttp in Dependencies/CMakeLists.txt
I see now though that this logic definitely doesn't belong at the top, but now I'm trying to figure out how to get the ARCH without just copying the logic...

Core/AppRuntime/CMakeLists.txt
```diff
@@ -1,37 +1,72 @@
...
+            "Source/v8/V8InspectorAgent.cpp"
+            "Source/v8/V8InspectorSocket.h"
+            "Source/v8/V8InspectorSocket.cpp"
+            "Source/v8/V8InspectorSocketServer.h"
+            "Source/v8/V8InspectorSocketServer.cpp"
+            "Source/v8/V8InspectorTCP.h"
+            "Source/v8/V8InspectorTCP.cpp"
+            "Source/v8/V8InspectorUtils.h"
+            "Source/v8/V8InspectorUtils.cpp")
+    endif()
```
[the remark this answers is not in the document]
rgerd: This felt more like runtime code since it's doing more server communication stuff than actually passing data over the engine boundary...
Although perhaps it could go into Dependencies instead?

Apps/Playground/CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ set(BABYLONSCRIPTS
     "../BabylonScripts/meshwriter.min.js"
     "../BabylonScripts/recast.js")
 
+set(BABYLONSCRIPTMAPS
```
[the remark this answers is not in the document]
rgerd: As far as devtools support goes I'm pretty sure we just need the source maps to be able to debug in typescript. Do you know if the declaration files offer more info to the inspector?

CMakeLists.txt
rgerd: I moved the logic into a function in `Dependencies/CMakeExtensions/CMakeLists.txt`. It's slightly nicer, but doesn't feel perfect. Any suggestions on how to improve this would be appreciated.

Core/AppRuntime/Source/AppRuntimeV8.cpp
```diff
@@ -1,73 +1,84 @@
...
+
+        // Use the isolate within a scope.
+        {
+            v8::Isolate::Scope isolate_scope{isolate};
+            v8::HandleScope isolate_handle_scope{isolate};
+            v8::Local<v8::Context> context = v8::Context::New(isolate);
+            v8::Context::Scope context_scope{context};
+
+#ifdef V8_USE_INSPECTOR
+            Babylon::V8InspectorAgent inspector{ *Module::Instance().Platform(), isolate, context, "BabylonNative", 5643 };
```
tymlipari, line 70: >5643 [](start = 116, length = 4)

This was a completely random port I picked when I was testing this out, we should consider making this a command line flag.

Apps/Playground/Win32/App.cpp
```diff
@@ -105,11 +105,12 @@ namespace
         inputBuffer = std::make_unique<InputManager<Babylon::AppRuntime>::InputBuffer>(*runtime);
 
         runtime->Dispatch([width, height, hWnd](Napi::Env env) {
+#ifndef V8_USE_INSPECTOR
```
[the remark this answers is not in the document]
tymlipari: Instead of gating entirely behind macros, could this be conditionally enabled at runtime? For node I believe they take like an `--inspect` flag or something to signal that the inspector support should be enabled.

---
In reply to: [580411539](https://github.com/BabylonJS/BabylonNative/pull/607#discussion_r580411539) [](ancestors = 580411539)

Dependencies/napi/napi-direct/source/js_native_api_v8.cc
```diff
@@ -2963,9 +2964,15 @@ napi_status napi_run_script(napi_env env,
...
-                            napi_value* result) {
-  // TODO: figure out what to do with source_url
-  return napi_run_script(env, script, result);
+                            napi_value* result) {    
+    // Append the source URL so V8 can locate the file.
+    std::ostringstream source_url_comment;
+    source_url_comment << std::endl << "//# sourceURL=" << source_url << std::endl;
+
+    const auto v8_script_string = v8::Local<v8::String>::Cast(v8impl::V8LocalValueFromJsValue(script));
+    const auto source_with_comment = v8::String::Concat(env->isolate, v8_script_string, OneByteString(env->isolate, source_url_comment.str().c_str(), source_url_comment.str().size()));
```
tymlipari, line 2973: >str() [](start = 169, length = 5)

nit: looks like `str()` returns a copy every time. I'd recommend caching the result of `source_url_comment.str()` in an intermediate so you don't do two copies of the underlying string.

Apps/Playground/Win32/App.cpp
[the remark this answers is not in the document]
bghgary: Perhaps we can do both. I would suspect not everyone would want the overhead of V8 inspector in their production apps.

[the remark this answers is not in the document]
tymlipari: I think that's reasonable, though what're your thoughts on the default behavior? If the runtime enablement is gated behind a flag, the overhead is only the binary size increase.  In that case I think if `NAPI_JAVASCRIPT_ENGINE=V8` is set, we should include it with an additional opt-out flag (`V8_NO_INSPECTOR`?) to strip the code for places where binary size matters.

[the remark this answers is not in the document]
bghgary: That's an option. Another option is that the V8 inspector is a separate static library dependency and if you choose not to bring it in, then it won't be included in the binary size.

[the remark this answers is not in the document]
rgerd: We don't currently have a way to pass information between the app and the runtime, I cleaned up the logic to reduce the usage of V8_USE_INSPECTOR to just here. You can conditionally include the inspector library when running CMake with `-D V8_INCLUDE_INSPECTOR=[ON|OFF]` (default is ON).

I've [filed this issue](https://github.com/BabylonJS/BabylonNative/issues/640) to track addition of an `AppRuntimeOptions` struct that we can use to communicate to the runtime to initialize the V8 inspector (assuming the static library is included), and then the App code here can just not polyfill the console if it chooses to initialize the V8 inspector.

Dependencies/napi/CMakeLists.txt
rgerd: I think it makes the most sense feature-wise to make V8 with inspector the default. If folks think otherwise I can remove the line.

Core/AppRuntime/Source/AppRuntimeV8.cpp
rgerd: I'm going to punt this until we have `AppRuntimeOptions`. Please see my other comment.

chrisfromwork:
> /* harmony import */ var _Maths_math_scalar__WEBPACK_IMPORTED_MODULE_5__ = __webpack_require__(/*! ../../Maths/math.scalar */ "./Maths/math.scalar.ts");

is there a @babylonjs/core npm package published with these changes? Did the process for compiling this file change?

---
Refers to: Apps/BabylonScripts/babylon.max.js:14695 in 418cbcd. [](commit_id = 418cbcd4a327157d4b0d2d280317696cd7e4cdae, deletion_comment = False)

Apps/Playground/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 set(BABYLONSCRIPTS
     "../BabylonScripts/babylon.glTF2FileLoader.js"
     "../BabylonScripts/babylon.max.js"
+    "../BabylonScripts/babylon.max.js.map"
```
chrisfromwork, line 4: Is this generated with normal babylon.max.js compilation? Is there any way through github actions to test that babylon.max.js.map matches the babylon.max.js file checked in?

Dependencies/napi/CMakeLists.txt
bghgary: I don't think we should unless all platforms do the same.

Apps/Playground/Win32/App.cpp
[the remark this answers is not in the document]
bghgary: Hmm, I'm not sure having AppRuntimeOptions is right for this. It's specific to V8. Would a static option specific to V8 be okay? I don't see why anyone would want two AppRuntimes and have them behave differently in terms of V8 inspector support.

Apps/Playground/CMakeLists.txt
rgerd: I'm not sure there's an easy way to test if the sourcemap matches the .js file. If we start consuming babylon.js via npm then we can just use the sourcemap file next to the .js file from inside the npm module...

I guess for now we could have @bghgary modify the CODEOWNERS file to include the .js.map file once this goes in, so we have a manual check to make sure both are updated together.

rgerd:
> > /* harmony import _/ var _Maths_math_scalar__WEBPACK_IMPORTED_MODULE_5__ = **webpack_require**(/_! ../../Maths/math.scalar */ "./Maths/math.scalar.ts");
> 
> is there a @babylonjs/core npm package published with these changes? Did the process for compiling this file change?
> 
> Refers to: Apps/BabylonScripts/babylon.max.js:14695 in [418cbcd](https://github.com/BabylonJS/BabylonNative/commit/418cbcd4a327157d4b0d2d280317696cd7e4cdae). [](commit_id = [418cbcd](https://github.com/BabylonJS/BabylonNative/commit/418cbcd4a327157d4b0d2d280317696cd7e4cdae), deletion_comment = False)

Updated Babylon.js to 5.0.0-alpha.14

Apps/Playground/CMakeLists.txt
bghgary: The CODEOWNERS file covers all the files in the BabylonScripts dir. It should already cover them.

Apps/Playground/Win32/App.cpp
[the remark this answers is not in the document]
rgerd: Cleaned it up using a V8-specific header. Feels better: https://github.com/BabylonJS/BabylonNative/pull/607/commits/ac13b7d56bfd18ebbc9991475b41d88414e3d18e

And also hits @tymlipari's comment about specifying the port.

Apps/Playground/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(BABYLONSCRIPTS
     "../BabylonScripts/babylon.glTF2FileLoader.js"
     "../BabylonScripts/babylon.max.js"
     "../BabylonScripts/babylon.max.js.map"
+    "../BabylonScripts/babylon.module.d.ts"
```
bghgary, line 5: Thanks for adding this. There are more `d.ts` files though (e.g. for the glTF loader). We can add this later though.

Apps/Playground/Win32/App.cpp
```diff
@@ -10,6 +10,11 @@
 #include <Shared/InputManager.h>
 
 #include <Babylon/AppRuntime.h>
+#if __has_include(<Babylon/AppRuntimeV8.h>)
+#include <Babylon/AppRuntimeV8.h>
+#define V8_ENGINE
+#endif
```
bghgary, line 16: I don't like these pragma conditions in this file. This specific one using Microsoft specific pragmas feels wrong to me. Also, what happens if someone uses V8 with Android? Does the inspector code work with Android?

Apps/Playground/Win32/App.cpp
```diff
@@ -107,18 +112,25 @@ namespace
...
-            Babylon::Polyfills::Console::Initialize(env, [](const char* message, auto) {
-                OutputDebugStringA(message);
-            });
-#endif
+            
+            // If we've initialized the inspector server, console messages 
+            // will be sent to the DevTools console.
+            if (!usingV8Inspector)
+            {
+                Babylon::Polyfills::Console::Initialize(env, [](const char* message, auto) {
```
bghgary, line 130: Can this be detected inside the Console initialize? For example, in the WIndow polyfill, we check to see if the window already has the method before overriding it. Can we do the same here?

Core/AppRuntime/Source/AppRuntimeV8.cpp
```diff
@@ -87,4 +93,19 @@ namespace Babylon
...
+
+    bool AppRuntimeV8::EnableInspector(unsigned short port)
+    {
+        s_inspectorPort = port;
+#ifdef V8_INSPECTOR_INCLUDED
+        return true;
+#else
+        return false;
+#endif
+    }
```
bghgary, line 105: This contract feels odd to me. Users can always call this function but it may not work if the code is not available?

Apps/Playground/CMakeLists.txt
rgerd: Yeah I think we should also bring in the corresponding map files in a separate PR.

Apps/Playground/Win32/App.cpp
rgerd: Checking if `console` is defined feels like the better option here. One thing I'll call out is that V8 defines the console object no matter what, so even if we don't use the inspector we won't see messages printed to the console. Maybe I can add another optional parameter called `overrideEngine` that defaults to false, just in case you really want your console callback called?

Apps/Playground/Win32/App.cpp
rgerd: That's a good question-- I'll have to set up Android Studio to find out, but shouldn't be hard at all to test.

I'm not really sure how we can keep App.cpp engine-agnostic without using this conditional include or some sort of options object in the AppRuntime layer. Is there a different pattern I can use here to call V8-specific functions? Perhaps we could do something like [ReactNativeHost](https://github.com/microsoft/react-native-windows/blob/26eac15f92dca279f78e9280ebe0aa4db9d81749/vnext/Microsoft.ReactNative/ReactApplication.cpp#L135-L146)'s instance settings.

bghgary: yeah, i'm not sure there is a good solution. @syntheticmagus @ryantrem any thoughts?

Apps/Playground/Win32/App.cpp
bghgary: I think if there is already a console object, then we should shim it instead of replace it. The console.log calls the callback and continue to call the existing console.log function? JavaScriptCore also defines a console object. I wonder if this will make it work in all cases.

Core/AppRuntime/Source/AppRuntimeV8.cpp
rgerd: Yeah, I wasn't sure if throwing an exception would be more appropriate here, since there isn't a clean way to tell from App.cpp if the static library is included or not, and since we might not have the inspector on Android...

Perhaps it would be best to test on Android and then just have the inspector code always included in the V8 build to avoid the extra logic at the cost of a small binary size bump...

chrisfromwork (approved):

Apps/Playground/Win32/App.cpp
rgerd: Added commit https://github.com/BabylonJS/BabylonNative/pull/607/commits/8b6ebb6d7e1d3d1d1ac127dd9a0a74ac2d53c70d shimming the console functions

syntheticmagus (commented):
Chatted with Robert about this offline, so just summarizing that conversation briefly. I think it might be cleaner to componentize the approach by moving responsibilities from `AppRuntimeV8` into a new plugin, `V8Inspector` or something like that. This would move away from a compiler-flag-based inclusion/exclusion strategy to a component-based strategy more closely matching what we do for other Babylon Native capabilities. It'll also bring a handful of other benefits to separation of concerns, V8 inspector reusability, etc. The only potential complication that can foresee will be if the inspector continues to need the V8 `Platform`; but even then, this can be solved by just storing a `Napi::External` to the `Platform` where the plugin can find it, much like we do with `Graphics` and `JsRuntime`.

Dependencies/napi/napi-direct/CMakeLists.txt
```diff
@@ -26,48 +26,50 @@ endif()
...
+    target_include_directories(v8 INTERFACE "${NUGET_PATH}/v8-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/include")
+
+    set(NAPI_JAVASCRIPT_RUNTIME_OUTPUT_ARTIFACTS
+        "${NUGET_PATH}/v8.redist-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/lib/Release/icudtl.dat"
+        "${NUGET_PATH}/v8.redist-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/lib/Release/icui18n.dll"
+        "${NUGET_PATH}/v8.redist-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/lib/Release/icuuc.dll"
+        "${NUGET_PATH}/v8.redist-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/lib/Release/v8.dll"
+        "${NUGET_PATH}/v8.redist-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/lib/Release/v8_libbase.dll"
+        "${NUGET_PATH}/v8.redist-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/lib/Release/v8_libplatform.dll"
+        "${NUGET_PATH}/v8.redist-${VS_PLAT_TOOLSET}-${ARCH}.${V8_VERSION}/lib/Release/zlib.dll"
```
bghgary, line 54: maybe I missed something, but where does `${ARCH}` get set now?

Dependencies/v8inspector/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
+# ------------------------------- asio -------------------------------
+add_library(asio INTERFACE)
+target_include_directories(asio INTERFACE "Dependencies/asio/asio/include")
+set_property(TARGET asio PROPERTY FOLDER Dependencies)
+
+# ------------------------------ llhttp ------------------------------
+# Install node to use npm
+set(NODE_VERSION "v14.15.5")
+set_win32_arch()
```
bghgary, line 9: does this mean this v8inspector module won't work with Android?

Dependencies/v8inspector/Include/V8InspectorAgent.h
```diff
@@ -0,0 +1,42 @@
...
+
+        bool IsStarted();
+        bool IsConnected();
+        void WaitForDisconnect();
+        void FatalException(
+            v8::Local<v8::Value> error,
+            v8::Local<v8::Message> message);
+
+    private:
+        std::shared_ptr<AgentImpl> impl;
```
bghgary, line 40: Why shared_ptr and not unique_ptr? Do we need V8InspectorAgent to be copyable?

Dependencies/v8inspector/Source/V8InspectorAgent.cpp
```diff
@@ -0,0 +1,759 @@
...
+
+        v8::Platform& platform_;
+
+        friend class ChannelImpl;
+        friend class DispatchOnInspectorBackendTask;
+        friend class SetConnectedTask;
+        friend class V8NodeInspector;
+        friend void InterruptCallback(v8::Isolate*, void* agent);
+
+    public:
```
bghgary, line 122: ?

Dependencies/v8inspector/Source/V8InspectorAgent.cpp
```diff
@@ -0,0 +1,759 @@
...
+    {
+        return script_name_.empty() ? GetProcessTitle() : script_name_;
+    }
+
+    std::string InspectorAgentDelegate::GetTargetUrl(const std::string& /*id*/)
+    {
+        return "file://" + script_path_;
+    }
+
+} // namespace inspector
```
bghgary, line 759:
```suggestion
} // namespace Babylon
```

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+        Napi::Value SupportsInspector(const Napi::CallbackInfo& info)
+        {
+            return Napi::Boolean::From(info.Env(), true);
+        }
+
+        void StartInspector(const Napi::CallbackInfo &info)
+        {
+            if (info.Length() == 0)
+                throw Napi::Error::New(info.Env(), "You must at least specify a port to start the DevTools inspector.");
```
bghgary, line 29:
```suggestion
            if (info.Length() == 0)
            {
                throw Napi::Error::New(info.Env(), "You must at least specify a port to start the DevTools inspector.");
            }
```

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+            if (m_inspector->IsStarted())
+            {
+                StopInspector();
+            }
+
+            const auto port = static_cast<unsigned short>(info[0].As<Napi::Number>().Int32Value());
```
bghgary, line 36: This should probably happen before stopping the inspector.

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+            std::string appName;
+            if (info.Length() > 1 && info[1].IsString())
+                appName = info[1].As<Napi::String>().Utf8Value();
```
bghgary, line 40:
```suggestion
            if (info.Length() > 1 && info[1].IsString())
            {
                appName = info[1].As<Napi::String>().Utf8Value();
            }
```

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -0,0 +1,94 @@
...
+            
+            m_inspector->start(port, appName);
+        }
+
+        void StopInspector(const Napi::CallbackInfo& info)
+        {
+            (void)info;
+            StopInspector();
+        }
```
bghgary, line 49:
```suggestion
        void StopInspector(const Napi::CallbackInfo&)
        {
            StopInspector();
        }
```

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+        void StopInspector()
+        {
+            m_inspector->stop();
+        }
+
+        ~ChromeDevToolsV8()
+        {
+            if (m_inspector != nullptr)
+                StopInspector();
```
bghgary, line 59:
```suggestion
            if (m_inspector != nullptr)
            {
                StopInspector();
            }
```

Please update all single line if statements or for loops to use braces.

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -0,0 +1,94 @@
...
+        }
+    protected:
+        std::unique_ptr<Babylon::V8InspectorAgent> m_inspector = nullptr;
```
bghgary, line 62:
```suggestion
        std::unique_ptr<Babylon::V8InspectorAgent> m_inspector{};
```

bghgary (commented):
Initial comments. Can do another pass later.

Polyfills/Console/Source/Console.h
```diff
@@ -20,7 +20,9 @@ namespace Babylon::Polyfills::Internal
         void Warn(const Napi::CallbackInfo& info);
         void Error(const Napi::CallbackInfo& info);
         void InvokeCallback(const Napi::CallbackInfo& info, Babylon::Polyfills::Console::LogLevel logLevel) const;
+        void InvokeEngineCallback(const std::string functionName, const Napi::CallbackInfo& info);
 
+        std::unique_ptr<Napi::FunctionReference> m_engineConsole;
```
syntheticmagus, line 25: nit:
```suggestion
        std::unique_ptr<Napi::FunctionReference> m_engineConsole{};
```

syntheticmagus (approved):
A few relatively minor comments, but overall it looks great to me. V8 debugging is something we've wanted in Babylon Native for almost as long as the project's been open source, so this is awesome!

Dependencies/napi/napi-direct/CMakeLists.txt
rgerd: The function takes `${ARCH}` as a parameter, supplied on line 64 via the `${WIN32_ARCH}` variable set by `set_win32_arch()` on line 62

Dependencies/v8inspector/CMakeLists.txt
rgerd: Good point-- this line doesn't exclude the ability to work on android, it just won't build on OSX or Linux, but this should be simple to fix in a follow-up PR, should just need to download the OSX/Linux version of node and run the correct command.

Apps/Playground/CMakeLists.txt
```diff
@@ -111,6 +113,7 @@ target_include_directories(Playground PRIVATE "Source" ".")
 target_link_to_dependencies(Playground
     PRIVATE AppRuntime
     PRIVATE NativeCapture
+    PRIVATE ChromeDevTools
```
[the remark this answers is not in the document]
rgerd: `ChromeDevToolsNull` provides a none-supported implementation so that you don't need to conditionally include for the sake of simplicity.

Dependencies/CMakeExtensions/CMakeLists.txt
```diff
@@ -109,3 +109,18 @@ function(disable_warnings target)
...
+    elseif(${CMAKE_CXX_COMPILER} MATCHES "x64/cl.exe$")
+        set(WIN32_ARCH "x64" PARENT_SCOPE)
+    elseif(${CMAKE_CXX_COMPILER} MATCHES "arm/cl.exe$")
+        set(WIN32_ARCH "ARM" PARENT_SCOPE)
+    elseif(${CMAKE_CXX_COMPILER} MATCHES "arm64/cl.exe$")
+        set(WIN32_ARCH "ARM64" PARENT_SCOPE)
+    else()
+        message(FATAL_ERROR "Unrecognized compiler: ${CMAKE_CXX_COMPILER}")
+    endif()
+endfunction()
```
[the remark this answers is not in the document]
rgerd: Isn't it already in the CMakeExtensions folder?

Dependencies/v8inspector/CMakeLists.txt
rgerd: I'm just going to limit ChromeDevTools to WIN32 for now so we don't run into any issues with Android until we have time to test its compatibility. In theory it should just work after some small fixes, but I wouldn't be confident without spending some time checking.

Core/AppRuntime/Source/AppRuntimeV8.cpp
```diff
@@ -1,73 +1,89 @@
...
+        {
+            v8::Isolate::Scope isolate_scope{isolate};
+            v8::HandleScope isolate_handle_scope{isolate};
+            v8::Local<v8::Context> context = v8::Context::New(isolate);
+            v8::Context::Scope context_scope{context};
+
+            Napi::Env env = Napi::Attach(context);
+            Dispatch([](Napi::Env env) {
+                Module::Instance().AddPlatformToJavascript(env);
+            });
```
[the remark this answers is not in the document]
rgerd: Hmmm I was getting a memory access exception when trying to access the JSRuntime NativeObject. I thought it was specifically because I wasn't on the JS thread so I used Dispatch and it worked. Maybe it could be an order of operations thing?

Dependencies/v8inspector/Include/V8InspectorAgent.h
rgerd: I'm honestly not totally sure. Switching to unique_ptr doesn't raise any compile or init-time issues, but an exception gets raised on teardown, so I assume there's something happening with references getting passed around

Dependencies/CMakeExtensions/CMakeLists.txt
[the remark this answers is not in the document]
bghgary: Yes. Yes it is. 😳

Plugins/ChromeDevTools/Source/ChromeDevToolsImpl.h
```diff
@@ -0,0 +1,17 @@
...
+    {
+        class Impl;
+    public:
+        static void Initialize(Napi::Env env);
+        ChromeDevTools(const Napi::CallbackInfo& info);
+        static constexpr auto JS_INSTANCE_NAME = "chromeDevTools";
+        Napi::Value SupportsInspector(const Napi::CallbackInfo& info);
+        void StartInspector(const Napi::CallbackInfo &info);
+        void StopInspector(const Napi::CallbackInfo& info);
+        std::unique_ptr<Impl> m_impl;
```
bghgary, line 15: should this be public?

Apps/Playground/Scripts/experience.js
```diff
@@ -51,6 +53,10 @@ function CreateInputHandling(scene) {
     });
 }
 
+if (chromeDevTools.supportsInspector()) {
+    chromeDevTools.startInspector(5643, "experience.js");
+}
```
bghgary, line 58: I'm not sure how I feel about this. I kind of think this code should be in native.

rgerd: Do you mean in `_native` or in `App.cpp`?

bghgary: Latter.

PR #608, opened 2021-02-18
createImageBitmap and resize image impl

### LINKED ISSUE #524: createImageBitmap unimplemented
Follow up on this thread: https://forum.babylonjs.com/t/babylon-native-on-ubuntu/15615/76
`createImageBitmap` is unimplemented which prevent to lo create ground from heightmap
`var ground = new BABYLON.Mesh.CreateGroundFromHeightMap("ground", "myimage.png", 500, 500, 5, 0, 10, scene, false);`

Some infos here:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap

CedricGuillemet:
fixes #524
babylonjs at commit https://github.com/BabylonJS/Babylon.js/pull/9933/commits/06c6b04d9f2f9fceebb6c8a7f3eac5aa5cc916f6
I also use this PR as a learning and experiment on JS abstraction for future dev(Canvas)

Works in conjunction with https://github.com/BabylonJS/Babylon.js/pull/9933
2 new functions added in Engine:
- createImageBitmap
- resizeImageBitmap

For the later, I'm using a 3rd party function to resize.
New validation test scene.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1532,4 +1536,98 @@ namespace Babylon
     {
         return Napi::Value::From(info.Env(), m_graphicsImpl.GetHardwareScalingLevel());
     }
+
+    Napi::Value NativeEngine::CreateImageBitmap(const Napi::CallbackInfo& info)
+    {
+        auto env = info.Env();
```
ryantrem, line 1542: Can this be `const` or no?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1532,4 +1536,98 @@ namespace Babylon
...
+        {
+            throw Napi::Error::New(env, "CreateImageBitmap first parameter must be an array.");
+        }
+        const auto parameters = info[0].As<Napi::Array>();
+        if (!parameters.Length())
+        {
+            throw Napi::Error::New(env, "CreateImageBitmap parameters are empty.");
+        }
+
+        const auto blob = parameters[0u].As<Napi::Array>();
```
ryantrem, line 1553: Should we be checking that this is actually an array (like we do for `info[0]`)?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1532,4 +1536,98 @@ namespace Babylon
...
+            throw Napi::Error::New(env, "CreateImageBitmap blob parameter is empty.");
+        }
+
+        const auto data = blob[0u].As<Napi::ArrayBuffer>();
+        if (!data.ByteLength())
+        {
+            throw Napi::Error::New(env, "CreateImageBitmap array buffer is empty.");
+        }
+
+        bimg::ImageContainer* image = bimg::imageParse(&m_allocator, data.Data(), static_cast<uint32_t>(data.ByteLength()));
```
ryantrem, line 1565: What is the image format in this case?

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Yes! Setting a few others as const as well.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: I'm not sure to understand. image here will contain decoded data from coded data

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Yes, I'm adding a check.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1532,4 +1536,103 @@ namespace Babylon
     {
         return Napi::Value::From(info.Env(), m_graphicsImpl.GetHardwareScalingLevel());
     }
+
+    Napi::Value NativeEngine::CreateImageBitmap(const Napi::CallbackInfo& info)
+    {
+        const auto env = info.Env();
```
bghgary, line 1542: nit: we are trying to use initializer lists everywhere and also don't use auto unless it's blatantly obvious what the type is
```suggestion
        const Napi::Env env{info.Env()};
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1532,4 +1536,103 @@ namespace Babylon
...
+        {
+            throw Napi::Error::New(env, "CreateImageBitmap first parameter must be an array.");
+        }
+        const auto parameters = info[0].As<Napi::Array>();
+        if (!parameters.Length())
+        {
+            throw Napi::Error::New(env, "CreateImageBitmap parameters are empty.");
+        }
+
+        const auto blob = parameters[0u].As<Napi::Array>();
```
bghgary, line 1553: Why do you need a `u`?

bghgary, line 1553: Is the intention to support everything that browser can support? Why pass in a blob at all? If we only want to support arraybuffer, do the blob checking in JS and pass in an arraybuffer.

CedricGuillemet: operator [] needs an unsigned value. 0 is signed int.

Plugins/NativeEngine/CMakeLists.txt
```diff
@@ -26,7 +26,8 @@ set(SOURCES
 
 add_library(NativeEngine ${SOURCES})
 
-target_include_directories(NativeEngine PUBLIC "Include")
+target_include_directories(NativeEngine PUBLIC "Include"
+    PRIVATE "${BIMG_DIR}/3rdparty")
```
syntheticmagus, line 30: This seems a little odd. Could we instead either expose STB as a target from bgfx.cmake? Or, if not there, could we add a bespoke target in dependencies/CMakeLists.txt like we do for base-n? I just think it might be best to remove the responsibility of understanding the structure of bimg from NativeEngine.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -11,6 +11,8 @@
 #include <bimg/decode.h>
 #include <bimg/encode.h>
 
+#define STB_IMAGE_RESIZE_IMPLEMENTATION
```
syntheticmagus, line 14: If we add a CMake target as described in my other comment, this could be defined in that target so that we don't have to worry about it here.

Plugins/NativeEngine/CMakeLists.txt
CedricGuillemet: 3rdparty directory already added in bgfx.cmake. I don't know why it didn't work 1st time I tried to include the .h. Anyway, it's fixed now.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: What's annoying with STB include file is only 1 .cpp file in the solution must define `STB_IMAGE_RESIZE_IMPLEMENTATION` or any other include will popup link error. I prefer to let this define clearly visible just before the #include statement.

syntheticmagus: I see. Can you note this nuance in a comment just above the define? That'll just help future developers know that we usually handle compiler defines differently and that this one is the exception rather than the rule.

syntheticmagus (approved):

PR #609, opened 2021-02-18
Introduce mechanism for obtaining XrContext off of XRSession

chrisfromwork:
For BabylonReactNative custom extensions we will need access to the underlying OpenXR XrInstance as well as other types. This pr demonstrates how we can expose a void pointer to a helper class specific to the xr session to obtain these types.

* IXrContext.h defines an ABI safe interface (work in progress) definition for the OpenXR XRSession.nativeExtension object
* Different xr platform implementations can populate their own underlying XRSession.nativeExtension and XRSession.nativeExtensionType values
* IXrContext.h and its dependency headers should be distributed in some form so that custom react-native-windows plugins can consume them

TODOs:
* we still need to define a model for enabling custom OpenXR extensions. In this review I have added XR_MSFT_PERCEPTION_ANCHOR_INTEROP_EXTENSION_NAME as an additional extension to enable surface reconstruction scenarios in custom react-native-windows plugins. In an ideal world, the custom plugin would be able to enable this OpenXR extension itself compared to requiring a change to BabylonNative
* we want to pick up napi changes from this review: https://github.com/BabylonJS/BabylonNative/pull/621


tymlipari:
I generally like the direction this is moving in. One comment though, I think we'll need this abstraction for more than just OpenXR, likely also for ARKit and ARCore too as we look into doing things like Azure service integration.

An approach that would work here would be to have `IXrContext` be a generic interface for all platforms, and then platform/framework-specific interfaces to get at the guts of each XR framework. So something like:

```cpp
struct IXrSession
{
    virtual bool TryGetContext(Uuid interfaceId, void** interfaceImpl) = 0;
};

struct IXrContextOpenXR
{
    virtual bool OPENXR_CONTEXT_INTERFACE_API IsInitialized() const = 0;
    ...
};

struct IXrContextARKit
{
    ...
};
```

If we pursue that path, another design piece worth considering is how to identify the different interface kinds. We _could_ just use an enum:

```cpp
enum class XrContextKind
{
    Unknown,
    OpenXR,
    ARKit
};

struct IXrSession
{
    virtual void TryGetContext(XrContextKind kind, void** contextImpl) = 0;
};
```

The challenge with this is if we ever need to add additional functions to the interfaces in the future. Because of ABI stability issues, you'd generally want to do subsequent additions as a separate versioned interface, making the enum path harder.

chrisfromwork:
> I generally like the direction this is moving in. One comment though, I think we'll need this abstraction for more than just OpenXR, likely also for ARKit and ARCore too as we look into doing things like Azure service integration.
> 
> An approach that would work here would be to have `IXrContext` be a generic interface for all platforms, and then platform/framework-specific interfaces to get at the guts of each XR framework. So something like:
> 
> ```c++
> struct IXrSession
> {
>     virtual bool TryGetContext(Uuid interfaceId, void** interfaceImpl) = 0;
> };
> 
> struct IXrContextOpenXR
> {
>     virtual bool OPENXR_CONTEXT_INTERFACE_API IsInitialized() const = 0;
>     ...
> };
> 
> struct IXrContextARKit
> {
>     ...
> };
> ```
> 
> If we pursue that path, another design piece worth considering is how to identify the different interface kinds. We _could_ just use an enum:
> 
> ```c++
> enum class XrContextKind
> {
>     Unknown,
>     OpenXR,
>     ARKit
> };
> 
> struct IXrSession
> {
>     virtual bool TryGetContext(XrContextKind kind, void** contextImpl) = 0;
> };
> ```
> 
> The challenge with this is if we ever need to add additional functions to the interfaces in the future. Because of ABI stability issues, you'd generally want to do subsequent additions as a separate versioned interface, making the enum path harder.

I'm not a huge fan of the enum path. If we end up exposing any amount of this information in the javascript layer it'll likely end up a string.

Do we really need different platforms to reference a shared header? XrContext.h is only included in openxr/uwp cmake generated BabylonNative builds currently. We will always have platform specific content referenced in IXrContext* definitions. we are exposing a nativeExtensionType string that can be used to infer what platform the void pointer corresponds to.

Dependencies/xr/Source/OpenXR/XrSupportedExtensions.h
```diff
@@ -28,6 +28,9 @@ namespace xr
             DepthExtensionSupported = TryEnableExtension(XR_KHR_COMPOSITION_LAYER_DEPTH_EXTENSION_NAME);
             UnboundedRefSpaceSupported = TryEnableExtension(XR_MSFT_UNBOUNDED_REFERENCE_SPACE_EXTENSION_NAME);
             SpatialAnchorSupported = TryEnableExtension(XR_MSFT_SPATIAL_ANCHOR_EXTENSION_NAME);
+            SpatialAnchorInteropSupported =
+                TryEnableExtension(XR_MSFT_PERCEPTION_ANCHOR_INTEROP_EXTENSION_NAME) ||
+                TryEnableExtension("XR_MSFT_perception_anchor_interop_preview");
```
chrisfromwork, line 33: TODO: revert

Dependencies/xr/Source/OpenXR/include/IXrContext.h
```diff
@@ -0,0 +1,30 @@
...
+    #define OPENXR_CONTEXT_INTERFACE_API __stdcall
+#elif defined(__MACH__) || defined(__ANDROID__) || defined(__linux__) || defined(LUMIN)
+    #define OPENXR_CONTEXT_INTERFACE_API
+#else
+    #define OPENXR_CONTEXT_INTERFACE_API
+#endif
+
+typedef struct IXrContext
+{
+    virtual bool OPENXR_CONTEXT_INTERFACE_API IsInitialized() const = 0;
```
yl-msft, line 19: will the consumer side use this bool differently than simply test "Instance() == 0"?

yl-msft (approved):

Dependencies/xr/Source/OpenXR/include/IXrContext.h
chrisfromwork: we currently use this to ensure the session is running

Dependencies/xr/Source/OpenXR/include/XrContextHelper.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "jsi/jsi.h"
+#include "IXrContext.h"
+
+namespace BabylonReactNative
+{
+    bool TryGetXrContext(facebook::jsi::Runtime& jsiRuntime, const facebook::jsi::Value& xrSession, IXrContext*& xrContext)
```
tymlipari, line 8: >bool TryGetXrContext [](start = 4, length = 20)

Is this header meant to be the entry point for native modules? If so, perhaps we should use a `#if __has_include("jsi/jsi.h")` guard against the JSI stuff (thinking of a future where we might want an equivalent NAPI version or something).

Dependencies/xr/Source/OpenXR/XrRegistry.h
```diff
@@ -1,25 +1,31 @@
...
-        bool IsInitialized() const;
-        XrInstance Instance() const;
-        XrSystemId SystemId() const;
-        XrTime DisplayTime() const;
-        const std::unique_ptr<XrSupportedExtensions>& Extensions() const;
-        const XrSession Session() const;
-        const XrSessionState State() const;
-        const XrSpace Space() const;
+
+        virtual bool OPENXR_CONTEXT_INTERFACE_API IsInitialized() const override;
```
tymlipari, line 15: >virtual [](start = 8, length = 7)

nit: `override` implies `virtual` so this is extraneous. Not sure what the style preferences are for this repo though.

Dependencies/xr/Source/OpenXR/include/XrContextHelper.h
chrisfromwork: we assume that this header will only be used when jsi is present. It isn't built in any BabylonNative lib currently. this safe guard seems like a cool addition though, so i will look at to add it :)

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -60,21 +60,28 @@ namespace xr
...
+    bool OPENXR_CONTEXT_INTERFACE_API XrSessionContext::IsInitialized() const
     {
         return ContextImpl->Session.Get() != XR_NULL_HANDLE &&
             ContextImpl->SceneSpace.Get() != XR_NULL_HANDLE &&
             ContextImpl->State != XrSessionState::XR_SESSION_STATE_UNKNOWN;
     }
-    XrInstance XrSessionContext::Instance() const { return ContextImpl->Instance.Get(); }
-    XrSystemId XrSessionContext::SystemId() const { return ContextImpl->SystemId; }
-    XrTime XrSessionContext::DisplayTime() const { return ContextImpl->DisplayTime; }
+    xr::ExtensionDispatchTable* OPENXR_CONTEXT_INTERFACE_API XrSessionContext::ExtensionDispatchTable() const { return ContextImpl->Extensions.get(); }
```
tymlipari, line 69: >r::ExtensionDis [](start = 5, length = 15)

Maybe space these out a little?

Dependencies/xr/Source/OpenXR/include/IXrContext.h
```diff
@@ -0,0 +1,30 @@
...
+    #define OPENXR_CONTEXT_INTERFACE_API __stdcall
+#elif defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(_WIN64) || defined(WINAPI_FAMILY)
+    #define OPENXR_CONTEXT_INTERFACE_API __stdcall
+#elif defined(__MACH__) || defined(__ANDROID__) || defined(__linux__) || defined(LUMIN)
+    #define OPENXR_CONTEXT_INTERFACE_API
+#else
+    #define OPENXR_CONTEXT_INTERFACE_API
+#endif
+
+typedef struct IXrContext
```
tymlipari, line 17: >IXrContext [](start = 15, length = 10)

Discussed offline, let's disambiguate this name to be OpenXR-specific

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2310,7 +2322,7 @@ namespace Babylon
                         InstanceMethod("trySetFeaturePointCloudEnabled", &XRSession::TrySetFeaturePointCloudEnabled),
                         InstanceMethod("trySetPreferredPlaneDetectorOptions", &XRSession::TrySetPreferredPlaneDetectorOptions),
                         InstanceMethod("trySetMeshDetectorEnabled", &XRSession::TrySetMeshDetectorEnabled),
-                        InstanceMethod("trySetPreferredMeshDetectorOptions", &XRSession::TrySetPreferredMeshDetectorOptions),
+                        InstanceMethod("trySetPreferredMeshDetectorOptions", &XRSession::TrySetPreferredMeshDetectorOptions)
```
tymlipari, line 2325: >) [](start = 123, length = 1)

nit: may want to restore the `,` it's useful to leave a trailing one so that future diffs and merge conflicts don't have issues

tymlipari (approved):
:shipit:

Dependencies/xr/Source/OpenXR/include/IXrContext.h
chrisfromwork: sounds good

Dependencies/xr/Source/OpenXR/Include/XrContextHelperOpenXR.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#if __has_include("jsi/jsi.h")
```
ryantrem, line 3: It didn't click for me earlier when we chatted that this would be in BabylonNative. Would it make more sense to just include this file in BabylonReactNative? It's a bit weird to split it that way, but it's also a bit weird to have any publicly exposed interface that uses jsi directly in BabylonNative.

Dependencies/xr/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ else()
     endif()
 
     set(SOURCES ${SOURCES}
+        "Source/OpenXR/Include/IXrContextOpenXR.h"
+        "Source/OpenXR/Include/XrContextHelperOpenXR.h"
```
syntheticmagus, line 30: This line should be excluded, right? This isn't for Babylon Native, it's for other consumers, unless I misunderstand its purpose.

chrisfromwork: this is only added for easing the developer process. This file doesn't get included in any static libs, the helper function also isn't enabled by default in BabylonNative. it does however allow you to view and edit the file in the cmake generated visual studio solution

chrisfromwork: moving to BabylonReactNative

Dependencies/xr/Source/OpenXR/Include/XrContextHelperOpenXR.h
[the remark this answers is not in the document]
chrisfromwork: moving to BabylonReactNative

chrisfromwork: moving to BabylonReactNative

PR #612, opened 2021-02-19
Split GraphicsImpl into its own cpp file

syntheticmagus:
Split GraphicsImpl into its own cpp file. Also created platform-specific headers to avoid ifdefs.


ryantrem (approved):

Core/Graphics/CMakeLists.txt
```diff
@@ -1,15 +1,18 @@
 set(SOURCES
     "Include/Babylon/Graphics.h"
+    "Source/PlatformSpecificHeaders/${BABYLON_NATIVE_PLATFORM}/GraphicsPlatform.h"
```
bghgary, line 3: @Drigax is already doing [something similar](https://github.com/BabylonJS/BabylonNative/pull/582/files?file-filters%5B%5D=.txt#diff-1c56f1023b5f001b105f47bb306a1daf4de97640a11e11d339fe5db298e9df17) in his PR. I much like the pattern in his PR than this pattern. Thoughts?

syntheticmagus: Spoke offline, restructured folders per recommendation.

Core/Graphics/Source/Android/GraphicsPlatform.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <bgfx/bgfx.h>
+
+// MSAA is disabled on Android.
+// See issue https://github.com/BabylonJS/BabylonNative/issues/494#issuecomment-731135918
+// for explanation
+#define BGFX_RESET_FLAGS (BGFX_RESET_VSYNC | BGFX_RESET_MAXANISOTROPY)
```
bghgary, line 8: Not for this PR, but we probably shouldn't use #define for this. It should be a constexpr like the code below.

bghgary (approved):

PR #613, opened 2021-02-19
Plumbing hand controller data through XR.cpp

rickfromwork:
Hand poses were already being passed through for raycasting, this adds the hand selection states as well, so we can then query the controller associated with each hand in BJS for select and squeeze.


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1178,6 +1234,17 @@ namespace xr
 
             return true;
         }
+
+        void UpdatePoseData(xr::Pose& targetPose, XrPosef& sourcePose)
```
tymlipari, line 1238: Doesn't seem to be a reason for this to be a member function. Consider making this `static` or a free standing function.

tymlipari (approved):

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I made it a member function since it fits with the utility of the other member functions (like PopulateDepthInfoView)

rickfromwork: I think I'll keep it as-is to fit with the others.

PR #614, opened 2021-02-19
Native Camera 

CedricGuillemet:
- [x] iOS functional
- [x] Android functional
- [x] Camera front/back Android
- [x] Camera front/back iOS
- [x] Wait and test with Gary's threading PR (regarding UI access on ios and opengl context on android)
- [ ] Texture size limitation for iOS (not mandatory, is it?)

Apple texture override to be done in other PR


Plugins/NativeCamera/Source/NativeCamera.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+#include <bgfx/bgfx.h>
+
+namespace Babylon::Plugins::Internal
+{
+    // Move this struct to Graphics
+    struct TextureData final
```
CedricGuillemet, line 7: This structure will be moved into Graphics once Gary's PR is in.

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+        android::graphics::SurfaceTexture surfaceTexture;
+        android::view::Surface surface;
+
+        constexpr GLint GetTextureUnit(GLenum texture)
+        {
+            return texture - GL_TEXTURE0;
+        }
+
+        GLuint LoadShader(GLenum shader_type, const char* shader_source)
```
CedricGuillemet, line 87: This code is the same as NativeXR/ARCore. I don't know where I can move it? New dep library?

Plugins/NativeCamera/Source/NativeCamera.h
```diff
@@ -0,0 +1,30 @@
...
+        uint32_t Height{ 0 };
+        uint32_t Flags{ 0 };
+        uint8_t AnisotropicLevel{ 0 };
+    };
+
+    struct CameraInterface
+    {
+        virtual ~CameraInterface() = default;
+        static CameraInterface* CreateInterface(uint32_t width, uint32_t height, bool frontCamera);
+        virtual void UpdateCameraTexture(bgfx::TextureHandle textureHandle) = 0;
```
syntheticmagus, line 28: Are the virtuality, staticness, and raw pointer necessary? Unless I'm misunderstanding, there's never more than one implementation of this at a time (one per platform). If so, could you just declare an ordinary constructor and methods, then have platform-specific implementations of them without virtuality/staticness/raw pointers?

```
CameraInterface(uint32_t, uint32_t, bool);
~CameraInterface();
void UpdateCameraTexture(bgfx::TextureHandle);
```

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+        ~NativeCamera()
+        {
+        }
+
+    private:
+        Napi::Value CreateVideo(const Napi::CallbackInfo& info)
+        {
+            const auto env = info.Env();
+            Napi::Object video = Napi::Object::New(env);
```
syntheticmagus, line 76: nit: Would it make sense to split this out into an `ObjectWrap`, similar to what we do in a few cases in NativeXr? Since this thing has methods, it just seems a little hefty to be constructing programmatically, though it probably ultimately resolves to the same thing and is thus more of a stylistic choice than anything else.

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,166 @@
...
+            video.Set("readyState", Napi::Number::New(env, 10).As<Napi::Value>());
+            video.Set("HAVE_CURRENT_DATA", Napi::Number::New(env, 1).As<Napi::Value>());
+            video.Set("isNative", Napi::Boolean::New(env, true).As<Napi::Value>());
+
+            video.Set("setAttribute", Napi::Function::New(env,[](const Napi::CallbackInfo& /*info*/){
+                }));
+
+            video.Set("addEventListener", Napi::Function::New(env, [](const Napi::CallbackInfo& info) {
+                auto callback = info[1].As<Napi::Function>();
+                info.This().ToObject().Set("_cb", callback.As<Napi::Object>().As<Napi::Value>());
```
syntheticmagus, line 110: I'm not sure I understand this pattern. It looks like it's only possible for there to be a single callback here regardless of the argument at `info[0]`. Is this the intent?

syntheticmagus, line 110: I'm not sure I understand this pattern. It looks like it's only possible for there to be a single callback here regardless of the argument at `info[0]`. Is this the intent?

Plugins/NativeCamera/Source/NativeCameraApple.mm
```diff
@@ -0,0 +1,149 @@
...
+            [cameraInterfaceApple->avCaptureSession startRunning];
+        });
+        return cameraInterfaceApple;
+    }
+
+    void CameraInterfaceApple::UpdateCameraTexture(bgfx::TextureHandle textureHandle)
+    {
+        if (textureBGRA)
+        {
+            bgfx::overrideInternal(textureHandle, reinterpret_cast<uintptr_t>(textureBGRA));
```
syntheticmagus, line 137: Is this going to work when the JavaScript thread is no longer the graphics thread?

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,166 @@
...
+            auto videoObject = info[1].As<Napi::Object>();
+
+            auto cameraInterfaceObject = videoObject.Get(JS_NATIVECAMERA_DATA);
+            if (!cameraInterfaceObject.IsExternal())
+            {
+                const auto width = videoObject.Get("videoWidth").As<Napi::Number>().Uint32Value();
+                const auto height = videoObject.Get("videoHeight").As<Napi::Number>().Uint32Value();
+                const bool frontCamera = videoObject.Get("frontCamera").As<Napi::Boolean>().Value();
+
+                cameraInterfaceObject = Napi::External<CameraInterface>::New(info.Env(), CameraInterface::CreateInterface(width, height, frontCamera));
```
syntheticmagus, line 151: The lifecycle of this thing is kind of confusing. If it's destroyed on pause, should it be created on play?

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
```diff
@@ -0,0 +1,411 @@
...
+            glTexParameteri(GL_TEXTURE_EXTERNAL_OES, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+            glBindSampler(0, 0);
+
+            // Draw the quad
+            glDrawArrays(GL_TRIANGLE_STRIP, 0, 3);
+
+            glUseProgram(0);
+            glBindFramebuffer(GL_FRAMEBUFFER, 0);
+
+            bgfx::overrideInternal(textureHandle, cameraRGBATextureId);
```
syntheticmagus, line 403: As on the iOS one, is this going to encounter problems if we're not on the graphics thread?

syntheticmagus (commented):
Looks pretty awesome overall, but I have a few concerns particularly regarding how this will interact and scale with multithreading. I think, with a few changes, this can probably go in fairly quickly if people are waiting for it since (I think) it doesn't really expose a new JS-facing API. Longer-term (or shorter-term if we want), I think it might end up making sense to split this functionality across two components, one native to the graphics thread and one a JS projection of that, in a somewhat similar way to how `NativeCapture` is structured. Thoughts?

Plugins/NativeCamera/Source/NativeCamera.cpp
syntheticmagus: How are there two of this comment?

Plugins/NativeCamera/Source/NativeCamera.h
CedricGuillemet: That was my first guess but the allocation also contains specific data per platform. It's not like the shader compiler where only code per platform is different and data the same.

Plugins/NativeCamera/Source/NativeCamera.cpp
CedricGuillemet: yes, only 1 cb needed.

Plugins/NativeCamera/Source/NativeCameraApple.mm
CedricGuillemet: That's something I need to test after Gary's PR is in. I have big concerns about both ios and android platforms. UI access on ios and opengl context on android. I won't merge this PR before I know it works with new threading model.

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
CedricGuillemet: Same comment as above. Wait to test with new threading model.

Plugins/NativeCamera/Source/NativeCamera.h
CedricGuillemet: Actually, I'll work on moving that struct later.

Plugins/NativeCamera/Source/NativeCamera.cpp
CedricGuillemet: yes, might be less confusing.

Plugins/NativeCamera/Source/NativeCamera.h
```diff
@@ -0,0 +1,31 @@
...
+        uint32_t Width{ 0 };
+        uint32_t Height{ 0 };
+        uint32_t Flags{ 0 };
+        uint8_t AnisotropicLevel{ 0 };
+    };
+
+    struct CameraInterface
+    {
+        virtual ~CameraInterface() = default;
+        static CameraInterface* CreateInterface(Napi::Env env, uint32_t width, uint32_t height, bool frontCamera);
```
syntheticmagus, line 28: Could this return the `Napi::External` directly instead of a raw pointer? I think we conventionally try to avoid returning raw pointer so we can keep the logic for creation and destruction closer together.

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
syntheticmagus: AndroidExtensions, maybe? I don't think we ever intend to support OpenGL for any platform other than Android.

Plugins/NativeCamera/Source/NativeCamera.h
syntheticmagus: As an alternative, you could probably do that with a pImpl-like pattern, I think. Forward declare `Impl`, or `CameraInterfaceState`, or something like that in the header, give `CameraInterface` a unique pointer to it, then define the struct only in the platform-specific cpp file. I haven't written it out, but I think that might give you the best of both worlds.

CedricGuillemet: If I follow you, I create a new CameraInterface structure which will contain a pointer to an Impl. Then this impl is created by a function dependant on the platform ? Almost what's here except with another level of abstraction. I don't see the point here.

Apps/BabylonScripts/babylon.max.js
```diff
@@ -47401,7 +47401,7 @@ var ThinEngine = /** @class */ (function () {
          */
         // Not mixed with Version for tooling purpose.
         get: function () {
-            return "babylonjs@5.0.0-alpha.12";
+            return "babylonjs@5.0.0-alpha.13";
```
chrisfromwork, line 47404: Do we have a @babylonjs/core npm package published that contains these changes?

Plugins/NativeCamera/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+    "Source/NativeVideo.cpp"
+    "Source/NativeVideo.h")
+
+if(APPLE)
+    set(SOURCES ${SOURCES} "Source/NativeCamera${BABYLON_NATIVE_PLATFORM}.mm")
+else()
+    set(SOURCES ${SOURCES} "Source/NativeCamera${BABYLON_NATIVE_PLATFORM}.cpp")
+endif()
+
+if (APPLE OR ANDROID OR (WIN32 AND NOT WINDOWS_STORE))
```
syntheticmagus, line 14: nit: Not even really a nit, just an interesting option I didn't know about. Apparently CMake has a [`return`](https://cmake.org/cmake/help/v3.12/command/return.html) command, in case you wanted to early-out instead of encapsulating all the functionality in a condition.

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,117 @@
...
+                navigator = global.Get(JS_NAVIGATOR_NAME).As<Napi::Object>();
+            }
+            else
+            {
+                navigator = Napi::Object::New(env);
+                global.Set(JS_NAVIGATOR_NAME, navigator);
+            }
+
+            // append media devices to navigator
+            Napi::Object mediaDevices = Napi::Object::New(env);
```
syntheticmagus, line 48: nit: Should we check to make sure this doesn't exist before creating a new one? Probably doesn't matter right now, just thinking this might get confusing if something else ever tries to make media devices too.

Plugins/NativeCamera/Source/NativeCamera.h
syntheticmagus: nit: It should be the same amount of indirection, just without a static factory and with pImpl instead of virtuality. So the `CameraInterface` declaration becomes
```
class CameraInterface
{
public:
    CameraInterface(Napi::Env, uint32_t, uint32_t, bool);
    void UpdateCameraTexture(bgfx::textureHandle);

private:
    struct Impl;
    std::unique_ptr<Impl> m_impl{};
}
```
and the definitions of everything, most importantly `Impl`, are platform-specific. At runtime, this should be basically the same as what you already have, so I don't think this is a huge deal either way. This approach just sticks closer to the patterns we've used most commonly elsewhere (avoiding both virtuality and staticness) and may help clarify from a contract standpoint that there's never actually more than one kind of `CameraInterface` in any given build. So, while I do think pImpl is better here, I'm not really concerned about it; the raw pointer was the only thing that originally made this comment more than a nit.

syntheticmagus (approved):

Plugins/NativeCamera/Source/NativeCamera.h
ryantrem: I would echo @syntheticmagus's comments here, but probably a bit more strongly. I think the inheritance pattern here is actually misleading, because it suggests there is a *runtime* abstraction, when in fact there is only a *compile* time abstraction. I think it would be much easier to quickly understand the code if it followed the same patterns we use elsewhere, both for the directory structure for platform specific vs. shared files and the pimpl pattern.

Apps/Playground/Android/app/build.gradle
```diff
@@ -18,7 +18,7 @@ if (project.hasProperty("JSEngine")) {
     }
 }
 
-def platformVersion = 21
+def platformVersion = 24
```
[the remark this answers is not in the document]
ryantrem: Do we need API level 24 for the NativeCamera stuff? Currently Power Apps has a min SDK version of 21, so it would probably impact usage there.

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
```diff
@@ -259,6 +261,18 @@ namespace android::content::res
...
 }
 
+namespace android::graphics
+{
+    class SurfaceTexture : public java::lang::Object
+    {
+    public:
+        SurfaceTexture();
+
+        void initWithTexture(int texture);
```
ryantrem, line 271: These wrappers are meant to map to the Java types 1:1, so I would expect a constructor that takes an `int texName`. Is there any reason you have a separate `initWithTexture` function?

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
```diff
@@ -259,6 +261,18 @@ namespace android::content::res
...
+        void updateTexture() const;
```
ryantrem, line 272: This should match the Java API naming:
```suggestion
        void updateTextImage() const;
```

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
```diff
@@ -277,6 +291,15 @@ namespace android::view
 
         Display getDefaultDisplay();
     };
+
+    class Surface : public java::lang::Object
+    {
+    public:
+        Surface();
+
+        void initWithSurfaceTexture(android::graphics::SurfaceTexture surfaceTexture);
```
ryantrem, line 300: Similar comment, convert this into a constructor to match the Java API?

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
```diff
@@ -277,6 +291,15 @@ namespace android::view
...
+        ANativeWindow* getNativeWindow();
```
ryantrem, line 301: I would remove this since it is not part of the Java API, and the caller can easily do this by just passing the `Surface` directly into `ANativeWindow_fromSurface` since it has an implicit conversion to the underlying `jobject`.

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -388,3 +407,23 @@ namespace android::net
...
+    }
+
+    void SurfaceTexture::initWithTexture(int texture)
+    {
+        m_object = m_env->NewObject(m_class, m_env->GetMethodID(m_class, "<init>", "(I)V"), texture);
+    }
+
+    void SurfaceTexture::updateTexture() const
+    {
+        if (m_object) {
```
ryantrem, line 425: If you make this set by the constructor (as described in previous comments), then you don't need this check, which will then also more closely match what the rest of the JavaWrappers code does.

Plugins/NativeCamera/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+set(SOURCES 
+    "Include/Babylon/Plugins/NativeCamera.h"
+    "Source/NativeCamera.cpp"
+    "Source/NativeCamera.h"
+    "Source/NativeVideo.cpp"
+    "Source/NativeVideo.h")
+
+if(APPLE)
+    set(SOURCES ${SOURCES} "Source/NativeCamera${BABYLON_NATIVE_PLATFORM}.mm")
```
ryantrem, line 9: I think this is captured elsewhere in feedback, but I believe the naming and directory structure we've been settling on is to have one directory per platform, have the file names all be the same, and put platform agnostic code in a `shared` folder.

Plugins/NativeCamera/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+if (APPLE OR ANDROID OR (WIN32 AND NOT WINDOWS_STORE))
+    add_library(NativeCamera ${SOURCES})
+    warnings_as_errors(NativeCamera)
+
+    target_include_directories(NativeCamera INTERFACE "Include")
+    target_include_directories(NativeCamera PRIVATE "Include/Babylon/Plugins")
+
+    if(ANDROID)
+        set(EXTENSIONS AndroidExtensions
```
ryantrem, line 22: Nit: Format this is all libs on separate lines, or all one one line?

Plugins/NativeCamera/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+        set(EXTENSIONS "-framework CoreMedia -framework AVFoundation")
+    endif()
+
+    target_link_to_dependencies(NativeCamera
+        PUBLIC JsRuntime
+        PRIVATE GraphicsInternal
+        PRIVATE ${EXTENSIONS})
+
+    set_property(TARGET NativeCamera PROPERTY FOLDER Plugins)
+    source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${SOURCES})
```
ryantrem, line 35: I don't recognize this. Do we use this elsewhere?

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
CedricGuillemet: I first have to create a transition opengl context, create a texture within that context and init surface texture with that texture.

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,117 @@
...
+            mediaDevices.Set("getUserMedia", Napi::Function::New(env, [](const Napi::CallbackInfo& info) {
```
ryantrem, line 49: I'm not familiar with this API, but when I looked it up it says it is deprecated. Do we use this in Babylon.js?

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
CedricGuillemet: See previous comment.

Plugins/NativeCamera/CMakeLists.txt
CedricGuillemet: yes, in most (if not, all) other cmake files.

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,117 @@
...
+                auto deferred{ Napi::Promise::Deferred::New(env) };
+                auto promise{ deferred.Promise() };
+               
+                auto& jsRuntime{ JsRuntime::GetFromJavaScript(env) };
+                    jsRuntime.Dispatch([deferred{ std::move(deferred) }](Napi::Env env) {
+                        deferred.Resolve(env.Null());
+                    });
+                
+                return promise;
+                }));
```
ryantrem, line 60:
```suggestion
            }));
```

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,117 @@
...
+            Napi::Object mediaDevices = Napi::Object::New(env);
+            mediaDevices.Set("getUserMedia", Napi::Function::New(env, [](const Napi::CallbackInfo& info) {
+                auto env = info.Env();
+                auto deferred{ Napi::Promise::Deferred::New(env) };
+                auto promise{ deferred.Promise() };
+               
+                auto& jsRuntime{ JsRuntime::GetFromJavaScript(env) };
+                    jsRuntime.Dispatch([deferred{ std::move(deferred) }](Napi::Env env) {
+                        deferred.Resolve(env.Null());
+                    });
```
ryantrem, line 57: Why do we need this dispatch? We're already on the JS thread, so this is just waiting for the next JS frame to resolve the promise. Do we need to do that?

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,117 @@
...
+        {
+        }
+
+    private:
+
+        Napi::Value CreateVideo(const Napi::CallbackInfo& info)
+        {
+            auto constraints = info[0].As<Napi::Object>();
+            uint32_t maxWidth{256}, maxHeight{256};
+            std::string facingMode;
```
ryantrem, line 79:
```suggestion
            std::string facingMode{};
```

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+            auto maxWidthValue = constraints.Get("maxWidth");
+            auto maxHeightValue = constraints.Get("maxHeight");
+            auto facingModeValue = constraints.Get("facingMode");
```
ryantrem, line 83:
```suggestion
            auto maxWidthValue{constraints.Get("maxWidth")};
            auto maxHeightValue{constraints.Get("maxHeight")};
            auto facingModeValue{constraints.Get("facingMode")};
```

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+    private:
+
+        std::string getCamId(bool frontCamera);
+
+        Graphics::Impl &m_graphicsImpl;
+        JsRuntimeScheduler m_runtimeScheduler;
+
+        uint32_t width;
+        uint32_t height;
```
ryantrem, line 70:
```suggestion
        uint32_t width{};
        uint32_t height{};
```

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
```diff
@@ -0,0 +1,404 @@
...
+        EGLContext context{};
+        EGLDisplay display{};
+    };
+
+    std::string CameraInterfaceAndroid::getCamId(bool frontCamera)
+    {
+        ACameraIdList *cameraIds = nullptr;
+        ACameraManager_getCameraIdList(cameraManager, &cameraIds);
+
+        std::string cameraId;
```
ryantrem, line 97:
```suggestion
        std::string cameraId{};
```

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
```diff
@@ -0,0 +1,404 @@
...
+        .onCaptureStarted = nullptr,
+        .onCaptureProgressed = nullptr,
+        .onCaptureCompleted = onCaptureCompleted,
+        .onCaptureFailed = onCaptureFailed,
+        .onCaptureSequenceCompleted = onCaptureSequenceCompleted,
+        .onCaptureSequenceAborted = onCaptureSequenceAborted,
+        .onCaptureBufferLost = nullptr,
+    };
+
+    arcana::task<void, std::exception_ptr> CheckCameraPermissionAsync()
```
ryantrem, line 186: Should we factor this out of xr.cpp like you did some of the OpenGL related stuff?

Plugins/NativeCamera/Source/NativeVideo.cpp
```diff
@@ -0,0 +1,156 @@
...
+    Napi::Value NativeVideo::GetHaveCurrentData(const Napi::CallbackInfo& /*info*/)
+    {
+        return Napi::Value::From(Env(), 1u);
+    }
+
+    void NativeVideo::UpdateTexture(bgfx::TextureHandle textureHandle)
+    {
+        if (!m_cameraInterface)
+        {
+            m_cameraInterface = CameraInterface::CreateInterface(Env(), m_width, m_height, m_frontCamera);
```
ryantrem, line 88: From what I can tell, the Android and iOS implementations both create the camera interface asynchronously, but then we call UpdateCameraTexture immediately after this. How is this supposed to work? Is UpdateCameraTexture just supposed to be a no-op during async creation?

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
```diff
@@ -0,0 +1,404 @@
...
+            ACameraCaptureSession_setRepeatingRequest(textureSession, &captureCallbacks, 1, &request, nullptr);
+
+            if (eglMakeCurrent(display, 0/*surface*/, 0/*surface*/, currentContext) == EGL_FALSE)
+            {
+                throw std::runtime_error{"Unable to restore GL context for camera texture init."};
+            }
+        });
+    }
+
+    CameraInterfaceAndroid::~CameraInterfaceAndroid()
```
ryantrem, line 337: Bad things will happen if the destructor is called while async construction is still in flight, right? Should we do something to defend against that?

Plugins/NativeCamera/Source/NativeVideo.h
```diff
@@ -0,0 +1,49 @@
...
+#include <GraphicsImpl.h>
+#include <vector>
+#include <algorithm>
+#include <unordered_map>
+
+namespace Babylon::Plugins::Internal
+{
+    class NativeVideo : public Napi::ObjectWrap<NativeVideo>
+    {
+        static constexpr auto JS_CLASS_NAME = "NativeVideo";
```
ryantrem, line 14: Could move this to the .cpp.

Apps/Playground/Android/app/src/main/AndroidManifest.xml
```diff
@@ -53,4 +53,5 @@
   <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
   <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
   <uses-permission android:name="android.permission.READ_INTERNAL_STORAGE" />
+  <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
```
bghgary, line 56: nit
```suggestion
  <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
```

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -101,6 +102,7 @@ extern "C"
                 g_nativeXr.emplace(Babylon::Plugins::NativeXr::Initialize(env));
                 g_nativeXr->SetSessionStateChangedCallback([](bool isXrActive){ g_isXrActive = isXrActive; });
 
+                Babylon::Plugins::NativeCamera::Initialize(env);
```
bghgary, line 105: nit

```suggestion
                Babylon::Plugins::NativeCamera::Initialize(env);

```

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
bghgary: I don't understand this. Can we sync up offline?

Dependencies/AndroidExtensions/Include/AndroidExtensions/OpenGLHelpers.h
```diff
@@ -0,0 +1,77 @@
+#pragma once
```
bghgary, line 1: Not sure this code belongs as exposed method for this library.

Plugins/NativeCamera/CMakeLists.txt
bghgary: We have two structures. I don't think we have to settle to exact one format.

Plugins/NativeCamera/CMakeLists.txt
bghgary: This just makes it so that VS puts the files in filters when files are in folders.

Plugins/NativeCamera/Source/NativeCameraAndroid.cpp
```diff
@@ -0,0 +1,404 @@
...
+        uint32_t height;
+        ACameraManager *cameraManager{};
+        ACameraDevice *cameraDevice{};
+        ACameraOutputTarget *textureTarget{};
+        ACaptureRequest *request{};
+        ANativeWindow *textureWindow{};
+        ACameraCaptureSession *textureSession{};
+        ACaptureSessionOutput *textureOutput{};
+        ACaptureSessionOutput *output{};
+        ACaptureSessionOutputContainer *outputs{};
```
bghgary, line 79:
```suggestion
        ACameraManager* cameraManager{};
        ACameraDevice* cameraDevice{};
        ACameraOutputTarget* textureTarget{};
        ACaptureRequest* request{};
        ANativeWindow* textureWindow{};
        ACameraCaptureSession* textureSession{};
        ACaptureSessionOutput* textureOutput{};
        ACaptureSessionOutput* output{};
        ACaptureSessionOutputContainer* outputs{};
```

bghgary (commented):
I made some initial comments.

Dependencies/AndroidExtensions/Include/AndroidExtensions/JavaWrappers.h
ryantrem: +1

Plugins/NativeCamera/CMakeLists.txt
syntheticmagus: Historically, I think we've done the `Shared` folder thing mainly with headers, handling platforms in `Source` folders more akin to how it's done here. We might not have been as consistent about that as I'm thinking, though.

Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
```diff
@@ -0,0 +1,27 @@
...
+        Camera(const Camera& other) = default;
+        Camera(Camera&&) = default;
+        ~Camera() = default;
+
+        // Initialization with overrideCameraTexture set to true means the caller
+        // is expected to override the camera source texture with a native texture.
+        // In that case, the HW camera will not be open.
+        static Camera Initialize(Napi::Env env, bool overrideCameraTexture = false);
+
+        void SetTextureOverride(void* texturePtr);
```
ryantrem, line 21: Can we make this strongly typed instead of `void*`? Like add a `using TextureType = int` to `GraphicsPlatform.h` for Android for example and then use that header here?

CedricGuillemet: I agree. I will do that in a second PR once the GraphicsContext PR is in as well.

CedricGuillemet: This proved to be more difficult than anticipated. Using Apple types (id<texture>) in .h/.cpp brings a lot of build issues. I think a 3rd (or 4th) refactor is needed. I will do it later.

PR #631, opened 2021-03-12
Fix ARCore double render (and rework NativeXr lifetime management)

### LINKED ISSUE #568: XR mode is double rendered on Android until tracking is obtained
I believe this behavior started with https://github.com/BabylonJS/BabylonNative/commit/2b6686db003b6f9430126efae028eb0aa8f6091d. After XR has started, but before we have obtained tracking, there are two calls per frame to `eglSwapBuffers`, which results in the back buffer being swapped with the front buffer, but then they are swapped again, and so you see the current frame then the previous frame. This makes what is being rendered look "jittery."

The reason this happens is that the off screen XR frame buffer is not bound until we have tracking. This is a result of Babylon.js's `WebXRCamera` doing an early out when the pose is null (haven't obtained tracking yet). This leaves the default (on screen) buffer bound, which makes bgfx think it needs to present, so it calls `eglSwapBuffers`. But, the Android/ARCore xr.cpp also explicitly calls `eglSwapBuffers` because during XR it is in charge of presenting. It's still unclear why this was not previously broken as well.

Ideally we would bind the off screen frame buffer even before we have obtained tracking (which would let you put something on the screen, such as UI telling the user they don't have tracking and what they need to do to get tracking), but this isn't really possible on binocular HMDs, because we don't have a pose yet which means we don't have views yet which means we don't have view transforms yet which means we couldn't possibly position the cameras correctly relative to each other (e.g. correct spacing between eyes).

I'm not sure what the right solution to this is yet. At the least, bgfx should not be trying to present to the screen when we have entered XR mode, even if we don't have tracking yet.

ryantrem:
Ultimately this change fixes #568, but I had to do a fair bit of restructuring to solve the problem. Previously, the ARCore XR.cpp implementation just grabbed the egl surface from the current thread (the one bgfx manages for Babylon Native) and just takes over it. This isn't great, but didn't cause any issues until the most recent bgfx upgrade (see #568 for details).

The solution here is to have XR render to its own window/surface so it doesn't stomp on the core Babylon Native rendering. For iOS, we were actually already doing this by having XR.mm create a child `MTKView` under the main view, but on Android this is not possible (we don't have enough context in XR.cpp to create a child `SurfaceView`). So now the Babylon Native API consumer is responsible for providing an XR rendering surface (optionally - so we'll do this for iOS and Android, but not for "exclusive" XR modes like HoloLens). To do this though, I needed a way for the "native" (C++/Java/ObjC) to talk to the XR system (to pass an XR window, and to know whether XR is active so it can show/hide the XR render surface). This is what caused the biggest part of the change. Here are the high level aspects of the changes:

- Rather than an `Initialize` function in a `NativeXr` namespace, there is now a static `Initialize` function on a `NativeXr` class.
- The new `NativeXr::Initialize` returns a `NativeXr`, which provides the new native contract to the XR system.
- `NativeXr` provides a way to know when XR is active, and also a way to provide the `windowPtr` to which XR should render.
- `NativeXr` internally holds a `shared_ptr` to it's underlying `impl`, and as such `NativeXr` itself is very light weight and copyable/movable (just a regular C++ object).
- `NativeXr::Impl` is actually the pre-existing `NativeXr` class inside `NativeXr.cpp`, with some updates.
- The pre-existing `NativeXr` (now `NativeXr::Impl`) is now a long lived object, where ownership is shared among any `NativeXr` instances, the JS `XR` object, and a JS `XRSession` instance when it is alive.
- To retain the previous behavior (where the old `NativeXr` was owned by an `XRSession` instance) and not explode this change, `NativeXR::Impl` wraps most of its state into a `SessionState` struct, and the `SessionState` lifetime is tied to an instance of an XR session (we can factor more out of this session lifetime scope later, like the `XRSystem` instance for example).
- `NativeXr::Initialize` instantiates the `NativeXr::Impl`, which is wrapped by the returned `NativeXr` instance, and is also passed along to `XR::Initialize`.
- `XR` holds a reference to `NativeXr::Impl` and also passes it along to `XRSession` when one is created.

There are some additional improvements that should probably be made in a future PR:
- As mentioned above, probably factor out `XRSystem` to be longer lived than a session (which conceptually makes more sense, I think).
- Remove all usages of Napi from `NativeXr::Impl`. I think this is fundamentally unsafe since the object performs multi-threaded async operations (any Napi constructs could die while async operations are happening on the render thread).
- `void* windowPtr` should be replaced with platform specific types (just follow the pattern @Drigax is establishing for Graphics).
- Possibly make `NativeXr` behave more like Graphics in that it could be long lived (across multiple JS engine instances) with an `AddToJavaScript` function (just like Graphics). If `NativeXr::Impl` does not hold Napi references, then this becomes possible and may be a better pattern.

I think this pattern overall (shared ref counted impl with no napi dependencies, native C++ interface that is a simple copyable/movable object with a reference to the shared impl, JS interface that is composed of Napi objects with a reference to the shared impl) should be adopted for other parts of Babylon Native (such as `NativeInput`) as I believe it will be more robust in a multi-threaded environment and allows us to have both a native C++ contract and a JS contract for the underlying Babylon Native components.

Beyond the core changes described above, there are both up-stream and down-stream changes:
- XR specific rendering surfaces are provided and are shown/hidden depending on XR state (by the Playground apps, as well as associated Babylon React Native changes).
- XR.cpp/XR.mm (Android/iOS) now use a dedicated rendering surface.


Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -211,4 +217,24 @@ extern "C"
             g_graphics->StartRenderingCurrentFrame();
         }
     }
+
+    JNIEXPORT void JNICALL
+    Java_BabylonNative_Wrapper_xrSurfaceChanged(JNIEnv* env, jclass clazz, jobject surface)
+    {
+        if (g_nativeXr)
+        {
+            ANativeWindow *window{};
```
ryantrem, line 226:
```suggestion
            ANativeWindow* window{};
```

Plugins/NativeXr/Include/Babylon/Plugins/NativeXr.h
```diff
@@ -2,7 +2,24 @@
...
-namespace Babylon::Plugins::NativeXr
+namespace Babylon::Plugins
 {
-    void Initialize(Napi::Env env);
+    class NativeXr final
+    {
+    public:
+        class Impl;
+
+        NativeXr(std::shared_ptr<Impl> impl);
```
ryantrem, line 12: Move to private?

ryantrem: `Impl` is public because it is also used by the Napi JS wrappers, but I moved the constructor.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -301,296 +301,342 @@ namespace
...
+                }
+
+                Graphics::Impl& graphicsImpl;
+                Napi::FunctionReference createRenderTexture{};
+                Napi::FunctionReference destroyRenderTexture{};
+                std::map<void*, FrameBuffer*> textureToFrameBufferMap{};
+                std::map<FrameBuffer*, Napi::ObjectReference> frameBufferToJsTextureMap{};
+                std::vector<void*> activeTextures{};
+                xr::System system{};
+                std::shared_ptr<xr::System::Session> session{};
```
chrisfromwork, line 397: should the system be owned by the session state or should it also exist for the duration of the js engine like the xr napi component?

chrisfromwork: this comment/ask shouldn't be considered blocking. it does seem like now if we've wanted to expose xr system specific functionality with no tie to a session (like enabling OpenXR extensions prior to the xr session creation), that the system could now be owned by NativeXr::Impl

---
In reply to: [593313639](https://github.com/BabylonJS/BabylonNative/pull/631#discussion_r593313639) [](ancestors = 593313639)

chrisfromwork: talked offline, sounds like we'll plan for this in a follow up review

ryantrem: I think it should not be part of the session state, and should live on the `Napi::Impl` directly, but this needs to be tested on HoloLens since `XR::System` actually does something on that platform (on iOS and Android it basically does nothing). Discussed more offline, and @chrisfromwork will test this out for a separate PR maybe next week.

chrisfromwork (approved):
validated that we are still able to enter into xr mode on hololens

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -301,296 +301,342 @@ namespace
...
+            void* m_windowPtr{};
+
+            struct SessionState
+            {
+                explicit SessionState(Graphics::Impl& graphicsImpl)
+                    : graphicsImpl{ graphicsImpl }
+                {
+                }
+
+                Graphics::Impl& graphicsImpl;
```
ryantrem, line 390: Make these all upper camel case (e.g. `GrahpicsImpl`).

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -301,296 +301,342 @@ namespace
...
         {
-            return m_session->TrySetPreferredMeshDetectorOptions(options);
+            m_windowPtr = windowPtr;
         }
 
-        uintptr_t GetNativeXrContext()
+        void NativeXr::Impl::SetSessionStateChangedCallback(std::function<void(bool)> callback)
         {
-            return m_session->GetNativeXrContext();
+            m_sessionStateChangedCallback = std::move(callback);
```
ryantrem, line 428: Add a mutex for this.

bghgary (approved):

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -301,296 +301,342 @@ namespace
...
+                    m_sessionState->frame = m_sessionState->session->GetNextFrame(shouldEndSession, shouldRestartSession);
+                    m_sessionState->frame.reset();
+                } while (!shouldEndSession);
 
-    void NativeXr::ScheduleFrame(std::function<void(const xr::System::Session::Frame&)>&& callback)
-    {
-        if (m_frameScheduled)
-        {
-            return;
+                m_sessionState->session.reset();
```
ryantrem, line 491: Remove.

PR #639, opened 2021-03-17
Ensure XR async lifecycle operations are sequential

ryantrem:
This change addresses the following problems in NativeXR:
- If `BeginSession` is called multiple times, it dies with a failed assertion. According to the WebXR spec, it should return a promise in a rejected state. It specifically wants a `DOMException` but Babylon.js catches any exception and logs it to the console as a warning. So the change for this is to track when we have an active XR session and return a faulted task if `BeginSession` is called again.
- If `EndSession` is called before `BeginSession` completes, it dies with a failed assertion. The WebXR spec does not indicate what should happen in this case, so the behavior I implemented is to have `EndSession` await completion of `BeginSession` before executing.
- If `BeginSession` is called before a previous `EndSession` completes, it does with a failed assertion. The WebXR spec does not indicate what should happen in this case, so the behavior I implemented is to have `BeginSession` await completion of the previous `EndSession` before executing.

The WebXR spec seems to imply that `EndSession` should just return the promise for ending the session, and it does not specify any exceptions should be thrown from `EndSession`, so I think this means it *should* just return the same promise. I did not implement this because Babylon.js prevents it from happening. We can revisit it in the future if needed.

### DIFF
--- a/Plugins/NativeXr/Source/NativeXr.cpp
+++ b/Plugins/NativeXr/Source/NativeXr.cpp
@@ -380,6 +380,8 @@ namespace Babylon
             std::mutex m_sessionStateChangedCallbackMutex{};
             std::function<void(bool)> m_sessionStateChangedCallback{};
             void* m_windowPtr{};
+            std::optional<arcana::task<void, std::exception_ptr>> m_beginTask{};
+            arcana::task<void, std::exception_ptr> m_endTask{arcana::task_from_result<std::exception_ptr>()};
 
             struct SessionState final
             {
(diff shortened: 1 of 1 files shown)


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -468,18 +476,27 @@ namespace Babylon
...
+
             m_sessionState->CancellationSource.cancel();
 
             m_sessionState->FrameBufferToJsTextureMap.clear();
             m_sessionState->TextureToFrameBufferMap.clear();
             m_sessionState->ActiveTextures.clear();
 
-            return m_sessionState->FrameTask.then(m_sessionState->GraphicsImpl.AfterRenderScheduler(), arcana::cancellation::none(), [this, thisRef{shared_from_this()}](const arcana::expected<void, std::exception_ptr>&) {
+            // Don't try to end the session while it is still starting.
+            m_endTask = m_beginTask->then(arcana::inline_scheduler, arcana::cancellation::none(), [this, thisRef{shared_from_this()}] {
```
bghgary, line 496: when_all?

bghgary (approved):
minor comment

Plugins/NativeXr/Source/NativeXr.cpp
ryantrem: I considered it, but I feel like the logic is more clear as is. If you feel strongly I can change it, but in a separate PR since I already tested it as is in multiple apps.

chrisfromwork (approved):
:shipit:

PR #641, opened 2021-03-17
Improve immersive xr fast refresh scenarios for BabylonReactNative/OpenXR

chrisfromwork:
Right now we never drop out of immersive mode when ending an xr session because we never release the session handle. This change releases the session handle when destroying the session after releasing hand tracking extension resources.

Note:
1. There are still issues related to NativeEngine texture creation that can cause crashes during reload.
2. OpenXR does not currently activate the original view when dropping out of immersive xr mode. Developers/users will have to touch the view on a HoloLens to get it to start rendering again. I have synced internally with the microsoft OpenXR team and they are looking into fixing this issue.


bghgary (approved):

tymlipari (approved):
:shipit:

chrisfromwork:
This review was the initial work for https://github.com/BabylonJS/BabylonReactNative/issues/176

PR #643, opened 2021-03-22
Updating hand input flow to be closer to how it is presented in Babylon.JS

### LINKED ISSUE #635: XR hand tracking should rely on xr session events for select/grasp events instead of declaring a gamepad property
Right now a xr hand input source exposes a gamepad property that causes it to register as a motion controller. xr hand input should avoid exposing a gamepad property in favor of raising select/grasp events through the xr session.

rickfromwork:
This PR adds a few new things:
- Hand input is now sent to Babylon.JS with the correct profiles (generic-hand-select*)
- Controller data (motion controllers, hands, etc.) is now collected based on the active input profile. This improves on the previous strategy of trying every profile and seeing what works
- The Session polyfill has been updated to handle select and squeeze events registered to it

This matches the behaviour Babylon.JS expects more closely, as it was using select event callbacks to get the Select gesture off of each hand, but we did not support it. This also removes the gamepad component from hands, which wasn't desired. While it had allowed us to pass Select through, it did not match browser behaviour, and caused Babylon.JS to treat it like a motion controller, and try and create controller models that did not exist.

Resolves #635


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1573,101 +1610,132 @@ namespace xr
                     }
                 }
 
-                // Get gamepad data 
+                // Get interaction data based on input profile
+                if (!InputSources[idx].interactionProfileName.empty())
                 {
```
satyapoojasama, line 1615: nit: its a good idea to test the controllers as well to make sure they are not broken with this change :)

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -804,6 +809,16 @@ namespace xr
                 }
             }
 
+            // Create controller get trigger value action and suggested bindings=
```
satyapoojasama, line 812: nit: update comment to select

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2523,18 +2541,67 @@ namespace Babylon
...
+                        // inputSource already exists, find the corresponding gamepad object if enabled and set to correct values
                         if (inputSourceVal.Has("gamepad"))
                         {
                             auto gamepadObject = inputSourceVal.Get("gamepad").As<Napi::Object>();
                             SetXRGamepadObjectData(inputSourceVal, gamepadObject, inputSource);
                         }
                     }
+
+                    // Handle gestures. Sources that do not support a gesture will always return false
+                    bool isSelected = inputSource.GamepadObject.Buttons[0].Pressed;
```
satyapoojasama, line 2553: is this gamepad object referring to motion controller?

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I've tested in simulation and can see the select gesture going through. Do you have any specifics to test?

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: Ah, I was tempted to rename it. I'm using the Gamepad object to hold Select, Squeeze for all inputs, not just the motion controller it was originally used for.

rickfromwork: I didn't want to cloud XR.h with different types specific to OpenXR's options.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2491,6 +2505,10 @@ namespace Babylon
                 std::set<xr::System::Session::Frame::InputSource::Identifier> added{};
```
tymlipari, line 2505: >std::set [](start = 16, length = 8)

Unless we deliberately want to keep these sorted for when we insert them into the NAPI arrays, these would probably be better served as a `std::deque` or `std::vector`.

Plugins/NativeXr/Source/NativeXr.cpp
tymlipari, line 2553: >bool [](start = 20, length = 4)

nit: `const` these

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2523,18 +2541,67 @@ namespace Babylon
...
+                    if (isSelected && (m_activeSelects.find(inputSource.ID) == m_activeSelects.end()))
+                    {
+                        selectStarts.insert(inputSource.ID);
+                    }
+                    else if (!isSelected && (m_activeSelects.find(inputSource.ID) != m_activeSelects.end()))
+                    {
+                        selectEnds.insert(inputSource.ID);
+                    }
+
+                    if (isSqueezed && (m_activeSqueezes.find(inputSource.ID) == m_activeSqueezes.end()))
```
tymlipari, line 2564: >isSqueezed [](start = 24, length = 10)

Is it possible to be both squeezed and pressed? And if so, is it safe to fire events for both?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2610,6 +2680,113 @@ namespace Babylon
...
+            {
+                auto inputSourceEvent = Napi::Object::New(env);
+                inputSourceEvent.Set("frame", m_jsXRFrame.Value());
+                inputSourceEvent.Set("inputSource", inputSource);
+
+                return inputSourceEvent;
+            }
+
+            void UpdateInputSourceEventValues(
+                std::set<xr::System::Session::Frame::InputSource::Identifier> selectStarts,
```
tymlipari, line 2693: >selectStarts [](start = 78, length = 12)

These are likely to be expensive copies, take by `const&` instead.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2610,6 +2680,113 @@ namespace Babylon
...
+
+                            if (name == JS_EVENT_NAME_SQUEEZE)
+                            {
+                                callback.Call({inputSourceEvent});
+                            }
+                        }
+
+                        m_activeSqueezes.erase(id);
+                    }
+                }
```
tymlipari, line 2787: These are all basically the same actions, just with different event names and source collections. Can this be extracted out into a separate function or at least a common lambda?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2610,6 +2680,113 @@ namespace Babylon
...
+                        Napi::Object inputSourceEvent = GenerateXRInputSourceEvent(inputSourceVal, env);
+
+                        for (const auto& [name, callback] : m_eventNamesAndCallbacks)
+                        {
+                            if (name == JS_EVENT_NAME_SELECT_END)
+                            {
+                                callback.Call({inputSourceEvent});
+                            }
+
+                            if (name == JS_EVENT_NAME_SELECT)
```
tymlipari, line 2754: >if [](start = 28, length = 2)

nit: `name` will never be more than one of these options, so I'd recommend making these `else if` to avoid subsequent string comparisons.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2610,6 +2680,113 @@ namespace Babylon
...
+                            {
+                                callback.Call({inputSourceEvent});
+                            }
+                        }
+
+                        m_activeSelects.erase(id);
```
tymlipari, line 2760: >id [](start = 46, length = 2)

nit: rather than using `id` here, use `inputSourceFound`. If you use the `std::map<>::key_type` overload it's likely going to have to internally execute a `find` to locate the node to erase. Passing in the iterator position should be slightly faster.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2610,6 +2680,113 @@ namespace Babylon
...
+                            }
+                        }
+
+                        m_activeSqueezes.insert(id);
+                    }
+                }
+
+                for (const auto& id : selectEnds)
+                {
+                    auto inputSourceFound = m_idToInputSource.find(id);
```
tymlipari, line 2741: >inputSourceFound [](start = 25, length = 16)

nit: consider renaming to like `inputSourceItr` or something to capture that this is an iterator and not a bool.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -157,6 +157,11 @@ namespace
             handJointCollection.Set("size", static_cast<int>(HAND_JOINT_NAMES.size()));
 
             jsInputSource.Set("hand", handJointCollection);
+
+            auto profiles = Napi::Array::New(env, 1);
```
chrisfromwork, line 161: >1 [](start = 50, length = 1)

does this create an array of size one that we then add two things to?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2523,18 +2541,67 @@ namespace Babylon
...
+
+                    if (isSqueezed && (m_activeSqueezes.find(inputSource.ID) == m_activeSqueezes.end()))
+                    {
+                        squeezeStarts.insert(inputSource.ID);
+                    }
+                    else if (!isSqueezed && (m_activeSqueezes.find(inputSource.ID) != m_activeSqueezes.end()))
+                    {
+                        squeezeEnds.insert(inputSource.ID);
+                    }
                 }
```
chrisfromwork, line 2572: >} [](start = 16, length = 1)

nit an additional line may make this more legible

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2523,18 +2541,67 @@ namespace Babylon
...
+
+                        if (m_activeSqueezes.find(id) != m_activeSqueezes.end())
+                        {
+                            for (const auto& [name, callback] : m_eventNamesAndCallbacks)
+                            {
+                                if (name == JS_EVENT_NAME_SQUEEZE_END)
+                                {
+                                    callback.Call({inputSourceEvent});
+                                }
+                            }
```
chrisfromwork, line 2601: nit: it may be worth adding helper functions for firing these callbacks

Dependencies/xr/Source/OpenXR/XR.cpp
satyapoojasama: I see, that should be fine but for the sake of sanity, could you test all the functionalities like trigger, thumbstick and trackpad as well?

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: Yeah, I was just matching the usage below

Plugins/NativeXr/Source/NativeXr.cpp
satyapoojasama: nvm, i just realized this is the gamepadobject on the native end.

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: It should be possible. Babylon.JS doesn't seem to use squeeze events right now, though, so nothing will break in any case.

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: I looked at the webXR API again, and since I need to fire select events before I fire selectend, I need to do two separate iterations.

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I tested motion controllers with simulation, and hit the expected breakpoints in WebXRControllerComponent.prototype.update for when a button or axis update, so the controller data is being piped through properly still.

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: Thhe shell fires both when you do an air tap, by the way. There's also no documentation saying both wont be fired together.

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: That's an iterator for a different object, unless I read this wrong

Dependencies/xr/Include/XR.h
```diff
@@ -241,6 +241,8 @@ namespace xr
                     bool TrackedThisFrame{};
                     bool JointsTrackedThisFrame{};
                     bool GamepadTrackedThisFrame{};
+                    bool HandTrackedThisFrame{};
+                    std::string interactionProfileName{""};
```
chrisfromwork, line 245: >interactionProfileName [](start = 32, length = 22)

nit: probably should capitalize this for consistency

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1573,101 +1610,132 @@ namespace xr
...
+                    XrCheck(xrStringToPath(m_impl->sessionImpl.HmdImpl.Context.Instance(), inputSource.interactionProfileName.data(), &interactionProfilePath));
+                    switch (interactionProfilePath)
                     {
-                        const auto& controllerInfo = sessionImpl.ControllerInfo;
-                        auto& gamepadObject = InputSources[idx].GamepadObject;
-
-                        // Get interaction data
-                        if ((m_impl->TryUpdateControllerBooleanAction(actionResources.HandGetSelectAction, session, gamepadObject.Buttons[controllerInfo.TRIGGER_BUTTON].Pressed)) &&
-                            (m_impl->TryUpdateControllerBooleanAction(actionResources.HandGetSqueezeAction, session, gamepadObject.Buttons[controllerInfo.SQUEEZE_BUTTON].Touched)))
+                        case 31:// MICROSOFT_XR_INTERACTION_PROFILE
```
chrisfromwork, line 1621: >31 [](start = 29, length = 2)

nit: do we have constants declaring these case values? if not, these comments are good

satyapoojasama (approved):

chrisfromwork (approved):
:shipit:

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: I searched and couldn't find any. I'd really rather use constants, though.

PR #644, opened 2021-03-23
NativeCapture functional in XR mode on Android and iOS

ryantrem:
This change introduces a new mechanism for off screen frame buffer capture into `NativeCapture`. The original mechanism is still used for on screen / default frame buffer capture (I don't *think* it is possible to use the same mechanism for both on screen and off screen frame buffers). While I tried to not break existing usage of the API, in the end there are a few things that are in fact breaking changes:
1. While in XR on Android, frame capture is RGBA8 instead of BGRA8 (I'm not sure what bgfx does internally to capture the on screen frame buffer as BGRA8 on Android since as far as I can tell GLES doesn't directly support it; maybe there is a way to make it BGRA8 which would make things a little simpler for consumers).
2. The `pitch` parameter is not being used by current consumers and I wasn't sure how to properly calculate it for the off screen frame buffer case, so I just removed it. I figure we can add it back later if needed.
3. When you want to capture an off screen frame buffer, you need to pass the frame buffer into the `NativeCapture` constructor. For XR, this would be obtained from `webXRExperienceHelper.camera.rigCameras[0]` (for Android and iOS since XR is monocular; for binocular XR devices like HoloLens, the caller can pick which camera to capture from).

I will have another PR with some small updates in Babylon React Native for this.

FYI @dyleatha, @bonnyMahajan10, @mkatzour


Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -4,15 +4,123 @@
...
+                    bgfx::blit(bgfx::getCaps()->limits.maxViews - 1, thisRef->m_blitTextureHandle, 0, 0, thisRef->m_frameBufferTextureHandle);
+
+                    // Reading the texture is an async operation, but everything that needs to be done prior to future write operations on that texture is completed synchronously,
+                    // so we kick off the read for the next frame prior to the read for the current frame completes.
+                    return arcana::when_all(thisRef->m_graphicsImpl.ReadTextureAsync(thisRef->m_blitTextureHandle, thisRef->m_textureBuffer).then(arcana::inline_scheduler, thisRef->m_cancellationToken, [thisRef]{
+                        thisRef->m_frameCallback(thisRef->m_textureInfo.Width, thisRef->m_textureInfo.Height, thisRef->m_textureInfo.Format, bgfx::getCaps()->originBottomLeft, thisRef->m_textureBuffer);
+                    }),
+                    thisRef->ReadTextureAsync()).then(arcana::inline_scheduler, arcana::cancellation::none(), [](const std::tuple<arcana::void_placeholder, arcana::void_placeholder>&){
```
bghgary, line 92: nit: since you don't care about the parameters, you could do this:
```suggestion
                    thisRef->ReadTextureAsync()).then(arcana::inline_scheduler, arcana::cancellation::none(), [](const auto&){
```

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -4,15 +4,123 @@
...
+    using FrameProviderTicket = gsl::final_action<FrameProviderCleanup>;
+
+    FrameProviderTicket BeginFrameCapture(Babylon::Graphics::Impl& graphicsImpl, bgfx::FrameBufferHandle frameBufferHandle, FrameCallback callback)
+    {
+        class DefaultBufferFrameProvider final : public std::enable_shared_from_this<DefaultBufferFrameProvider>
+        {
+        public:
+            static FrameProviderTicket Create(Babylon::Graphics::Impl& graphicsImpl, FrameCallback callback)
+            {
+                std::shared_ptr<DefaultBufferFrameProvider> frameProvider{new DefaultBufferFrameProvider(graphicsImpl, std::move(callback))};
```
bghgary, line 24: does make_shared not work?

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -4,15 +4,123 @@
...
+            FrameCallback m_frameCallback{};
+        };
+
+        class OffScreenBufferFrameProvider final : public std::enable_shared_from_this<OffScreenBufferFrameProvider>
+        {
+        public:
+            static FrameProviderTicket Create(Babylon::Graphics::Impl& graphicsImpl, bgfx::FrameBufferHandle frameBufferHandle, FrameCallback callback)
+            {
+                std::shared_ptr<OffScreenBufferFrameProvider> frameProvider{new OffScreenBufferFrameProvider(graphicsImpl, frameBufferHandle, std::move(callback))};
+                frameProvider->ReadTextureAsync();
```
bghgary, line 57: Might be good to comment here saying this is starting the capture and this function will call itself.

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -34,19 +142,34 @@ namespace Babylon::Plugins::Internal
...
+                frameBufferHandle = frameBuffer.Handle();
+            }
+
+            m_frameProviderTicket.emplace(BeginFrameCapture(graphicsImpl, frameBufferHandle, std::move(frameCallback)));
         }
 
         ~NativeCapture()
         {
-            if (m_ticket != nullptr)
+            if (m_frameProviderTicket.has_value())
```
bghgary, line 170: Is this check necessary?

Plugins/NativeCapture/Source/NativeCapture.cpp
```diff
@@ -34,19 +142,34 @@ namespace Babylon::Plugins::Internal
...
             jsData.Set("data", Napi::ArrayBuffer::New(info.Env(), 0));
+
+            FrameCallback frameCallback{[this](uint32_t width, uint32_t height, bgfx::TextureFormat::Enum format, bool yFlip, gsl::span<const uint8_t> data){
+                CaptureDataReceived(width, height, format, yFlip, data);
+            }};
+
+            bgfx::FrameBufferHandle frameBufferHandle{bgfx::kInvalidHandle};
+
+            // For an off screen frame buffer, the frame buffer must be passed into the constructor. Otherwise, we capture the on screen / default frame buffer.
+            if (info.Length() > 0 && info[0].IsExternal())
```
bghgary, line 159: The IsExternal should probably throw. Otherwise, we are just letting invalid arguments go through.

bghgary (approved):

Plugins/NativeCapture/Source/NativeCapture.cpp
ryantrem: It works if I make the constructor public, but that doesn't express the intent well, and we instantiate these types infrequently, so the extra cost is not worth the extra confusion in my opinion.

Plugins/NativeCapture/Source/NativeCapture.cpp
ryantrem: Perhaps not, but that's the way it was originally implemented. Do you feel strongly about changing it?

bghgary: nope, just don't understand the check

Plugins/NativeCapture/Source/NativeCapture.cpp
bghgary: okay

PR #645, opened 2021-03-24
graphics init from context

CedricGuillemet:
- Add the ability to create Graphics from an existing context

### DIFF
--- a/Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
+++ b/Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
@@ -46,9 +46,9 @@ extern "C"
 
         g_nativeXr.reset();
         g_scriptLoader.reset();
-        g_graphics.reset();
         g_inputBuffer.reset();
         g_runtime.reset();
+        g_graphics.reset();
 
         g_isXrActive = false;
     }
@@ -76,7 +76,7 @@ extern "C"
             g_runtime = std::make_unique<Babylon::AppRuntime>();
             g_inputBuffer = std::make_unique<InputManager<Babylon::AppRuntime>::InputBuffer>(*g_runtime);
 
-            g_runtime->Dispatch([javaVM, window, width, height](Napi::Env env)
+            g_runtime->Dispatch([javaVM](Napi::Env env)
             {
                 g_graphics->AddToJavaScript(env);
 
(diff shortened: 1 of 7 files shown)


ryantrem:
So with these changes, you either create graphics with a window or with a context?

Core/Graphics/Include/Babylon/Graphics.h
```diff
@@ -16,9 +16,15 @@ namespace Babylon
         template<typename... Ts>
         static std::unique_ptr<Graphics> CreateGraphics(Ts...);
 
+        template<typename... Ts>
+        static std::unique_ptr<Graphics> CreateGraphicsFromContext(Ts...);
```
syntheticmagus, line 20: This pattern is going away anyway in @Drigax's change (which this change should probably wait for to use the new pattern), but if we stop using `void*` then this shouldn't be necessary. The point of the template trick was to allow specializations based on the arguments so that if you just passed it typed pointer to a context, it would automatically look for a specialization that would take that context and create `Graphics` from it, without increasing the surface area of the API.

Core/Graphics/Include/Babylon/Graphics.h
```diff
@@ -16,9 +16,15 @@ namespace Babylon
...
+
         template<typename... Ts>
         void UpdateWindow(Ts...);
 
+        template<typename... Ts>
+        void UpdateContext(Ts...);
```
syntheticmagus, line 26: Maybe once the new pattern comes in, it might make sense to consolidate this, `UpdateWindow`, and other functions with related purposes into a `Reinitialize(const GraphicsConfiguration&)` method, or something like that.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -132,6 +132,17 @@ extern "C"
         }
     }
 
+    JNIEXPORT void JNICALL
+    Java_BabylonNative_Wrapper_surfaceResize(JNIEnv* env, jclass clazz, jint width, jint height)
+    {
+        if (g_runtime)
+        {
+            g_runtime->Dispatch([width = static_cast<size_t>(width), height = static_cast<size_t>(height)](auto env) {
+                g_graphics->UpdateSize(width, height);
```
Drigax, line 141: Why do we want to dispatch this to the JS thread?

CedricGuillemet:
Closing. Discussions on how to do it in progress.

PR #646, opened 2021-03-24
Update openxr scene understanding to use the second version of the scene understanding preview extension

chrisfromwork:
This change updates the OpenXR xr component to use XR_MSFT_SCENE_UNDERSTANDING_PREVIEW2_EXTENSION_NAME. This involved some substantial code changes/a rewrite of the SceneUnderstanding component.

This change also removes Planes and Meshes vectors from the xr frame that aren't used anymore.

There is one pull request that should be completed before checking in this change:

- [x] https://github.com/microsoft/OpenXR-MixedReality/pull/80

There is one pull request related but not required for this change:

- [x] https://github.com/BabylonJS/Babylon.js/pull/10095


Alex-MSFT (approved):

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -71,500 +73,463 @@ SceneUnderstanding::SceneUnderstanding::InitOptions::InitOptions(
...
+            const auto& xrMeshId = static_cast<xr::su::SceneMesh::Id>(xrMeshIds.at(i));
+            const auto& meshId = m_meshIds.at(xrMeshId);
+            auto& mesh = m_meshes.at(meshId);
+            const auto& location = xrMeshLocations.at(i);
+
+            // TODO: in preview 3 of the OpenXR scene understanding extension, plane's will also have associated meshes
+            xr::SceneMeshBuffers meshBuffers;
+            xr::ReadMeshBuffers(m_scene->Handle(), args.Extensions, mesh->MeshBufferId, meshBuffers);
+
+            assert(sizeof(XrVector3f) == sizeof(xr::Vector3f));
```
dgough, line 362: This should probably be `static_assert` instead of `assert`.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -71,500 +73,463 @@ SceneUnderstanding::SceneUnderstanding::InitOptions::InitOptions(
...
+    std::unique_ptr<xr::su::Scene> m_scene;
+    xr::SceneBounds m_sceneBounds{};
+    ScanState m_scanState{ ScanState::Idle };
+    XrTime m_updateInterval{ 0 };
+    XrTime m_lastUpdateTime{ 0 };
+
+    std::unordered_map<xr::su::SceneObject::Id, SceneObject::Identifier> m_objectIds{};
+    std::unordered_map<SceneObject::Identifier, std::unique_ptr<SceneObject>> m_objects{};
+    std::unordered_map<xr::su::SceneMesh::Id, Mesh::Identifier> m_meshIds{};
+    std::unordered_map<Mesh::Identifier, std::unique_ptr<Mesh>> m_meshes{};
```
ryantrem, line 512: What is the purpose of having the values in these maps be wrapped in a `unique_ptr`?

chrisfromwork: it lets me get access to the struct generated id before inserting into the maps. If you have a better suggestion, definitely let me know :)

dgough: The auto generated id is only auto generated when you use the default constructor of `Mesh`. If you use the copy constructor or explicitly set all members then you could set the ID to whatever you want. I think you should be able to create a local `Mesh`, get its auto generated ID then insert or emplace the Mesh into the `unordered_map` using that ID. It should invoke the copy constructor of `Mesh` without incrementing the static `NEXT_ID`.

Change this:

```c++
auto mesh = std::make_unique<Mesh>();
const auto meshId = mesh->ID;
m_meshIds[xrMesh.id] = meshId;
m_meshes[meshId] = std::move(mesh);
```

to:

```c++
// Assuming m_meshes is updated to use Mesh instead of unique_ptr<Mesh>:
std::unordered_map<Mesh::Identifier, Mesh> m_meshes;

Mesh mesh; // auto generates new ID
const auto meshId = mesh.ID; // get the auto generated ID
m_meshes.emplace(meshId, mesh); // should call copy constructor of Mesh. NEXT_ID shouldn't change.
m_meshIds[xrMesh.id] = meshId;
```

Although I haven't tried this since I'm not setup to build this repo. 

You need to use insert or emplace instead of `operator[]` because of how `std::unordered_map::operator[]` behaves. `operator[]` will actually call the default constructor before then calling the copy assignment operator. The STL is full of weird stuff like this.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
```diff
@@ -71,500 +73,463 @@ SceneUnderstanding::SceneUnderstanding::InitOptions::InitOptions(
...
+            object->Type = c_objectTypeMap.at(xrObject.kind);
         }
 
-        args.UpdatedPlanes.clear();
-        for (const auto xrKey : m_updatedPlanes)
+        std::unordered_set<xr::su::SceneObject::Id> xrObjectIdsToRemove{};
+        for (const auto& [xrObjectId, objectId] : m_objectIds)
         {
-            if (m_babylonPlanes.count(xrKey) > 0)
+            if (observedXRObjectIds.find(xrObjectId) == observedXRObjectIds.end())
```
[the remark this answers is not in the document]
chrisfromwork: i think this ones a little bit different because we are inserting into a different set for removal. The set here we are checking against may or may not contain the key.

Dependencies/xr/Source/OpenXR/SceneUnderstanding.cpp
chrisfromwork: this worked. thank you for the suggestion @dgough!

satyapoojasama (approved):

PR #650, opened 2021-03-29
Expose nativeXrContext off of the xr system and native xr component to allow enabling openxr extensions prior to creating a xr session

chrisfromwork:
For OpenXR, we need to be able to enable extensions prior to creating the xr session. This change exposes the xr context off of the global navigator.xr component. This allows us to access the xr context without creating a xr session.

This review requires the following changes for BabylonReactNative: https://github.com/BabylonJS/BabylonReactNative/pull/185


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -463,15 +463,15 @@ namespace Babylon
 
                     m_sessionState = std::make_unique<SessionState>(graphicsImpl);
 
-                    if (!m_sessionState->System.IsInitialized())
+                    if (!m_system.IsInitialized())
```
ryantrem, line 466: So the pattern is that `m_system` has the scope of the `XR` object, *but* it is still lazily initialized by the first `XRSession`?

ryantrem (approved):

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: this seems less likely to introduce any regressions for the different platforms. The system component we are accessing prior to the xr session being created doesn't require the system to be initialized.

chrisfromwork:
closing and reopening to reset ci

PR #655, opened 2021-04-01
Hand joint rotation fix

rickfromwork:
OpenXR and WebXR used different rotations for their joints, and have a delta of 90 degrees between them. This doesn't affect joint positions, but it does distort the hand mesh/anything else that might rely on the joint direction.

### DIFF
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -1733,25 +1733,25 @@ namespace xr
                                         rotationDelta.y = -rotationDelta.w;
                                         rotationDelta.z = 0;
 
-                                        rotationDelta.w = joint.pose.orientation.w * rotationDelta.w -
-                                                          joint.pose.orientation.x * rotationDelta.x -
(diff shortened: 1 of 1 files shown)


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1720,6 +1722,40 @@ namespace xr
...
+
+                                        newRotation.y = joint.pose.orientation.w * rotationDelta.y -
+                                                        joint.pose.orientation.x * rotationDelta.z +
+                                                        joint.pose.orientation.y * rotationDelta.w +
+                                                        joint.pose.orientation.z * rotationDelta.x;
+
+                                        newRotation.z = joint.pose.orientation.w * rotationDelta.z +
+                                                        joint.pose.orientation.x * rotationDelta.y -
+                                                        joint.pose.orientation.y * rotationDelta.x +
+                                                        joint.pose.orientation.z * rotationDelta.w;
```
tymlipari, line 1754: This feels like the kind of thing that we should have a library for already...

If not, consider pulling this out into a helper header

PR #656, opened 2021-04-02
Touch-ups following up on graphics contract change PR

syntheticmagus:
Starting as a draft so I can fire CI to make sure I did the changes on many platforms sanely.

### DIFF
--- a/Core/Graphics/Include/Shared/Babylon/Graphics.h
+++ b/Core/Graphics/Include/Shared/Babylon/Graphics.h
@@ -1,18 +1,17 @@
 #pragma once
 
 #include <Babylon/JsRuntime.h>
+#include <Babylon/GraphicsPlatform.h>
 
 #include <memory>
 
 namespace Babylon
 {
-    struct GraphicsConfiguration;
-
     class Graphics
     {
-    public:
         class Impl;
 
+    public:
         ~Graphics();
 
         static std::unique_ptr<Graphics> CreateGraphics(const GraphicsConfiguration& config);
--- a/Core/Graphics/Include/UWP/CoreWindow/Babylon/GraphicsPlatform.h
+++ b/Core/Graphics/Include/UWP/CoreWindow/Babylon/GraphicsPlatform.h
@@ -1,15 +1,17 @@
 #pragma once
 
-#include <Babylon/Graphics.h>
 #include <windows.ui.core.h>
 #include <winrt/windows.ui.core.h>
 #include <winrt/Windows.ApplicationModel.h>
 
-using WindowType = winrt::Windows::UI::Core::ICoreWindow*;
-
-struct Babylon::GraphicsConfiguration
+namespace Babylon
 {
-    WindowType WindowPtr;
-    size_t Width;
-    size_t Height;
-};
+    using WindowType = winrt::Windows::UI::Core::ICoreWindow*;
+
+    struct GraphicsConfiguration
+    {
+        WindowType WindowPtr;
+        size_t Width;
+        size_t Height;
+    };
+}
--- a/Core/Graphics/Source/Unix/GraphicsPlatform.cpp
+++ b/Core/Graphics/Source/Unix/GraphicsPlatform.cpp
@@ -1,15 +1,14 @@
-#include <Babylon/Graphics.h>
 #include <Babylon/GraphicsPlatform.h>
-#include "../GraphicsImpl.h"
+#include <GraphicsImpl.h>
 
-constexpr float MILLIMETERS_TO_INCHES = 0.03937;
+constexpr float MILLIMETERS_TO_INCHES{0.03937f};
 
 namespace Babylon
 {
     void GraphicsImpl::ConfigureBgfxPlatformData(const GraphicsConfiguration& config, bgfx::PlatformData& pd)
     {
         pd.ndt = nullptr;
-        pd.nwh = (void*)config.WindowPtr;
+        pd.nwh = static_cast<void*>(config.WindowPtr);
         pd.context = nullptr;
         pd.backBuffer = nullptr;
         pd.backBufferDS = nullptr;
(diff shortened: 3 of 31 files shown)


syntheticmagus:
Looks like it's only failing on the Android CI that's currently experiencing intermittent problems.

Core/Graphics/Include/Shared/Babylon/Graphics.h
```diff
@@ -1,18 +1,17 @@
 #pragma once
 
 #include <Babylon/JsRuntime.h>
+#include <Babylon/GraphicsPlatform.h>
```
Drigax, line 4: Now everything that consumes Graphics will consume platform headers, is this your intention?

Core/Graphics/Include/UWP/CoreWindow/Babylon/GraphicsPlatform.h
```diff
@@ -1,15 +1,17 @@
...
-struct Babylon::GraphicsConfiguration
+namespace Babylon
 {
-    WindowType WindowPtr;
-    size_t Width;
-    size_t Height;
-};
+    using WindowType = winrt::Windows::UI::Core::ICoreWindow*;
+
+    struct GraphicsConfiguration
```
Drigax, line 11: I don't see too much issue in leaving our config struct in `Babylon` namespace either.

Core/Graphics/Source/Unix/GraphicsPlatform.cpp
```diff
@@ -1,15 +1,14 @@
...
-constexpr float MILLIMETERS_TO_INCHES = 0.03937;
+constexpr float MILLIMETERS_TO_INCHES{0.03937f};
 
 namespace Babylon
 {
     void GraphicsImpl::ConfigureBgfxPlatformData(const GraphicsConfiguration& config, bgfx::PlatformData& pd)
     {
         pd.ndt = nullptr;
-        pd.nwh = (void*)config.WindowPtr;
+        pd.nwh = static_cast<void*>(config.WindowPtr);
```
Drigax, line 11: As you mentioned before, we're still using `void*` for our linux window type, is there any clever casting that we can do in the shared code to allow for us to use an `unsigned long` "pointer"

Drigax (approved):
Most of these changes look good, thanks for clarifying the #include hierarchy :)

Core/Graphics/Include/UWP/CoreWindow/Babylon/GraphicsPlatform.h
syntheticmagus: Not sure I understand; the struct is in the Babylon namespace here, isn't it? The change is that I'm declaring and implementing it here now, not just implementing it, and that `WindowType` is now in the Babylon namespace, too.

Drigax: This was me agreeing with your change.

Core/Graphics/Include/Shared/Babylon/Graphics.h
syntheticmagus: That's the idea, though perhaps I've misunderstood something. Graphics.h exports the public contract of the Graphics component, which by definition (now that we've introduced `GraphicsConfiguration` contains platform-specific information. Anyone who wants consume the public Graphics contract should have to include Graphics.h, and only Graphics.h. But this public contract has nothing to do with the internal contract, which has many different pieces, only some of which are platform specific. Internal contract consumers, generally speaking, should not include Graphics.h at all. (They can -- it's possible -- but I don't believe anyone has reason to in the current repo, and I suspect I would regard any component that wanted to consume both the public and internal contracts of Graphics simultaneously as a "code smell" indicative of an architecture problem, most likely that the consumer was really two concepts being mistakenly implemented as one.) Instead, consumers of the internal contract should include only the parts of the internal API that they need, typically this will be just GraphicsImpl.h; and those who also need GraphicsPlatform.h are, by virtue of the fact that they need that, certainly doing something platform-specific already.

That's the idea, anyway; let's definitely follow up if I've missed something, though.

Core/Graphics/Source/Unix/GraphicsPlatform.cpp
syntheticmagus: _Hmm_, interesting... I'm not sure if this PR is the right place to change that, but this brings up a good point. Where actually is the shared code that would have to deal with this cast? In other words, if you change `WindowType` to be `unsigned long` and change this line to cast from that to the `void*` that bgfx is expecting, where does that break? I can check myself when I'm back on Monday and can fire up my Linux VM to try it, was just curious if you happened to know off the top of your head.

Core/Graphics/Include/UWP/CoreWindow/Babylon/GraphicsPlatform.h
syntheticmagus: Gotcha. Yeah, it was less clear to me when looking at the PR here on GitHub, but in the code I could see that the `Babylon` namespace was a better fit than the `Babylon::Graphics` namespace -- partly because there isn't a `Babylon::Graphics` namespace. 😆

Core/Graphics/Source/Unix/GraphicsPlatform.cpp
Drigax: This seems only relevant when we're casting to/from void* when setting the native window handle that we provide to bgfx:

https://github.com/BabylonJS/BabylonNative/blob/8371b79805f2647dc7c43aa1f3e0a5552f23e2f6/Core/Graphics/Source/GraphicsImpl.cpp#L45

Drigax: I don't think this minor detail should hold up the rest of these changes, however.

Core/Graphics/Include/Shared/Babylon/Graphics.h
Drigax: I see, I definitely needed to see this to fully contextualize what you were going for, thanks for the explanation.

Drigax:
Are we waiting for more review?

ryantrem (approved):

Core/Graphics/Include/Shared/Babylon/Graphics.h
bghgary: This is kind of a side note, but we shouldn't depend on the include path to include dependent headers. This requires that our consumers use include directories that might not be what they want to do. Anyways, something to keep in mind. If we ever create a packaged solution for Babylon Native, this will become an issue.

bghgary (approved):

PR #665, opened 2021-04-06
Fix crash when rotating screen in XR mode

bghgary:


bghgary:
@CedricGuillemet Can you test this with your devices?

CedricGuillemet:
> @CedricGuillemet Can you test this with your devices?

Tested OK on device. No crash when rotating screen.

Dependencies/xr/Include/XR.h
```diff
@@ -333,7 +333,7 @@ namespace xr
             // Move to private when changing to unique_ptr.
             Session(System& system, void* graphicsDevice, std::function<void*()> windowProvider);
 
-            std::unique_ptr<Frame> GetNextFrame(bool& shouldEndSession, bool& shouldRestartSession, std::function<void(void* texturePointer)> deletedTextureCallback = [](void*){});
+            std::unique_ptr<Frame> GetNextFrame(bool& shouldEndSession, bool& shouldRestartSession, std::function<arcana::task<void, std::exception_ptr>(void*)> deletedTextureAsyncCallback = [](void*){ return arcana::task_from_result<std::exception_ptr>(); });
```
ryantrem, line 336: Can we default this to `nullptr`?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -1080,19 +1080,15 @@ namespace xr
             }
         }
 
-        void DestroyDisplayResources(std::function<void(void* texturePointer)> deletedTextureCallback = [](void*){})
+        void DestroyDisplayResources(std::function<arcana::task<void, std::exception_ptr>(void* texturePointer)> deletedTextureAsyncCallback = [](void*){ return arcana::task_from_result<std::exception_ptr>(); })
```
ryantrem, line 1083: Why do we need to specify the default value a second time here? Sorry can't remember.

Dependencies/xr/Include/XR.h
bghgary: We can, but I don't like it. The logic is more complicated.

Dependencies/xr/Source/ARCore/XR.cpp
bghgary: `DestroyDisplayResources` is called without the last parameter in the destructor.

ryantrem (approved):

PR #669, opened 2021-04-07
Add symbol replacement case for children of TInterimSelection nodes

### LINKED ISSUE #554: Shader compiler traverser doesn't handle conditionals with uniforms for D3D
If you have a shader with something like:
```glsl
uniform bool isFoo;
...
vec4 outColor = isFoo ? foo : bar;
// or
vec4 outColor;
if (isFoo)
    outColor = foo;
else
    outColor = bar;
```

This error gets thrown:
https://github.com/BabylonJS/BabylonNative/blob/2227c1ce4f197c585505f555595274b206d51bad/Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp#L70

This is because the traverser hits a parent node of type `TInterimSelection`, which has a condition that contains the uniform that needs to be replaced with a struct member. 

Possible solutions include finding a way to properly replace the symbol in `makeReplacements` (by adding something like `else if (auto* branch = parent->getAsSelectionNode())`), or potentially adding another special case in [`visitSymbol`](https://github.com/BabylonJS/BabylonNative/blob/2227c1ce4f197c585505f555595274b206d51bad/Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp#L112).

Drigax:
This PR adds an additional symbol replacement case for when symbols to be replaces are used as the condition of a branching expression. e.g:

Dependent on https://github.com/KhronosGroup/glslang/pull/2600

```
      ...
            if (symbolToBeReplaced){
                value = trueValue;
            } else {
                value = falseValue;
            }
      ...
```

Fixes #554


Drigax:
Marking as draft until glslang PR is merged.

rgerd (approved):
:shipit:

Drigax:
Android build failures have been resolved. glslang has some functionality to support pre-compiled headers, which we explicitly disable in the Playground app, but haven't been disabled in the ValidationTests app.

Current MacOS build failures appear related to:
https://github.com/BabylonJS/BabylonNative/issues/683

Drigax:
Looks good to go. @bghgary can you merge whenever you get the time?

PR #671, opened 2021-04-08
Fix initialization when openxr resources aren't available

chrisfromwork:
Right now an xrcheck throws crashing the app when openxr resources aren't installed. this change should fail silently.


syntheticmagus (approved):

bghgary (approved):
Ok for now. We really should try to remove the global XR session context such that this only runs when XR is used.

PR #673, opened 2021-04-08
Set cmake version for Mac and Linux

CedricGuillemet:


bghgary (approved):

azure-pipelines.yml
```diff
@@ -304,7 +312,8 @@ jobs:
       git submodule update --init --recursive
     displayName: 'Checkout dependencies'
   - script: |
-      brew install ninja
+      brew uninstall cmake
+      brew install ninja http://homebrew.bintray.com/bottles/cmake-$(CMAKE_VERSION).mojave.bottle.tar.gz
```
bghgary, line 316: Add `cmake --version` to ensure cmake was installed?

CedricGuillemet: brew already displays the version installed ...and also a warning that a newer 3.20 version is available.

CedricGuillemet: Adding it anyway, to be consisten with other jobs.

bghgary: I think only the last command error will fail the build. I found that interim command failures don't stop the script. I don't know what brew does in terms of returning errors for scripts when it fails. It's safer to call cmake to make sure everything worked. Ideally, we can even check that the output version is actually the one we want. Maybe we can put all of these steps into some scripts so that we can use powershell or something. Then we can make sure they work locally first.

CedricGuillemet: Also, do we want to set the cmake version for Windows build?

bghgary: Probably just to be safe, but it's not currently a problem.

PR #674, opened 2021-04-08
Create Graphics from a native Context

CedricGuillemet:


Core/Graphics/Include/Shared/Babylon/Graphics.h
```diff
@@ -15,8 +15,10 @@ namespace Babylon
         ~Graphics();
 
         static std::unique_ptr<Graphics> CreateGraphics(const GraphicsConfiguration& config);
+        static std::unique_ptr<Graphics> CreateGraphics(const ContextConfiguration& config);
 
         void UpdateWindow(const GraphicsConfiguration& config);
+        void UpdateContext(const ContextConfiguration& config);
```
ryantrem, line 21: If you call `CreateGraphics` with a `GraphicsConfiguration` and then call `UpdateContext`, does that work? Or does bgfx blow up?

Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -73,6 +74,7 @@ namespace Babylon
         template<typename WindowT>
         WindowT GetNativeWindow();
```
ryantrem, line 75: Should this be private now since it is only used within this class, and sometimes will be `null`, depending on how you've configured the class?

Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -73,6 +74,7 @@ namespace Babylon
...
         void SetNativeWindow(const GraphicsConfiguration& config);
+        void SetNativeContext(const ContextConfiguration& config);
```
ryantrem, line 77: Should these both just be called `UpdateConfiguration` to be more symmetrical with creation?

Core/Graphics/Include/Android/Babylon/GraphicsPlatform.h
```diff
@@ -1,15 +1,23 @@
...
+    using ContextType = EGLContext;
 
     struct GraphicsConfiguration
     {
         WindowType WindowPtr;
         size_t Width;
         size_t Height;
     };
+
+    struct ContextConfiguration
```
bghgary, line 18: This is minor, but I feel like `GraphicsConfiguration` and `ContextConfiguration` are not good names. It seems like it should be `Window` vs `Context` and not `Graphics` vs. `Context`. Can we rename `GraphicsConfiguration` to `WindowConfiguration` maybe?

Core/Graphics/Source/Graphics.cpp
```diff
@@ -25,6 +30,13 @@ namespace Babylon
         return graphics;
     }
 
+    std::unique_ptr<Graphics> Graphics::CreateGraphics(const ContextConfiguration& config)
+    {
+        std::unique_ptr<Graphics> graphics{ new Graphics() };
```
bghgary, line 35: nit: formatting (in a few places)

```suggestion
        std::unique_ptr<Graphics> graphics{new Graphics()};
```

Core/Graphics/Source/GraphicsImpl.h
bghgary: Similar for the parent functions (UpdateWindow and UpdateContext) in Graphics.h

Core/Graphics/Include/Android/Babylon/GraphicsPlatform.h
Drigax: I originally intended for Graphics configuration to be _the_ config struct, do we want to have multiple structs for different uses?

CedricGuillemet: I have no strong feelings for 1 function with more members in the struct or 2 more explicit functions. Maybe 2 functions make the contract more clear and it simplifies checking logic behind (having both window and context set for example).

ryantrem: I think two makes the contract a lot more self explanatory since you have to pick one or the other.

Core/Graphics/Include/Shared/Babylon/Graphics.h
CedricGuillemet: Actually, it doesn't make sense to have `UpdateContext` as a public method. Once bgfx is created with a context, you can't change it. You can change the output window but all GPU resources are linked with a context. Moving `UpdateContext`  to private.

Core/Graphics/Source/GraphicsImpl.h
CedricGuillemet: yes, now private.

Core/Graphics/Source/GraphicsImpl.cpp
```diff
@@ -45,14 +45,22 @@ namespace Babylon
...
-    void GraphicsImpl::SetNativeWindow(const GraphicsConfiguration& config)
+    void GraphicsImpl::UpdateWindow(const WindowConfiguration& config)
     {
         std::scoped_lock lock{m_state.Mutex};
         m_state.Bgfx.Dirty = true;
         ConfigureBgfxPlatformData(config, m_state.Bgfx.InitState.platformData);
         UpdateDevicePixelRatio();
     }
 
+    void GraphicsImpl::UpdateContext(const ContextConfiguration& config)
```
ryantrem, line 56: Didn't you say this can't be called a second time? If that's the case, then it seems like we should remove the `UpdateContext` function and just put this logic directly in `CreateGraphics(const ContextConfiguration& config)`.

Core/Graphics/Include/Shared/Babylon/Graphics.h
```diff
@@ -14,9 +14,10 @@ namespace Babylon
     public:
         ~Graphics();
 
-        static std::unique_ptr<Graphics> CreateGraphics(const GraphicsConfiguration& config);
+        static std::unique_ptr<Graphics> CreateGraphics(const WindowConfiguration& config);
+        static std::unique_ptr<Graphics> CreateGraphics(const ContextConfiguration& config);
 
-        void UpdateWindow(const GraphicsConfiguration& config);
+        void UpdateWindow(const WindowConfiguration& config);
```
ryantrem, line 20: I'm still unclear, can you call this function after having created a graphics object with a `ContextConfiguration`? And if so, what exactly does that do?

Core/Graphics/Include/Android/Babylon/GraphicsPlatform.h
syntheticmagus: Is the expectation then that the chosen struct type must be comprehensively populated? If so, unless I misunderstand, that seems like it goes against one of the primary arguments for using structs in the first place, which (as I recall) was that using such a config struct would make it easier to update the available contracts by simply extending the struct and allowing consumers to leave new fields empty if they didn't want to use them. If we're going to have a different struct for every use case, is there an advantage to having structs at all versus just having overloads?

Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -21,7 +21,8 @@
 
 namespace Babylon
 {
-    struct GraphicsConfiguration;
+    struct WindowConfiguration;
+    struct ContextConfiguration;
```
syntheticmagus, line 25: Is this sustainable? Is every platform going to have a comparable configuration struct for every construction scenario, or is it possible that some scenarios will be highly platform-specific such that we might end up with things like `AppleHeadlessNoContextConfiguration` for specialized scenarios? If the latter, that seems like an argument to keep the one-struct paradigm so that the platform specifics can be pushed out above and below this contract layer without exposing them here.

Core/Graphics/Include/Shared/Babylon/Graphics.h
CedricGuillemet: native window is only used at bgfx context creation (bgfx::init). So updateWindow will do nothing but update framebuffer with backbuffer ratio.

Core/Graphics/Source/GraphicsImpl.cpp
CedricGuillemet: Between createGraphics being static and m_impl, I don't think I can put the code in there.

bghgary (approved):

PR #675, opened 2021-04-08
Add ability to obtain native anchor through xr component

chrisfromwork:
Some services (Azure Spatial Anchors) need access to native anchors from XRAnchors to work correctly. This pr does the following:

1. An XRAnchor can now be passed to the the global xr component to obtain a void native anchor pointer as a uintptr_t
2. The OpenXR xr component now exposes the XrSpatialAnchorMSFT type instead of a pointer to the anchor space

Note: I don't have a great way to test this change currently. There may be subsequent follow up prs with fixes.


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -3059,6 +3060,24 @@ namespace Babylon
 
                 return info.Env().Undefined();
             }
+
+            Napi::Value GetNativeAnchor(const Napi::CallbackInfo& info)
+            {
+                if (info.Length() != 1 ||
+                    !info[0].IsObject())
+                {
+                    throw std::exception(/*invalid arguments*/);
```
ryantrem, line 3069: Add some message?
```throw std::runtime_error{"A single object argument is required."};```

**Review by ryantrem — APPROVED** [review]

File: Dependencies/xr/Source/OpenXR/XR.cpp [hunk 2]
```diff
@@ -587,7 +587,7 @@ namespace xr
             anchorSpaceCreateInfo.poseInAnchorSpace = xr::math::Pose::Identity();
             CHECK_XRCMD(apiExtensions.xrCreateSpatialAnchorSpaceMSFT(session, &anchorSpaceCreateInfo, anchorSpace->Space.Put()));
 
-            const auto nativeAnchorPtr = reinterpret_cast<NativeAnchorPtr>(anchorSpace.get());
+            const auto nativeAnchorPtr = reinterpret_cast<NativeAnchorPtr>(anchorSpace->Anchor.Get());
```
chrisfromwork, line 590: @satyapoojasama  and @rgerd  we need to confirm that we should expose the anchor compared to a space here.

rgerd: This looks correct to me. There's currently a 1-1-1 mapping of Anchor, Identity Pose Space, and AnchorSpace. I think I just forgot and did the intuitive (albeit incorrect) thing.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -3059,6 +3060,24 @@ namespace Babylon
...
+                }
+
+                auto xrAnchor{ XRAnchor::Unwrap(info[0].ToObject()) };
```
rgerd, line 3072: nit: These could be `const auto`

rgerd (approved):

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: updated

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: updated

PR #676, opened 2021-04-09
Add a xr context definition for ARCore

chrisfromwork:
This change moves ARCore resources to a location accessible to external plugins.

- [x] Note: we don't have any usage of this logic right now. Upcoming work for azure spatial anchors will validate this change.


rgerd (approved):

Alex-MSFT (approved):

Dependencies/xr/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(SOURCES "Include/XR.h"
 
 if (ANDROID)
     set(SOURCES ${SOURCES}
+        "Source/ARCore/Include/IXrContextARCore.h"
```
ryantrem, line 10: Since this is already in the ARCore folder it seems like it should just be named IXrContext. That would be more similar to the rest of the code.

Dependencies/xr/Source/ARCore/Include/IXrContextARCore.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <arcore_c_api.h>
+
+typedef struct IXrContextARCore
+{
+    virtual bool IsInitialized() const = 0;
+    virtual ArSession* XrSession() const = 0;
```
ryantrem, line 8: Should this just be named Session? The xr prefix seems redundant.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -40,8 +42,31 @@ namespace xr
     // Permission request ID used to uniquely identify our request in the callback when calling requestPermissions.
     const int PERMISSION_REQUEST_ID{ 8435 };
 
+    struct XrContextARCore : public IXrContextARCore {
```
ryantrem, line 45: Similar comment, seems like this could just be XrContext, or even just Context.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -1313,14 +1340,12 @@ namespace xr
 
     uintptr_t System::GetNativeXrContext()
     {
-        // TODO
-        return 0;
+        return reinterpret_cast<uintptr_t>(&m_impl->XrContext);
```
ryantrem, line 1343: Not new, but it seems a little weird to return this as an opaque pointer here since it is NativeXR that has knowledge of napi and makes the decision to expose the opaque pointer. Would it make more sense to return an IContext and then let NativeXr choose to cast to an opaque pointer?

Dependencies/xr/CMakeLists.txt
tymlipari: I'm assuming this is a header we'll publish in which case I'd argue we should keep the names distinct so that a consuming module can be built to target multiple platforms (OpenXR, ARCore, etc.). While we could require that they end up with an include path that's like `#include <ARCore/IXrContext.h>` or something like that, I think keeping the files named separately is still good.

Dependencies/xr/Source/ARCore/XR.cpp
tymlipari, line 45: Any reason this needs to be a separate object and not an interface directly off of the `Impl`?

Dependencies/xr/CMakeLists.txt
chrisfromwork: talked with @ryantrem offline. If we want to support oculus quest we could have both IXrContextARCore.h and IXrContextOpenXR.h distributed with android. This naming choice is primarily to avoid file renames/breaking changes in the future if different platforms support multiple xr backends.

Dependencies/xr/Source/ARCore/Include/IXrContextARCore.h
chrisfromwork: the prefix was to prevent collisions with the underlying Session variable. I don't have a preference here. if theres a good way to have functions and fields have the same name in a struct we could go that route instead.

Dependencies/xr/Source/ARCore/XR.cpp
chrisfromwork: we don't currently have a universal IContext definition. if we want a global interface we could pass around a pointer. Otherwise IContext likely will just be a typedef of a similar opaque pointer

Dependencies/xr/Source/ARCore/XR.cpp
chrisfromwork: no reason, can try that out as well

chrisfromwork: looking back on this, we want this XrContextARCore to exist independent of a xr session. we could have the System::Impl extend this interface, but I don't think we can have the session::impl implement this interface. Probably still worth decreasing the number of structs, so ill update the system impl to implement IXrContextARCore

chrisfromwork: actually, System::Impl already has an is initialized function that does not have the same usage as the IXrContextARCore::IsInitialized function. I think we should keep this XrContextARCore as a separate struct

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -1029,7 +1056,7 @@ namespace xr
         }
 
     private:
-        bool isInitialized{false};
+        bool& isInitialized;
```
tymlipari, line 1059: >bool& isInitialized; [](http://example.com/codeflow?start=8&length=20)

Rather than store these directly as references, would it just make more sense to remove the member itself and always call into the context object instead? Member references are often difficult to reason about object lifetimes for.

chrisfromwork: seems reasonable, let me update this code :)

chrisfromwork: fwiw, the system impl is already obtained through a reference like this so going through it seems to have this same potential risk. I'm going to change the context to a shared pointer so the session and system can have their own individual references

chrisfromwork: the change is much larger now in scope, but we now have the value backed by a shared pointer that will exist for the whole session lifecycle even if the system gets torn down for some reason

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -40,8 +42,33 @@ namespace xr
...
+        {
+            return Session;
+        }
+
+        ArFrame* XrFrame() const override
+        {
+            return Frame;
+        }
+
+        virtual ~XrContextARCore(){}
```
tymlipari, line 65: nit: `virtual ~XrContextARCore() = default`

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -40,8 +42,33 @@ namespace xr
     // Permission request ID used to uniquely identify our request in the callback when calling requestPermissions.
     const int PERMISSION_REQUEST_ID{ 8435 };
 
+    struct XrContextARCore : public IXrContextARCore {
+        bool Initialized{false};
+        ArSession* Session{nullptr};
```
tymlipari, line 47: Where do we destroy the `ArSession` object? Since this object is now the holder/owner of it, would it make sense to move that ownership into the `XrContextARCore` destructor?

tymlipari (approved):

Dependencies/xr/Source/ARCore/XR.cpp
chrisfromwork: its destroyed in the session impl destructor. I'll make sure its set to a nullptr, but i don't think we should change this life cycle behavior while folks are actively investigating xr life cycle bugs.

ryantrem (approved):

PR #677, opened 2021-04-09
Add a pr validation job to test building the jsi flavor of napi + all components used in BabylonReactNative

chrisfromwork:
This should help us more quickly identify compile time regressions related to using napi logic that isn't implemented by the jsi flavor of napi

- [x] Note: we need to wait for https://github.com/BabylonJS/BabylonNative/pull/680


azure-pipelines.yml
```diff
@@ -300,6 +300,31 @@ jobs:
...
+    vmImage: 'windows-latest'
+
+  steps:
+  - script: |
+      git submodule update --init --recursive
+    displayName: 'Checkout dependencies'
+
+  - script: |
+      cd Validation/napi-jsi
+      npm install
```
chrisfromwork, line 314: not sure if folks know whether npm is installed by default on windows agents?

chrisfromwork: or if there is a better way to flag this needed dependency

ryantrem: Yes, it is preinstalled (it's part of Node).
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md

ryantrem: More generally: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software

chrisfromwork: sounds good, im going to look into also using the npm install azure devops task

bghgary:
I think this test should be inside napi?

chrisfromwork:
@bghgary sounds good. Any preference on folder naming/structure? Also are there any concerns around putting this test in the napi folder if it references napi-consuming components?

bghgary:
Oh, I'm dumb. This is testing Babylon Native compiled with napi-jsi. Never mind then.

chrisfromwork:
@bghgary yeah most of the compile time regressions we have seen have been related to NativeXr's usage of napi-jsi. This should allow us to build everything.

bghgary:
Ok, in that case, I think we should add functionality to the existing ValidationTests for say Win32 or UWP to support compiling against JSI. Then we can actually run a validation test against it also.

chrisfromwork:
> Ok, in that case, I think we should add functionality to the existing ValidationTests for say Win32 or UWP to support compiling against JSI. Then we can actually run a validation test against it also.

Do you know of examples for populating the jsi runtime outside of a react-native/react-native-windows application? I'm not sure how to call into this napi jsi with a standard uwp/win32 babylonnative application.

bghgary:
I did a local prototype of this with napi-jsi against the Chakra version at some point when I was implementing napi-jsi. I don't think I pushed it anywhere though as it was mostly hacks. All you really need is a `jsiRuntime&` and then you are good to go to create a napi environment using napi-jsi.

chrisfromwork:
@bghgary well track napi-jsi usage tests here: https://github.com/BabylonJS/BabylonNative/issues/678

@bghgary @ryantrem @syntheticmagus @tymlipari do folks have a preference on where this initial napi-jsi cmake setup lives? Do folks like Validation\napi-jsi? Would they prefer Apps\NapiJsiTest or Apps\ValidationTests\napi-jsi instead?

bghgary:
I vote against Validation\napi-jsi since adding another top level folder is too much in my opinion. I would expect this folder to go away when we do something better (i.e. add unit tests for napi or integrate with existing ValidationTests app).

chrisfromwork:
@bghgary I've moved these files to Apps/napi-jsi for the time being.

Apps/napi-jsi/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+# cmake 3.15.3+ because of a c++ regression when swift is enable
+cmake_minimum_required(VERSION 3.15.3)
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+project(napi-jsi)
+
+# Shared component source will be built in this UWP flavor
+# We can support other flavors if the need arises, but a UWP build should get us the first wave of napi-jsi build coverage
+set(BABYLON_NATIVE_PLATFORM "UWP")
+set(CMAKE_SYSTEM_NAME "WindowsStore")
```
tymlipari, line 10: Should we predefine the `CMAKE_SYSTEM_NAME`? Isn't this usually inferred or overridden from the command line?

azure-pipelines.yml
```diff
@@ -300,6 +300,36 @@ jobs:
       msbuildArgs: '/p:AppxPackageSigningEnabled=false'
     displayName: 'Build UWP_x86'
 
+- job: uwp_napi_jsi_x64
+  pool:
+    vmImage: 'windows-latest'
+
+  steps:
+  - script: |
```
tymlipari, line 308: You don't need to manually invoke the submodule sync. As part of the Azure Pipelines schema you can specify how you want the repo to be cloned:

```yml
steps:
  checkout: self
  clean: true
  submodules: true

  - task: ...
```

Apps/napi-jsi/CMakeLists.txt
chrisfromwork: Updated the pipeline definition. it seems like UWP doesn't get resolved on its own, but if we specify it to cmake as an argument we dont need these definitions.

azure-pipelines.yml
chrisfromwork: this should now be fixed

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -33,7 +31,7 @@ namespace
                 return bgfx::TextureFormat::D24S8;
 
             default:
-                throw std::runtime_error{"Unsupported texture format"};
+                throw std::exception{/* Unsupported texture format */};
```
chrisfromwork, line 34: todo: this needs to be reverted/merged the wrong master branch

azure-pipelines.yml
chrisfromwork: fwiw it needs to be:

  - checkout: self
    clean: true
    submodules: recursive

.github/jobs/napijsi.yml
```diff
@@ -0,0 +1,36 @@
+parameters:
```
CedricGuillemet, line 1: I would name this yml `uwp_napijsi.yml` to be more consistent with other templates.

chrisfromwork:
closing and reopening based on failed win32 validation test (unrelated)

rgerd (approved):

PR #680, opened 2021-04-12
[CI] Refactor pipelines

CedricGuillemet:
- Refactor azure-pipelines.yml to get rid of copy/pasted jobs
- Install CMake MacOS without Homebrew direct link as Bintray is stopping the service


.github/jobs/macos_cmake.yml
```diff
@@ -0,0 +1,14 @@
+parameters:
+  vmImage: ''
+
+steps:
+  - script: |
+      brew uninstall cmake
```
Drigax, line 6: Looks like this should workaround https://github.com/BabylonJS/BabylonNative/issues/683, thanks!

CedricGuillemet: Same for ninja which is also on homebrew

.github/jobs/macos_cmake.yml
```diff
@@ -0,0 +1,15 @@
+parameters:
+  vmImage: ''
+
+steps:
+  - script: |
+      wget https://github.com/Kitware/CMake/releases/download/v$(CMAKE_VERSION)/cmake-$(CMAKE_VERSION)-macos-universal.tar.gz
```
Drigax, line 6: I noticed that my dev machine does not have wget, adding some snippets so all our bases are covered for installation and reinstallation, unless we're already doing this in a previous step.
```suggestion
        brew install wget
        wget https://github.com/Kitware/CMake/releases/download/v$(CMAKE_VERSION)/cmake-$(CMAKE_VERSION)-macos-universal.tar.gz
```

.github/jobs/macos_cmake.yml
```diff
@@ -0,0 +1,15 @@
...
+      sudo mkdir -p /opt/cmake
```
Drigax, line 7: do we want to remove /opt/cmake if it already exists?
```suggestion
      sudo rm -rf /opt/cmake
      sudo mkdir -p /opt/cmake
```

.github/jobs/macos_cmake.yml
CedricGuillemet: wget is available by default on all VMs. Do it locally only.

.github/jobs/macos_cmake.yml
CedricGuillemet: /opt should be empty by default in the VMs

.github/jobs/macos_cmake.yml
```diff
@@ -0,0 +1,15 @@
...
+      sudo tar xvzf cmake-$(CMAKE_VERSION)-macos-universal.tar.gz -C /opt/cmake
+      sudo ln -s /opt/cmake/cmake-$(CMAKE_VERSION)-macos-universal/CMake.app/contents/bin/cmake /usr/local/bin/cmake
```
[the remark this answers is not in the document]
CedricGuillemet: yes, done.

.github/jobs/macos_cmake.yml
Drigax: cool, not needed.

.github/jobs/macos_cmake.yml
Drigax: Ok, I wasn't sure what the lifecycle for a VM was, if it was recreated from an image for each pipeline run, or only periodically

bghgary (approved):

PR #682, opened 2021-04-12
Add validation script to ensure that babylon.max.js changes match the most recent npm preview package of babylonjs

chrisfromwork:
This review contains the following:

1. Helper scripts to obtain babylon.max.js and other babylon.js resources from npm packages
2. A package.json is added to the Apps\BabylonScripts folder so that folks can more easily determine what Babylon.js npm package BabylonNative is currently using
3. A pr validation step to ensure that the babylon.max.js/peer scripts checked in match the specified npm package


chrisfromwork:
@bghgary do you know whether its possible to get an additional ci pipeline defined for azure-pipelines-babylonscripts.yml?

bghgary:
> @bghgary do you know whether its possible to get an additional ci pipeline defined for azure-pipelines-babylonscripts.yml?

I don't know. But a quick search found this: https://github.com/Microsoft/azure-pipelines-agent/issues/1885. Is this helpful?

rgerd:
Is there any reason we shouldn't just make people run npm install, and stop checking the files in directly?

Apps/BabylonScripts/ValidateFiles.ps1
```diff
@@ -0,0 +1,32 @@
+Push-Location $PSScriptRoot
+Write-Host "Installing npm packages for comparison..."
+npm install
```
rgerd, line 3: Is there any reason we shouldn't just make people npm install and stop checking the files in directly?

chrisfromwork: that would require all developers to start using npm. Right now in this current form npm is only used for validation and isn't required for development if you aren't changing these script files

ryantrem:
Does this mean that as soon as the Babylon.js NPM packages are updated on npmjs.org that Babylon Native builds will start failing until we update babylon.max.js etc.?

bghgary (changes requested):
Let's discuss these changes before moving further.

chrisfromwork:
@bghgary sounds good. This pr was mainly an example of something we could do to to ensure babylon.max.js matches an npm package.

chrisfromwork:
closing and reopening to see if it restarts pr ci, for some reason its not getting queued for the pipeline

bghgary (approved):

Apps/BabylonScripts/Build/DownloadAndCopyFiles.ps1
```diff
@@ -0,0 +1,24 @@
+Push-Location $PSScriptRoot
+Write-Host "Installing npm packages..."
+npm install
+
+$files = "babylon.gui.js",
```
rgerd, line 5: I wonder if you could do some `Get-ChildItem` fanciness and glob *.js, *.js.map, *.d.ts instead of needing to explicitly list the files.

rgerd (approved):

Apps/BabylonScripts/Build/DownloadAndCopyFiles.ps1
chrisfromwork: I don't think we'll ever get away from needing to list specific file requirements. I think this is okay for the time being.

PR #684, opened 2021-04-12
Change cmake install link to point to formula for cmake 3.19.7

Drigax:
Instead of pointing to the brew hosted tar binary for cmake 3.19, use an older formula to retrieve the cmake 3.19.7 bottle.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,7 +7,7 @@ pr:
 variables:
 - name: CMAKE_VERSION
   value: 3.19.6
-  
+
 jobs:
 - job: macOS
   pool:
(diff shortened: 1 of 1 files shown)


Drigax:
This approach won't work on its own. Currently, Homebrew persists a short list of stable releases, subject to change. Using a formula for a previous version of a package will not work without maintaining our own custom tap/package repository to point the homebrew client towards.

Drigax:
PR has been modified to install CMake and Ninja manually from the github-hosted packages.

chrisfromwork (approved):

chrisfromwork:
I am in favor of this change but would suggest syncing with other BabylonNative owners to ensure this is the correct fix

Drigax:
#680 does all this and is generally a better solution. Lets merge that instead.

PR #690, opened 2021-04-15
Linux URL path fix

CedricGuillemet:
Small fixes when doing a quick test on Linux CI/Validation tests

- fix escape character for URL path
- update texture cache test snippet
- increase LOD test tolerance

Deadlock issue is still present.


Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -174,7 +174,58 @@ namespace UrlLib
             if (curl)
             {
                 data.clear();
-                curl_easy_setopt(curl, CURLOPT_URL, m_url.data());
+
+                CURLUcode rc;
+                CURLU *url = curl_url();
```
bghgary, line 179: nit in a few places:

```suggestion
                CURLU* url = curl_url();
```

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -174,7 +174,58 @@ namespace UrlLib
...
+                rc = curl_url_set(url, CURLUPART_URL, m_url.c_str(), 0);
+                if (rc != CURLUE_OK)
+                {
+                    throw std::runtime_error("CURL: Unable to build URL.");
```
bghgary, line 183: nit again in a few places:
```suggestion
                    throw std::runtime_error{"CURL: Unable to build URL."};
```

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -174,7 +174,58 @@ namespace UrlLib
...
+                    curl_easy_setopt(curl, CURLOPT_URL, urlEscaped);
+                    curl_free(urlEscaped);
+                }
+                else
+                {
+                    curl_easy_setopt(curl, CURLOPT_URL, m_url.c_str());
+                }
+
+                curl_free(scheme);
+                curl_url_cleanup(url);
```
bghgary, line 227: Because of the throws earlier, it's probably good to put these clean up statements in gsl::finally so that they will get called even if an exception is thrown.

bghgary: Applies to all the `curl_free` and other clean up functions.

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -174,7 +174,58 @@ namespace UrlLib
...
+                    char *urlEscaped;
+                    rc = curl_url_get(url, CURLUPART_URL, &urlEscaped, 0);
+                    if (rc != CURLUE_OK)
+                    {
+                        throw std::runtime_error("CURL: Unable to get URL string.");
+                    }
+
+                    curl_easy_setopt(curl, CURLOPT_URL, urlEscaped);
+                    curl_free(urlEscaped);
+                }
```
bghgary, line 220: Maybe a comment to explain why this is necessary?

bghgary (changes requested):

bghgary (approved):

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -174,7 +174,61 @@ namespace UrlLib
             if (curl)
             {
                 data.clear();
-                curl_easy_setopt(curl, CURLOPT_URL, m_url.data());
+
+                // libCurl doesn't escape url strings automatically.
```
bghgary, line 178: nit: it might be good to put this in a private helper function to make the code easier to read.

CedricGuillemet: Will update in my next PR.

PR #698, opened 2021-04-19
Changing hand data to align with latest webXR changes

### LINKED ISSUE #700: Send hand select/squeeze information in a gamepad object to match webXR
webXR switched to providing hand select/squeeze in a gamepad object, so we need to update the Native code to match this.

Related issue in Babylon.JS: https://github.com/BabylonJS/Babylon.js/issues/10144

rickfromwork:
The webXR hand now passes through select/squeeze via the gamepad component. As the hand squeeze is a special input, it shouldn't be hooked up to things like the XRSession's onsqueeze events, which listen to the primary squeeze event.

Resolves #700


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1664,14 +1665,17 @@ namespace xr
...
+                                (m_impl->TryUpdateControllerBooleanAction(actionResources.HandGetSqueezeAction, session, gamepadObject.Buttons[controllerInfo.SPECIAL_BUTTON_1].Pressed)))
                             {
                                 gamepadObject.Buttons[controllerInfo.TRIGGER_BUTTON].Value = (gamepadObject.Buttons[controllerInfo.TRIGGER_BUTTON].Pressed);
                                 gamepadObject.Buttons[controllerInfo.TRIGGER_BUTTON].Touched = (gamepadObject.Buttons[controllerInfo.TRIGGER_BUTTON].Pressed);
-                                gamepadObject.Buttons[controllerInfo.SQUEEZE_BUTTON].Value = (gamepadObject.Buttons[controllerInfo.SQUEEZE_BUTTON].Pressed);
-                                gamepadObject.Buttons[controllerInfo.SQUEEZE_BUTTON].Touched = (gamepadObject.Buttons[controllerInfo.SQUEEZE_BUTTON].Pressed);
+                                gamepadObject.Buttons[controllerInfo.SPECIAL_BUTTON_1].Value = (gamepadObject.Buttons[controllerInfo.SPECIAL_BUTTON_1].Pressed);
```
chrisfromwork, line 1677: >SPECIAL_BUTTON_1 [](http://example.com/codeflow?start=69&length=16)

nit: Can we rename special to something like unique/custom and avoid having numbers in constexpr names here? 

maybe just call this CUSTOM_HARDWARE_BUTTON for now even if we need to add more in the future

Dependencies/xr/Include/XR.h
```diff
@@ -209,7 +210,10 @@ namespace xr
                     };
 
                     std::array<float, DEFAULT_CONTROLLER_AXES> Axes;
-                    std::array<Button, DEFAULT_CONTROLLER_BUTTONS> Buttons;
+                    std::array<Button, DEFAULT_CONTROLLER_BUTTONS + SPECIAL_CONTROLLER_BUTTONS> Buttons;
+
+                    size_t AxesUsed{ DEFAULT_CONTROLLER_AXES };
```
chrisfromwork, line 215: >AxesUsed [](http://example.com/codeflow?start=27&length=8)

nit: AxesCount

Dependencies/xr/Include/XR.h
```diff
@@ -209,7 +210,10 @@ namespace xr
...
+                    size_t ButtonsUsed{ DEFAULT_CONTROLLER_BUTTONS };
```
chrisfromwork, line 216: >ButtonsUsed [](http://example.com/codeflow?start=27&length=11)

nit ButtonCount or NumButtons

Dependencies/xr/Include/XR.h
```diff
@@ -177,6 +177,7 @@ namespace xr
         static constexpr float DEFAULT_DEPTH_NEAR_Z{ 0.5f };
         static constexpr float DEFAULT_DEPTH_FAR_Z{ 1000.f };
         static constexpr uint32_t DEFAULT_CONTROLLER_BUTTONS{ 4 };
```
chrisfromwork, line 179: >DEFAULT_CONTROLLER_BUTTONS [](http://example.com/codeflow?start=34&length=26)

nit: can we get these changed to DEFAULT_CONTROLLER_BUTTON_COUNT etc so that folks know this is a length. right now it reads like the value is a button object

chrisfromwork: or NUM_DEFAULT_CONTROLLER_BUTTONS I don't personally have a strong preference on how the fact this is a length/count is expressed

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -140,38 +140,47 @@ namespace
...
-            profiles.Set(uint32_t{0}, Napi::String::New(env, "generic-hand-select-grasp"));
-            profiles.Set(uint32_t{1}, Napi::String::New(env, "generic-hand-select"));
-            jsInputSource.Set("profiles", profiles);
+                jsInputSource.Set("hand", handJointCollection);
+
+            }
+            else
+            {
+                // If hand joints aren't available on a hand input, send a null hand object
+                jsInputSource.Set("hand", env.Null());
```
chrisfromwork, line 173: >Null [](http://example.com/codeflow?start=46&length=4)

nit: should this be null or undefined?

Dependencies/xr/Include/XR.h
```diff
@@ -209,7 +210,10 @@ namespace xr
                     };
 
                     std::array<float, DEFAULT_CONTROLLER_AXES> Axes;
-                    std::array<Button, DEFAULT_CONTROLLER_BUTTONS> Buttons;
+                    std::array<Button, DEFAULT_CONTROLLER_BUTTONS + SPECIAL_CONTROLLER_BUTTONS> Buttons;
```
chrisfromwork, line 213: does this really need to be a static size or could we use a vector with the correct sizes and avoid having an additional count value?

chrisfromwork (approved):
:shipit:

Dependencies/xr/Include/XR.h
rgerd: or AxisCount, to match ButtonCount

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: Null, same as when tracking is not supported
https://www.w3.org/TR/webxr-hand-input-1/#xrinputsource-interface

satyapoojasama (approved):

Issue #678, opened 2021-04-09
Setup a UWP validation test and run through napi-jsi validation scenarios


chrisfromwork:
We should setup uwp tests to run through initialization and basic xr scenarios. At the same time we could benefit from having test coverage of napi jsi

CedricGuillemet:
@chrisfromwork When building and running UWP Playground, I get this error : 
`Uncaught Error: 'chromeDevTools' is not defined`
Any idea on how to fix this issue?

bghgary:
This is because we added `chromeDevTools` to the playground. We should check if `chromeDevTools` exists first before using it.

Personally, I don't think this should be a JS contract. See https://github.com/BabylonJS/BabylonNative/issues/689.

chrisfromwork:
@bghgary is there a suggested fix for the playground? are we missing a line in experience.js? @CedricGuillemet are you unblocked?

bghgary:
The fix is simple. Change this:

https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Scripts/experience.js#L56-L58

To something like this:

```javascript
if (typeof chromeDevTools !== "undefined" && chromeDevTools.supportsInspector()) {
    chromeDevTools.startInspector(5643, "experience.js");
}
```

rgerd:
Once we get #854 closed we can close this one out too.

PR #699, opened 2021-04-20
Allow native access to ChromeDevTools API

### LINKED ISSUE #689: Allow native access to ChromeDevTools plugin API
(#607) introduced Chromium DevTools support as a javascript API, but it feels like it would make more sense as a native API.

rgerd:
We originally made the ChromeDevTools API accessible through JavaScript, but it makes more sense for it to be a native API. This also fixes issues related to not initializing the plugin and hitting an exception in experience.js (noted here https://github.com/BabylonJS/BabylonNative/issues/678#issuecomment-822288897)

fixes #689


Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -68,24 +61,35 @@ namespace Babylon::Plugins::Internal
...
     {
+        s_instance = std::make_unique<ChromeDevTools>(env);
     }
-    
-    Napi::Value ChromeDevTools::SupportsInspector(const Napi::CallbackInfo& info)
+
+    ChromeDevTools* ChromeDevTools::GetInstance()
+    {
+        return s_instance.get();
+    }
```
ryantrem, line 79: Is there any specific reason for the singleton pattern for `ChromeDevTools`? I don't think we use this pattern anywhere else currently. Most recently a pattern we've been using for cases where we want Babylon Native components to have a native contract is to have `Initialize` return an object with the native interface, where the lifetime of that object is managed by the caller. You can see this in [NativeXr](https://github.com/BabylonJS/BabylonNative/blob/master/Plugins/NativeXr/Include/Babylon/Plugins/NativeXr.h#L16) and in [NativeCamera](https://github.com/BabylonJS/BabylonNative/blob/fa042cfae7081b54c9519cc7a130858162d966d0/Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h#L19). I think it would be great to start converging on a common pattern to give the Babylon Native API a more consistent feel.

chrisfromwork:
@rgerd is there a follow up/backlog item that we could file for validation tests that could have found this issue at the time of pr?

rgerd:
@chrisfromwork I filed #702

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -1,91 +1,81 @@
...
 
-        void StopInspector(const Napi::CallbackInfo&)
-        {
-            StopInspector();
-        }
+            ~Impl()
+            {
+                if (m_inspector != nullptr)
+                    m_inspector->stop();
+            }
```
bghgary, line 50: nit: This should be next to the constructor?

Plugins/ChromeDevTools/Source/ChromeDevToolsV8.cpp
```diff
@@ -1,91 +1,81 @@
...
-        }
 
-        void StopInspector(const Napi::CallbackInfo&)
-        {
-            StopInspector();
-        }
+            ~Impl()
+            {
+                if (m_inspector != nullptr)
+                    m_inspector->stop();
```
bghgary, line 49: nit:
```suggestion
                if (m_inspector != nullptr)
                {
                    m_inspector->stop();
                }
```

bghgary (approved):
A couple of minor comments. Seems good to me.

PR #710, opened 2021-04-23
[ANDROID] Test disabling ARM7 builds on CI

CedricGuillemet:
Disable CI ARM7 build for Android. Only Arm64.


Apps/ValidationTests/Android/app/build.gradle
```diff
@@ -44,7 +44,11 @@ android {
             }
         }
         ndk {
-            abiFilters "arm64-v8a", "armeabi-v7a", "x86"
+            if (project.hasProperty("NoArm7")) {
```
bghgary, line 47: nit: `NoArm7` also removes x86? Maybe `Arm64Only`?

bghgary (approved):

Apps/ValidationTests/Android/app/build.gradle
CedricGuillemet: yes, more explicit.

PR #716, opened 2021-04-26
Add DevTools debugging documentation

### LINKED ISSUE #687: Need documentation for using chromium inspector
Now that we have (#607), we have script debugging with Chrome DevTools!

We need documentation outlining how to open the inspector to use it for profiling / setting breakpoints / etc.

rgerd:
This page will serve as documentation for script debugging using Chromium DevTools. Right now the page is explicit about only supporting Win32, with the intention of removing the restriction as the feature grows to support all of our offered platforms.

The rendered page on my private fork: [DebugJavascriptDevTools.md](https://github.com/rgerd/BabylonNative/blob/dev-tools-docs/Documentation/DebugJavascriptDevTools.md)

fixes #687


Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,29 @@
+# BabylonNative Script Debugging on Win32
+You can use Chromium DevTools to debug your JavaScript while running BabylonNative on Win32 (x86/x64) with the V8 javascript engine.
```
chrisfromwork, line 2: >on Win32 (x86/x64) [](http://example.com/codeflow?start=83&length=18)

nit: on Win32 sounds kind of strange. Can this be "You can use chromium devtools to debug JavaScript while running a Win32 BabylonNative application (x86/x64) with the V8 JavaScript engine"?

chrisfromwork, line 2: >javascript [](http://example.com/codeflow?start=114&length=10)

nit: this should have the same capitalization as earlier in the sentence

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Set up BabylonNative to Use V8
+When running CMake, be sure to specify the following flag:
+`-D NAPI_JAVASCRIPT_ENGINE=V8`
```
chrisfromwork, line 6: nit: probably should call out that you are specifying this flag to use the v8 engine which isn't used by default for win32. Although it seems somewhat obvious whats going on in the command, itd still be good to confirm that this enables v8 in the comment

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,29 @@
...
+For example:
+```powershell
+mkdir Build
+cd Build
+cmake -A Win32 -D NAPI_JAVASCRIPT_ENGINE=V8 ..
+```
+
+## Add the Remote Target URL
+Depending on which Chromium-based browser you're using (Edge or Chrome), navigate to `edge://inspect` or `chrome://inspect`.
+Click `Configure...` and in the target discovery settings add an entry `localhost:{port}`, where `{port}` is consistent with the one supplied to the call to [`StartInspector`](https://github.com/BabylonJS/BabylonNative/blob/2db465f5569d8eb833245a41030c5219cfa3fe59/Apps/Playground/Win32/App.cpp#L141-L144).
```
chrisfromwork, line 17: >StartInspector [](http://example.com/codeflow?start=159&length=14)

nit: this is a cool hyper link but we should say what start inspector does in this comment

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,29 @@
...
+
+![Target discovery settings](Images/DevTools/chrome-targets.png)
+
+## Click Inspect
+Open the Playground application and wait for the DevTools page to recognize the inspector server spun up by the app. Once you see the app pop up in the list of Remote Targets, you may click inspect to open up a DevTools window.
```
chrisfromwork, line 22: >Open [](http://example.com/codeflow?start=0&length=4)

nit: Run

chrisfromwork (approved):
:shipit:

Documentation/DebugJavascriptDevTools.md
```diff
@@ -0,0 +1,29 @@
...
+
+![Ready to inspect](Images/DevTools/chrome-playground.png)
+
+## View Logs, Set Breakpoints, Profile, and More
+Once you have the DevTools window open, you can use it in the same way as if you were inspecting a web page.
+
```
chrisfromwork, line 28: nit: if theres an overview page for chromium debugging, we should link to it here

PR #720, opened 2021-04-28
No bgfx::touch just before submit + default state

CedricGuillemet:
bgfx::touch calls bgfx::submit under the hood. 
In a case where there is no clear call but only submit calls, the first submit will call bgfx::touch from the NewView method and will discard states (shaders, vertex buffer,...). More over, no bgfx::touch is needed for the view if there is at least one submit.

webgl default state has face culling disabled ( https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/cullFace ) which is different than bgfx default state. Setting to webgl default state so rendering is the same when no call to engine.setState is performed.

depth test is disabled by default as well : https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/depthFunc


Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -106,6 +106,9 @@ namespace Babylon
 
         // This dummy draw call is here to make sure that the view is cleared
         // if no other draw calls are submitted to the view.
-        encoder->touch(m_viewId.value());
+        if (doTouch)
+        {
+            encoder->touch(m_viewId.value());
+        }
```
bghgary, line 112: This is an odd way to do this. Maybe just move this code into another function and call it only when touch is warranted? Or maybe just call the code without another function since it's so small anyways.

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -227,7 +227,7 @@ namespace Babylon
         bool m_requestAnimationFrameCallbacksScheduled{};
 
         bx::DefaultAllocator m_allocator{};
-        uint64_t m_engineState{BGFX_STATE_DEFAULT};
+        uint64_t m_engineState{BGFX_STATE_WRITE_RGB | BGFX_STATE_WRITE_A | BGFX_STATE_WRITE_Z | BGFX_STATE_MSAA};
```
bghgary, line 230: Comment here might be good.

bghgary (approved):

Core/Graphics/Source/FrameBuffer.cpp
CedricGuillemet: I agree it's a bit odd. After some more thoughts, I still think it's the best way here. I'm against copy/paste to a new function or inlined code because of code divergence that can lead to difficult times when debugging. And the boolean parameter is determined at compile time. I believe moving that boolean to a parameter will give the same binary code with Release/O3 but I have no guarantee.

bghgary: I don't mean copy and paste to a new function. I mean split the function in half. There should be no code divergence. Anyways, it's not a big deal.

PR #721, opened 2021-04-28
Update BabylonNative to consume Babylon.js through npm

chrisfromwork:
This review contains the following changes:

1. The apps folder now has a package.json definition for resolving es5 babylon.js npm packages
2. The readme and pipeline definitions are updated to call out requiring an npm install in the Apps folder prior to development
3. Cmake files are updated to look for babylon.js script files in npm packages
4. We remove the ci step for ensuring babylon.js script files match content in npm packages


CedricGuillemet:
Is it possible to also install npm packages for Android (V8 and JSC) with Apps/package.json?

chrisfromwork:
@CedricGuillemet it should be possible to do this all with cmake, but the ask for the time being was to have a manual npm install step. 

That all going said, it does seem reasonable to obtain android dependencies through this single package.json, let me look into doing so

chrisfromwork:
@CedricGuillemet I have updated us to use a single package.json to pull down android js dependencies. Does the validation test app for android work or require anything to run on device? The playground app on android is working as expected after these changes but the validation test app crashes for me.

chrisfromwork:
@bghgary and @CedricGuillemet I have been able to verify that windows and android builds work locally. Will build ci suffice for validating mac/linux/ios?

.github/workflows/nightly.yml
```diff
@@ -14,20 +14,25 @@ jobs:
       uses: actions/checkout@master
       with:
         submodules: 'recursive'
+    - task: Npm@1
+      inputs:
+        command: 'install'
+        workingDir: 'Apps'
+      displayName: 'Install Babylon.js NPM packages'
```
bghgary, line 21: Why install npm packages at all? It seems like we don't need them.

Apps/Playground/Android/package.json
```diff
@@ -1,16 +0,0 @@
...
-    "jsc-android": "^241213.1.0",
-    "v8-android": "^7.8.2"
-  },
-  "devDependencies": {},
-  "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
-  },
-  "author": "",
-  "license": "MIT"
-}
```
bghgary, line 16: Hmm, this is used only by Android. I'm not sure it's best to move this to the Apps root.

.github/workflows/nightly.yml
chrisfromwork: was just doing this right now to insure the directories existed. It doesn't seem like I can trigger this ci easily in my feature branch

Apps/Playground/Android/package.json
chrisfromwork: @CedricGuillemet @bghgary happy to align with whatever the group decides on

chrisfromwork: fwiw, these two packages add ~300mb of files. But in the current form we have to install twice/separately for validationtests vs playground so we would go from ~600mb added to ~300mb, but we would be obligating all babylonnative developers to download android dependencies for using this repos apps

.github/workflows/nightly.yml
bghgary: There is got to be a way to create directories without install npm packages. :-)

chrisfromwork: this should now be fixed :)

.github/workflows/nightly.yml
```diff
@@ -14,13 +14,13 @@ jobs:
...
-      displayName: 'Install Babylon.js NPM packages'
     - name: Make Solution
-      run: mkdir buildWin32_x64 && 
+      run: mkdir Apps\node_modules &&
+        mkdir Apps\node_modules\babylonjs &&
+        mkdir Apps\node_modules\babylonjs-gui &&
+        mkdir Apps\node_modules\babylonjs-loaders &&
+        mkdir Apps\node_modules\babylonjs-materials &&
+        mkdir buildWin32_x64 && 
         cd buildWin32_x64 &&
```
bghgary: `&&` isn't necessary

```suggestion
      run: mkdir Apps\node_modules
        mkdir Apps\node_modules\babylonjs
        mkdir Apps\node_modules\babylonjs-gui
        mkdir Apps\node_modules\babylonjs-loaders
        mkdir Apps\node_modules\babylonjs-materials
        mkdir buildWin32_x64
        cd buildWin32_x64
```

bghgary: Might also consider separating this into individual steps before calling curl. It's kind of weird to have these commands in the Make solution step.

chrisfromwork: i was able to test this out in my fork, we needed to split this up so that we have the curl downloaded files prior to cmake generation too

.github/workflows/nightly.yml
```diff
@@ -14,25 +14,36 @@ jobs:
...
         mkdir Apps\node_modules\babylonjs &&
         mkdir Apps\node_modules\babylonjs-gui &&
         mkdir Apps\node_modules\babylonjs-loaders &&
-        mkdir Apps\node_modules\babylonjs-materials &&
-        mkdir buildWin32_x64 && 
-        cd buildWin32_x64 &&
-        cmake -G "Visual Studio 16 2019" -A x64 -DBGFX_CONFIG_MEMORY_TRACKING=ON -DBGFX_CONFIG_DEBUG=ON ..
-    - name: Add msbuild to PATH
-      uses: microsoft/setup-msbuild@v1.0.2
+        mkdir Apps\node_modules\babylonjs-materials
```
bghgary: I just looked this up. Curl has a `--create-dirs` option. That should make this easier.

chrisfromwork: this should now be fixed

.github/workflows/nightly.yml
```diff
@@ -14,20 +14,30 @@ jobs:
       uses: actions/checkout@master
       with:
         submodules: 'recursive'
+    - name: Download babylon.max.js
+      run: curl.exe -o Apps/node_modules/babylonjs/babylon.max.js --create-dirs https://raw.githubusercontent.com/BabylonJS/Babylon.js/master/dist/preview%20release/babylon.max.js
```
bghgary, line 18: This isn't from your change, but we shouldn't be using raw github urls to get files.

```suggestion
      run: curl.exe -o Apps/node_modules/babylonjs/babylon.max.js --create-dirs https://preview.babylonjs.com/babylon.max.js
```

Apps/Playground/CMakeLists.txt
```diff
@@ -1,15 +1,17 @@
...
-    "../BabylonScripts/babylon.max.js"
-    "../BabylonScripts/babylon.max.js.map"
-    "../BabylonScripts/babylonjs.materials.js"
-    "../BabylonScripts/babylonjs.materials.js.map"
-    "../BabylonScripts/babylon.gui.js"
-    "../BabylonScripts/babylon.gui.js.map"
-    "../BabylonScripts/ammo.js"
-    "../BabylonScripts/meshwriter.min.js"
-    "../BabylonScripts/recast.js")
+    "../node_modules/babylonjs-loaders/babylonjs.loaders.js"
```
bghgary, line 2: You also need to update these lines:

https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Scripts/experience.js#L1-L4

Apps/Playground/CMakeLists.txt
```diff
@@ -1,15 +1,17 @@
...
+    "../node_modules/babylonjs/babylon.max.js.map"
+    "../node_modules/babylonjs-materials/babylonjs.materials.js"
+    "../node_modules/babylonjs-materials/babylonjs.materials.js.map"
+    "../node_modules/babylonjs-gui/babylon.gui.js"
+    "../node_modules/babylonjs-gui/babylon.gui.js.map")
+
+set(DEPENDENCIES
+    "../Dependencies/ammo.js"
+    "../Dependencies/meshwriter.min.js"
+    "../Dependencies/recast.js")
```
bghgary, line 14: nit: I would not bother creating a separate list of these. Conceptually, they are still babylon scripts. They just don't come from NPM.

bghgary (approved):

PR #725, opened 2021-04-29
Reject promise on xr initialization failure

chrisfromwork:
Right now we have a never ending while loop for xr initialization on pc when no vr headset is plugged in. After syncing with babylon.js webxr folks, it sounds like the following is the expected behavior:

* navigator.xr exists whether or not a vr headset is plugged in
* creating the default experience should succeed even if the xr session is unsupported
* attempting to enterXRAsync when the session is not supported should result in a rejected promise/a thrown exception when using await.

This change has us reject the requestSession promise when initialization fails.

I have tested both pc with no vr headset and hololens 2. If we need to test other platforms, please let me know. I am not sure if other initilization logic was relying on this repeated synchronous query for system initialization


Alex-MSFT (approved):

rgerd (approved):

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -479,12 +479,10 @@ namespace Babylon
...
 
-                    if (!m_system.IsInitialized())
+                    if (!m_system.IsInitialized() &&
+                        !m_system.TryInitialize())
                     {
-                        while (!m_system.TryInitialize())
-                        {
-                            // do nothing
-                        }
+                        return arcana::task_from_error<void>(std::make_exception_ptr(std::runtime_error{"Failed to initialize xr system."}));
```
bghgary, line 485:
```suggestion
                        throw std::runtime_error{"Failed to initialize xr system."};
```

bghgary (commented):
I can't comment on whether this change is right or not since I don't know how this API is supposed to work.

bghgary:
If we don't need to poll to initialize, `TryInitialize` doesn't really make sense any more. We should perhaps rename it.

Plugins/NativeXr/Source/NativeXr.cpp
chrisfromwork: @ryantrem with these arcana tasks if we are trying to reject the promise, is it better to return the task_from_error or just throw here?

chrisfromwork: @bghgary was this commit suggestion related to not knowing how the api works? or is this the correct way to reject this requestSession promise?

chrisfromwork:
> If we don't need to poll to initialize, `TryInitialize` doesn't really make sense any more. We should perhaps rename it.

I personally prefer when things returning booleans have a Try* prefix. That going said, I'm not sure if that aligns with BabylonNative naming standards

Plugins/NativeXr/Source/NativeXr.cpp
bghgary: A throw here will be converted to a task_from_error automatically in arcana.

bghgary:
We don't have strict naming standards. I typically see Try functions returning boolean meaning it may not do the work or can't find something (i.e. it is expected that it may return false under normal conditions), which I distinguish from something failing (i.e. it is not expected under normal conditions). Maybe just me.

PR #728, opened 2021-04-30
Move js v8 debugging to supported list

rgerd:


bghgary (approved):

README.md
```diff
@@ -32,6 +32,7 @@ UWP, macOS, iOS, Android.
 - Network requests (including accessing local files using the `file://` protocol) 
 made throught the Babylon.js APIs.
 - Extending JavaScript functionality using N-API and native plugins.
+- Debugging JavaScript with V8 on the following platforms: Win32.
```
bghgary, line 35: Since we only say Win32, maybe we should leave the unsupported line with UWP, Android, etc.?

PR #739, opened 2021-05-11
stencil + fullscreen YFlip

### LINKED ISSUE #559: RenderTargets and order of rendering
framebuffers (for shadows, rtt,..) have a viewId superior to the backbuffer viewid.
Thus, drawcalls in FB are performed after drawcalls in backbuffer.
So, for example, the validation test scene 'point light shadows' needs 1 frame for warming up before the shadow map is rendered and usable.

CedricGuillemet:
Related PR on babylonjs : https://github.com/BabylonJS/Babylon.js/pull/10420

fixes #559 

- Expose Stencil functions
- Create a rendertarget used as back buffer when a fullscreen rendering is needed for YFlip

# Back buffer YFlip
Texture coordinates Y is flipped between API (webgl/opengl Vs Metal/D3D/webgpu).
After some chat with Alexis and Angle tests, this PR does the same thing as them: Flip gl_position.y for all shaders, then do a fullscreen pass to do the last flip.
The shader flip is done in NativeEngine when creating a new program ( `CreateProgram` ). The fullscreen pass in done in NativeEngine.
What's getting tricky is how to handle the backbuffer. With OpenGL, a drawcall on the backbuffer happens in the backBuffer. With D3D, a drawcall on the backBuffer happens in a renderTarget and that renderTarget is used at the end of the frame for the fullscreen pass on the real backbuffer.

In `FrameBufferManager.cpp `, a renderTarget is created and becomes a substitute for `DefaultFrameBuffer`. When the Graphics is resized, the FrameBuffer is also resized. Then 2 new methods are  exposed : `loadBackbufferTexture` and `getFinalBackbuffer`.
 `loadBackbufferTexture`  set the texture Handle of that transient backBuffer texture to the web version.
`getFinalBackbuffer` returns the real back buffer.
At the end of the engine frame, before calling bgfx::frame, a full screen pass is performed with the backBufferTexture as a source into the real back buffer.

This simplifies the code a little bit with the removed YFlip flag and simpler Submit call. Y derivative shader traversal is removed.


Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -112,6 +112,8 @@ namespace Babylon
         using CaptureCallbackTicketT = arcana::ticketed_collection<std::function<void(const BgfxCallback::CaptureData&)>>::ticket;
         CaptureCallbackTicketT AddCaptureCallback(std::function<void(const BgfxCallback::CaptureData&)> callback);
 
+        const std::unique_ptr<FrameBufferManager>& GetFrameBufferManager() const { return m_frameBufferManager; }
```
bghgary, line 115: Where is this used?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1552,62 +1670,26 @@ namespace Babylon
...
+            encoder->setUniform({ it.first }, value.Data.data(), value.ElementLength);
+        }
 
+        if (!bgfx::getCaps()->originBottomLeft)
+        {
             uint64_t engineStateYFlipped = m_engineState;
-            engineStateYFlipped &= ~BGFX_STATE_CULL_MASK;
-            engineStateYFlipped |= (cullCW | cullCCW) << BGFX_STATE_CULL_SHIFT;
-
+            engineStateYFlipped ^= BGFX_STATE_FRONT_CCW;
```
bghgary, line 1682: What does this XOR do?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1508,6 +1599,33 @@ namespace Babylon
...
+        const auto depthOpPass{info[3].As<Napi::Number>().Uint32Value()};
+        const auto func{info[4].As<Napi::Number>().Uint32Value()};
+        const auto ref{info[5].As<Napi::Number>().Uint32Value()};
+        
+        m_stencilState = BGFX_STENCIL_FUNC_RMASK(0xFF); //  always 0xFF
+        m_stencilState |= stencilOpFail;
+        m_stencilState |= depthOpFail;
+        // bgfx write mask is always 0xFF, to not change stencil value when writemask is 0
+        // its value is kept unchanged.
+        // https://github.com/bkaradzic/bgfx/blob/2c21f68998595fa388e25cb6527e82254d0e9bff/src/renderer_d3d11.cpp#L2874
```
bghgary, line 1616: I don't understand this comment. Can you explain more?

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -228,6 +227,7 @@ namespace Babylon
 
         bx::DefaultAllocator m_allocator{};
         uint64_t m_engineState{BGFX_STATE_DEFAULT};
+        uint32_t m_stencilState{ BGFX_STENCIL_TEST_ALWAYS | BGFX_STENCIL_FUNC_REF(0) | BGFX_STENCIL_FUNC_RMASK(0xFF) | BGFX_STENCIL_OP_FAIL_S_KEEP | BGFX_STENCIL_OP_FAIL_Z_KEEP | BGFX_STENCIL_OP_PASS_Z_REPLACE };
```
bghgary, line 230: nit
```suggestion
        uint32_t m_stencilState{BGFX_STENCIL_TEST_ALWAYS | BGFX_STENCIL_FUNC_REF(0) | BGFX_STENCIL_FUNC_RMASK(0xFF) | BGFX_STENCIL_OP_FAIL_S_KEEP | BGFX_STENCIL_OP_FAIL_Z_KEEP | BGFX_STENCIL_OP_PASS_Z_REPLACE};
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -27,6 +27,29 @@ namespace Babylon
...
+            precision highp float;
+            in vec2 vUV;
+            uniform sampler2D backBuffer;
+            out vec4 glFragColor;
+            void main(void)
+            {
+                vec3 color = texture(backBuffer, vUV).xyz;
+                glFragColor = vec4(color, 1.0);
+            }
+        )";
```
bghgary, line 51: Maybe I'm being dense, but I don't see how this shader flips the y axis. How does this work?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -507,9 +567,23 @@ namespace Babylon
...
 
-    Napi::Value NativeEngine::HomogeneousDepth(const Napi::CallbackInfo& info)
+    void NativeEngine::LoadYFlipProgram()
     {
-        return Napi::Value::From(info.Env(), bgfx::getCaps()->homogeneousDepth);
+        if (!bgfx::getCaps()->originBottomLeft)
+        {
+            // This is temporary until a Shader plugin is developed.
+            // Shader is loaded/created here, then passed to graphics Impl for use and destruction.
+            // Idealy, this shader would be created and managed by a 3rd entity (Shader plugin)
```
bghgary, line 576: I think this comment needs to be on the SetYFlipProgram function also.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -628,6 +702,25 @@ namespace Babylon
...
+            const auto lastClosingCurly = vertexSource.find_last_of('}');
+            patchedVertexSource = vertexSource.substr(0, lastClosingCurly);
+            if (!caps->originBottomLeft)
+            {
+                patchedVertexSource += "gl_Position.y *= -1.;\n";
+            }
+            patchedVertexSource += "gl_Position.z = (gl_Position.z + gl_Position.w) / 2.0; }";
+            return patchedVertexSource;
+        }
+        return vertexSource;
```
bghgary, line 721: this ideally should be done with glslang, but i'm okay with a comment saying as such for now.

bghgary (commented):
It would be good to go over this PR in person virtually.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Flip bit

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: bgfx stencil write mask is always 0xFF. when bjs set write mask to 0x0, depth pass op changes to keep so stencil value is not modified....just like having a 0x00 write mask.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: The shader itself doesn't flip. It only copies. UV are handled so texture tap and pixel output make the flip.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: I agree.

Core/Graphics/Source/GraphicsImpl.h
CedricGuillemet: Fixed.

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: done.

Core/Graphics/Source/FrameBufferManager.h
```diff
@@ -33,5 +35,6 @@ namespace Babylon
         std::list<FrameBuffer> m_frameBuffers{};
 
         FrameBuffer m_default;
+        FrameBuffer m_backBuffer;
```
bghgary, line 38: Only minor thing left is the naming of these. It's a bit unclear what these two frame buffers mean now. It would be good to add some comments and make the naming consistent between the variable and the getter function.

bghgary (approved):

Core/Graphics/Source/GraphicsImpl.h
```diff
@@ -129,6 +133,7 @@ namespace Babylon
         bgfx::Encoder* GetEncoderForThread();
         void EndEncoders();
         void CaptureCallback(const BgfxCallback::CaptureData&);
+        void YFlip();
```
syntheticmagus, line 136: I don't think the Graphics component should have this feature/responsibility, at least not long-term. It is `NativeEngine`, not `Graphics`, which wants to render things upside-down, so tethering core behaviors of the `Graphics` backbuffer to `NativeEngine` logic seems like it might be breaking separation of concerns; and, more importantly, I think it might disrupt certain scenarios.

Suppose someone was writing some sort of CAD or other editor-like thing using Babylon Native for 3D rendering and a native UI system for the interface. Babylon Native would render to a portion of the backbuffer and the UI would render to the rest, both in their own separate threads coordinated by the Graphics component. In this situation, I don't think it makes sense for the UI to have to worry about the backbuffer getting flipped; it should just be able to render normally without consideration for what `NativeEngine` may or may not be doing.

Even if my thinking is correct on this, I don't know of any currently-forecast use cases where this would be an issue, so it's not the end of the world if we break this in the short term. If we do that, though, and this does become important, then we'll probably end up having to walk this back eventually.

PR #744, opened 2021-05-13
Update babylon.js version to pick up slider 3d controls

chrisfromwork:
Update babylon.js version


chrisfromwork:
will pick up these changes in https://github.com/BabylonJS/BabylonNative/pull/743

PR #748, opened 2021-05-18
Initialize pose members for XRRigidTransform

### LINKED ISSUE #736: Undefined pose in WebXR polyfill on initialization of the XR experience
**For questions, please use the [forum](https://forum.babylonjs.com)**.

**Describe the bug**
WebXR polyfill in Babylon native can give undefined values on initialization for the camera pose.

A hack has been made on the main BJS repo : https://github.com/BabylonJS/Babylon.js/blob/02eba45a6b3eec0b0eaa56f0b29fde45e7c2e739/src/XR/webXRCamera.ts#L176

**To Reproduce**
1. Set the `hololens` var to `true` in `experience.js`
2. Add a breakpoint in `babylon.max.js` on the line corresponding to the above link
3. `pose.transform.orientation` should be an empty object prototype, thus `pose.transform.orientation.x` etc are undefined. Same for `pose.transform.position`

**Expected behavior**
`pose.transform.orientation` should be initialized with an identity Quaternion.
`pose.transform.position` should be initialized with a 0,0,0 Vector.

**Screenshots**

**Other**
 - Platform: [e.g. Win32, UWP, HoloLens, Android, iOS]

rgerd:
There's a [check](https://github.com/BabylonJS/Babylon.js/blob/02eba45a6b3eec0b0eaa56f0b29fde45e7c2e739/src/XR/webXRCamera.ts#L176) in Babylon.js to fight an issue where the transform on the pose returned from `WebXRSessionManager.getViewerPose` is defined but has position and orientation defined as empty objects, which doesn't conform to the WebXR spec.

fixes #736


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -853,6 +853,18 @@ namespace Babylon
...
+                    position.Set("x", 0.f);
+                    position.Set("y", 0.f);
+                    position.Set("z", 0.f);
+                    position.Set("w", 1.f);
+
+                    auto orientation = m_orientation.Value();
+                    orientation.Set("x", 0.f);
+                    orientation.Set("y", 0.f);
+                    orientation.Set("z", 0.f);
+                    orientation.Set("w", 1.f);
```
bghgary, line 867: nit: slightly more efficient to do this

```suggestion
                    auto position{Napi::Object::New(info.Env())};
                    position.Set("x", 0.f);
                    position.Set("y", 0.f);
                    position.Set("z", 0.f);
                    position.Set("w", 1.f);
                    m_position = Napi::Persistent(position);

                    auto orientation{Napi::Object::New(info.Env())};
                    orientation.Set("x", 0.f);
                    orientation.Set("y", 0.f);
                    orientation.Set("z", 0.f);
                    orientation.Set("w", 1.f);
                    m_orientation = Napi::Persistent(orientation);
```

bghgary (approved):

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -857,8 +857,19 @@ namespace Babylon
...
                 else
                 {
-                    m_position = Napi::Persistent(Napi::Object::New(info.Env()));
-                    m_orientation = Napi::Persistent(Napi::Object::New(info.Env()));
+                    auto position{Napi::Object::New(info.Env())};
+                    position.Set("x", 0.f);
+                    position.Set("y", 0.f);
+                    position.Set("z", 0.f);
+                    position.Set("w", 1.f);
+                    m_position = Napi::Persistent(position);
```
bghgary, line 865: Oops, I think I typed this wrong in the suggestion. There should be no `w` in the position.

rgerd: Actually the w is fine / intentional.
See here:
https://github.com/BabylonJS/BabylonNative/blob/0b5676fa5d7b74988fdbee781d81e007759c9b94/Plugins/NativeXr/Source/NativeXr.cpp#L864-L871

A w of 1.0 on a vector like this just means that it's a position and not a normal.

bghgary: Gotcha.

PR #750, opened 2021-05-18
Don't set folder property on asio library

### LINKED ISSUE #747: cmake doesn't generate build solution on windows 10
Cmake doesn't generate build solution after installing the new master branch on windows 10

powershell_XV5op7YRbv

rgerd:
asio is a header-only library included as an INTERFACE target, and CMake doesn't allow the FOLDER property to get set on INTERFACE's, so we'll just remove the line.

fixes #747

### DIFF
--- a/Dependencies/v8inspector/CMakeLists.txt
+++ b/Dependencies/v8inspector/CMakeLists.txt
@@ -1,7 +1,6 @@
 # ------------------------------- asio -------------------------------
 add_library(asio INTERFACE)
 target_include_directories(asio INTERFACE "Dependencies/asio/asio/include")
-set_property(TARGET asio PROPERTY FOLDER Dependencies)
 
 # ------------------------------ llhttp ------------------------------
 # Install node to use npm


chrisfromwork (approved):

bghgary:
How did this work before?

rgerd:
> How did this work before?

I'm trying to figure that out with the issue OP, but it seems like they might have a weird version of CMake? Or there might be some settings that enable more pedantic error messages?

bghgary:
I see that.

You say above:
> Make doesn't allow the FOLDER property to get set on INTERFACE

It also sounds like this currently works correctly. From the issue you reference in OP.
>  I couldn't get a local repro

I guess I'm a bit confused why it is currently working before this fix.

rgerd:
It only "works" in that it doesn't throw an error when running CMake. I haven't been able to get asio to actually show up in the VS project with that line. So it technically never worked, but it never stopped CMake from running. Tbh I probably just forgot about the line after fiddling with it and just merged it in.

PR #752, opened 2021-05-19
Fix crash on empty ray

### LINKED ISSUE #751: NativeXR's XRRay class causes crashes if appropriate options aren't provided
XRRay, if not provided with an origin and direction on creation, will trigger a crash during the call to GetNativeRay due to the origin and direction not being properly populated. An example of this happens in BabylonJS's WebXRHitTest class, if the offsetRay or transientOffsetRay is not provided in the options:
```
const offsetRay = new XRRay(this.options.offsetRay || {});
...
const offsetRay = new XRRay(this.options.transientOffsetRay || {});
```

rickfromwork:
XRRay causes a crash if the ray doesn't get passed in with an origin or direction. This change overhauls the constructor to match the webXR XRRay constructor, as seen here, which also initializes the ray data, avoiding crashes later: https://immersive-web.github.io/hit-test/#xrray-interface

Fixes #751


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1234,16 +1234,23 @@ namespace Babylon
             xr::Ray GetNativeRay()
             {
                 xr::Ray nativeRay{{0, 0, 0}, {0, 0, -1}};
+
                 auto originObject = m_origin.Value();
-                nativeRay.Origin.X = originObject.Get("x").ToNumber().FloatValue();
-                nativeRay.Origin.Y = originObject.Get("y").ToNumber().FloatValue();
-                nativeRay.Origin.Z = originObject.Get("z").ToNumber().FloatValue();
+                if (originObject.Has("x"))
```
rickfromwork, line 1239: I'm assuming it's alright to check just x, but I can individually check each value or check all 3 beforehand.

bghgary: https://github.com/BabylonJS/Babylon.js/blob/master/src/LibDeclarations/webxr.d.ts#L280-L285

```typescript
interface DOMPointInit {
    w?: number;
    x?: number;
    y?: number;
    z?: number;
}
```

It looks like from the type declarations that any of the x, y, z, w can be not set. It probably makes more sense to check each one individually. Although maybe it's better to do this in the constructor?

bghgary (approved):

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1202,48 +1202,75 @@ namespace Babylon
             XRRay(const Napi::CallbackInfo& info)
                 : Napi::ObjectWrap<XRRay>{info}
             {
-                bool originSet = false;
-                bool directionSet = false;
-                bool matrixSet = false;
-                if (info[0].IsObject())
+                auto argLength = info.Length();
```
bghgary, line 1205: nit:
```suggestion
                auto argLength{info.Length()};
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1202,48 +1202,75 @@ namespace Babylon
...
                     auto argumentObject = info[0].As<Napi::Object>();
-                    auto originValue = argumentObject.Get("origin");
-                    if (originValue.IsObject())
+
+                    XRRigidTransform* transform = XRRigidTransform::Unwrap(argumentObject);
+                    if (transform != nullptr)
                     {
-                        originSet = true;
-                        m_origin = Napi::Persistent(originValue.As<Napi::Object>());
+                        // The value passed in to the contructor is an XRRigidTransform
```
bghgary, line 1221: nit:
```suggestion
                        // The value passed in to the constructor is an XRRigidTransform
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1202,48 +1202,75 @@ namespace Babylon
...
+                    if (argumentObject.Has("y"))
                     {
-                        matrixSet = true;
-                        m_matrix = Napi::Persistent(matrixValue.As<Napi::Float32Array>());
+                        tempVals.Origin.Y = argumentObject.Get("y").ToNumber().FloatValue();
+                    }
+                    if (argumentObject.Has("z"))
+                    {
+                        tempVals.Origin.Z = argumentObject.Get("z").ToNumber().FloatValue();
                     }
```
bghgary, line 1242: should this be in the else of the transform if statement?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1202,48 +1202,75 @@ namespace Babylon
...
-                if (info[0].IsObject())
+                auto argLength = info.Length();
+                xr::Ray tempVals{};
+                tempVals.Direction.Z = -1.0;
+
+                m_origin = Napi::Persistent(Napi::Object::New(info.Env()));
+                m_direction = Napi::Persistent(Napi::Object::New(info.Env()));
+                m_matrix = Napi::Persistent(Napi::Float32Array::New(info.Env(), MATRIX_SIZE));
+
+                // The constructor is either sent a BABYLON.Vector3, {}, an XRRigidTransform, or {x,y,z,w},{x,y,z,w}
```
bghgary, line 1213: Is it expected that the `w` component is ignored completely?

Plugins/NativeXr/Source/NativeXr.cpp
bghgary: Some places below as well.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1202,48 +1202,75 @@ namespace Babylon
...
-                bool directionSet = false;
-                bool matrixSet = false;
-                if (info[0].IsObject())
+                auto argLength = info.Length();
+                xr::Ray tempVals{};
+                tempVals.Direction.Z = -1.0;
+
+                m_origin = Napi::Persistent(Napi::Object::New(info.Env()));
+                m_direction = Napi::Persistent(Napi::Object::New(info.Env()));
+                m_matrix = Napi::Persistent(Napi::Float32Array::New(info.Env(), MATRIX_SIZE));
```
bghgary, line 1211: `m_matrix` isn't being set anymore. Is that expected?

rickfromwork: We didn't have any code in BJS that passed in a matrix on creation, and it doesn't look like there's any code that tries to access the matrix either. The docs have it initialized on query, so I think since nothing queries it it should be fine.

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: What is the general guideline for using "{}" instead of "=" in Babylon code?

bghgary: This is not guidelines for Babylon code but C++ in general: 
https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives

Plugins/NativeXr/Source/NativeXr.cpp
bghgary: It's okay to not implement, but it might be good to assert or throw an exception when it is passed in, just in case the Babylon.js side changes. Also, if it really isn't needed, then `m_matrix` isn't necessary.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1201,49 +1201,55 @@ namespace Babylon
...
+                        {
+                            tempVals.Origin.Y = argumentObject.Get("y").ToNumber().FloatValue();
+                        }
+                        if (argumentObject.Has("z"))
+                        {
+                            tempVals.Origin.Z = argumentObject.Get("z").ToNumber().FloatValue();
+                        }
+                        if (argumentObject.Has("w") && argumentObject.Get("w").ToNumber().FloatValue() != 1.0)
+                        {
+                            Napi::Error::New(info.Env(), "TypeError: w-axis provided for XRRay's Origin is not 1").ThrowAsJavaScriptException();
```
bghgary, line 1246: nit: alternative way
```suggestion
                            throw Napi::Error::New(info.Env(), "TypeError: w-axis provided for XRRay's Origin is not 1");
```

rickfromwork: Sounds good, I'll use that instead, since it's cleaner.

Plugins/NativeXr/Source/NativeXr.cpp
rickfromwork: I've updated the code to throw an error if the w component isn't what is expected, as per xrray docs

PR #758, opened 2021-05-21
Throw C++ exceptions if Java exceptions occur in URL+URI classes

### LINKED ISSUE #594: UrlRequest on Android throws an exception an crashes if the URI doesn't have a protocol
While testing out loading different file types on Android I noticed an exception was being thrown in the implementation of URLRequest::SendAsync when the requested file didn't include a protocol (such as just a relative path). The exception is thrown in [UrlRequest.cpp:68](https://github.com/BabylonJS/BabylonNative/blob/acf1015a8c579269bdd9dc942d924b3a1caf006e/Dependencies/UrlLib/Source/Android/UrlRequest.cpp#L68):
```c++
std::string scheme{uri.getScheme()};
```

The issue is that uri.getScheme() may return null if the URI has no scheme. That null value is then passed to the [JavaWrappers.cpp String::operator](https://github.com/BabylonJS/BabylonNative/blob/b9883c71e307c00250b43a2e649d342fed3864c4/Dependencies/AndroidExtensions/Source/JavaWrappers.cpp#L111) where it passes it to the JNI environment resulting in a SIGABRT.

From the route I was testing the exception being thrown was also not catchable and crashed the app. The code I had calling into BabylonNative was this:
```ts
try {
  rootNode = (await SceneLoader.ImportMeshAsync("", "noprotocolurl.glb")).meshes[0];
} catch(e) {
    console.error("Any exceptions should be caught here from ImportMeshAsync... but this one escapes");
}
```

darraghjburke:
Changes the following:
- Adds a check to URLRequest in case the URL scheme is null, ensuring we don't attempt to compare a null to a string.
- Makes the JavaWrapper URL and URI classes throw if Java exceptions are detected in any of their methods.
- If there's an error in SendAsync, throws it to JS using Napi::Error.

Closes #594. If you attempt to load "noprotocolurl.glb" now, you will be able to catch the exception using a try/catch block in your Javascript. Currently, you will just get a status code 0 and an unspecific error message (Error status: 0 undefined - Unable to load noprotocol.gltf). This should be addressed in a future PR that can tackle HTTPRequest error handling in general.


Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -108,6 +108,9 @@ namespace java::lang
 
     String::operator std::string() const
     {
+        if (m_string == nullptr) {
+            throw std::runtime_error("Tried to implicitly convert null Java String to C++ String");
```
bghgary, line 112: nit:
```suggestion
            throw std::runtime_error{"Tried to implicitly convert null Java String to C++ String"};
```

bghgary (approved):

bghgary:
@CoPrez @ryantrem Can you do a quick pass just to double check?

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -108,6 +108,9 @@ namespace java::lang
 
     String::operator std::string() const
     {
+        if (m_string == nullptr) {
+            throw std::runtime_error{"Tried to implicitly convert null Java String to C++ String"};
+        }
```
ryantrem, line 113: Since we don't provide a way to check if the underlying java string is null, would it make sense to provide a conversion operator to `std::optional<std::string>` instead of or in addition to the operator that tries to go straight to `std::string`?

Dependencies/UrlLib/Source/Android/UrlRequest.cpp
```diff
@@ -65,8 +65,8 @@ namespace UrlLib
                 try
                 {
                     Uri uri{Uri::Parse(m_url.data())};
-                    std::string scheme{uri.getScheme()};
-                    if (scheme == "app")
+                    String scheme{uri.getScheme()};
+                    if (scheme != nullptr && (std::string)scheme == "app")
```
ryantrem, line 69: If we did what I mentioned above, this could could look like:
```c++
std::optional<std::string> scheme{uri.getScheme()};
if (scheme == "app")
```
Maybe this is better? @bghgary thoughts?

ryantrem (approved):

darraghjburke:
@bghgary @ryantrem Added a comment to the original solution explaining what to do if the Java string might be null. The PR should be done now

bghgary:
Looks good to me. Feel free to merge when the CI passes.

PR #761, opened 2021-05-24
Entering and exiting MR seems to leak about 20mb per iteration

jp833:
**Issue**? There is a ~9MB growth every time user exits XR.
**How**? A retain on the texture was causing the texture to remain in memory. A few more objects were not released at the end of XR session. The same issue can be repro'ed by adding another texture object i.e. the reference is not held by any other object in the session.
**Tests**? Verified in BabylonNative/BabylonReactNative Playground app.

Before:
image

After:
image


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -703,7 +707,6 @@ void UpdateXRView(MTKView* activeXRView) {
                     MTLTextureDescriptor *textureDescriptor = [MTLTextureDescriptor texture2DDescriptorWithPixelFormat:MTLPixelFormatBGRA8Unorm width:width height:height mipmapped:NO];
                     textureDescriptor.usage = MTLTextureUsageRenderTarget | MTLTextureUsageShaderRead;
                     id<MTLTexture> texture = [metalDevice newTextureWithDescriptor:textureDescriptor];
-                    [texture retain];
```
Alex-MSFT, line 706: Is there no danger that this could be purged from underneath us while it's still actively being used from the void* pointer below?

Could we alternatively just add cleaning up the ColorTexturePointer to the destructor?

jp833: image

**Old approach** : That second line is the extra retain we were doing.
If we weren't keeping a reference at the class level [i.e. std::vector<Frame::View> ActiveFrameViews{ {} };], it would be purged.

ryantrem: In the screenshot above, it looks to me like the ref count goes to 1 when the object is created (`newTextureWithDescriptor`), goes to 2 when we explicitly call `retain`, and then goes back to 1 when the `id<MTLTexture>` goes out of scope. If you remove the `retain`, then isn't the ref count going to 0 is soon as `id<MTLTexture>` goes out of scope?

jp833: Realized that the above screenshot was for another object (shader library).

Alex-MSFT (approved):

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -564,9 +564,9 @@ bool TryInitialize() {
 
             [configuration release];
 
-            id<MTLLibrary> lib = CompileShader(metalDevice, shaderSource);
-            id<MTLFunction> vertexFunction = [lib newFunctionWithName:@"vertexShader"];
-            id<MTLFunction> fragmentFunction = [lib newFunctionWithName:@"fragmentShader"];
+            lib = CompileShader(metalDevice, shaderSource);
+            vertexFunction = [lib newFunctionWithName:@"vertexShader"];
+            fragmentFunction = [lib newFunctionWithName:@"fragmentShader"];
```
ryantrem, line 569: Discussed offline, but keep these as local variables and just `release` them after they are assigned to the `pipelineStateDescriptor` (e.g. down on line 577).

ryantrem (approved):

Dependencies/xr/Source/ARKit/XR.mm
jp833: We found that the object here created with new (newTextureWithDescriptor) is retained and not auto released. The explicit retain is not required and needs to be explicitly released (done in destructor).

PR #763, opened 2021-05-26
Render XR using texture arrays

### LINKED ISSUE #705: Use texture arrays in OpenXR
In order to maximize performance on HoloLens, we need to use texture arrays in our color and depth swapchains as opposed to using one swapchain per eye. This will allow us to avoid an expensive texture copy step that happens inside of the Windows Mixed Reality OpenXR Runtime when the textures are composited. 

openxr-errors

https://github.com/BabylonJS/BabylonNative/blob/2db465f5569d8eb833245a41030c5219cfa3fe59/Dependencies/xr/Source/OpenXR/XR.cpp#L527-L550

rgerd:
The initial implementation of stereoscopic XR rendering used per-eye swapchains. Because the WMR OpenXR runtime expects to receive texture arrays, we were causing the runtime to use the "TexMapping" compositor type, meaning that the color and depth buffers of every frame (5,391,360 pixels on HoloLens) were getting copied into a texture array as a post-processing step before getting sent to the system compositor. 

**The following app timings come from release builds of the default playground experience on a HoloLens 2 (just rendering a cube) with hand tracking and pointer selection disabled.**

The OpenXR frame timings were showing around 1-3ms post CPU and 6-8ms post GPU time, which represents the amount of extra time added per-frame almost solely due to these texture copies, giving us a maximum framerate of about 50 FPS:
<image src="https://user-images.githubusercontent.com/4724014/119605396-9612cc80-bda5-11eb-894c-0d6ce76745ab.jpg" width="512"/>

Rendering into texture arrays shows a significant speed-up on HoloLens, with the post-app GPU time almost completely eliminated, bringing us up to the target framerate of 60 FPS (now limited by xrWaitFrame):
<image src="https://user-images.githubusercontent.com/4724014/119605394-94490900-bda5-11eb-8e96-b059bb3c8ed4.jpg" width="512"/>

We're left a nice budget for actual app time, this is a debug build with two animated cubes, for example:
<image src="https://user-images.githubusercontent.com/4724014/119581395-51703c80-bd77-11eb-97e3-26dc30470019.jpg" width="512"/>

Most of the main logic is left intact. The changes here involve the use of per-texture view configurations over per-eye views, since there's no longer a 1-1 mapping between textures and eyes.

fixes #705


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -411,9 +424,9 @@ namespace Babylon
                 GraphicsImpl& GraphicsImpl;
                 Napi::FunctionReference CreateRenderTexture{};
                 Napi::FunctionReference DestroyRenderTexture{};
-                std::map<void*, FrameBuffer*> TextureToFrameBufferMap{};
-                std::map<FrameBuffer*, Napi::ObjectReference> FrameBufferToJsTextureMap{};
-                std::vector<void*> ActiveTextures{};
+                std::vector<ViewConfiguration*> ActiveViewConfigurations{};
+                std::unordered_map<ViewConfiguration*, uint32_t> ViewConfigurationStartViewIdx{};
+                std::unordered_map<void*, ViewConfiguration> TextureToViewConfigurationMap{};
```
chrisfromwork, line 429: >                 [](http://example.com/codeflow?start=93&length=16)

nit: remove trailing spaces

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -609,68 +623,99 @@ namespace Babylon
...
+                    m_sessionState->ViewConfigurationStartViewIdx[&viewConfig] = viewIdx;
+
+                    viewConfig.ColorTexturePointer = view.ColorTexturePointer;
+                    viewConfig.DepthTexturePointer = view.DepthTexturePointer;
+                    viewConfig.ViewTextureSize = view.ColorTextureSize;
+
                     // If a texture width or height is 0, bgfx will assert (can't create 0 sized texture). Asserting here instead of deeper in bgfx rendering.
                     assert(view.ColorTextureSize.Width != 0);
                     assert(view.ColorTextureSize.Height != 0);
+                    assert(view.ColorTextureSize.Depth != 0);
```
chrisfromwork, line 648: >Depth [](http://example.com/codeflow?start=49&length=5)

nit: update the above comment or call out why depth can't be zero either?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -609,68 +623,99 @@ namespace Babylon
...
-                        m_sessionState->TextureToFrameBufferMap[view.ColorTexturePointer] = &frameBuffer;
-
-                        auto jsWidth{Napi::Value::From(m_env, view.ColorTextureSize.Width)};
-                        auto jsHeight{Napi::Value::From(m_env, view.ColorTextureSize.Height)};
-                        auto jsFrameBuffer{Napi::External<FrameBuffer>::New(m_env, &frameBuffer)};
-                        m_sessionState->FrameBufferToJsTextureMap[&frameBuffer] = Napi::Persistent(m_sessionState->CreateRenderTexture.Call({jsWidth, jsHeight, jsFrameBuffer}).As<Napi::Object>());
+                    auto depthTexture = bgfx::createTexture2D(textureWidth, textureHeight, false, textureLayers, depthTextureFormat, BGFX_TEXTURE_RT);
+                    m_sessionState->GraphicsImpl.AddTexture(depthTexture, textureWidth, textureHeight, false, textureLayers, depthTextureFormat);
+
+                    arcana::make_task(m_sessionState->GraphicsImpl.AfterRenderScheduler(), arcana::cancellation::none(), [colorTexture, depthTexture, &viewConfig]() {
```
chrisfromwork, line 668: >viewConfig [](http://example.com/codeflow?start=151&length=10)

i might be over thinking this, but if view config is a reference declared here in this function scope, is there any way to guarantee that it still exists when this task executes? should we just be feeding in a copy of the reference compared to a reference to the reference?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -609,68 +623,99 @@ namespace Babylon
...
+                        bgfx::overrideInternal(colorTexture, reinterpret_cast<uintptr_t>(viewConfig.ColorTexturePointer));
+                        bgfx::overrideInternal(depthTexture, reinterpret_cast<uintptr_t>(viewConfig.DepthTexturePointer));
+                    }).then(m_runtimeScheduler, m_sessionState->CancellationSource, [this, thisRef{shared_from_this()}, colorTexture, depthTexture, &viewConfig]() {
```
chrisfromwork, line 671: >this, thisRef{shared_from_this()} [](http://example.com/codeflow?start=85&length=33)

is there a reason we pass both a this ref and shared_from_this ref? Are we just trying to guarantee that this continues to exist? Would it be better if this was actually a weak reference that we then attempted to resolve when executing the llambda?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -609,68 +623,99 @@ namespace Babylon
...
+
+                            viewConfig.FrameBuffers[eyeIdx] = &frameBuffer;
+
+                            auto jsWidth{Napi::Value::From(m_env, viewConfig.ViewTextureSize.Width)};
+                            auto jsHeight{Napi::Value::From(m_env, viewConfig.ViewTextureSize.Height)};
+                            auto jsFrameBuffer{Napi::External<FrameBuffer>::New(m_env, &frameBuffer)};
+                            viewConfig.JsTextures[&frameBuffer] = Napi::Persistent(m_sessionState->CreateRenderTexture.Call({jsWidth, jsHeight, jsFrameBuffer}).As<Napi::Object>());
+
+                            viewConfig.Initialized = true;
+                        }
```
chrisfromwork, line 698: it seems like maybe this should go outside the foor loop?

chrisfromwork (approved):
:shipit:

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -1178,7 +1176,53 @@ namespace xr
...
+            const bool depthSupported,
+            const bool isPrimaryObserver)
+        {
+            const auto& colorSwapchain = renderResource.ColorSwapchain;
+            const auto& depthSwapchain = renderResource.DepthSwapchain;
+            const auto colorSwapchainImageIndex = AquireAndWaitForSwapchainImage(colorSwapchain.Handle);
+            const auto depthSwapchainImageIndex = AquireAndWaitForSwapchainImage(depthSwapchain.Handle);
+
+            uint32_t viewIdx = 0;
+            for(auto viewIter = viewsStart; viewIter < viewsEnd; viewIter++, viewIdx++)
```
ryantrem, line 1192:
```suggestion
            for (auto viewIter = viewsStart; viewIter < viewsEnd; viewIter++, viewIdx++)
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -609,68 +623,99 @@ namespace Babylon
...
                     assert(view.ColorTextureSize.Width != 0);
                     assert(view.ColorTextureSize.Height != 0);
+                    assert(view.ColorTextureSize.Depth != 0);
                     assert(view.ColorTextureSize.Width == view.DepthTextureSize.Width);
                     assert(view.ColorTextureSize.Height == view.DepthTextureSize.Height);
+                    assert(view.ColorTextureSize.Depth == view.DepthTextureSize.Depth);
+
+                    const uint16_t textureWidth = static_cast<uint16_t>(view.ColorTextureSize.Width);
+                    const uint16_t textureHeight = static_cast<uint16_t>(view.ColorTextureSize.Height);
+                    const uint16_t textureLayers = static_cast<uint16_t>(view.ColorTextureSize.Depth);
```
ryantrem, line 655: `auto` to be consistent with similar use cases (and since the type is obvious from the case)?
```suggestion
                    const auto textureWidth = static_cast<uint16_t>(view.ColorTextureSize.Width);
                    const auto textureHeight = static_cast<uint16_t>(view.ColorTextureSize.Height);
                    const auto textureLayers = static_cast<uint16_t>(view.ColorTextureSize.Depth);
```

ryantrem (commented):
I can't really comment on the OpenXR related changes (I don't have enough past experience), but from a quick look at the changes to the shared code it seems reasonable. Has this been tested on Android/iOS yet?

Plugins/NativeXr/Source/NativeXr.cpp
rgerd: Seems like @bghgary may have written those lines? I'm not really sure when it comes to this.

Plugins/NativeXr/Source/NativeXr.cpp
rgerd: Good catch, looks like bgfx accepts 0 layers

Plugins/NativeXr/Source/NativeXr.cpp
tymlipari: This is often a common pattern you see people do when trying to ensure a shared pointer lives through the duration of the lambda. You capture both `this` which is itself a non-owning pointer and `shared_from_this()` which extends the lifetime.

If you wanted to, you could omit the `this` and always reference through the `thisRef`, it just adds boilerplate code.

Plugins/NativeXr/Source/NativeXr.cpp
rgerd: The capture syntax here implies a reference to the original memory, in this case an entry in `m_sessionState->TextureToViewConfigurationMap`. As long as this lambda executes while the session is still alive (and the map hasn't been cleared), this will work.

So this will work as long as nothing super weird happens, but I could potentially see something happening where the session is ended before the textures get overridden, so maybe as an extra measure I could through the session state cancellation source into the first task as well. @syntheticmagus might have more thoughts.

PR #764, opened 2021-05-26
Use local reference images for validation tests

bghgary:


Apps/ValidationTests/CMakeLists.txt
```diff
@@ -125,18 +127,28 @@ if(APPLE)
     endif()
 endif()
 
-foreach(script ${SCRIPTS} ${BABYLONSCRIPTS})
-    get_filename_component(SCRIPT_NAME "${script}" NAME)
+foreach(SCRIPT ${SCRIPTS} ${BABYLON_SCRIPTS})
```
chrisfromwork, line 130: >SCRIPT [](http://example.com/codeflow?start=8&length=6)

nit: did this capitalization need to change?

chrisfromwork (approved):
it'd be great if we avoided cmake variable renames to reduce the scope of this change. But in general the change seems good.

bghgary:
I renamed it because I add a name that made the code inconsistent. There are some weird stuff going on in these cmake files that we should clean up.

PR #774, opened 2021-06-04
Fix iOS Deployment Target 13 (still working for iOS 12)

EricBeetsOfficial-Opuscope:


CedricGuillemet:
LGTM! Thank you

CedricGuillemet (approved):

CedricGuillemet:
@EricBeetsOfficial-Opuscope  Is it ok to merge ?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -160,14 +160,12 @@ - (void) UnlockPlanes {
...
-        return window.windowScene.interfaceOrientation;
-    }
-    else {
-        return [[UIApplication sharedApplication] statusBarOrientation];
-    }
+    UIApplication* sharedApplication = [UIApplication sharedApplication];
+#if (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_13_0)
+    return [[[[sharedApplication windows] firstObject] windowScene] interfaceOrientation];
+#else
+    return [sharedApplication statusBarOrientation];
```
Alex-MSFT, line 167: Should we keep the runtime @available check in the else case since we might still be running on an iOS version >= to 13?

ryantrem: Previously, we required at least the iOS 13 SDK to compile, but at runtime it would work on iOS 13 (using `window.windowScene.interfaceOrientation`) or iOS <13 (using `[[UIApplication sharedApplication] statusBarOrientation]`). Is that right?
Is this change trying to make it so you can compile the code against an older SDK that doesn't know anything about `sharedApplication.windows.firstObject` or `window.windowScene.interfaceOrientation`?

EricBeetsOfficial-Opuscope: I changed the @available for a conditional compilation because on iOS 13.0, it didn't compile: `'statusBarOrientation' was deprecated in iOS 13.0` error. (I would have liked keep `@available` which was more elegant :p)

EricBeetsOfficial-Opuscope: Im afraid that `if(!@available(..))` is not allowed. I suggest you this :

 ```
   UIApplication* sharedApplication = [UIApplication sharedApplication];
#if (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_13_0)
    return [[[[sharedApplication windows] firstObject] windowScene] interfaceOrientation];
#else
    if (@available(iOS 13.0, *)) {
        return [[[[sharedApplication windows] firstObject] windowScene] interfaceOrientation];
    }
    else {
        return [sharedApplication statusBarOrientation];
    }
#endif

```

ryantrem: Ok I think that is reasonable. I was trying to avoid duplicating that line, but I think my logic was actually wrong anyway. Thanks for making this contribution!

ryantrem (approved):

Dependencies/xr/Source/ARKit/XR.mm
EricBeetsOfficial-Opuscope: Yeap, I would also have liked to avoid duplicating these lines :|

PR #776, opened 2021-06-04
promote int streams to float for d3d and vulkan

### LINKED ISSUE #738: GLBs packed via GLTFPack do not display correctly on Windows
After packing a GLB with [GLTFPack](https://github.com/Alex-MSFT/meshoptimizer) with default settings the GLB does not display correctly in Windows, but does on Android, and Mac OSX, I have tested this with several GLBs and all display the same behavior.  The file appears to load successfully, and you can see vertex buffers being allocated and loaded, but no content is display in the playground.

Repro branch with an example Playground: https://github.com/Alex-MSFT/BabylonNative/tree/gltfpack_repro

CedricGuillemet:
- convert integer streams to float on the fly for d3d and vulkan

fixes #738


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -604,13 +630,53 @@ namespace Babylon
...
+            case bgfx::AttribType::Int16:
+                vertexBufferData->PromoteToFloats<int16_t>(numElements, byteStride);
+                break;
+            case bgfx::AttribType::Uint16:
+                vertexBufferData->PromoteToFloats<uint16_t>(numElements, byteStride);
+                break;
+            case bgfx::AttribType::Uint10: // is supported by any format ?
+            default:
+                throw std::runtime_error("Unable to promote vertex stream to a float array.");
+            }
```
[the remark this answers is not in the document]
bghgary: I think non-zero `byteOffset` will be broken also.

bghgary (changes requested):

Plugins/NativeEngine/Source/NativeEngine.cpp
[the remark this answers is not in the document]
CedricGuillemet: Indeed. Fixed both issues.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -604,12 +638,77 @@ namespace Babylon
...
+                case bgfx::AttribType::Int16:
+                    vertexBufferData->PromoteToFloats<int16_t, true>(numElements, byteStride, 32767.f);
+                    break;
+                case bgfx::AttribType::Uint16:
+                    vertexBufferData->PromoteToFloats<uint16_t, true>(numElements, byteStride, 65535.f);
+                    break;
+                case bgfx::AttribType::Uint10: // is supported by any format ?
+                default:
+                    throw std::runtime_error("Unable to promote vertex stream to a float array.");
+                }
```
bghgary, line 678: I don't think you need this switch. You can use templates with [std::numeric_limits<T>::max](https://en.cppreference.com/w/cpp/types/numeric_limits/max) to achieve the same effect and pass normalized as a regular parameter.

CedricGuillemet: Added numeric_limits. The normalized boolean template is useful for compile time check instead of runtime branching.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -261,6 +261,35 @@ namespace Babylon
...
+            const size_t count = m_bytes.size() / byteStride;
+            const size_t destinationSize = count * numElements * sizeof(float);
+            const float maxRange = static_cast<float>(std::numeric_limits<sourceType>::max());
+            if (destinationSize != m_bytes.size()) // ensure both vectors have different size
+            {
+                std::vector<uint8_t> bytes(destinationSize);
+                float* destination = (float*)bytes.data();
+                for (size_t i = 0; i < count; i++)
+                {
+                    sourceType* source = (sourceType*)(m_bytes.data() + byteStride * i);
```
bghgary, line 275: nit
```suggestion
                    sourceType* source = reinterpret_cast<sourceType*>(m_bytes.data() + byteStride * i);
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -328,6 +357,12 @@ namespace Babylon
             DoForHandleTypes(nonDynamic, dynamic);
         }
 
+        // size in bytes
+        size_t Size() const 
+        { 
+            return m_bytes.size();
+        }
```
bghgary, line 364: Where is this used?

bghgary (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
CedricGuillemet: Fixed

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -261,6 +261,35 @@ namespace Babylon
...
 
+        template<typename sourceType, bool normalized> void PromoteToFloats(uint32_t numElements, uint32_t byteStride)
+        {
+            const size_t count = m_bytes.size() / byteStride;
+            const size_t destinationSize = count * numElements * sizeof(float);
+            const float maxRange = static_cast<float>(std::numeric_limits<sourceType>::max());
+            if (destinationSize != m_bytes.size()) // ensure both vectors have different size
+            {
+                std::vector<uint8_t> bytes(destinationSize);
+                float* destination = (float*)bytes.data();
```
bghgary, line 272: nit: missed another one
```suggestion
                float* destination = reinterpret_cast<float*>(bytes.data());
```

PR #780, opened 2021-06-07
Generate Reference Images

CedricGuillemet:
Changed validationtests script to generate images.
Added documentation on how to add a new scene.


Apps/ValidationTests/Scripts/validation_native.js
```diff
@@ -6,7 +6,7 @@ var justOnce;
 var saveResult = true;
 var testWidth = 600;
 var testHeight = 400;
-
+var generateReferences = false;
```
bghgary, line 9: I guess this is deviating even more from the Babylon.js version now? It would be good to figure out a way such that more code is shared between Babylon.js and this code or at least make it easy to update.

Apps/ValidationTests/Scripts/validation_native.js
```diff
@@ -51,15 +51,20 @@ function compare(test, renderData, referenceImage, threshold, errorRatio) {
     return error;
 }
 
-function evaluate(test, resultCanvas, result, referenceImage, index, waitRing, done) {
+function saveRenderedResult(test, renderData) {
+    TestUtils.writePNG(renderData, testWidth, testHeight, TestUtils.getOutputDirectory() + "/Results/" + test.referenceImage);
```
bghgary, line 55: I noticed this function writes out uncompressed PNG files. Are we okay with that? Especially since they are merged in the repo.

CedricGuillemet: Not ideal, indeed. A potential improvement is adding a more advanced png writer in validationtests app.

Apps/ValidationTests/Scripts/validation_native.js
CedricGuillemet: Yes, the code is deviating even more. Same with the datas. For this rather small script, I think trying to make it more shareable with bjs will make it more complex and more difficult to update. The script in bjs repo is not part of any npm, it does things differently (using canvas Vs using TestUtils). I don't see how benefical the time spent to make it more generic will be. I'm more interested in making the datas completely shareable between js and native. Test script does not change as much as the datas. And for now, I know PGs are updated on the JS side and they are not on the native side.

bghgary: I'm good with that. It would be good to clean up this file. I did this a while back and didn't merge it because we wanted to share this code with the JS side.

Apps/ValidationTests/Scripts/validation_native.js
bghgary: Might be something worth doing as to not bloat the repo with big pngs.

bghgary (approved):

Apps/ValidationTests/Scripts/validation_native.js
CedricGuillemet: I take note.

PR #786, opened 2021-06-08
More shader error messages that were missed from the prior change.

### LINKED ISSUE #733: Shader compilation exceptions not holding messages
**Describe the bug**
The following exceptions (for example):
https://github.com/BabylonJS/BabylonNative/blob/432f0db8e0b0ddd98fb9c7744572f87c9c16528b/Plugins/NativeEngine/Source/ShaderCompilerD3D.cpp#L23-L26
https://github.com/BabylonJS/BabylonNative/blob/432f0db8e0b0ddd98fb9c7744572f87c9c16528b/Plugins/NativeEngine/Source/ShaderCompilerD3D.cpp#L93-L96

Don't hold their error messages (the "what" field) when being rethrown here:
https://github.com/BabylonJS/BabylonNative/blob/432f0db8e0b0ddd98fb9c7744572f87c9c16528b/Plugins/NativeEngine/Source/NativeEngine.cpp#L637-L644

Seems like this is just a case of memory getting freed, so should (hopefully) be a pretty straightforward issue to fix.

**To Reproduce**
I triggered the linker exception by doing something like:
`shader.vertex.fx`
```glsl
uniform float _Radius_Top_Left_;
```
`shader.fragment.fx`
```glsl
mediump uniform float _Radius_Top_Left_;
```

And I triggered the parser exception by trying to compile a shader missing a paren.

**Expected behavior**
The exception being thrown should hold the error message and show up in the javascript console. Right now the error message is blank:
empty-error-bjs-shader

**Screenshots**
The "what" of the error here is "":
shader-error-what-none
But the info sink holds useful info:
shader-compiler-info-sink
Same here with the parser:
parse-error-info

**Other**
 - Platform: Win32

syntheticmagus:
Fixes #733. This is a very strange mechanism for associating bugs.


bghgary (approved):

syntheticmagus:
Fixes #733.

PR #795, opened 2021-06-10
Support WebXR fillPoses and fillJointRadii

### LINKED ISSUE #773: Support WebXR fillPoses and fillJointRadii
The WebXR hand tracking API explainer recommends using `fillPoses` and `fillJointRadii` as a way of reducing the number of javascript objects created per-frame while updating tracked hand models:
https://github.com/immersive-web/webxr-hand-input/blob/main/explainer.md#efficiently-obtaining-hand-poses

We currently only support `getJointPose`:
https://github.com/BabylonJS/BabylonNative/blob/befafc77497fceb8ea870f3f65f6420ac77f9392/Plugins/NativeXr/Source/NativeXr.cpp#L1987

Which forces BabylonJS to make a call for each joint on each hand, "for 25 joints per hand and two hands, this is 150-250 objects created per frame." We're currently seeing performance issues around hand tracking, and this seems to be a leading cause.

See https://github.com/BabylonJS/Babylon.js/issues/10454 for more info.

rgerd:
Adding functions and a small refactor to support more efficient querying of hand tracking data. Anecdotally, between the native and JS changes I've seen about a 10-15 FPS improvement with the more efficient hand tracking.

See https://github.com/BabylonJS/Babylon.js/pull/10481 for more info.

fixes #773


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -151,23 +151,26 @@ namespace
 
             if (inputSource.JointsTrackedThisFrame)
             {
-                auto handJointCollection = Napi::Array::New(env, HAND_JOINT_NAMES.size());
+                auto handJointCollection = jsInputSource.Get("hand").As<Napi::Array>();
+                if (!jsInputSource.Has("hand"))
```
chrisfromwork, line 155: if jsInputSource doesn't have a hand, won't the above line throw?

chrisfromwork: i feel like a cast to napi array for undefined will cause a runtime error

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2209,6 +2214,44 @@ namespace Babylon
...
+                if (spaces.Length() != radii.ElementLength()) 
+                {
+                    throw std::runtime_error{"Number of spaces doesn't match number of radii."};
+                }
+
+                for (uint32_t spaceIdx = 0; spaceIdx < spaces.Length(); spaceIdx++)
+                {
+                    const auto& jointSpace = *spaces[spaceIdx].As<Napi::External<xr::System::Session::Frame::JointSpace>>().Data();
+                    const auto jointRadius = jointSpace.PoseRadius;
+                    *(radii.Data() + spaceIdx) = jointRadius;
```
tymlipari, line 2249: This looks a lot less sketchy if you use array syntax: `radii.Data()[spaceIdx] = jointRadius;`

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2209,6 +2214,44 @@ namespace Babylon
...
+                if (spaces.Length() != transforms.ElementLength() >> 4)
+                {
+                    throw std::runtime_error{"Number of spaces doesn't match number of transforms * 16."};
+                }
+
+                for (uint32_t spaceIdx = 0; spaceIdx < spaces.Length(); spaceIdx++)
+                {
+                    const auto& jointSpace = *spaces[spaceIdx].As<Napi::External<xr::System::Session::Frame::Space>>().Data();
+                    const auto transformMatrix = CreateTransformMatrix(jointSpace, false);
+                    memcpy(transforms.Data() + (spaceIdx << 4), transformMatrix.data(), 4 /*Float32*/ * 16);
```
tymlipari, line 2230: nit: I personally hate constants like this. Can we either express these offsets in terms of `sizeof` results (I think you're trying to do `sizeof(float)`) or at least a named constant that explains what's going on?

Plugins/NativeXr/Source/NativeXr.cpp
rgerd: It doesn't seem to, but I'll change the logic to make it less scary

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -151,23 +151,26 @@ namespace
...
+                    handJointCollection = Napi::Array::New(env, HAND_JOINT_NAMES.size());
+
+                    auto jointGetter = [handJointCollection](const Napi::CallbackInfo& info) -> Napi::Value {
+                        return handJointCollection.Get(info[0].As<Napi::String>());
+                    };
+
+                    handJointCollection.Set("get", Napi::Function::New(env, jointGetter, "get"));
+                    handJointCollection.Set("size", static_cast<int>(HAND_JOINT_NAMES.size()));
+
+                    jsInputSource.Set("hand", handJointCollection);
```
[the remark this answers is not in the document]
rgerd: Lack of experience keeps me from knowing whether capture by reference or value is better here, but I do want the `hand` property to outlive `handJointCollection`, so I think that dependence is fine, right?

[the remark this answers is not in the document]
bghgary: You definitely don't want to capture by reference (at least not as a C++ reference). If anything, it would be to capture a Napi::Reference. I don't think it's necessary (as you say) hand should outlive handJointCollection.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2221,6 +2225,44 @@ namespace Babylon
...
+                if (spaces.Length() != (transforms.ElementLength() >> 4))
+                {
+                    throw std::runtime_error{"Number of spaces doesn't match number of transforms * 16."};
+                }
+
+                for (uint32_t spaceIdx = 0; spaceIdx < spaces.Length(); spaceIdx++)
+                {
+                    const auto& jointSpace = *spaces[spaceIdx].As<Napi::External<xr::System::Session::Frame::Space>>().Data();
+                    const auto transformMatrix = CreateTransformMatrix(jointSpace, false);
+                    memcpy(transforms.Data() + (spaceIdx << 4), transformMatrix.data(), sizeof(float) * 16);
```
bghgary, line 2241: sorry for the late comment

nit: inconsistency between `<< 4` and `* 16`

and also std::memcpy would be better

bghgary (approved):

PR #797, opened 2021-06-10
Add the ability to declare a xr anchor to the BabylonNative xr session that was created outside of WebXRAnchorSystem calls

chrisfromwork:
For azure spatial anchors support, detected anchors generate their own native anchors (For android, this is a ArAnchor*). We need a way to share these native anchors with the babylonnative xr session/xr frame so that the webxranchorsystem creates a webxranchor that corresponds to the detected azure spatial anchor. This change allows native modules/non-babylonnative code to declare a native anchor that they have created and have it registered in the WebXRAnchorSystem.

We will have helper headers in the BabylonReactNative repo that demonstrate how to access this functionality off of the NativeXr component on the global navigator. https://github.com/BabylonJS/BabylonReactNative/pull/244


Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -657,6 +657,27 @@ namespace xr
             return {pose, reinterpret_cast<NativeAnchorPtr>(arAnchor)};
         }
 
+        Anchor DeclareAnchor(NativeAnchorPtr anchor)
+        {
+            ArAnchor* arAnchor{reinterpret_cast<ArAnchor*>(anchor)};
+            const auto iterator = std::find(arCoreAnchors.begin(), arCoreAnchors.end(), arAnchor);
```
ryantrem, line 663: Isn't the code in NativeXR.cpp already guarding against this? How can this happen?

ryantrem (approved):

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1452,9 +1452,9 @@ namespace Babylon
                 env.Global().Set(JS_CLASS_NAME, func);
             }
 
-            static Napi::Object New(const Napi::CallbackInfo& info)
+            static Napi::Object New(const Napi::Env& env)
```
syntheticmagus, line 1455: nit: I'm not sure I ever fully understood the reason, but I seem to remember that the best practice for `Napi::Env` is to pass it by copy. @bghgary, is this still true (if it ever was), and if so, do you remember the reason?

syntheticmagus (approved):
As far as I can tell, seems comfortably in line with what's already being done. There seems to be some growing "blurriness" between standard functionality and nonstandard native functionality that makes me jittery, but I think that's a subject for refactors, not this feature.

chrisfromwork:
> As far as I can tell, seems comfortably in line with what's already being done. There seems to be some growing "blurriness" between standard functionality and nonstandard native functionality that makes me jittery, but I think that's a subject for refactors, not this feature.

yeah, it should be easy to move folks to other apis based on the helper headers we have. Do you think we should open an issue to relook this api design for the native navigator? I'm not sure if we have anything tracking this currently.

Plugins/NativeXr/Source/NativeXr.cpp
bghgary: Env is intended to be a thin wrapper around the underlying C pointer. Passing by ref does nothing in this case. I don't think there is a big difference though.

chrisfromwork: would folks prefer if i change this here or should the reference suffice?

chrisfromwork: ill just go a head and change this for now

Dependencies/xr/Source/ARCore/XR.cpp
chrisfromwork: this is now removed

PR #798, opened 2021-06-11
Simplify memory management xr lib on iOS

### LINKED ISSUE #527: AppRuntime's Dispatch should wrap each work queue tick in an @autoreleasepool
With Objective C + Cocoa, it is standard for threads with a work queue to wrap each tick of the work queue in an `@autoreleasepool`. This is a core part of the way memory management works on platforms like iOS. In code, you can then do something like this:

```objc
[someObject retain];
[someObject autorelease];
```

That call to `autorelease` adds the object the closest scoped `@autoreleasepool` (it is stored as a thread local). When that `@autoreleasepool` "goes out of scope" (e.g. the stack is popped past the `@autoreleasepool` bounds), all objects registered with it via `autorelease` are released (e.g. it does a call `[someobject release]` automatically at that point).

Many object types in Objective C (including ARKit) internally use these mechanisms and rely on them. Other user code would expect the same. For example, if a consumer of Babylon Native calls `AppRuntime::Dispatch`, they would expect that they could use the `retain` `autorelease` pairing mentioned above (or use Cocoa objects that themselves do this internally), but this doesn't work right now.

I think what we need is for the platform specific parts of `AppRuntime` (e.g. `AppRuntimeApple`) to provide a `Dispatch` wrapper. On most platforms, this would just call the function we are trying to dispatch directly. On Apple, it should wrap it in an `@autoreleasepool`.

ryantrem:
We've hit memory management issues several times in the xr lib on iOS:
- Memory leaks from missing calls to `release`
- Memory leaks from missing @autoreleasepools
- Crashes from releasing objects too early

With this change, I'm simplifying the memory management in two main ways:
- Enable ARC (automatic reference counting) for the iOS compilation of the xr lib
- Wrap each "tick" of the JS work queue in a @autoreleasepool (this is analogous to what happens on the UI thread, and also to what happens on the JS thread in react native - this fixes #527)

With ARC enabled, I *mostly* just had to remove all the explicit calls to `retain`, `release`, and `dealloc`. The one complicated part is the bridge casting. When ownership is transferred to or from ARC, a bridge cast is needed. There are three types of bridge cast used in these changes:
**`__bridge`** - this is used to cast to/from `void*` and does not affect the ref count.
**`__bridge_retained`** - this is used to cast to `void*` when we need to keep the object alive (essentially increments the ref count).
**`__bridge_transfer`** - this is used to cast from `void*` when we that pointer no longer needs to keep the object alive (essentially does not increment the ref count when casting back to the Foundation object that ARC understands).

For the @autoreleasepool, I needed to add an extra hook in `AppRuntime` for this, and it is a no-op for most platforms. This is not used in Babylon React Native, but I did test this in the regular Babylon Native playground to make sure memory is cleaned up as expected.

### DIFF
--- a/Core/AppRuntime/Include/Babylon/AppRuntime.h
+++ b/Core/AppRuntime/Include/Babylon/AppRuntime.h
@@ -36,6 +36,7 @@ namespace Babylon
         void RunPlatformTier();
         void RunEnvironmentTier(const char* executablePath = ".");
         void Run(Napi::Env);
+        void Execute(std::function<void()> callback);
 
         static void DefaultUnhandledExceptionHandler(std::exception_ptr ptr);
 
--- a/Core/AppRuntime/Source/AppRuntime.cpp
+++ b/Core/AppRuntime/Source/AppRuntime.cpp
@@ -37,6 +37,8 @@ namespace Babylon
 
     void AppRuntime::Dispatch(std::function<void(Napi::Env)> func)
     {
-        m_workQueue->Append(std::move(func));
+        m_workQueue->Append([this, func{std::move(func)}] (Napi::Env env) {
+            Execute({[env, func{std::move(func)}] { func(env); }});
+        });
     }
 }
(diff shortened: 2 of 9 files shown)


bghgary (approved):
Seems okay to me, but I have no experience with ARC.

Issue #802, opened 2021-06-15
Integrate Mocha.js-based testing infrastructure


darraghjburke:
Parts of the Babylon Native codebase should be unit tested. We will use Mocha.js to test our polyfills from Javascript. We should build out the infrastructure to do unit testing, and then gradually add test for different parts of the codebase. The initial focus will be on testing XMLHTTPRequest.

Progress so far:

- [X] Design meaningful tests for XMLHTTPRequest
- [X] Create infrastructure to run unit tests
- [x] Add mechanism for setting an exit code from Mocha to enable automation 
- [ ] Implement clearTimeout() in window polyfill
- [ ] Fix console polyfill so we can use built-in Mocha test reporter
- [ ] Get unit tests building on non-Windows platforms (Linux and Android at a minimum)
- [x] Make unit tests run on each platform as part of the CI pipeline

(Note: removed references to Google Test and UrlLib. Testing is now using Mocha.js and focused on XMLHTTPRequest.)

darraghjburke:
Closed by #872

PR #804, opened 2021-06-16
Integrate GoogleTest framework to Babylon Native, build for Win32

darraghjburke:
Adds the Google Test framework into Babylon Native (see #802). Creates a file, UnitTests.cpp, which will hold future Unit Tests. Currently, Unit Tests will only be built for Win32. They can be manually invoked from `BabylonNative\build\win32\Apps\UnitTests\Release` with the command `UnitTests`.

Future work will involve:

- building for additional platforms (UWP and Unix probably next)
- writing good unit tests for at least UrlLib
- adding the unit tests to the CI pipeline
- rewriting WaitForCompletion in a cleaner way

Looking for feedback in particular on UnitTests\CMakeLists.txt. Also, does it make sense for UnitTests to be inside the `Apps` directory?


Apps/UnitTests/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+target_include_directories(UnitTests PRIVATE "Source" ".")
+
+target_link_libraries(UnitTests
+    gtest_main
+    UrlLib)
+
+
+add_test(NAME UnitTests COMMAND UnitTests)
+
+set_property(TARGET UnitTests PROPERTY FOLDER Apps)
```
bghgary, line 29: Probably good to add the googletest projects to a folder also, so they show up in the dependencies folder like other dependencies.

Apps/UnitTests/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  # Specify the commit you depend on and update it regularly.
+  URL https://github.com/google/googletest/archive/23ef29555ef4789f555f1ba8c51b4c52975f0907.zip
+)
+# For Windows: Prevent overriding the parent project's compiler/linker settings
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
```
bghgary, line 13: Where does the code end up using this method?

Apps/UnitTests/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+add_executable(UnitTests "UnitTests.cpp")
+
+target_compile_definitions(UnitTests PRIVATE UNICODE)
+target_compile_definitions(UnitTests PRIVATE _UNICODE)
+
+target_include_directories(UnitTests PRIVATE "Source" ".")
```
bghgary, line 20: Is this necessary?

Apps/UnitTests/UnitTests.cpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include "gtest/gtest.h"
+#include <UrlLib/UrlLib.h>
+#include <arcana/threading/task.h>
+
+namespace Babylon
+{
+
```
bghgary, line 8: nit: extra newline

Apps/UnitTests/UnitTests.cpp
```diff
@@ -0,0 +1,32 @@
...
+    int test(int argc, char** argv)
+    {
+        ::testing::InitGoogleTest(&argc, argv);
+        return RUN_ALL_TESTS();
+    }
```
bghgary, line 13: Who calls this function?

Apps/UnitTests/CMakeLists.txt
darraghjburke: It was not, just an artifact from copying over the ValidationTests CMakeLists :) Thanks!

Apps/UnitTests/CMakeLists.txt
darraghjburke: It seems to build fine on Windows without it. Here is the explanation for these lines: https://github.com/google/googletest/blob/master/googletest/README.md#visual-studio-dynamic-vs-static-runtimes

Doesn't seem to be necessary for us, not sure I understand why though.

Apps/UnitTests/UnitTests.cpp
darraghjburke: We don't need it since we are linked with gtest_main (which defines an entry point), nice catch! Removed

Apps/UnitTests/CMakeLists.txt
darraghjburke: Agreed, but I'm not sure how to put content from FetchContent into a subdirectory. I will keep looking into this

Apps/UnitTests/CMakeLists.txt
bghgary: I mean where on disk does the FetchContent unzip the file?

Apps/UnitTests/CMakeLists.txt
bghgary: This `set_property(TARGET UnitTests PROPERTY FOLDER Apps)` just puts the UnitTests project in a solution folder (or filter) in Visual Studio. I would expect you can do the same for the googletest projects.

darraghjburke: Yep, I was definitely way overthinking this  :man_facepalming: The GoogleTest projects now all go in Dependencies\GoogleTest

Apps/UnitTests/CMakeLists.txt
darraghjburke: Oh, it looks like the googletest source code is being placed in a `_deps` subdirectory, I'll see if I can get it to unzip to `Dependencies\GoogleTest` instead

bghgary: I would recommend putting the downloaded files in the build folder somewhere as to not pollute the source folders.

bghgary: If we do end up putting it in Dependencies, then we need to put this folder in `.gitginore`.

bghgary: To be clear, by default, this ends up in the <build folder>/_deps.

Apps/UnitTests/UnitTests.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <arcana/threading/task.h>
+
+namespace Babylon
+{
+    void WaitForCompletion(arcana::task<void, std::exception_ptr> task) {
+        bool done{false};
+        task.then(arcana::inline_scheduler, arcana::cancellation::none(), [&done](arcana::expected<void, std::exception_ptr> result) {
+            done = true;
+        });
+        while (!done);
```
bghgary, line 13: This is a temporary hack. Might be good to have a comment on this.

bghgary (approved):

Apps/UnitTests/UnitTests.cpp
darraghjburke: Done!

PR #805, opened 2021-06-17
Use the NativeEngine TextureData definition to avoid crashes unwrapping Napi::External<NativeEngine::TextureData> for jsi

chrisfromwork:
Defining a second TextureData definition for the Canvas seemed to resolve correctly in BabylonNative. But in BabylonReactNative the jsi wrapper unwrapping the external type threw exceptions. This change updates the Canvas polyfill to consume the TextureData struct defined by NativeEngine. It seems like this is what we should be doing to avoid getting into scenarios where the struct definition changes and we need to update multiple locations.

We also update loadTTF to only load a font once. Once a font is getting used if its updated on the js thread it can cause crashes for the graphics thread.

Note: this change requires having the Canvas polyfill reference internal headers for NativeEngine. It seems like this is already an expected practice. I'm not sure if polyfills are supposed to consume Core/Plugin definitions, but In this scenario we need to.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1116,12 +1116,6 @@ namespace Babylon
 
     void NativeEngine::CopyTexture(const Napi::CallbackInfo& info)
     {
-        if (!info[0].IsExternal() ||
-            !info[1].IsExternal())
-        {
-            return;
-        }
-
         // This code does not copy texture but overrides texture handle.
         // It will leak texture memory and should be fixed.
         // See task in this list : https://github.com/BabylonJS/BabylonNative/issues/616#issuecomment-831162396
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1116,6 +1116,12 @@ namespace Babylon
 
     void NativeEngine::CopyTexture(const Napi::CallbackInfo& info)
     {
+        if (!info[0].IsExternal() ||
+            !info[1].IsExternal())
+        {
+            return;
+        }
+
         // This code does not copy texture but overrides texture handle.
         // It will leak texture memory and should be fixed.
         // See task in this list : https://github.com/BabylonJS/BabylonNative/issues/616#issuecomment-831162396
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -5,10 +5,16 @@
 #include <functional>
 #include <sstream>
 #include <assert.h>
+#include <NativeEngine.h>
 
 namespace Babylon::Polyfills::Internal
 {
     static constexpr auto JS_CONSTRUCTOR_NAME = "NativeCanvas";
+
+    struct NativeCanvas::Impl
+    {
+        std::unique_ptr<Babylon::TextureData> m_textureData;
+    };
 
     void NativeCanvas::CreateInstance(Napi::Env env)
     {
@@ -32,6 +38,7 @@ namespace Babylon::Polyfills::Internal
     NativeCanvas::NativeCanvas(const Napi::CallbackInfo& info)
         : Napi::ObjectWrap<NativeCanvas>{info}
         , m_graphicsImpl{ Babylon::GraphicsImpl::GetFromJavaScript(info.Env()) }
+        , m_impl{std::make_unique<Impl>()}
     {
     }
 
@@ -99,31 +106,14 @@ namespace Babylon::Polyfills::Internal
         }
     }
 
-    struct TextureData final
-    {
-        ~TextureData()
-        {
-            if (bgfx::isValid(Handle))
-            {
-                bgfx::destroy(Handle);
-            }
-        }
-
-        bgfx::TextureHandle Handle{ bgfx::kInvalidHandle };
-        uint32_t Width{ 0 };
-        uint32_t Height{ 0 };
-        uint32_t Flags{ 0 };
-        uint8_t AnisotropicLevel{ 0 };
-    };
-
     Napi::Value NativeCanvas::GetCanvasTexture(const Napi::CallbackInfo& info)
     {
         assert(m_frameBufferHandle.idx != bgfx::kInvalidHandle);
-        auto data = new TextureData();
-        data->Handle = bgfx::getTexture(m_frameBufferHandle);
-        data->Width = m_width;
-        data->Height = m_height;
-        return Napi::External<TextureData>::New(info.Env(), data);
+        m_impl->m_textureData = std::make_unique<TextureData>();
+        m_impl->m_textureData->Handle = bgfx::getTexture(m_frameBufferHandle);
+        m_impl->m_textureData->Width = m_width;
+        m_impl->m_textureData->Height = m_height;
+        return Napi::External<TextureData>::New(info.Env(), m_impl->m_textureData.get());
     }
 
     void NativeCanvas::Dispose(const Napi::CallbackInfo& /*info*/)
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -19,7 +19,7 @@ namespace Babylon::Polyfills::Internal
             env,
             JS_CONSTRUCTOR_NAME,
             {
-                StaticMethod("loadTTF", &NativeCanvas::LoadTTF),
+                StaticMethod("loadTTFAsync", &NativeCanvas::LoadTTFAsync),
                 InstanceMethod("getContext", &NativeCanvas::GetContext),
                 InstanceAccessor("width", &NativeCanvas::GetWidth, &NativeCanvas::SetWidth),
                 InstanceAccessor("height", &NativeCanvas::GetHeight, &NativeCanvas::SetHeight),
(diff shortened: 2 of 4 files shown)


Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -5,11 +5,17 @@
...
 #include <assert.h>
+#include <NativeEngine.h>
 
 namespace Babylon::Polyfills::Internal
 {
     static constexpr auto JS_CONSTRUCTOR_NAME = "NativeCanvas";
 
+    struct NativeCanvas::Impl
+    {
+        std::unique_ptr<Babylon::TextureData> m_textureData;
```
ryantrem, line 16: For structs with public members, I think this should be `TextureData` instead of `m_textureData`.

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -5,11 +5,17 @@
 #include <functional>
 #include <sstream>
 #include <assert.h>
+#include <NativeEngine.h>
 
 namespace Babylon::Polyfills::Internal
 {
     static constexpr auto JS_CONSTRUCTOR_NAME = "NativeCanvas";
 
+    struct NativeCanvas::Impl
```
ryantrem, line 14:
```suggestion
    struct NativeCanvas::Impl final
```

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -99,31 +112,14 @@ namespace Babylon::Polyfills::Internal
...
-
     Napi::Value NativeCanvas::GetCanvasTexture(const Napi::CallbackInfo& info)
     {
         assert(m_frameBufferHandle.idx != bgfx::kInvalidHandle);
-        auto data = new TextureData();
-        data->Handle = bgfx::getTexture(m_frameBufferHandle);
-        data->Width = m_width;
-        data->Height = m_height;
-        return Napi::External<TextureData>::New(info.Env(), data);
+        m_impl->m_textureData = std::make_unique<TextureData>();
```
ryantrem, line 118: It looks like this gets deleted at the end of the function, but you are passing a pointer to the underlying data to the `Napi::External`. I think the default finalizer action for `Napi::External` is to delete the pointer, so it seems like it was correct the way it was, but maybe I'm missing something.

tymlipari: It looks like `Napi::External` only deletes if you provide your own finalizer. This was likely leaking before.

ryantrem: Hmm I wonder if we have leaks elsewhere…

chrisfromwork: id bet we definitely have leaks elsewhere

chrisfromwork: whatever was going on before, calling new just seems like a bad time for the TextureData when theres no subsequent delete call. To be fair though Cedric called out that this was leaking if i remember correctly.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: this should now be fixed.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: sounds good. i see this vary in different places. But am happy to start using capitalization throughout.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1116,6 +1116,12 @@ namespace Babylon
 
     void NativeEngine::CopyTexture(const Napi::CallbackInfo& info)
     {
+        if (!info[0].IsExternal() ||
+            !info[1].IsExternal())
+        {
+            return;
+        }
+
```
bghgary, line 1124: This is a bit weird. We just silently do nothing when the types are wrong? Is that the intended behavior?

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -5,11 +5,17 @@
...
 #include <assert.h>
+#include <NativeEngine.h>
 
 namespace Babylon::Polyfills::Internal
 {
     static constexpr auto JS_CONSTRUCTOR_NAME = "NativeCanvas";
 
+    struct NativeCanvas::Impl final
+    {
+        std::unique_ptr<Babylon::TextureData> TextureData;
```
bghgary, line 16: Do you need the unique_ptr?

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -5,11 +5,17 @@
 #include <functional>
 #include <sstream>
 #include <assert.h>
+#include <NativeEngine.h>
 
 namespace Babylon::Polyfills::Internal
 {
     static constexpr auto JS_CONSTRUCTOR_NAME = "NativeCanvas";
 
+    struct NativeCanvas::Impl final
```
bghgary, line 14: I don't understand this. Why do you need an Impl class?

Plugins/NativeEngine/Source/NativeEngine.cpp
chrisfromwork: more was a check because the external wasn't unwrapping correctly for jsi when it was a unique struct definition. I'll undo this check.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: I added an impl to have a data container so that i wouldn't have to declare the babylon/NativeEngine.h header in the the canvas header.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: no we don't, good call out.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: tking another look it seems like there are two canvas headers and we already declare bgfx types in the internal canvas header. I'll remove this container and will just declare the texturedata as a property of the canvas struct for now.

Plugins/NativeEngine/Source/NativeEngine.cpp
chrisfromwork: i'm going to keep this check but will switch to throwing instead of returning

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: this should now be fixed.

Plugins/NativeEngine/Source/NativeEngine.cpp
chrisfromwork: this should now be fixed.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: this should now be fixed. impl is removed and we just declare the property for the native canvas.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: chatted with ryan offline. seems like a finalizer is what we should do here.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: see changes, this struct no longer exists.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: based on other conversations, this datas lifecycle is going to be maintained by the external through a finalizer.

bghgary: that would make a lot more sense. create it for javascript and finalize with javascript.

chrisfromwork: this should now be fixed

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -41,11 +42,15 @@ namespace Babylon::Polyfills::Internal
 
     void NativeCanvas::LoadTTF(const Napi::CallbackInfo& info)
     {
-        std::string fontName = info[0].As<Napi::String>().Utf8Value();
+        auto& graphicsImpl{Babylon::GraphicsImpl::GetFromJavaScript(info.Env())};
+        const auto cancellationSource = std::shared_ptr<arcana::cancellation_source>();
```
ryantrem, line 46: Unused, remove?

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -41,11 +42,15 @@ namespace Babylon::Polyfills::Internal
...
         const auto buffer = info[1].As<Napi::ArrayBuffer>();
-
-        fontsInfos[fontName] = std::vector<uint8_t>(buffer.ByteLength());
-        memcpy(fontsInfos[fontName].data(), (uint8_t*)buffer.Data(), buffer.ByteLength());
+        std::vector<uint8_t> fontBuffer{};
+        fontBuffer.resize(buffer.ByteLength());
+        memcpy(fontBuffer.data(), (uint8_t*)buffer.Data(), buffer.ByteLength());
+        arcana::make_task(graphicsImpl.BeforeRenderScheduler(), arcana::cancellation::none(), [fontName{ info[0].As<Napi::String>().Utf8Value() }, fontData{ std::move(fontBuffer) }]() {
```
ryantrem, line 51: If this is going to be async, then this function should return a promise right?

chrisfromwork: if i understand correctly this function doesn't need to resolve before declaring text. Given returning a promise would require updating babylon.js, i'm inclined to say that returning a promise here is beyond the scope of this review and should be addressed when we finalize the public api contract for loading fonts (fontface, etc.)

chrisfromwork: actually loadTTF isn't defined anywhere public. let me return a promise here.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: this should now be fixed.

Polyfills/Canvas/Source/Canvas.cpp
chrisfromwork: this should now be fixed

ryantrem (approved):

bghgary (approved):

PR #812, opened 2021-06-23
More VT tests + fixed loadRawTexture

CedricGuillemet:
12 more tests (4 GLTF tests)


Apps/ValidationTests/Scripts/config.json
```diff
@@ -1,6 +1,73 @@
...
+            "referenceImage": "KernelBlur.png"
+        },
+        {
+            "title": "Instanced Bones",
+            "playgroundId": "#2YLJ1L#2",
+            "referenceImage": "instancedBones.png"
+        },
+        {
+            "title": "GLTF Mesh Primitive Attribute",
+            "playgroundId": "#DS8AA7#27",
```
bghgary, line 48: This is not quite right. We are still using the old playground for glTF. I would prefer that we stay consistent. Also, how are these sorted? We are adding them to the top of the file right now. Maybe match the order in the source config?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1260,7 +1260,7 @@ namespace Babylon
             throw std::runtime_error{"The data size does not match width, height, and format"};
         }
 
-        bimg::ImageContainer* image{bimg::imageAlloc(&m_allocator, format, width, height, 1, 1, false, bytes)};
+        bimg::ImageContainer* image{bimg::imageAlloc(&m_allocator, format, width, height, 1, 1, false, false, bytes)};
```
bghgary, line 1263: uh, how did this compile before??

bghgary (approved):

Apps/ValidationTests/Scripts/config.json
CedricGuillemet: yes, many gltf tests this way in babylon.js VT.

bghgary: Babylon.js VT has support for sRGB, but Babylon Native does not yet. The new playground with the `__disableSRGBBuffers__` flag is not supported by Babylon Native yet. We should still be using `#DS8AA7#18` playground for this.

PR #813, opened 2021-06-23
Fix Missing Canvas Init VT iOS

CedricGuillemet:
Missing Canvas init for iOS ValidationTests


CedricGuillemet:
No need for review here. Merging.

PR #814, opened 2021-06-24
3rd version of YFlip. W/O fullscreen pass

CedricGuillemet:
I had to get back to the root of the problem and find a new way instead of trying to add a YFlip full target pass on XR. 
The root of the problem in not in the clip space coordinates but in the texture coordinates.
Everything is explained here:
https://www.puredevsoftware.com/blog/2018/03/17/texture-coordinates-d3d-vs-opengl/
as stated : "The important thing to notice is that if both your texture is upside down, and your UV coordinates are upside down, then they cancel each other out, and the end result is the same. " That's why we don't have any special code when loading texture between D3D/GL.
Clipspace is the same, no need to change that. but as previously explained, sampling a rendertarget texture is YFlipped.
The approach in this PR is to keep everything the same but change sampling coordinates to do a YFlip on the V uv coordinate.
And last step is to flip InvertY boolean in loadTexture to make the last YFlip. 

No fullscreen pass or flip in the vertex shader is needed. 

TLDR: This PR removes everything from the previous YFlip PR. Invert YInvert when loading textures for D3D and does a YFlip in texture sampling.


syntheticmagus (approved):
This, among other things, is likely to be heavily affected by the upcoming Graphics Context work; but if this addresses issues in the immediate term, it looks good to me.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -33,10 +33,10 @@ bool operator()(const ARPlaneAnchor* lhs, const ARPlaneAnchor* rhs) const {
      */
     static XRVertex vertices[] = {
         // 2D positions, UV,        camera UV
-        { { -1, -1 },   { 0, 0 },   { 0, 0} },
-        { { -1, 1 },    { 0, 1 },   { 0, 0} },
-        { { 1, -1 },    { 1, 0 },   { 0, 0} },
-        { { 1, 1 },     { 1, 1 },   { 0, 0} },
+        { { -1, -1 },   { 0, 1 },   { 0, 0} },
```
ryantrem, line 36: With this change, I think the whole big comment above should be removed?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -815,11 +772,40 @@ namespace Babylon
...
+    std::string ProcessSamplerFlip(const std::string& vertexSource)
+    {
+        const auto* caps = bgfx::getCaps();
+        // for d3d, vulkan, metal, flip the texture sampling on vertical axis
+        if (!caps->originBottomLeft)
+        {
+            std::string patchedVertexSource = vertexSource;
+
+            static const std::string shaderNameDefineStr = "#define SHADER_NAME";
+            static const std::string samplingStr = "texture(";
```
ryantrem, line 791: Unused?

Plugins/NativeEngine/Source/ShaderCompilerD3D.cpp
```diff
@@ -46,7 +46,7 @@ namespace Babylon
             }
 
             std::string hlsl = compiler->compile();
-
+
```
ryantrem, line 49: Nit:
```suggestion

```

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -731,6 +731,41 @@ namespace Babylon::ShaderCompilerTraversers
...
+                return true;
+            }
+
+        private:
+            InvertYDerivativeOperandsTraverser(TIntermediate* intermediate)
+                : m_intermediate{ intermediate }
+            {
+            }
+
+            TIntermediate* m_intermediate;
```
ryantrem, line 767: Init to `nullptr`?
```suggestion
            TIntermediate* m_intermediate{};
```

bghgary (approved):
Big shift in thinking. Are we sure we have it right this time? :-)

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
bghgary: Not necessary since the constructor initializes the variable, but it's not harmful either.

rgerd:
Orientation and perf look fixed on HoloLens 2 👍

rgerd (approved):

PR #820, opened 2021-07-01
Added functionality for producing ARKit LiDAR mesh geometry data

chiamaka-123:
Functionality for producing ARKit LiDAR mesh geometry data was added. ARKit uses the LiDAR scanner to create a polygonal model of the physical environment. This data will allow the user to more accurately locate points on real-world surfaces. 

- Available on devices with a LiDAR scanner that support scene reconstruction, running iOS 13.4 or later.

### DIFF
--- a/Apps/Playground/Scripts/experience.js
+++ b/Apps/Playground/Scripts/experience.js
@@ -12,6 +12,7 @@ var vr = false;
 var ar = false;
 var xrHitTest = false;
 var xrFeaturePoints = false;
+var meshDetection = false;
 var text = false;
 var hololens = false;
 var cameraTexture = false;
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
                     }, 5000);
                 }
 
+                // Showing visualization for ARKit LiDAR mesh data
+                if(meshDetection) {
+                    var mat = new BABYLON.StandardMaterial("mat", scene);
+                    mat.wireframe = true;
+                    mat.diffuseColor = BABYLON.Color3.Blue();
+                    const xrMeshes = xr.baseExperience.featuresManager.enableFeature(
+                    BABYLON.WebXRFeatureName.MESH_DETECTION,
+                    "latest",
+                    {convertCoordinateSystems: true});
+                    console.log("Enabled mesh detection.");
+                    const meshes = [];
+                    const meshID = [];
+                    
+                    // adding meshes
+                    xrMeshes.onMeshAddedObservable.add(mesh=> {
+                        try {
+                            console.log("Mesh added.");
+                            // create new mesh object
+                            var customMesh = new BABYLON.Mesh("custom", scene);
+                            var vertexData = new BABYLON.VertexData();
+                            vertexData.positions = mesh.positions;
+                            vertexData.indices = mesh.indices;
+                            vertexData.normals = mesh.normals;
+                            vertexData.applyToMesh(customMesh, true);
+                            customMesh.material = mat;
+                            // add mesh and mesh id to arrays
+                            meshes.splice(mesh.id, 0, (customMesh));
+                            meshID[mesh.id] = mesh.id;
+                        } catch (ex) {
+                            console.error(ex);
+                        }
+                    });
+                    
+                    // updating meshes
+                    xrMeshes.onMeshUpdatedObservable.add(mesh=> {
+                        try{
+                            console.log("Mesh updated.");
+                            if(mesh.id == meshID[mesh.id]) {
+                                var vertexData = new BABYLON.VertexData();
+                                vertexData.positions = mesh.positions;
+                                vertexData.indices = mesh.indices;
+                                vertexData.normals = mesh.normals;
+                                vertexData.applyToMesh(meshes[mesh.id], true);
+                            }
+                        } catch (ex) {
+                            console.error(ex);
+                        }
+                    });
+                    
+                    // removing meshes
+                    xrMeshes.onMeshRemovedObservable.add(mesh => {
+                        try{
+                            console.log("Mesh removed.");
+                            if(mesh.id == meshID[mesh.id]) {
+                                meshes[mesh.id].positions.dispose();
+                                meshes[mesh.id].normals.dispose();
+                                meshes[mesh.id].indices.dispose();
+                                scene.remove(meshes[mesh.id]);
+                                meshes[mesh.id].dispose();
+                                delete meshID[mesh.id];
+                            }
+                        } catch (ex) {
+                            console.error(ex);
+                        }
+                    });
+                }
+                
                 // Below is an example of how to process feature points.
                 if (xrFeaturePoints) {
                     // First we attach the feature point system feature the XR experience.
--- a/Apps/Playground/Scripts/experience.js
+++ b/Apps/Playground/Scripts/experience.js
@@ -179,18 +179,17 @@ CreateBoxAsync().then(function () {
                 }
 
                 // Showing visualization for ARKit LiDAR mesh data
-                if(meshDetection) {
+                if (meshDetection) {
                     var mat = new BABYLON.StandardMaterial("mat", scene);
                     mat.wireframe = true;
                     mat.diffuseColor = BABYLON.Color3.Blue();
                     const xrMeshes = xr.baseExperience.featuresManager.enableFeature(
-                    BABYLON.WebXRFeatureName.MESH_DETECTION,
-                    "latest",
-                    {convertCoordinateSystems: true});
+                        BABYLON.WebXRFeatureName.MESH_DETECTION,
+                        "latest",
+                        {convertCoordinateSystems: true});
                     console.log("Enabled mesh detection.");
-                    const meshes = [];
-                    const meshID = [];
-                    
+                    const meshMap = new Map();
+
                     // adding meshes
                     xrMeshes.onMeshAddedObservable.add(mesh=> {
                         try {
@@ -203,48 +202,43 @@ CreateBoxAsync().then(function () {
                             vertexData.normals = mesh.normals;
                             vertexData.applyToMesh(customMesh, true);
                             customMesh.material = mat;
-                            // add mesh and mesh id to arrays
-                            meshes.splice(mesh.id, 0, (customMesh));
-                            meshID[mesh.id] = mesh.id;
+                            // add mesh and mesh id to map
+                            meshMap.set(mesh.id, customMesh);
                         } catch (ex) {
                             console.error(ex);
                         }
                     });
-                    
+
                     // updating meshes
                     xrMeshes.onMeshUpdatedObservable.add(mesh=> {
-                        try{
+                        try {
                             console.log("Mesh updated.");
-                            if(mesh.id == meshID[mesh.id]) {
+                            if (meshMap.has(mesh.id)) {
                                 var vertexData = new BABYLON.VertexData();
                                 vertexData.positions = mesh.positions;
                                 vertexData.indices = mesh.indices;
                                 vertexData.normals = mesh.normals;
-                                vertexData.applyToMesh(meshes[mesh.id], true);
+                                vertexData.applyToMesh(meshMap.get(mesh.id), true);
                             }
                         } catch (ex) {
                             console.error(ex);
                         }
                     });
-                    
+
                     // removing meshes
                     xrMeshes.onMeshRemovedObservable.add(mesh => {
-                        try{
+                        try {
                             console.log("Mesh removed.");
-                            if(mesh.id == meshID[mesh.id]) {
-                                meshes[mesh.id].positions.dispose();
-                                meshes[mesh.id].normals.dispose();
-                                meshes[mesh.id].indices.dispose();
-                                scene.remove(meshes[mesh.id]);
-                                meshes[mesh.id].dispose();
-                                delete meshID[mesh.id];
+                            if (meshMap.has(mesh.id)) {
+                                meshMap.get(mesh.id).dispose();
+                                meshMap.delete(mesh.id);
                             }
                         } catch (ex) {
                             console.error(ex);
                         }
                     });
                 }
-                
+
                 // Below is an example of how to process feature points.
                 if (xrFeaturePoints) {
                     // First we attach the feature point system feature the XR experience.
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -1032,16 +1032,16 @@ namespace xr {
                 return;
             }
 
-            [sessionDelegate LockMeshes];
-            @try {
-                if (@available(iOS 13.4, *)) {
+            if (@available(iOS 13.4, *)) {
+                [sessionDelegate LockMeshes];
+                @try {
                     // Update all meshes that have been updated since the last frame
                     auto updatedARKitMeshes = [sessionDelegate GetUpdatedMeshes];
                     for (ARMeshAnchor* updatedMesh : *updatedARKitMeshes) {
-                        const auto geometry = updatedMesh.geometry;
-                        const auto faces = geometry.faces;
-                        const auto vertices = geometry.vertices;
-                        const auto normals = geometry.normals;
+                        const auto& geometry = updatedMesh.geometry;
+                        const auto& faces = geometry.faces;
+                        const auto& vertices = geometry.vertices;
+                        const auto& normals = geometry.normals;
                         // get vertices data for vertex buffer
                         meshVertexBuffer.clear();
                         meshVertexBuffer.resize(vertices.count);
@@ -1125,9 +1125,9 @@ namespace xr {
 
                     // Clear the list of removed frames to start building up for the next mesh update.
                     removedARKitMeshes->clear();
+                } @finally {
+                    [sessionDelegate UnlockMeshes];
                 }
-            } @finally {
-                [sessionDelegate UnlockMeshes];
             }
         }
 
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -24,10 +24,13 @@ namespace {
         bool operator()(const ARPlaneAnchor* lhs, const ARPlaneAnchor* rhs) const {
             return lhs.identifier < rhs.identifier;
         }
+
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
         API_AVAILABLE(ios(13.4))
         bool operator()(const ARMeshAnchor* lhs, const ARMeshAnchor* rhs) const {
             return lhs.identifier < rhs.identifier;
         }
+#endif
     };
 
     /**
@@ -96,10 +99,12 @@ namespace {
     std::set<ARPlaneAnchor*,ARAnchorComparer> updatedPlanes;
     std::vector<ARPlaneAnchor*> deletedPlanes;
     bool planeDetectionEnabled;
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
     API_AVAILABLE(ios(13.4))
     std::set<ARMeshAnchor*,ARAnchorComparer> updatedMeshes;
     API_AVAILABLE(ios(13.4))
     std::vector<ARMeshAnchor*> deletedMeshes;
+#endif
     bool meshDetectionEnabled;
 
     CVMetalTextureCacheRef textureCache;
@@ -142,12 +147,14 @@ namespace {
     return &updatedPlanes;
 }
 
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
 /**
  Returns the set of all updated meshes since the last time we consumed mesh updates.
  */
 - (std::set<ARMeshAnchor*, ARAnchorComparer>*) GetUpdatedMeshes  API_AVAILABLE(ios(13.4)){
     return &updatedMeshes;
 }
+#endif
 
 /**
  Returns the vector containing all deleted planes since the last time we consumed plane updates.
@@ -161,16 +168,25 @@ namespace {
 }
 
 - (bool) TrySetMeshDetectorEnabled:(const bool)enabled {
-    meshDetectionEnabled = enabled;
-    return true;
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
+    if (@available(iOS 13.4, *)) {
+        if([ARWorldTrackingConfiguration supportsSceneReconstruction: ARSceneReconstructionMesh]) {
+            meshDetectionEnabled = enabled;
+            return true;
+        }
+    }
+#endif
+    return false;
 }
 
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
 /**
  Returns the vector containing all deleted meshes since the last time we consumed mesh updates.
  */
 - (std::vector<ARMeshAnchor*>*) GetDeletedMeshes  API_AVAILABLE(ios(13.4)){
     return &deletedMeshes;
 }
+#endif
 
 /**
  Initializes this session delgate with the given frame views and metal graphics context.
@@ -187,9 +203,9 @@ namespace {
     updatedPlanes = {};
     deletedPlanes = {};
     planeLock = [[NSLock alloc] init];
-    
+
     updatedMeshes = {};
-    deletedMeshes= {};
+    deletedMeshes = {};
     meshLock = [[NSLock alloc] init];
     return self;
 }
@@ -398,6 +414,7 @@ namespace {
         [self UnlockPlanes];
     }
 
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
     if (@available(iOS 13.4, *)) {
         if (meshDetectionEnabled) {
             [self LockMeshes];
@@ -406,10 +423,11 @@ namespace {
                     updatedMeshes.insert((ARMeshAnchor*)newAnchor);
                 }
             }
-        }
 
-        [self UnlockMeshes];
+            [self UnlockMeshes];
+        }
     }
+#endif
     return;
 }
 
@@ -425,6 +443,7 @@ namespace {
         [self UnlockPlanes];
     }
 
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
     if (@available(iOS 13.4, *)) {
         if (meshDetectionEnabled) {
             [self LockMeshes];
@@ -433,10 +452,11 @@ namespace {
                     updatedMeshes.insert((ARMeshAnchor*)updatedAnchor);
                 }
             }
-        }
 
-        [self UnlockMeshes];
+            [self UnlockMeshes];
+        }
     }
+#endif
     return;
 }
 
@@ -452,6 +472,7 @@ namespace {
         [self UnlockPlanes];
     }
 
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
     if (@available(iOS 13.4, *)) {
         if (meshDetectionEnabled) {
             [self LockMeshes];
@@ -460,10 +481,11 @@ namespace {
                     deletedMeshes.push_back((ARMeshAnchor*)removedAnchor);
                 }
             }
-        }
 
-        [self UnlockMeshes];
+            [self UnlockMeshes];
+        }
     }
+#endif
     return;
 }
 
@@ -627,11 +649,13 @@ namespace xr {
             // Create the ARSession enable plane detection, include scene reconstruction mesh if supported, and disable lighting estimation.
             session = [ARSession new];
             auto configuration = [ARWorldTrackingConfiguration new];
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
             if (@available(iOS 13.4, *)) {
-                if([ARWorldTrackingConfiguration supportsSceneReconstruction: ARSceneReconstructionMesh]) {
+                if ([ARWorldTrackingConfiguration supportsSceneReconstruction: ARSceneReconstructionMesh]) {
                     configuration.sceneReconstruction = ARSceneReconstructionMesh;
                 }
             }
+#endif
             configuration.planeDetection = ARPlaneDetectionHorizontal | ARPlaneDetectionVertical;
             configuration.lightEstimationEnabled = false;
             configuration.worldAlignment = ARWorldAlignmentGravity;
@@ -1031,17 +1055,17 @@ namespace xr {
             if (![sessionDelegate TrySetMeshDetectorEnabled: true]) {
                 return;
             }
-
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
             if (@available(iOS 13.4, *)) {
                 [sessionDelegate LockMeshes];
                 @try {
                     // Update all meshes that have been updated since the last frame
                     auto updatedARKitMeshes = [sessionDelegate GetUpdatedMeshes];
                     for (ARMeshAnchor* updatedMesh : *updatedARKitMeshes) {
-                        const auto& geometry = updatedMesh.geometry;
-                        const auto& faces = geometry.faces;
-                        const auto& vertices = geometry.vertices;
-                        const auto& normals = geometry.normals;
+                        const auto& geometry{ updatedMesh.geometry };
+                        const auto& faces{ geometry.faces };
+                        const auto& vertices{ geometry.vertices };
+                        const auto& normals{ geometry.normals };
                         // get vertices data for vertex buffer
                         meshVertexBuffer.clear();
                         meshVertexBuffer.resize(vertices.count);
@@ -1129,6 +1153,7 @@ namespace xr {
                     [sessionDelegate UnlockMeshes];
                 }
             }
+#endif
         }
 
         void UpdateFeaturePointCloud() {
@@ -1444,9 +1469,9 @@ namespace xr {
     Anchor System::Session::Frame::CreateAnchor(Pose pose, NativeTrackablePtr) const {
         return m_impl->sessionImpl.CreateAnchor(pose);
     }
-    
+
     Anchor System::Session::Frame::DeclareAnchor(NativeAnchorPtr /*anchor*/) const {
-        throw std::runtime_error("not implemented"); 
+        throw std::runtime_error("not implemented");
     }
 
     void System::Session::Frame::UpdateAnchor(xr::Anchor& anchor) const {
@@ -1464,7 +1489,7 @@ namespace xr {
     System::Session::Frame::Plane& System::Session::Frame::GetPlaneByID(System::Session::Frame::Plane::Identifier planeID) const {
         return m_impl->sessionImpl.GetPlaneByID(planeID);
     }
-    
+
     System::Session::Frame::Mesh& System::Session::Frame::GetMeshByID(System::Session::Frame::Mesh::Identifier meshID) const {
         return m_impl->sessionImpl.GetMeshByID(meshID);
     }
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -24,8 +24,7 @@ namespace {
         bool operator()(const ARPlaneAnchor* lhs, const ARPlaneAnchor* rhs) const {
             return lhs.identifier < rhs.identifier;
         }
-
-#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130400)
         API_AVAILABLE(ios(13.4))
         bool operator()(const ARMeshAnchor* lhs, const ARMeshAnchor* rhs) const {
             return lhs.identifier < rhs.identifier;
@@ -99,7 +98,7 @@ namespace {
     std::set<ARPlaneAnchor*,ARAnchorComparer> updatedPlanes;
     std::vector<ARPlaneAnchor*> deletedPlanes;
     bool planeDetectionEnabled;
-#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130400)
     API_AVAILABLE(ios(13.4))
     std::set<ARMeshAnchor*,ARAnchorComparer> updatedMeshes;
     API_AVAILABLE(ios(13.4))
@@ -147,7 +146,7 @@ namespace {
     return &updatedPlanes;
 }
 
-#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130400)
 /**
  Returns the set of all updated meshes since the last time we consumed mesh updates.
  */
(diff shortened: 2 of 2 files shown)


chiamaka-123:
Also I believe Xcode 11.4+ would be needed for this, as earlier versions do not have support for developing apps for iOS 13.4 and do not have ARKit 3.5 library's update.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,50 +387,88 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
+    if (planeDetectionEnabled) {
+        [self LockPlanes];
+        for (ARAnchor* updatedAnchor : anchors) {
+            if ([updatedAnchor isKindOfClass:[ARPlaneAnchor class]]) {
+                updatedPlanes.insert((ARPlaneAnchor*)updatedAnchor);
+            }
+        }
+        
+        [self UnlockPlanes];
+
```
jomcclan, line 426: nit: empty line before closing curly bracket

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -161,6 +202,14 @@ - (void) UnlockPlanes {
     [planeLock unlock];
 }
 
+- (void) LockMeshes {
+    [planeLock lock];
```
jomcclan, line 206: Should this be using the meshLock?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -161,6 +202,14 @@ - (void) UnlockPlanes {
...
+}
+
+- (void) UnlockMeshes {
+    [planeLock unlock];
```
jomcclan, line 210: Should this be using the meshLock?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,50 +387,88 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
         }
+        
+        [self UnlockMeshes];
     }
-
-    [self UnlockPlanes];
+    return;
 }
 
+
```
jomcclan, line 471: nit: empty line

chiamaka-123:
https://user-images.githubusercontent.com/86322959/124304730-21644800-db32-11eb-8aa2-6bcb77e6344f.mov

Visualization of scene reconstruction!

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
                     }, 5000);
                 }
 
+                // Showing visualization for ARKit LiDAR mesh data
+                if(meshDetection) {
```
Alex-MSFT, line 182:
```suggestion
                if (meshDetection) {
```

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
...
+                            meshes.splice(mesh.id, 0, (customMesh));
+                            meshID[mesh.id] = mesh.id;
+                        } catch (ex) {
+                            console.error(ex);
+                        }
+                    });
+                    
+                    // updating meshes
+                    xrMeshes.onMeshUpdatedObservable.add(mesh=> {
+                        try{
```
Alex-MSFT, line 216:
```suggestion
                        try {
```

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
...
+                            console.log("Mesh updated.");
+                            if(mesh.id == meshID[mesh.id]) {
```
Alex-MSFT, line 218:
```suggestion
                            if (mesh.id == meshID[mesh.id]) {
```

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
...
+                                vertexData.applyToMesh(meshes[mesh.id], true);
+                            }
+                        } catch (ex) {
+                            console.error(ex);
+                        }
+                    });
+                    
+                    // removing meshes
+                    xrMeshes.onMeshRemovedObservable.add(mesh => {
+                        try{
```
Alex-MSFT, line 232:
```suggestion
                        try {
```

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
...
+                            console.log("Mesh removed.");
+                            if(mesh.id == meshID[mesh.id]) {
```
Alex-MSFT, line 234:
```suggestion
                            if (mesh.id == meshID[mesh.id]) {
```

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
...
 
+                // Showing visualization for ARKit LiDAR mesh data
+                if(meshDetection) {
+                    var mat = new BABYLON.StandardMaterial("mat", scene);
+                    mat.wireframe = true;
+                    mat.diffuseColor = BABYLON.Color3.Blue();
+                    const xrMeshes = xr.baseExperience.featuresManager.enableFeature(
+                    BABYLON.WebXRFeatureName.MESH_DETECTION,
+                    "latest",
+                    {convertCoordinateSystems: true});
```
Alex-MSFT, line 189:
```suggestion
                    const xrMeshes = xr.baseExperience.featuresManager.enableFeature(
                        BABYLON.WebXRFeatureName.MESH_DETECTION,
                        "latest",
                        {convertCoordinateSystems: true});
```

Alex-MSFT, line 189:
```suggestion
                    const xrMeshes = xr.baseExperience.featuresManager.enableFeature(
                        BABYLON.WebXRFeatureName.MESH_DETECTION,
                        "latest",
                        {convertCoordinateSystems: true});
```

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
...
+                    console.log("Enabled mesh detection.");
+                    const meshes = [];
+                    const meshID = [];
```
Alex-MSFT, line 192: Using these two arrays seems a little odd.  Maybe consider just using a map of mesh IDs to meshes.

eg.
const map = new Map();

// Add mesh to map (onMeshAddedObservable)
meshMap.set(meshId, customMesh);

// get mesh from map (onMeshUpdatedObservable)
if (meshMap.has(mesh.id)) {
   const mesh = meshMap[mesh.id];
}

// Delete mesh from map onMeshRemovedObservable
meshMap[mesh.id].dispose();
meshMap.delete(mesh.id);

chiamaka-123: Sounds good, will make the change

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
-
-    [self LockPlanes];
-    for (ARAnchor* newAnchor : anchors) {
-        if ([newAnchor isKindOfClass:[ARPlaneAnchor class]]) {
-            updatedPlanes.insert((ARPlaneAnchor*)newAnchor);
+    
+    if (meshDetectionEnabled) {
+        [self LockMeshes];
+        for (ARAnchor* newAnchor : anchors) {
+            if (@available(iOS 13.4, *)) {
```
Alex-MSFT, line 404: Might as well put this in the meshDetectionEnabled if statement, so we don't need to grab the lock if the API is unavailable.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
     }
 
-    [self LockPlanes];
-    for (ARAnchor* updatedAnchor : anchors) {
-        if ([updatedAnchor isKindOfClass:[ARPlaneAnchor class]]) {
-            updatedPlanes.insert((ARPlaneAnchor*)updatedAnchor);
+    if (meshDetectionEnabled) {
+        [self LockMeshes];
+        for (ARAnchor* updatedAnchor : anchors) {
+            if (@available(iOS 13.4, *)) {
```
Alex-MSFT, line 431: Same here

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
 
-    [self LockPlanes];
-    for (ARAnchor* removedAnchor : anchors) {
-        if ([removedAnchor isKindOfClass:[ARPlaneAnchor class]]) {
-            deletedPlanes.push_back((ARPlaneAnchor*)removedAnchor);
+    if (meshDetectionEnabled) {
+        [self LockMeshes];
+        for (ARAnchor* removedAnchor : anchors) {
+            if (@available(iOS 13.4, *)) {
+                if ([removedAnchor isKindOfClass:[ARMeshAnchor class]]) {
```
Alex-MSFT, line 459: And here

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
     }
 
-    [self LockPlanes];
-    for (ARAnchor* removedAnchor : anchors) {
-        if ([removedAnchor isKindOfClass:[ARPlaneAnchor class]]) {
-            deletedPlanes.push_back((ARPlaneAnchor*)removedAnchor);
+    if (meshDetectionEnabled) {
+        [self LockMeshes];
+        for (ARAnchor* removedAnchor : anchors) {
+            if (@available(iOS 13.4, *)) {
```
Alex-MSFT, line 458: And here

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -539,9 +624,14 @@ bool TryInitialize() {
...
 
-            // Create the ARSession enable plane detection, and disable lighting estimation.
+            // Create the ARSession enable plane detection, include scene reconstruction mesh if supported, and disable lighting estimation.
             session = [ARSession new];
             auto configuration = [ARWorldTrackingConfiguration new];
+            if (@available(iOS 13.4, *)) {
+                if([ARWorldTrackingConfiguration supportsSceneReconstruction: ARSceneReconstructionMesh]) {
+                    configuration.sceneReconstruction = ARSceneReconstructionMesh;
+                }
+            }
```
Alex-MSFT, line 634: Does this have a performance/memory impact?  Can we add it after the session is already started, and only enable it when the user requests mesh detection?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -39,6 +43,14 @@ bool operator()(const ARPlaneAnchor* lhs, const ARPlaneAnchor* rhs) const {
...
     };
 
+    /**
+     Defines the three-index combination that forms a unique triangle, or face. The index refers to that vertex's position in the vertices array.
+     */
+    struct Index3
+    {
+        uint32_t x, y, z;
+    };
+
```
chrisfromwork, line 53: nit: remove empty line spaces for easier merging in the future

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1295,10 +1550,10 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
         return false;
     }
 
-    bool System::Session::TrySetMeshDetectorEnabled(const bool)
+    bool System::Session::TrySetMeshDetectorEnabled(const bool enabled)
     {
-        // TODO
-        return false;
+        m_impl->TrySetMeshDetectorEnabled(enabled);
+        return enabled;
```
chrisfromwork, line 1556: >enabled [](http://example.com/codeflow?start=15&length=7)

this should probably always return true compared to return the value provided. returning true when setting the mesh detector to disabled will probably be a utilized scenario in the future. Right now it would return false.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1078,6 +1303,35 @@ void UpdatePlane(std::vector<Frame::Plane::Identifier>& updatedPlanes, Frame::Pl
...
+                static_assert(sizeof(Frame::Mesh::IndexType) == sizeof(uint32_t));
+                mesh.Indices.resize(indexBuffer.size());
+                memcpy(mesh.Indices.data(), indexBuffer.data(), indexBuffer.size() * sizeof(uint32_t));
+                mesh.HasNormals = true;
+                mesh.Normals.resize(normalsBuffer.size());
+                mesh.Normals.swap(normalsBuffer);
+                updatedMeshes.push_back(mesh.ID);
+            }
+        }
+
```
chrisfromwork, line 1334: nit: remove empty spaces in this line

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1078,6 +1303,35 @@ void UpdatePlane(std::vector<Frame::Plane::Identifier>& updatedPlanes, Frame::Pl
...
+            } else {
+                for (size_t i = 0; i < vertexBuffer.size(); i++) {
+                    if (abs(mesh.Positions[i].X - vertexBuffer[i].X) > 0.02f
+                        || abs(mesh.Positions[i].Y - vertexBuffer[i].Y) > 0.02f
+                        || abs(mesh.Positions[i].Z - vertexBuffer[i].Z) > 0.02f) {
+                        checkIfMeshUpdated = true;
+                    }
+                }
+            }
+
```
chrisfromwork, line 1320: nit: remove unused spaces in this line

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1078,6 +1303,35 @@ void UpdatePlane(std::vector<Frame::Plane::Identifier>& updatedPlanes, Frame::Pl
...
         }
 
+        void UpdateMesh(std::vector<Frame::Mesh::Identifier>& updatedMeshes, Frame::Mesh& mesh, std::vector<Vector3f>& vertexBuffer, std::vector<uint32_t>& indexBuffer, std::vector<Vector3f>& normalsBuffer) {
+            // Check if mesh was actually updated
+            BOOL checkIfMeshUpdated = false;
+            if(mesh.Positions.size() != vertexBuffer.size()) {
+                checkIfMeshUpdated = true;
+            } else {
+                for (size_t i = 0; i < vertexBuffer.size(); i++) {
+                    if (abs(mesh.Positions[i].X - vertexBuffer[i].X) > 0.02f
```
chrisfromwork, line 1313: it seems like this check might be more expensive than the below copy operation. Could we use something like memcmp on the data vectors instead?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1009,6 +1222,13 @@ void SetPlaneDetectionEnabled(bool enabled)
             [sessionDelegate SetPlaneDetectionEnabled:enabled];
         }
 
+        bool TrySetMeshDetectorEnabled(const bool enabled)
+        {
+            meshDetectionEnabled = enabled;
+            [sessionDelegate TrySetMeshDetectorEnabled:enabled];
+            return enabled;
+        }
+
```
chrisfromwork, line 1231: nit: remove unused spaces on this line

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -135,6 +160,18 @@ - (void) SetPlaneDetectionEnabled:(bool)enabled {
     planeDetectionEnabled = enabled;
 }
 
+- (bool) TrySetMeshDetectorEnabled:(const bool)enabled {
+    meshDetectionEnabled = enabled;
+    return enabled;
```
chrisfromwork, line 165: >enabled [](http://example.com/codeflow?start=11&length=7)

these TrySetMeshDetectorEnabled should always return true not necessarily enabled/disabled value.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1009,6 +1222,13 @@ void SetPlaneDetectionEnabled(bool enabled)
             [sessionDelegate SetPlaneDetectionEnabled:enabled];
         }
 
+        bool TrySetMeshDetectorEnabled(const bool enabled)
+        {
+            meshDetectionEnabled = enabled;
```
chrisfromwork, line 1227: >meshDetectionEnabled [](http://example.com/codeflow?start=12&length=20)

do we need to cache this meshDetectionEnabled value twice? It seems like we could always maintain this value in the session delegate and grab it when needed from the session delegate.

I do not know the history behind why plane detection doesn't do this/but double caching this same value seems like itd be prone to getting out of sync/generating regressions

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -981,6 +1180,20 @@ void UpdateFeaturePointCloud() {
             throw std::runtime_error{"Tried to get non-existent plane."};
         }
 
+        Frame::Mesh& GetMeshByID(Frame::Mesh::Identifier meshID)
+        {
+            // Loop over the mesh vector and find the correct mesh.
+            for (Frame::Mesh& mesh : Meshes)
+            {
```
chrisfromwork, line 1187: We should consider using an unordered_map here instead of a vector. js/ts developers can directly call logic that will call this funciton. moving from n to log(n) would be great here.

I am not sure why plane detection uses a vector/do not have full context on that choice.

Dependencies/xr/Source/ARKit/XR.mm
chrisfromwork: actually looks like all the underlying TrySetMeshDetectorEnabled functions also return bools. We should consider just returning the function call here compared to an additional value.

chrisfromwork:
>     }

this just seems fundamentally broken and might explain the confusion below related to returning enabled.

---
Refers to: Dependencies/xr/Source/ARKit/XR.mm:1545 in aa91489. [](commit_id = aa91489dab4e33969c3ade56cfa88711bae198cd, deletion_comment = False)

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1009,6 +1222,13 @@ void SetPlaneDetectionEnabled(bool enabled)
             [sessionDelegate SetPlaneDetectionEnabled:enabled];
         }
 
+        bool TrySetMeshDetectorEnabled(const bool enabled)
+        {
+            meshDetectionEnabled = enabled;
+            [sessionDelegate TrySetMeshDetectorEnabled:enabled];
+            return enabled;
```
chrisfromwork, line 1229: consider returning the result of this function here compared to enabled

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -933,6 +1024,114 @@ void UpdatePlanes(std::vector<Frame::Plane::Identifier>& updatedPlanes, std::vec
...
+            [sessionDelegate LockMeshes];
+            @try {
+                if (@available(iOS 13.4, *)) {
+                    // Update all meshes that have been updated since the last frame
+                    auto updatedARKitMeshes = [sessionDelegate GetUpdatedMeshes];
+                    for (ARMeshAnchor* updatedMesh : *updatedARKitMeshes) {
+                        auto geometry = updatedMesh.geometry;
+                        auto faces = geometry.faces;
+                        auto vertices = geometry.vertices;
+                        auto normals = geometry.normals;
```
chrisfromwork, line 1045: it seems like we may want to access these vertices, normals, faces by reference. I think right now you have  copy constructor declared. If we do something like

const auto& vertices = geometry.vertices;

or

auto& vertices = geometry.vertices;

i think you should be able to reference this data in its current location. This will be more performant than having to regenerate this data in the new vertices/normals/faces vectors you are creating.

chrisfromwork: that all going said, im not 100% what data format these different values are in wrapped by the geometry. if this is only a reference copy to the data object it should be fine.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -933,6 +1024,114 @@ void UpdatePlanes(std::vector<Frame::Plane::Identifier>& updatedPlanes, std::vec
...
+                            auto vertexTransform = matrix_identity_float4x4;
+                            vertexTransform.columns[3] = simd_float4{vertex.X, vertex.Y, vertex.Z, 1};
+                            auto transform = matrix_multiply(updatedMesh.transform, vertexTransform);
+                            auto position = simd_float3{transform.columns[3].x ,transform.columns[3].y, transform.columns[3].z};
+                            meshVertexBuffer[i].X = position[0];
+                            meshVertexBuffer[i].Y = position[1];
+                            meshVertexBuffer[i].Z = position[2];
+                        }
+                        // get indices data for index buffer
+                        auto indexCount = faces.count * faces.indexCountPerPrimitive;
```
chrisfromwork, line 1063: >auto [](http://example.com/codeflow?start=24&length=4)

nit: consider with some of these autos you are creating declaring them const if they aren't ever reupdated. const specifiers should be better for performance

Apps/Playground/Scripts/experience.js
```diff
@@ -177,6 +178,73 @@ CreateBoxAsync().then(function () {
...
+
```
chrisfromwork, line 193: nit: we should probably go through and remove all unused spaces on empty lines. Some editors can have their preferences set to do this by default.

FWIW, its not that the spaces do anything substantial to the code or impact the application, but adding empty characters like this has historically caused weird empty space merge conflicts for folks continuing to iterate on the code.

chrisfromwork (approved):
:shipit:

chrisfromwork:
signing off but please get sign off from folks actively working in the ios space before checking in. it looks like the iOS ci build is failing with this change so i think there may also be some clean up items to make sure all newer types are wrapped by the correct API if available stubs.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: The reason that we're using an std::Vector instead of a map is because IDs are created at initialization time for Planes/Meshs in xr.h, which is in line with how we handle native IDs for other XR objects like InputSource, SceneObject, etc.  So we don't have a way to get the ID before the object is created when calling emplace_back (or emplace if using a map).

We could switch to managing the IDs ourselves, but this would be a change in how we manage these IDs for other objects as well.  At the time we made the decision that the number of planes was small enough that it wasn't worth the tradeoff, but as we start introducing more and more of these it might make sense to make that change.

Eg.
`                struct Mesh
                {
                    using Identifier = size_t;
                private:
                    static inline Identifier NEXT_ID{ 0 };
                };
`

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Agreed, this was actually just a holdover of how it was handled in Android where we don't have the notion of a separate sessionDelegate, and the *DetectionEnabled property on SessionImpl probably should have been factored out.

Dependencies/xr/Source/ARKit/XR.mm
chiamaka-123: Because faces, vertices, and normals are in MTL Buffers, I was having issues with accessing the data and converting it from bytes to the formats that I need (Vector3f & uint32_t) correctly. The numbers were always off. But I will look into this some more

Dependencies/xr/Source/ARKit/XR.mm
chiamaka-123: Based on the documentation that I've been referencing, the scene reconstruction was always added to the config before the session starts. I'm not entirely sure how I would enable this after the session has started.

Alex-MSFT:
> > ```
> > }
> > ```
> 
> this just seems fundamentally broken and might explain the confusion below related to returning enabled.
> 
> Refers to: Dependencies/xr/Source/ARKit/XR.mm:1545 in [aa91489](https://github.com/BabylonJS/BabylonNative/commit/aa91489dab4e33969c3ade56cfa88711bae198cd). [](commit_id = [aa91489](https://github.com/BabylonJS/BabylonNative/commit/aa91489dab4e33969c3ade56cfa88711bae198cd), deletion_comment = False)

The reason this returns a boolean was to allow for us a straight forward way to allow for platform specific feature enablement in BabylonNative.  So this method returns the state of whether the feature is disabled/enabled after the call to trySet*Enabled.  If you take a look at how this is used in Babylon.JS we always pass true in when initializing the feature, then decide whether it was actually enabled or not based on the result:
```
    private _init() {
        if (!this._xrSessionManager.session.trySetFeaturePointCloudEnabled || !this._xrSessionManager.session.trySetFeaturePointCloudEnabled(true)) {
            // fail silently
            return;
        }

        this._enabled = true;
    }
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -24,6 +24,10 @@ @interface SessionDelegate : NSObject <ARSessionDelegate, MTKViewDelegate>
         bool operator()(const ARPlaneAnchor* lhs, const ARPlaneAnchor* rhs) const {
             return lhs.identifier < rhs.identifier;
         }
+        API_AVAILABLE(ios(13.4))
+        bool operator()(const ARMeshAnchor* lhs, const ARMeshAnchor* rhs) const {
+            return lhs.identifier < rhs.identifier;
+        }
```
ryantrem, line 30: If this code is compiled with an older version of XCode, which uses an older version of the iOS SDK, then the `ARMeshAnchor` type is unknown. The CI builds this code with both older and newer versions of XCode, and when using newer versions of XCode also builds with different target iOS versions. To handle the case where we are compiling with an older version of XCode/iOS SDK, I think we also need to conditionally remove this from even being compiled. I think something like:
```suggestion
#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
        API_AVAILABLE(ios(13.4))
        bool operator()(const ARMeshAnchor* lhs, const ARMeshAnchor* rhs) const {
            return lhs.identifier < rhs.identifier;
        }
#endif
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -80,9 +92,15 @@ @implementation SessionDelegate {
...
 
     NSLock* planeLock;
+    NSLock* meshLock;
     std::set<ARPlaneAnchor*,ARAnchorComparer> updatedPlanes;
     std::vector<ARPlaneAnchor*> deletedPlanes;
     bool planeDetectionEnabled;
+    API_AVAILABLE(ios(13.4))
+    std::set<ARMeshAnchor*,ARAnchorComparer> updatedMeshes;
+    API_AVAILABLE(ios(13.4))
+    std::vector<ARMeshAnchor*> deletedMeshes;
```
ryantrem, line 102: I think in this case, since this is just part of the implementation details, you can just do:
```suggestion
#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)
    std::set<ARMeshAnchor*,ARAnchorComparer> updatedMeshes;
    std::vector<ARMeshAnchor*> deletedMeshes;
#endif
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1216,9 +1464,9 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
     System::Session::Frame::Plane& System::Session::Frame::GetPlaneByID(System::Session::Frame::Plane::Identifier planeID) const {
         return m_impl->sessionImpl.GetPlaneByID(planeID);
     }
-
-    System::Session::Frame::Mesh& System::Session::Frame::GetMeshByID(System::Session::Frame::Mesh::Identifier) const {
-        throw std::runtime_error("not implemented");
+
```
ryantrem, line 1467: Nit (extra white space presumably inadvertently added):
```suggestion

```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -135,6 +160,18 @@ - (void) SetPlaneDetectionEnabled:(bool)enabled {
     planeDetectionEnabled = enabled;
 }
 
+- (bool) TrySetMeshDetectorEnabled:(const bool)enabled {
+    meshDetectionEnabled = enabled;
+    return true;
```
ryantrem, line 165: It seems like this needs to do a runtime check. My understanding is mesh detection is only available on iOS 13.4 or higher, so I would think this would be:
```suggestion
if (@available(iOS 13.4, *)) {
    meshDetectionEnabled = enabled;
    return true;
}

return false;
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
+        }
+
+        [self UnlockPlanes];
     }
 
-    [self LockPlanes];
-    for (ARAnchor* newAnchor : anchors) {
-        if ([newAnchor isKindOfClass:[ARPlaneAnchor class]]) {
-            updatedPlanes.insert((ARPlaneAnchor*)newAnchor);
+    if (@available(iOS 13.4, *)) {
```
ryantrem, line 401: If you do this check where `meshDetectionEnabled` is set (see my other comment), then I don't think you need to check it here (and the caller of `TrySetMeshDetectorEnabled` will see that enabling it failed because it returned false).

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
+            for (ARAnchor* newAnchor : anchors) {
+                if ([newAnchor isKindOfClass:[ARMeshAnchor class]]) {
+                    updatedMeshes.insert((ARMeshAnchor*)newAnchor);
+                }
+            }
         }
-    }
 
-    [self UnlockPlanes];
+        [self UnlockMeshes];
```
ryantrem, line 411: I think the `[self UnlockMeshes]` is meant to be inside the `if (meshDetectionEnabled)` conditional block?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -124,6 +142,13 @@ @implementation SessionDelegate {
     return &updatedPlanes;
 }
 
+/**
+ Returns the set of all updated meshes since the last time we consumed mesh updates.
+ */
+- (std::set<ARMeshAnchor*, ARAnchorComparer>*) GetUpdatedMeshes  API_AVAILABLE(ios(13.4)){
```
ryantrem, line 148: Everywhere you use `ARMeshAnchor` I think needs to be wrapped in a `#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)` check.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -150,6 +187,10 @@ - (id)init:(std::vector<xr::System::Session::Frame::View>*)activeFrameViews meta
     updatedPlanes = {};
     deletedPlanes = {};
     planeLock = [[NSLock alloc] init];
+    
+    updatedMeshes = {};
+    deletedMeshes= {};
```
ryantrem, line 192:
```suggestion
    deletedMeshes = {};
```

Dependencies/xr/Source/ARKit/XR.mm
ryantrem: I think you do need to wrap this block instead though in a `#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_4)` though since it is again using a type that is only available in newer SDKs.

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
+            for (ARAnchor* updatedAnchor : anchors) {
+                if ([updatedAnchor isKindOfClass:[ARMeshAnchor class]]) {
+                    updatedMeshes.insert((ARMeshAnchor*)updatedAnchor);
+                }
+            }
         }
-    }
 
-    [self UnlockPlanes];
+        [self UnlockMeshes];
```
ryantrem, line 438: Similar comment from above - move this inside the `if (meshDetectionEnabled)` block?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -338,48 +387,84 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
+            for (ARAnchor* removedAnchor : anchors) {
+                if ([removedAnchor isKindOfClass:[ARMeshAnchor class]]) {
+                    deletedMeshes.push_back((ARMeshAnchor*)removedAnchor);
+                }
+            }
         }
-    }
 
-    [self UnlockPlanes];
+        [self UnlockMeshes];
```
ryantrem, line 465: Same thing here :)

Dependencies/xr/Source/ARKit/XR.mm
ryantrem: Interesting, it would be good to understand the perf impact if this is basically always enabled. @chrisfromwork could it make sense to also allow enabling mesh detection either at the `System` level or as a parameter when starting an XR session? It could still be available as an API at the session level for platforms that support dynamically enabling/disabling it, but it would provide a way to enable/disable at a higher level as well. For HoloLens for example this would mean you could enable mesh detection before the session actually starts, and you could disable it after the session starts, whereas on iOS if you try to enable/disable after the session starts it would just return false (it can only be enabled/disabled before the session starts). Would something like that make sense?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -933,6 +1024,113 @@ void UpdatePlanes(std::vector<Frame::Plane::Identifier>& updatedPlanes, std::vec
...
 
+        /**
+         Updates existing meshes in place, gets the list of updated/created mesh IDs and removed mesh IDs.
+         */
+        void UpdateMeshes(std::vector<Frame::Mesh::Identifier>& updatedMeshes, std::vector<Frame::Mesh::Identifier>& deletedMeshes) {
+            if (![sessionDelegate TrySetMeshDetectorEnabled: true]) {
+                return;
+            }
+
+            if (@available(iOS 13.4, *)) {
```
ryantrem, line 1035: Shouldn't we not even get into this code if we aren't running at least iOS 13.4? Seems like this check is unnecessary here?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -933,6 +1024,113 @@ void UpdatePlanes(std::vector<Frame::Plane::Identifier>& updatedPlanes, std::vec
...
+                [sessionDelegate LockMeshes];
+                @try {
+                    // Update all meshes that have been updated since the last frame
+                    auto updatedARKitMeshes = [sessionDelegate GetUpdatedMeshes];
+                    for (ARMeshAnchor* updatedMesh : *updatedARKitMeshes) {
+                        const auto& geometry = updatedMesh.geometry;
```
ryantrem, line 1041: I think this file uses uniform initialization everywhere else, so it would be ideal to make this new code consistent. For this line for example, it would be:
```suggestion
                        const auto& geometry{ updatedMesh.geometry };
```
(and that would be applied everywhere)

ryantrem (commented):
Super awesome to see these changes coming in!

Dependencies/xr/Source/ARKit/XR.mm
chiamaka-123: I was thinking the same but Xcode views it as an error if I don't include @available whenever I reference ARMeshAnchor (or anything mesh related from ARKit) and the build fails without it

Dependencies/xr/Source/ARKit/XR.mm
chiamaka-123: I also realized that mesh detection would only be available if scene reconstruction is supported on the device, so I went ahead and added that check in as well. In the case that they enable mesh detection but device doesn't support it, should return false.

chiamaka-123:
Requesting approval for the PR & any additional feedback/comments you all may have :) @Alex-MSFT @jomcclan @ryantrem

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -933,6 +1048,114 @@ void UpdatePlanes(std::vector<Frame::Plane::Identifier>& updatedPlanes, std::vec
             }
         }
 
+#if (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130400)
+        /**
+         Updates existing meshes in place, gets the list of updated/created mesh IDs and removed mesh IDs.
+         */
+        void UpdateMeshes(std::vector<Frame::Mesh::Identifier>& updatedMeshes, std::vector<Frame::Mesh::Identifier>& deletedMeshes) {
+            if (![sessionDelegate TrySetMeshDetectorEnabled: true]) {
```
jomcclan, line 1056: Should this be setting the mesh detector to enabled, or just checking if it's enabled? It seems odd that the UpdateMeshes function would directly enable the mesh detector, but I might be missing something

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1078,6 +1326,32 @@ void UpdatePlane(std::vector<Frame::Plane::Identifier>& updatedPlanes, Frame::Pl
             updatedPlanes.push_back(plane.ID);
         }
 
+        void UpdateMesh(std::vector<Frame::Mesh::Identifier>& updatedMeshes, Frame::Mesh& mesh, std::vector<Vector3f>& vertexBuffer, std::vector<uint32_t>& indexBuffer, std::vector<Vector3f>& normalsBuffer) {
+            // Check if mesh was actually updated
+            BOOL checkIfMeshUpdated = false;
```
ryantrem, line 1331:
```suggestion
            bool checkIfMeshUpdated = false;
```

jomcclan (approved):

chrisfromwork:
@ryantrem that suggestion doesn't really work well with how WebXR Features are enabled today at the js level. The webxr features manager typically allows webxr features to be enabled before creating an xr session or while an xr session is running. for hololens, we always declare that we may want to use the scene understanding openxr feature, but we only start detecting meshes once the webxr mesh/plane detection feature has been enabled.

I don't have a strong preference on requiring that all webxr features need to be able to be turned on before/during xr sessions, but we need to sync with babylon.js owners because its a deviation from how the apis are used today. IMO, this change would be beyond the scope of this review because its also going to require babylon.js changes.

Dependencies/xr/Source/ARKit/XR.mm
chiamaka-123: So, to measure the execution time I added NSDate *methodStart = [NSDate date]; at  line 1452, and NSDate *methodFinish = [NSDate date]; at line 1553 right before the call to GetNextFrame. Also added NSTimeInterval executionTime = [methodFinish timeIntervalSinceDate:methodStart]; to get the total execution time. It doesn’t seem like theres a huge difference between the timing. Let me know if there was a different way you wanted me to measure this @ryantrem 
tempImage3Tn0kQ
tempImagecrtLXJ

Alex-MSFT (approved):
I'm approving, but I would like to see some more investigation into how much always enabling mesh reconstruction impacts performance.  If it is significant it would be great if we could find a way to dynamically enable it only when enabled.

One way we might be able to do this would be to delay the initialization of the ARSession until we enter the render loop, but this would be a relatively large change, and outside of the scope of this change.

ryantrem (approved):

PR #826, opened 2021-07-09
GRAPHICS_API switch for Win32/Android/Linux

CedricGuillemet:
Select Graphics API this way:
```
cmake .. -G "Visual Studio 16 2019" -DGRAPHICS_API=D3D12
```

Need https://github.com/BabylonJS/BabylonNative/pull/950 to be merged before this one.

Vulkan status:
Current state: Playground runs and displays a green cube. Uniforms for VS are properly bound. For FS, it looks like uniforms are shuffled.
image

Cube appears green because vAmbientColor is full green. No idea why it happens.

image


Dependencies/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(BGFX_USE_DEBUG_SUFFIX OFF CACHE BOOL "Add 'd' suffix to debug output targets
 add_subdirectory(bgfx.cmake)
 target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_DEBUG_UNIFORM=0)
 target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_MULTITHREADED=1)
-target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_MAX_VERTEX_STREAMS=32)
+target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_MAX_VERTEX_STREAMS=16)
```
bghgary, line 28: why?

Apps/ValidationTests/Scripts/validation_native.js
```diff
@@ -84,6 +84,7 @@ function evaluate(test, resultCanvas, result, referenceImage, index, waitRing, d
 function processCurrentScene(test, resultCanvas, result, renderImage, index, waitRing, done, compareFunction) {
     currentScene.useConstantAnimationDeltaTime = true;
     var renderCount = test.renderCount || 1;
+    renderCount++;
```
bghgary, line 87: Why increment here?

CedricGuillemet:
@bghgary @darraghjburke Regarding D3D12/D3D11 shader compilation. I've put everything in 1 file and just added an `#include` per platform. Not great but does the job. Do you see anything better that can be done?

Plugins/NativeEngine/Source/ShaderCompilerD3D.h
```diff
@@ -96,6 +98,7 @@ namespace Babylon
         }
 
         ShaderCompilerTraversers::IdGenerator ids{};
+        auto cutScope = ShaderCompilerTraversers::ChangeUniformTypes(program, ids);
```
CedricGuillemet, line 101: uniform bloc is the same for the VS and FS. FS uniforms definition start with an offset in that buffer. For D3D12, this offset must be 16bytes aligned. So, with vec4/mat4, it is. D3D11 uniform definition is different but making it aligned anyway.

Apps/ValidationTests/Scripts/validation_native.js
CedricGuillemet: removed.

Dependencies/CMakeLists.txt
CedricGuillemet: The crash happens here with 32 streams but works ok with 16. https://github.com/bkaradzic/bgfx/blob/16ee93b88d7ec056a4cc5a7a7d9679af1e8cfefe/src/renderer_d3d12.cpp#L4292 I tried to find some documentation/limits but I didn't find anything.

CedricGuillemet: I (eventually) had idea to use dxcpl with validation layer and got this:
```
D3D12 ERROR: ID3D12CommandList::IASetVertexBuffers: StartSlot(0) + NumDescriptors(33) must not be greater than 32. [ EXECUTION ERROR #726: SET_VERTEX_BUFFERS_INVALID]
```
BGFX_CONFIG_MAX_VERTEX_STREAMS=31 should be ok.

CedricGuillemet:
gltf2 samples (models and generated) are passing with mandatory `BABYLON.MorphTargetManager.EnableTextureStorage = false;`

Dependencies/CMakeLists.txt
bghgary: We talked offline and verified 16 is okay. Plus, we limit to 16 on the JS side right now anyways.

bghgary (approved):

PR #832, opened 2021-07-19
Enable NAPI-JSI build on top of V8JSI

### LINKED ISSUE #828: Add a NAPI-JSI build to pre-test Babylon React Native
We just need to fill in the gaps to implement an AppRuntimeJSI and add some CMake to consume the proper headers and libraries from the [v8-jsi project](https://github.com/microsoft/v8-jsi). 

This will help us catch issues in the future like these:
https://github.com/BabylonJS/BabylonNative/pull/822
https://github.com/BabylonJS/BabylonNative/pull/805

And could help diagnose this issue:
https://github.com/BabylonJS/BabylonReactNative/issues/257

rgerd:
Adding support for building on JSI will let us test our changes before pushing them to BabylonReactNative. A follow-up PR will enable CI validation testing for the JSI build.

* Added a JSI target for Win32 + UWP by pulling in the V8JSI NuGet package.
* Cleaned up some CMake logic for pulling in NuGet packages and including NAPI artifacts.
* Fixed a couple inconsistencies between napi-direct and napi-jsi at the contract layer which were blocking compilation, for example Console was using a version of Napi::Function::Call only available on napi-direct.
* Added Chrome DevTools support on UWP (including on HoloLens!)

Thanks to @bghgary for helping me whip up the POC minutes before his vacation.
fixes #828


Apps/Playground/Win32/App.cpp
```diff
@@ -140,7 +140,7 @@ namespace
             chromeDevTools = std::make_unique<Babylon::Plugins::ChromeDevTools>(Babylon::Plugins::ChromeDevTools::Initialize(env));
             if (chromeDevTools->SupportsInspector())
             {
-                chromeDevTools->StartInspector(5643, "BabylonNative Playground");
+                chromeDevTools->StartInspector(4653, "BabylonNative Playground");
```
bghgary, line 143: Just curious, why did the port change?

Core/AppRuntime/Source/AppRuntimeJSI.cpp
```diff
@@ -0,0 +1,47 @@
...
+                shared_task2->run();
+            });
+        }
+
+    private:
+        TaskRunnerAdapter(const TaskRunnerAdapter&) = delete;
+        TaskRunnerAdapter& operator=(const TaskRunnerAdapter&) = delete;
+
+        WorkQueue& m_workQueue;
+    };
```
bghgary, line 30: This should go in an anonymous namespace.

Core/AppRuntime/Source/AppRuntimeJSI.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    void AppRuntime::RunEnvironmentTier(const char*)
+    {   
+        v8runtime::V8RuntimeArgs args{};
+        args.inspectorPort = 4653;
+        args.foreground_task_runner = std::make_shared<TaskRunnerAdapter>(*m_workQueue);
+        
+        const auto runtime = v8runtime::makeV8Runtime(std::move(args));
+        Napi::Env env = Napi::Attach<facebook::jsi::Runtime&>(*runtime);
```
bghgary, line 39: nit: use brace initialization when possible

Dependencies/napi/OnLinkedAsDependency.cmake
```diff
@@ -1,27 +1,31 @@
...
-    if (NAPI_JAVASCRIPT_ENGINE STREQUAL "V8")
-
-        # Propagate this file to the target so that it will be transitively available to targets that
-        # link to that one, too.
-        propagate_on_linked_as_dependency_cmake_file(napi ${target})
-    
+    # Propagate this file to the target so that it will be transitively available to targets that
+    # link to that one, too.
+    propagate_on_linked_as_dependency_cmake_file(napi ${target})
+    if (DEFINED NAPI_JAVASCRIPT_RUNTIME_OUTPUT_ARTIFACTS)
```
bghgary, line 7: nit: extra spaces at the end
```suggestion
    if (DEFINED NAPI_JAVASCRIPT_RUNTIME_OUTPUT_ARTIFACTS)
```

Dependencies/CMakeExtensions/CMakeLists.txt
```diff
@@ -123,4 +123,20 @@ function(set_win32_arch)
     else()
         message(FATAL_ERROR "Unrecognized compiler: ${CMAKE_CXX_COMPILER}")
     endif()
+endfunction()
+
+# Uses the nuget.config and packages.config files in the current directory to download packages from NuGet.
+# NUGET_PATH will be set to the directory where the packages are installed.
+function (nuget_download)
```
bghgary, line 130: nit: most functions put the verb first. maybe this?
```suggestion
function (download_nuget)
```

Dependencies/napi/napi-direct/include/napi/napi-inl.h
```diff
@@ -1781,6 +1781,24 @@ inline Value Function::Call(napi_value recv, const std::vector<napi_value>& args
...
+        argv = heapArgs.data();
+    }
+
+    for (size_t i = 0; i < argc; ++i) {
+        argv[i] = args[i];
+    }
+
+    return Call(recv, argc, argv);
+}
+
```
bghgary, line 1801: This file is mostly unmodified from the node repo. If this is actually necessary, then we should try to upstream it to node. We typically don't change the N-API contract.

Dependencies/napi/napi-direct/nuget.config
```diff
@@ -0,0 +1,13 @@
...
+    <add key="repositoryPath" value="packages" />
+  </config>
+  <packageSources>
+    <clear />
+    <add key="Nuget.org" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+  <disabledPackageSources>
+    <clear />
+  </disabledPackageSources>
+</configuration>
```
bghgary, line 13: what's this for?

Dependencies/napi/napi-jsi/include/napi/js_native_api.h
```diff
@@ -0,0 +1,419 @@
+#ifndef SRC_JS_NATIVE_API_H_
```
bghgary, line 1: I'm not sure what is going on with this. We can't include the low-level C N-API files for napi-jsi which is an adapter for the C++ layer. Can we chat offline?

Dependencies/napi/napi-jsi/include/napi/napi-inl.h
```diff
@@ -1222,7 +1222,7 @@ namespace details
   struct Function {
     static inline jsi::Value Callback(napi_env env, const jsi::Value& thisVal, const jsi::Value* args, size_t count, void* data, Callable cb) {
       CallbackInfo callbackInfo{env, thisVal, args, count, {}, data};
-      return {env->rt, cb(callbackInfo)};
+      return jsi::Value(env->rt, cb(std::move(callbackInfo)));
```
bghgary, line 1225: Why is this necessary?

Dependencies/napi/napi-jsi/include/napi/napi.h
```diff
@@ -1161,6 +1130,8 @@ namespace Napi {
     CallbackInfo(napi_env env, const jsi::Value& thisVal, const jsi::Value* args, size_t argc, const jsi::Value& newTarget, void* data);
     ~CallbackInfo();
 
+    CallbackInfo(CallbackInfo &&) = default;
```
bghgary, line 1133: This is going to cause `_dynamicArgs` to be deleted twice I think. If we must have a move constructor, it needs to be implemented properly. Also, it's weird to have just a move constructor and no move assignment.

Dependencies/napi/napi-jsi/source/jsi.cpp
```diff
@@ -0,0 +1,484 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
```
bghgary, line 6: Why is this in source control? How do we make sure it is in sync? If we must put this in source control, we must also add this to the NOTICES file at the root.

Plugins/ChromeDevTools/CMakeLists.txt
```diff
@@ -1,9 +1,14 @@
 set(SOURCES 
     "Include/Babylon/Plugins/ChromeDevTools.h")
 
-set(CHROME_DEVTOOLS_SUPPORTED_ENGINES "V8")
+if (WIN32)
+    set(CHROME_DEVTOOLS_SUPPORTED_ENGINES ${CHROME_DEVTOOLS_SUPPORTED_ENGINES} "JSI")
+    if (NOT WINDOWS_STORE)
```
bghgary, line 6: Is V8 CDT not working with UWP?

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -66,18 +66,7 @@ namespace Babylon
...
-                    auto promise{ deferred.Promise() };
-
-                    auto& jsRuntime{ JsRuntime::GetFromJavaScript(env) };
-                    jsRuntime.Dispatch([deferred{ std::move(deferred) }](Napi::Env env) {
-                        deferred.Resolve(env.Null());
-                    });
-
-                    return promise;
-                }));
+                mediaDevices.Set("getUserMedia", Napi::Function::New(env, &NativeCamera::GetUserMedia, "getUserMedia"));
```
bghgary, line 69: This seems like an unrelate change?

bghgary (changes requested):
Main issue is the change to N-API contract.

Dependencies/napi/napi-direct/nuget.config
rgerd: It defines the upstream(s) for nuget.exe when it's downloading packages from packages.config, and tells nuget.exe where to put the packages. This is a boilerplate version, the napi-jsi one is more interesting, but it's used in the download_nuget command, and I believe it's best practice to have one with the packages.config file.

Plugins/ChromeDevTools/CMakeLists.txt
rgerd: No, only because we don't have a NuGet package for V8 compiled for UWP. I think going forward we're going to want to just remove V8 CDT in favor of using v8jsi, since that's where we get the inspector code from anyway.

bghgary: Gotcha. Was the logic was missing the `WINDOWS_STORE` check before?

Dependencies/napi/napi-jsi/include/napi/napi-inl.h
rgerd: I was getting compilation issues with the callback usage in the Console polyfill and this change seemed to satisfy the compiler errors-- the compiler doesn't yell anymore when I back it out so I'm just going to revert the change. Thanks for pointing it out!

Plugins/NativeCamera/Source/NativeCamera.cpp
rgerd: I was trying to satisfy the following compilation errors:
```
error C2440: 'return': cannot convert from 'initializer list' to 'facebook::jsi::Value'
napi-inl.h(1225,41): message : No constructor could take the source type, or constructor overload resolution was ambiguous
```
I put it in a more familiar shape and the errors went away, but it just occurred to me that I could just stick a `static_cast` on the promise, so that's what I'll do instead

Plugins/ChromeDevTools/CMakeLists.txt
rgerd: It's possible. I think when I originally wrote it I was under the false assumption that `WINDOWS_STORE` and `WIN32` were mutually exclusive, and I never hit this path on UWP since V8 isn't available for UWP in the first place.

bghgary: Yeah, I didn't know this either before. `WIN32` is for all windows platforms (Win32 and UWP). `WINDOWS_STORE` is an additional thing to check to differentiate.

Dependencies/napi/napi-jsi/source/jsi.cpp
bghgary: Actually, now that I think about it. We should add JSI to the notices regardless. We are still pulling it down.

Dependencies/napi/napi-jsi/include/napi/napi.h
```diff
@@ -39,6 +39,8 @@ typedef enum {
   napi_uint32_array,
   napi_float32_array,
   napi_float64_array,
+  napi_bigint64_array,
+  napi_biguint64_array,
```
bghgary, line 43: These weren't included since JSI doesn't support bigint. Maybe I should have commented them out with a comment explaining.

bghgary (approved):

Dependencies/napi/napi-direct/include/napi/napi-inl.h
rgerd: Filed a PR here: https://github.com/nodejs/node-addon-api/pull/1026

PR #833, opened 2021-07-20
Implement XrContext for ARKit

chrisfromwork:
This review contains the following changes:

1. XR resources for ARKit are moved to a shared location (IXrContextARKit) that can be accessed by external plugins through the javascript runtime
2. Functionality is added so that ARKit anchors created outside the BabylonNative XR system can be declared to the XR system.

Note: these changes are needed for iOS XR service scenarios such as Azure Spatial Anchors support.


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1480,9 +1516,9 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
     Anchor System::Session::Frame::CreateAnchor(Pose pose, NativeTrackablePtr) const {
         return m_impl->sessionImpl.CreateAnchor(pose);
     }
-
-    Anchor System::Session::Frame::DeclareAnchor(NativeAnchorPtr /*anchor*/) const {
-        throw std::runtime_error("not implemented");
+
```
rgerd, line 1519: nit: spaces

Dependencies/napi/napi-jsi/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ add_library(napi ${SOURCES})
 target_include_directories(napi
     PUBLIC "include")
 
-target_link_to_dependencies(napi
+target_link_libraries(napi
```
rgerd, line 16: Was this change intentional?

chrisfromwork: yes depending on make version target_link_to_dependencies doesn't exist as a valid cmake command. I'm not sure why we've been using it here.

chrisfromwork: actually looks like this was a custom command. Do we have more context around why we've been using this custom command:

https://github.com/BabylonJS/BabylonNative/blob/master/Dependencies/CMakeExtensions/CMakeLists.txt

syntheticmagus: I believe this was originally introduced, among other reasons, to copy DLLs for V8 on Windows. Does that still work with this change?

chrisfromwork: I may be naive/wrong here, but it doesn't look like we are setting ON_LINKED_AS_DEPENDENCY_CMAKE_FILES for this library, so I don't think using the custom target_link_to_dependencies is necessary here?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -622,12 +624,37 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
+        {
+            return Frame;
+        }
+
+        virtual ~XrContextARKit() = default;
+    };
+
     struct System::Impl {
     public:
+        std::shared_ptr<XrContextARKit> XrContext{std::make_shared<XrContextARKit>()};
```
syntheticmagus, line 652: Is this ownership actually conceptually shared? If not, could this be a simpler type?

Dependencies/napi/napi-jsi/CMakeLists.txt
chrisfromwork: if folks want this reverted, I can look into updating our consuming projects to locate this target_link_to_dependencies function. That going said, if it doesn't serve any additional purpose it seems worth converting back to target_link_libraries so that external consumers of napi jsi don't have to include these additional cmake function definitions

Dependencies/xr/Source/ARKit/XR.mm
chrisfromwork: this should now be fixed

Dependencies/xr/Source/ARKit/XR.mm
chrisfromwork: I don't think it will be possible to hand a shared pointer across dll/lib boundaries. That going said, from a legibility standpoint id prefer this to stay a pointer compared to a memory reference to a declared struct. We could change this to a unique_ptr if that's preferred over shared_ptr?

Dependencies/napi/napi-jsi/CMakeLists.txt
chrisfromwork: given this could potentially have issues on UWP but I've been primarily testing on iOS, I'm just going to refer this change.

rgerd: We will be using `add_on_linked_as_dependency_cmake_file` for napi-jsi in order to pull in v8jsi dll's (see #832). I also thought CMakeLists globally enforced a minimum version across files, so I'm not sure why iOS specifically would have issues consuming the custom function, unless there's a file with a version setter that needs to get revved.

rgerd (approved):

ryantrem (approved):

Dependencies/xr/Source/ARKit/XR.mm
chrisfromwork: talked offline, opted to switch to unique_ptr. this should now be fixed.

PR #838, opened 2021-07-21
Add srgb buffers support

bghgary:
Corresponding PR: https://github.com/BabylonJS/Babylon.js/pull/10710


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -72,81 +72,138 @@ namespace Babylon
...
+            const uint8_t srcBytesPerPixel{static_cast<uint8_t>(bimg::getBitsPerPixel(src->m_format) >> 3)};
+            const uint8_t* srcDataEnd{srcData + src->m_size};
 
-            std::vector<uint8_t> buffer(rowPitch);
+            uint8_t* dstData{static_cast<uint8_t*>(dst->m_data)};
+            const uint8_t dstBytesPerPixel{static_cast<uint8_t>(bimg::getBitsPerPixel(dst->m_format) >> 3)};
+            const uint8_t* dstDataEnd{dstData + dst->m_size};
 
-            for (size_t row = 0; row < rowCount / 2; row++)
+            while (srcData < srcDataEnd && dstData < dstDataEnd)
```
CedricGuillemet, line 86: Moving from 3 memcpy per row to 1 call per pixel, how is the performance? I guess it will be way slower as memcpy are inlined and can use SIMD in release builds.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1617,33 +1639,31 @@ namespace Babylon
...
         {
             throw Napi::Error::New(env, "CreateImageBitmap array buffer is empty.");
         }
 
-        bimg::ImageContainer* image = bimg::imageParse(&m_allocator, data.Data(), static_cast<uint32_t>(data.ByteLength()));
-        if (image == nullptr)
-        {
-            throw Napi::Error::New(env, "Unable to decode image in createImageBitmap function.");
-        }
+        bimg::ImageContainer* image{ParseImage(m_allocator, gsl::make_span(static_cast<uint8_t*>(data.Data()), data.ByteLength()))};
```
CedricGuillemet, line 1655: nit: use auto everywhere suitable?

CedricGuillemet (approved):
LGTM!

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: My opinion is that we should move away from using auto unless it's completely obvious what the type is.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: This code moved. I didn't actually change FlipY which is now inside the PrepareImage function. The code here is a new function I added called TransformImage which does a transform per pixel.

PR #839, opened 2021-07-22
Introduce Eye Tracking to BabylonNative

rickfromwork:
This change queries the OpenXR runtime for eye tracking information, and if eye tracking has been enabled by the developer and accepted by the user, sends eye data down to babylon.JS using a new set of events on the XRSession object.

BabylonJS-side change: https://github.com/BabylonJS/Babylon.js/pull/10713


Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -184,6 +184,8 @@ namespace xr
         std::vector<Frame::Plane> Planes{};
         std::vector<Frame::Mesh> Meshes{};
         std::vector<FeaturePoint> FeaturePointCloud{};
+        std::pair<bool, Frame::Space> EyeTrackerSpace{};
```
syntheticmagus, line 187: nit: Is the Boolean here just intended to express whether a space was detected or not? If so, might it make sense to have this be a `std::optional` instead of a pair?

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -642,10 +653,13 @@ namespace xr
         void InitializeRenderResources(XrInstance instance, XrSystemId systemId)
         {
             // Read graphics properties for preferred swapchain length and logging, and hand tracking availability.
-            XrSystemHandTrackingPropertiesEXT handTrackingSystemProperties{ XR_TYPE_SYSTEM_HAND_TRACKING_PROPERTIES_EXT };
+            XrSystemEyeGazeInteractionPropertiesEXT eyeTrackingSystemProperties{ XR_TYPE_SYSTEM_EYE_GAZE_INTERACTION_PROPERTIES_EXT};
+            XrSystemHandTrackingPropertiesEXT handTrackingSystemProperties{ XR_TYPE_SYSTEM_HAND_TRACKING_PROPERTIES_EXT, &eyeTrackingSystemProperties };
```
syntheticmagus, line 657: This seems very odd. Why does hand tracking get to know anything about eye tracking?

Dependencies/xr/Include/XR.h
```diff
@@ -301,6 +301,8 @@ namespace xr
                 std::vector<InputSource>& InputSources;
                 std::vector<FeaturePoint>& FeaturePointCloud;
 
+                std::pair<bool,Space>& EyeTrackerSpace;
```
chrisfromwork, line 304: >,S [](http://example.com/codeflow?start=30&length=2)

nit: add a space between this , and Space

chrisfromwork, line 304: >EyeTrackerSpace [](http://example.com/codeflow?start=39&length=15)

should this be a struct compared to a pair? are there other properties that could eventually be added here that won't allow for it to be a std::pair? it feels like .first and .second accessors for this bool and space is worse for readability compared to a struct with field names pertaining to whatever these values are for

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2700,6 +2703,42 @@ namespace Babylon
                 std::vector<xr::System::Session::Frame::InputSource::Identifier> squeezeStarts{};
                 std::vector<xr::System::Session::Frame::InputSource::Identifier> squeezeEnds{};
 
+                if (frame.EyeTrackerSpace.first)
```
chrisfromwork, line 2706: >first [](http://example.com/codeflow?start=42&length=5)

i really feel like this should be a struct field, it is really unclear what this first value is for

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2700,6 +2703,42 @@ namespace Babylon
...
+                        m_eyeTrackedSource.second.Set("gazeSpace", Napi::External<xr::System::Session::Frame::Space>::New(env, &frame.EyeTrackerSpace.second));
+
+                        for (const auto& [name, callback] : m_eventNamesAndCallbacks)
+                        {
+                            if (name == JS_EVENT_NAME_EYE_TRACKING_START)
+                            {
+                                Napi::Object obj = m_eyeTrackedSource.second.Value();
+                                callback.Call({obj});
+                            }
+                        }
```
chrisfromwork, line 2721: this function is already huge, can you move this event firing logic to new functions?

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2700,6 +2703,42 @@ namespace Babylon
                 std::vector<xr::System::Session::Frame::InputSource::Identifier> squeezeStarts{};
                 std::vector<xr::System::Session::Frame::InputSource::Identifier> squeezeEnds{};
 
+                if (frame.EyeTrackerSpace.first)
+                {
```
chrisfromwork, line 2707: can you move all of this eye tracking logic to its own ProcessEyeTracking function? The length of ProcessInputSources is making it less and less legible.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2700,6 +2703,42 @@ namespace Babylon
...
+
+                    for (const auto& [name, callback] : m_eventNamesAndCallbacks)
+                    {
+                        if (name == JS_EVENT_NAME_EYE_TRACKING_END)
+                        {
+                            callback.Call({});
+                        }
+                    }
+                }
+
```
chrisfromwork, line 2741: nit: remove unneeded spaces

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -670,6 +670,7 @@ bool TryInitialize() {
         std::vector<Frame::Plane> Planes{};
         std::vector<Frame::Mesh> Meshes{};
         std::vector<FeaturePoint> FeaturePointCloud{};
+        std::pair<bool, Frame::Space> EyeTrackerSpace{};
```
chrisfromwork, line 673: >EyeTrackerSpace [](http://example.com/codeflow?start=38&length=15)

should this be referenced by the xr frame? right now although this struct is populated it doesn't seem like it would ever be correctly populated for the frame on iOS

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -184,6 +184,8 @@ namespace xr
...
+
         float DepthNearZ{ DEFAULT_DEPTH_NEAR_Z };
```
chrisfromwork, line 189: nit: remove extra line

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2638,6 +2640,7 @@ namespace Babylon
 
             Napi::Reference<Napi::Array> m_jsInputSources{};
             std::map<xr::System::Session::Frame::InputSource::Identifier, Napi::ObjectReference> m_idToInputSource{};
+            std::pair<bool, Napi::ObjectReference> m_eyeTrackedSource{};
```
chrisfromwork, line 2643: >bool [](http://example.com/codeflow?start=22&length=4)

does this bool serve any purpose? why is it used over a null check for the ObjectReference?

chrisfromwork: if the bool is needed, please create your own struct definition for the type thats being used here compared to relying on a pair with a .first .second fields that don't reflect anything about how these properties are used.

chrisfromwork (approved):
:shipit:

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -642,10 +653,13 @@ namespace xr
...
             XrSystemProperties systemProperties{ XR_TYPE_SYSTEM_PROPERTIES, &handTrackingSystemProperties };
             XrCheck(xrGetSystemProperties(instance, systemId, &systemProperties));
 
+            SupportsEyeTracking = !!eyeTrackingSystemProperties.supportsEyeGazeInteraction;
```
chrisfromwork, line 661: >!! [](http://example.com/codeflow?start=34&length=2)

nit: this isn't a standard c++ coding practice

Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -813,6 +828,29 @@ namespace xr
...
+            {
+                XrActionCreateInfo actionInfo{ XR_TYPE_ACTION_CREATE_INFO };
+                actionInfo.actionType = XR_ACTION_TYPE_POSE_INPUT;
+                strcpy_s(actionInfo.actionName, ActionResources.EYE_GET_POSE_ACTION_NAME);
+                strcpy_s(actionInfo.localizedActionName, ActionResources.EYE_GET_POSE_ACTION_LOCALIZED_NAME);
+                XrCheck(xrCreateAction(ActionResources.ActionSet, &actionInfo, &ActionResources.EyeGetPoseAction));
+
+                // Create suggested binding
+                std::string path{ "/user/eyes_ext" };
+                path.append(ActionResources.EYE_GET_POSE_PATH_SUFFIX);
```
rgerd, line 842: fwiw instead of doing multiple std::string allocations here you could just `sprintf` into a `char actionPath[XR_MAX_PATH_LENGTH]`

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2524,6 +2524,8 @@ namespace Babylon
             static constexpr auto JS_EVENT_NAME_SQUEEZE = "squeeze";
             static constexpr auto JS_EVENT_NAME_SQUEEZE_START = "squeezestart";
             static constexpr auto JS_EVENT_NAME_SQUEEZE_END = "squeezeend";
+            static constexpr auto JS_EVENT_NAME_EYE_TRACKING_START= "eyetrackingstart";
```
rgerd, line 2527: nit: spaces before =

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: The system properties are passed in as a chain of properties to look for. So each item has to hold a reference to the previous, then only one query is needed.
XrSystemProperties holds a link to XrSystemHandTrackingPropertiesEXT , which links to XrSystemEyeGazeInteractionPropertiesEXT

Dependencies/xr/Source/ARCore/XR.cpp
rickfromwork: I was worried about the overhead of creating/tearing down a Frame::Space repeatedly, since we lose tracking whenever the user closes their eyes

Dependencies/xr/Source/OpenXR/XR.cpp
rickfromwork: Yeah, this is kind of ugly, I didn't really clean it up after copying the way controllers were processed

PR #844, opened 2021-07-23
Implement async setTimeout

### LINKED ISSUE #843: Window::RecursiveWaitOrCall is performance-heavy
Taking a CPU profile on HoloLens running a release-mode build shows that Window::RecursiveWaitOrCall is fairly heavy on performance-- most likely due to the high number of dispatch calls. It would be beneficial to reimplement this function using a worker thread that sleeps for the given timeout and then dispatches the function.

recursive-wait-or-call

rgerd:
**Note this PR is a work in progress as I learn more about async programming**

While taking CPU profiles of BabylonNative I noticed that `Window::RecursiveWaitOrCall` (used by the `setTimeout` polyfill) was taking up a bunch of cycles:
recursivewaitorcall-desktop

This seemed to be due to the high number of `Dispatch` calls coming from the recursive loop.

Now we're using a single worker thread with a condition variable and a priority queue, which allows us to both queue up delayed functions and preempt them, reducing the CPU load significantly.

fixes #843

### DIFF
--- a/Polyfills/Window/Source/Window.cpp
+++ b/Polyfills/Window/Source/Window.cpp
@@ -6,6 +6,112 @@
 #include <chrono>
 #include <iterator>
 
+namespace
+{
+    struct TimeoutFunction
+    {
+        std::shared_ptr<Napi::FunctionReference> Function;
+        std::chrono::time_point<std::chrono::system_clock> TimePoint;
+
+        TimeoutFunction(std::shared_ptr<Napi::FunctionReference> func, std::chrono::time_point<std::chrono::system_clock> time)
+            : Function{std::move(func)}
+            , TimePoint{time}
+        { }
+
+        static bool Compare(const TimeoutFunction& tf1, const TimeoutFunction& tf2)
+        {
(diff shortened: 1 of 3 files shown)


Polyfills/Window/Source/Window.cpp
```diff
@@ -6,6 +6,112 @@
...
+    {
+    public:
+        TimeoutDispatcher(Babylon::JsRuntime& runtime)
+            : m_runtime(runtime)
+            , m_thread{&TimeoutDispatcher::WaitThenCallProc, this}
+        { }
+
+        void Dispatch(std::shared_ptr<Napi::FunctionReference> func, std::chrono::system_clock::time_point time) 
+        {   
+            if (time <= std::chrono::system_clock::now())
```
tymlipari, line 37: FYI, `system_clock` can roll backwards if the user adjusts their system clock. `steady_clock` is guaranteed to always progress forward. I'd recommend using `steady_clock` for everything here.

docEdub:
Now implemented in https://github.com/BabylonJS/BabylonNative/pull/1146.

PR #845, opened 2021-07-23
Canvas draw image 2

CedricGuillemet:


Polyfills/Canvas/Source/Image.cpp
```diff
@@ -65,13 +79,42 @@ namespace Babylon::Polyfills::Internal
...
         auto text{value.As<Napi::String>().Utf8Value()};
-        m_src = text;
+        UrlLib::UrlRequest request{};
+        request.Open(UrlLib::UrlMethod::Get, text);
+        request.ResponseType(UrlLib::UrlResponseType::Buffer);
+        request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{ info.Env() }, this, request{std::move(request)}](arcana::expected<void, std::exception_ptr> result) {
+            if (result.has_error())
+            {
+                throw Napi::Error::New(env, result.error());
+            }
```
bghgary, line 92: This doesn't seem right. There is no way for the JS side to catch this error. Is that how the browser version works?

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -65,13 +79,42 @@ namespace Babylon::Polyfills::Internal
...
 
-    void NativeCanvasImage::SetSrc(const Napi::CallbackInfo&, const Napi::Value& value)
+    void NativeCanvasImage::SetSrc(const Napi::CallbackInfo& info, const Napi::Value& value)
     {
         auto text{value.As<Napi::String>().Utf8Value()};
-        m_src = text;
+        UrlLib::UrlRequest request{};
+        request.Open(UrlLib::UrlMethod::Get, text);
+        request.ResponseType(UrlLib::UrlResponseType::Buffer);
+        request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{ info.Env() }, this, request{std::move(request)}](arcana::expected<void, std::exception_ptr> result) {
```
bghgary, line 88: nit: spacing
```suggestion
        request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{info.Env()}, this, request{std::move(request)}](arcana::expected<void, std::exception_ptr> result) {
```

bghgary, line 88: I think you must use a cancellationToken here. If the runtime goes down before this request completes, the continuation code will probably not be happy.

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -65,13 +79,42 @@ namespace Babylon::Polyfills::Internal
...
+            m_height = m_imageContainer->m_height;
+
+            m_onloadHandlerRefs.Call({});
+        });
+    }
+
+    void NativeCanvasImage::SetOnload(const Napi::CallbackInfo&, const Napi::Value& value)
+    {
+        Napi::Function eventHandler = value.As<Napi::Function>();
+        m_onloadHandlerRefs = Napi::Persistent(eventHandler);
```
bghgary, line 113: The name for this variable is a bit odd since there is only one ref.

```suggestion
        m_onloadHandlerRef = Napi::Persistent(eventHandler);
```

Polyfills/Canvas/Source/Image.cpp
bghgary: There are a bunch of other places too.

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -65,13 +79,42 @@ namespace Babylon::Polyfills::Internal
...
+        request.ResponseType(UrlLib::UrlResponseType::Buffer);
+        request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{ info.Env() }, this, request{std::move(request)}](arcana::expected<void, std::exception_ptr> result) {
+            if (result.has_error())
+            {
+                throw Napi::Error::New(env, result.error());
+            }
+
+            Dispose();
+
+            auto buffer = request.ResponseBuffer();
```
bghgary, line 96: nit: Use brace initialization (many places)

```suggestion
            auto buffer{request.ResponseBuffer()};
```

Polyfills/Canvas/Source/Image.cpp
bghgary: Same for the second throw below.

bghgary (changes requested):

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -85,32 +86,32 @@ namespace Babylon::Polyfills::Internal
         UrlLib::UrlRequest request{};
         request.Open(UrlLib::UrlMethod::Get, text);
         request.ResponseType(UrlLib::UrlResponseType::Buffer);
-        request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{ info.Env() }, this, request{std::move(request)}](arcana::expected<void, std::exception_ptr> result) {
+        request.SendAsync().then(m_runtimeScheduler, *m_cancellationSource, [env{info.Env()}, this, request{std::move(request)}](arcana::expected<void, std::exception_ptr> result) {
             if (result.has_error())
             {
-                throw Napi::Error::New(env, result.error());
+                Napi::Error::New(env, result.error()).ThrowAsJavaScriptException();
```
bghgary, line 92: This is the same. It will still be an uncaught exception. Does canvas image not have a mechanism for errors?

Polyfills/Canvas/Source/Image.h
```diff
@@ -41,7 +41,8 @@ namespace Babylon::Polyfills::Internal
         std::string m_src{};
 
         JsRuntimeScheduler m_runtimeScheduler;
-        Napi::FunctionReference m_onloadHandlerRefs;
+        Napi::FunctionReference m_onloadHandlerRef;
+        std::shared_ptr<arcana::cancellation_source> m_cancellationSource{};
```
bghgary, line 45: Who calls cancel on this? Maybe dispose should?

bghgary (changes requested):

Polyfills/Canvas/Source/Image.cpp
CedricGuillemet: I've found nothing in the doc or in the GUI image code.

Polyfills/Canvas/Source/Image.h
CedricGuillemet: Indeed!

PR #859, opened 2021-08-03
NAPI-JSI: Actually return the result in static class method

### LINKED ISSUE #854: NativeCanvas.loadTTFAsync returns null when using NAPI-JSI
Running the Win32 validation tests using NAPI-JSI, the following line throws an error because `loadTTFAsync` returns `undefined`. Moving the test start-up code outside of the promise callback reveals everything else working, so if we fix this then we can have validation tests running on NAPI-JSI. 

https://github.com/BabylonJS/BabylonNative/blob/954a6007b1b5ef6c4f45297e44e486fbad7dd177/Apps/ValidationTests/Scripts/validation_native.js#L360

This behavior also occurs when the function call is moved into experience.js.

bghgary:
Fixes #854


rgerd (approved):
:shipit:

PR #860, opened 2021-08-03
Add Win32 JSI build to CI

rgerd:
Adds Win32 JSI build and validation testing to our CI so we can automatically test JSI usage before sending updates to BabylonReactNative.


bghgary:
I only see published results for Win32_x86 and Win32_x64. Is there supposed to be JSI version?

rgerd:
@bghgary Good catch, fixed. I'm going to hold off on checking this in until I've tracked down the cause of the teardown bug in the validation tests. Seems like it's slightly non-deterministic and not local to my setup.

bghgary:
I've ran into a crash on shutdown when I fixed the napi-jsi bug.

rgerd:
Waiting on https://github.com/BabylonJS/BabylonNative/pull/868

rickfromwork (approved):
Code looks good and new JSI tests pass.

PR #863, opened 2021-08-05
Fix Dynamic texture flag for Canvas

CedricGuillemet:
Fix bug reported by @Alex-MSFT 
- Enable blit emulation for OpenGL (blit not natively supported on Android Emulator. Good catch Alex!)
- Fix BLIT flag for textures. Texture was recreated each frame and made the blit to fail when updating the canvas each frame.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.h
+++ b/Plugins/NativeEngine/Source/NativeEngine.h
@@ -42,6 +42,9 @@ namespace Babylon
         uint32_t Width{0};
         uint32_t Height{0};
         uint32_t Flags{0};
+        // CreationFlags contains flags used at texture creation
+        // regarding BLIT support and READBACK
+        uint64_t CreationFlags{0};
         uint8_t AnisotropicLevel{0};
     };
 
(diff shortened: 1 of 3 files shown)


Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -42,6 +42,9 @@ namespace Babylon
         uint32_t Width{0};
         uint32_t Height{0};
         uint32_t Flags{0};
+        // CreationFlags contains flags used at texture creation
+        // regarding BLIT support and READBACK
+        uint64_t CreationFlags{0};
```
ryantrem, line 47: Why is this new member needed? The logic looks identical, does Flags get overwritten at some point or something?

ryantrem (approved):

Plugins/NativeEngine/Source/NativeEngine.h
CedricGuillemet: `Flags` is used to texture properties when used for rendering (sampling mode, wrapping, anisotropic, ...) and changed by JS.
`CreationFlags` is the caps for the texture: will be used for blit, readback, is RT,... and should only change if underlying bgfx texture is destroyed and recreated.

Alex-MSFT (approved):

PR #864, opened 2021-08-05
Replace C++ GoogleTest tests with JS tests using Mocha and Chai

darraghjburke:
Removes GoogleTest dependency and C++ tests for UrlLib. In their place, adds basic Javascript tests for XMLHTTPRequest using Mocha.js and Chai.js.

Includes several hacks that should be cleaned up, marked with TODOs.

Creates base framework for #802.

### DIFF
--- a/Apps/UnitTests/Scripts/tests.js
+++ b/Apps/UnitTests/Scripts/tests.js
@@ -2,24 +2,24 @@
 function BabylonReporter(runner) {
     const stats = runner.stats;
 
-    runner.on("pass", function (test) {
+    runner.on("pass", test => {
         console.log(`Passed: ${test.fullTitle()}`);
     });
 
-    runner.on("fail", function (test, err) {
+    runner.on("fail", (test, err) => {
         console.log(`Failed: ${test.fullTitle()} with error: ${err.message}`);
     });
 
-    runner.on("end", function () {
+    runner.on("end", () => {
         console.log(`Total passed: ${stats.passes}/${stats.tests} tests`);
     });
 }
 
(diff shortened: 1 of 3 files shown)


bghgary (approved):
Looks okay overall. Obviously lots of TODOs still, but that's okay with me.

Apps/UnitTests/Scripts/tests.js
```diff
@@ -31,21 +31,21 @@ const expect = chai.expect;
...
-        const xhr = await makeXHR("GET", "https://babylonjs.com/invalid");
+        const xhr = await createRequest("GET", "https://babylonjs.com/invalid");
         expect(xhr.status).to.equal(404);
     })
 
 });
 
 
-mocha.run(function (failures) {
+mocha.run( failures => {
```
bghgary, line 49: nit: there typically isn't a space here.  might consider auto formatting the whole file.
```suggestion
mocha.run(failures => {
```

darraghjburke: fixed! thanks! would it be possible to add a linting step to the CI checks?

bghgary: We could. We probably should transition this to a TS file before doing that though.

darraghjburke: What should we use to compile TS files? As far as I know we are not using TS anywhere in the Babylon Native repo yet

bghgary: I'm not sure yet. Something similar to how Babylon.js does it probably.

PR #866, opened 2021-08-10
Allow specifying font size for fonts in Canvas Context Polyfill

Alex-MSFT:
This PR improves the argument parsing when setting the font for Canvas based text rendering to allow setting font size as well as font family name.  Instead of just assuming that the passed in argument for font only contains a font family name, we split the font options as a space delimited list of options, and parse each argument separately.   We maintain the same default font size of 16 when the argument is unspecified or invalid.  In the future we should likely also include support specifying for font styles as well.

For example specifying the font:
"72px droidsans" will match to a font size of 72 pixels and the font name "droidsans".

Below are some example screenshots of rendering at different font sizes in the playground.

12 px:

24 px:

36 px:

### DIFF
--- a/Polyfills/Canvas/Source/Context.cpp
+++ b/Polyfills/Canvas/Source/Context.cpp
@@ -2,7 +2,7 @@
 #include <map>
 #include <algorithm>
 #include <assert.h>
-#include <sstream>
+#include <regex>
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wpedantic"
 #include "nanovg/nanovg.h"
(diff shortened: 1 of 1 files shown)


Polyfills/Canvas/Source/Context.cpp
```diff
@@ -650,15 +651,37 @@ namespace Babylon::Polyfills::Internal
...
-        const auto fontName = value.ToString();
-        if (m_fonts.find(fontName) == m_fonts.end())
+        // Default font id, and font size values.
+        // TODO: Determine better way of signaling to user that font specified is invalid.
+        m_currentFontId = -1;
+        int fontSize = 16;
+
+        // Parse user specified font style to set font ID and font size if specified.
+        std::istringstream fontOptionsStream(value.ToString());
+        std::string fontOption;
```
ryantrem, line 661: Wouldn't it be easier (and more scalable to additional things like style as you mentioned) if it just used a regex?

Alex-MSFT: Good point, will look into using regex for pulling out arguments instead.

Alex-MSFT: Updated the PR to use regex for parsing tokens.  First time using C++ regex, so let me know if there's something smarter I could be doing instead.

ryantrem (approved):

PR #868, opened 2021-08-13
Update v8jsi nuget package to .23

rgerd:
Updating for a fix to an inspector teardown crash, commenting out recast.js since there seems to be a v8 bug that hangs the runtime when attaching a JS debugger.


Apps/Playground/Win32/App.cpp
```diff
@@ -150,7 +150,7 @@ namespace
         Babylon::ScriptLoader loader{*runtime};
         loader.Eval("document = {}", "");
         loader.LoadScript("app:///Scripts/ammo.js");
-        loader.LoadScript("app:///Scripts/recast.js");
+        // loader.LoadScript("app:///Scripts/recast.js");
```
chrisfromwork, line 153: please add a comment here for why this is commented out.

chrisfromwork (approved):

PR #872, opened 2021-08-19
XHR: Make open() throw synchronous errors for bad URLs

### LINKED ISSUE #759: Loading resources over HTTP should produce informative, consistent exceptions
Issue: we don't currently know what passing an invalid URL to XMLHTTPRequest will do across different platforms, and we're not sure what the correct behavior should be. We probably want to throw a JavaScript exception, but should that be thrown in the Send() method or the Open() method?

Things to consider:

- How closely should we match behavior on the web? [The spec does not explicitly specify where exceptions should be thrown](https://xhr.spec.whatwg.org/#the-open()-method), but typically browsers will throw DOMExceptions on open() if the path is invalid.
- Is it even possible to throw exceptions from open() on all platforms? That would require that the platform allows us to parse the URL as a separate action from the network request, which may not be the case for all platforms. (it is at least true for Android)
- What are we currently doing on Windows and iOS?

The solution here is to decide on a contract and ensure that it's implemented consistently across all platforms. It would be nice to have unit tests for these exceptions as well.

Feel free to add any further comments @bghgary

darraghjburke:
Makes UrlLib do URL parsing in open() instead of send(). It will throw an error if there's an issue. We'll catch it in XMLHttpRequest and rethrow to JavaScript. This matches the behavior of browsers.

Implemented on all platforms. Fixes #759 .

Also adds unit tests for XHR, and adds the unit tests to the CI pipeline for Win32, MacOS, and Ubuntu. Unit Tests are still not supported on Android/iOS/UWP.

Note: if you want to run the unit tests app, make sure to run `npm install` again to install Mocha and Chai.


Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -25,6 +25,92 @@ namespace UrlLib
...
+            if (m_curl)
+            {
+                curl_easy_reset(m_curl);
+            } else
+            {
+                m_curl = curl_easy_init();
+            }
+            if (m_curl)
+            {
+
```
bghgary, line 37: nit: extra line

Apps/UnitTests/Shared/Tests.cpp
```diff
@@ -15,10 +14,10 @@
 std::unique_ptr<Babylon::AppRuntime> runtime{};
 std::unique_ptr<Babylon::Graphics> graphics{};
 
-int main()
+int run()
 {
-    Babylon::ContextConfiguration graphicsConfig{};
-    graphics = Babylon::Graphics::CreateGraphics(graphicsConfig);
+   Babylon::ContextConfiguration graphicsConfig{};
+   graphics = Babylon::Graphics::CreateGraphics(graphicsConfig);
```
bghgary, line 20: weird spacing changes?

Dependencies/UrlLib/Source/Apple/UrlRequest.mm
```diff
@@ -27,6 +27,23 @@ void Open(UrlMethod method, std::string url)
...
+            NSString* urlString = [[NSString stringWithUTF8String:m_url.data()] stringByAddingPercentEncodingWithAllowedCharacters:NSCharacterSet.URLQueryAllowedCharacterSet];
+            m_nsURL = {[NSURL URLWithString:urlString]};
+            NSString* scheme{m_nsURL.scheme};
+            if ([scheme isEqual:@"app"])
+            {
+                NSString* path{[[NSBundle mainBundle] pathForResource:m_nsURL.path ofType:nil]};
+                if (path == nil)
+                {
+                    throw std::runtime_error{"Local path is invalid"};
+
```
bghgary, line 39: nit: extra line

Dependencies/UrlLib/Source/Apple/UrlRequest.mm
```diff
@@ -27,6 +27,23 @@ void Open(UrlMethod method, std::string url)
         {
             m_method = method;
             m_url = std::move(url);
```
bghgary, line 29: Do you need `m_url` anymore?

darraghjburke: Removed, thanks

Apps/UnitTests/Shared/Tests.cpp
darraghjburke: Not sure why this looks weird on GitHub - if you open the file, the spacing seems okay, unless I am missing something

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -178,25 +178,39 @@ namespace Babylon::Polyfills::Internal
...
     }
 
     void XMLHttpRequest::Send(const Napi::CallbackInfo& info)
     {
-        m_request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{info.Env()}, this](arcana::expected<void, std::exception_ptr> result) {
-            if (result.has_error())
+        if (m_readyState != ReadyState::Opened)
+        {
+            Napi::Error::New(info.Env(), "XMLHttpRequest must be opened before it can be sent").ThrowAsJavaScriptException();
+            return;
```
bghgary, line 198: let's try to be consistent with this

```suggestion
            throw Napi::Error::New(info.Env(), "XMLHttpRequest must be opened before it can be sent");
```

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -178,25 +178,39 @@ namespace Babylon::Polyfills::Internal
...
+        }
+        m_request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{info.Env()}, this](arcana::expected<void, std::exception_ptr> result)
             {
-                throw Napi::Error::New(env, result.error());
-            }
+                if (result.has_error())
+                {
+                    Napi::Error::New(env, result.error()).ThrowAsJavaScriptException();
+                }
```
bghgary, line 205: Right now, if urllib throws an exception, this will throw an uncatchable exception on the JavaScript side and the XHR will complete (i think successfully?). Is that what we want?

```suggestion
                if (result.has_error())
                {
                    Napi::Error::New(env, result.error()).ThrowAsJavaScriptException();
                }
```

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -178,25 +178,39 @@ namespace Babylon::Polyfills::Internal
...
     {
-        m_request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{info.Env()}, this](arcana::expected<void, std::exception_ptr> result) {
-            if (result.has_error())
+        if (m_readyState != ReadyState::Opened)
+        {
+            Napi::Error::New(info.Env(), "XMLHttpRequest must be opened before it can be sent").ThrowAsJavaScriptException();
+            return;
+        }
+        m_request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{info.Env()}, this](arcana::expected<void, std::exception_ptr> result)
             {
```
bghgary, line 201: Please use clang format to format this file. I believe clang format puts the brace on the same line for lambdas.

bghgary (changes requested):
Mostly good. Just some smallish comments that we should probably address before merging.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -186,7 +186,11 @@ namespace Babylon::Polyfills::Internal
         catch (const std::exception& e)
         {
             // If we have a parse error, catch and rethrow to JavaScript
-            throw Napi::Error::New(info.Env(), "Could not parse URL scheme: " + std::string{e.what()});
+            throw Napi::Error::New(info.Env(), "Error parsing URL scheme: " + std::string{e.what()});
```
bghgary, line 189: nit: this is probably slightly cheaper
```suggestion
            throw Napi::Error::New(info.Env(), std::string{"Error parsing URL scheme: "} + e.what());
```

bghgary (approved):

Dependencies/UrlLib/Source/Unix/UrlRequest.cpp
```diff
@@ -60,16 +60,16 @@ namespace UrlLib
                     // So, escaping only path part of the URL for every URL but file scheme.
 
                     CURLUcode rc;
-                    CURLU* url = curl_url();
-                    auto urlScopeGuard = gsl::finally([url] { curl_url_cleanup(url); });
-                    rc = curl_url_set(url, CURLUPART_URL, url.c_str(), 0);
+                    CURLU* cURL_url = curl_url();
```
bghgary, line 63: nit: this is a weird name. it's mixed style (camel and snake case).
```suggestion
                    CURLU* urlObject = curl_url();
```

bghgary: Or
```suggestion
                    CURLU* curlUrl = curl_url();
```

Issue #782, opened 2021-06-07
Unhandled exception handler in WorkQueue doesn't handle exceptions inside arcana taskchain continuations


darraghjburke:
**Describe the bug**
When an exception happens inside of an arcana task chain continuation, it is never caught by the WorkQueue's unhandled exception handler.

See https://github.com/BabylonJS/BabylonNative/blob/master/Core/AppRuntime/Source/WorkQueue.h#L30

**To Reproduce**
1. Throw an exception from inside an arcana task chain continuation
2. Place a breakpoint in Workqueue.h at line 30
3. Observe that the program will never reach your breakpoint

**Expected behavior**
An exception that is thrown inside an arcana task continuation and not handled by another should wind up in the WorkQueue's unhandled exception handler.

PR #882, opened 2021-09-14
Simplify Work Queue

syntheticmagus:
Attempt to simplify work queue to make handling errors a bit easier.

### DIFF
--- a/Core/AppRuntime/Source/WorkQueue.h
+++ b/Core/AppRuntime/Source/WorkQueue.h
@@ -19,17 +19,19 @@ namespace Babylon
         template<typename CallableT>
         void Append(CallableT callable)
         {
-            std::scoped_lock lock{m_appendMutex};
-            m_task = m_task.then(m_dispatcher, m_cancelSource, [this, callable = std::move(callable)]() mutable noexcept {
-                try
-                {
-                    callable(m_env.value());
-                }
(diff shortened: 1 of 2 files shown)


syntheticmagus:
For #782

Core/AppRuntime/Source/WorkQueue.h
```diff
@@ -19,32 +19,44 @@ namespace Babylon
...
-                try
-                {
-                    callable(m_env.value());
-                }
-                catch (...)
-                {
-                    m_unhandledExceptionHandler(std::current_exception());
-                }
-            });
+            if constexpr (std::is_copy_constructible<CallableT>::value)
```
bghgary, line 22: I don't understand this. Why do you need to copy?

syntheticmagus: The callable to be queued needs to be copyable in order for the dispatcher to make a copyable inplace function from it. I didn't dive all the way into why the work queue needs its inplace functions to be copyable; I feel like we've been down that road before, but I don't remember the particulars. What I encountered was that the compiler complained if I tried to put noncopyable callables onto the dispatcher directly, so I worked around it for scenarios where the callable has no copy constructor but does have a move constructor (for example, if the copyable captured a `unique_ptr`).

PR #884, opened 2021-09-16
ChromeDevTools for Android

### LINKED ISSUE #686: Test ChromeDevTools Android Support
Once #607 is merged, we can test whether it works on Android V8. This should be fairly straightforward (just run it and see what happens). There's nothing that inherently excludes the possibility of Android as long as the TCP/HTTP/WebSocket server doesn't conflict with any weird firewall policies.

One thing to note is the v8inspector dependency depends on a submodule `llhttp` that requires node to build, as it uses TypeScript-based code generation. We currently have the inclusion of the v8inspector dependency gated behind `WIN32` because we only download the windows version of node, so the build would break if we ran it on macOS or Linux. Part of testing Android should be making the node download work on all supported build platforms, which involves using the right download URL and running the right command.

CedricGuillemet:
fixes #686 

- Replaced `set_win32_arch` with `set_cpu_pLatform_arch` to expose building platform & arch
- node download/build for llhttp support for Mac and Linux
- replaced wstring with basic_string<char16_t> because wchar_t is platform dependant and not 16bits long on Android
- Disabled warnings for `v8inspector` because some API is deprecated with Clang
- Removed `OBJECT` from add_library for llhttp


Dependencies/CMakeExtensions/CMakeLists.txt
```diff
@@ -110,16 +110,29 @@ function(disable_warnings target)
     endif()
 endfunction()
 
-# Sets the WIN32_ARCH variable in the calling scope.
-function(set_win32_arch)
+# Sets CPU_ARCH and PLATFORM_ARCH variables in the calling scope.
+function(set_cpu_pLatform_arch)
```
bghgary, line 114:
```suggestion
function(set_cpu_platform_arch)
```

bghgary (approved):

rgerd (approved):
Glad to see CDT support on more than just windows!
:shipit:

PR #886, opened 2021-09-17
Improve sampler mode and anisotropy translation

bghgary:
Fixes the Babylon Native side for https://github.com/babylonjs/babylonreactnative/issues/256 (see [this comment](https://github.com/BabylonJS/BabylonReactNative/issues/256#issuecomment-921364716) for more context)

There are a few texture settings that will not translate into bgfx properly:
1. Texture sampling modes that ignore mips (e.g. NEAREST_NEAREST or LINEAR_LINEAR). Bgfx does not support this. This change uses MIP_POINT even though it's not technically correct. It was using MIP_LINEAR which is even worse.
2. Anisotropic levels are not supported by bgfx. Bgfx will either set no anisotropy or max anisotropy.


CedricGuillemet (approved):
LGTM!

syntheticmagus (approved):

PR #890, opened 2021-09-22
Update xcode and ios deployment targets

CedricGuillemet:


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -236,11 +236,14 @@ - (void) UnlockMeshes {
 }
 
 /**
- Returns the orientation of the app based on the current status bar orientation.
+ Returns the orientation of the app
 */
 - (UIInterfaceOrientation)orientation {
     UIApplication* sharedApplication = [UIApplication sharedApplication];
-#if (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_13_0)
+#if (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_15_0)
```
ryantrem, line 243: If I recall, we had to set this up this way because in iOS 13, `[sharedApplication statusBarOrientation]` was deprecated and could produce an error. I guess we don't have the same problem with 15... e.g. `[[[[sharedApplication windows] firstObject] windowScene] interfaceOrientation]` is not deprecated in 15, it's just that the new way of doing this is better somehow?

CedricGuillemet: `[[[[sharedApplication windows] firstObject] windowScene] interfaceOrientation]` is deprecated with iOS15

ryantrem: Should there be a similar runtime check to use the new API when rubbing on iOS 15, or should the patent be different than the others for this API?

CedricGuillemet: no, a deployment target is the minimal iOS version the program will run on. a binary with a deployment target iOS13 will run on iOS13, 14, 15. you can't do runtime checks as some iOS15 API do not exist in iOS13. We can avoid for now targetting iOS15 but in some months/years it will be the minimum required you can target with a new version of Xcode. And, for some reasons, a BN user may want to target to iOS15+ because of some needed api.

PR #891, opened 2021-09-23
Update bgfx

CedricGuillemet:
- update bgfx.cmake, bgfx, bimg, bx. module revision was 7 moths old
- add subrepo for stb as stb_image_resize has been removed from bgfx

### DIFF
--- a/Plugins/NativeEngine/CMakeLists.txt
+++ b/Plugins/NativeEngine/CMakeLists.txt
@@ -27,7 +27,8 @@ set(SOURCES
 
 add_library(NativeEngine ${SOURCES})
 
-target_include_directories(NativeEngine PUBLIC "Include")
+target_include_directories(NativeEngine PUBLIC "Include"
+    PRIVATE "${CMAKE_SOURCE_DIR}/Dependencies") # for stb
 
 target_link_to_dependencies(NativeEngine
     PUBLIC JsRuntime
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "Dependencies/llhttp"]
 	path = Dependencies/v8inspector/Dependencies/llhttp
 	url = https://github.com/nodejs/llhttp
+[submodule "Dependencies/stb"]
+	path = Dependencies/stb
+	url = https://github.com/nothings/stb.git


Apps/ValidationTests/Android/app/CMakeLists.txt
```diff
@@ -65,7 +65,9 @@ target_link_libraries(BabylonNativeJNI
         Window
         ScriptLoader
         bgfx
-        XMLHttpRequest)
+        XMLHttpRequest
+        bx
+        bimg)
```
bghgary, line 70: why do you need this now but not before?

Plugins/NativeEngine/CMakeLists.txt
```diff
@@ -27,7 +27,8 @@ set(SOURCES
 
 add_library(NativeEngine ${SOURCES})
 
-target_include_directories(NativeEngine PUBLIC "Include")
+target_include_directories(NativeEngine PUBLIC "Include"
+    PRIVATE "${BIMG_DIR}/3rdparty") # for stb
```
bghgary, line 31: why do you need this now but not before?

bghgary (approved):

Apps/ValidationTests/Android/app/CMakeLists.txt
CedricGuillemet: I guess these dependencies were not private before. I've merged https://github.com/bkaradzic/bgfx.cmake for latest updates on 3rd party with bimg. It must have changed at that time.

CedricGuillemet:
@ryantrem Do you need to update BabylonReactNative in the coming days? As this is a quite big update, I'd like a few days to be sure nothing is broken.

ryantrem:
> @ryantrem Do you need to update BabylonReactNative in the coming days? As this is a quite big update, I'd like a few days to be sure nothing is broken.

Are you asking if we can hold off on changes to BabylonReactNative for a few days while you test this change with BabylonReactNative as well?

CedricGuillemet:
@ryantrem yes, or I can wait merging until you don't need any update in BRN. I want to avoid merging this PR, getting it integrated in BRN and having to do a fix in emergency because of something we could detect in BN but didn't.

PR #896, opened 2021-09-24
Fix Resize for Apple

### LINKED ISSUE #877: Playground app crashes on MacOS when resizing window
**For questions, please use the [forum](https://forum.babylonjs.com)**.

**Describe the bug**
When resizing the playground window on MacOS, the app fails the following assertion:
_MTLValidateResolveTexture:281: failed assertion `MTLRenderPassDescriptor texture and resolveTexture size mismatch.`

**To Reproduce**
1. Build the playground on MacOS
2. Resize the window by dragging from the corner or maximizing the window
3. An assertion will fail

**Expected behavior**
You should be able to resize the window, and the canvas should resize to match the new dimensions. This is what happens on Windows (although the resize only happens when dragging ends)

**Screenshots**
Stack trace:
image
**Other**
 - Platform: MacOS

CedricGuillemet:
fixes #877 

This bug has 2 parts but are both related to a 1 frame delay of resize infos propagation.

1)

`StartRenderingCurrentFrame` and `FinishRenderingCurrentFrame` are called once per frame but do not concern the same frame. Frame is rendered using previous back buffer size and a new one is started just after. Scisor infos in bgfx were out of sync leading to an assert in Metal. 
This one doesn't concern BabylonReactNative:
https://github.com/BabylonJS/BabylonReactNative/blob/eea89150682a28155811d671666dfe553bf34d61/Modules/%40babylonjs/react-native/shared/BabylonNative.cpp#L111
OpenGL and D3D are more tolerant to view rect different from backbuffer dimension. Metal and Vulkan are more strict.

2)

In the next frame rendering after a resize, the view rect was using `bgfx::getStats()` to get the dimension. As the rendering/init didn't occur yet, it returned the former size. Because the viewRect is different from the backBuffer dimension, `clearWithRenderPass` is false:
https://github.com/bkaradzic/bgfx/blob/ad8e7eb974c943757c804f121ca52f60048d129c/src/renderer_mtl.mm#L3906
And the attachment store action is not set here:

https://github.com/bkaradzic/bgfx/blob/ad8e7eb974c943757c804f121ca52f60048d129c/src/renderer_mtl.mm#L3955

When this happens, it does nothing when MSAA is off. But assert and crashes with MSAA.
I think setting a storeAction here:
https://github.com/bkaradzic/bgfx/blob/ad8e7eb974c943757c804f121ca52f60048d129c/src/renderer_mtl.mm#L4002
is needed and I'll do a PR in bgfx but anyways, the size provided to bgfx is wrong. 
Before this PR, when MSAA is off and window is resized, there were some rendering artifacts in the border. This this PR, the size is correct and there is no more artifact.

### DIFF
--- a/Apps/Playground/iOS/LibNativeBridge.mm
+++ b/Apps/Playground/iOS/LibNativeBridge.mm
@@ -46,7 +46,6 @@ bool g_isXrActive{};
     graphicsConfig.Width = static_cast<size_t>(width);
     graphicsConfig.Height = static_cast<size_t>(height);
     graphics = Babylon::Graphics::CreateGraphics(graphicsConfig);
-    graphics->StartRenderingCurrentFrame();
     runtime = std::make_unique<Babylon::AppRuntime>();
     inputBuffer = std::make_unique<InputManager<Babylon::AppRuntime>::InputBuffer>(*runtime);
 
@@ -98,8 +97,8 @@ bool g_isXrActive{};
 {
     if (graphics)
     {
-        graphics->FinishRenderingCurrentFrame();
         graphics->StartRenderingCurrentFrame();
+        graphics->FinishRenderingCurrentFrame();
     }
 }
 
--- a/Apps/Playground/macOS/ViewController.mm
+++ b/Apps/Playground/macOS/ViewController.mm
@@ -32,8 +32,8 @@ std::unique_ptr<InputManager<Babylon::AppRuntime>::InputBuffer> inputBuffer{};
 - (void)drawInMTKView:(MTKView *)__unused view
 {
     if (graphics) {
-        graphics->FinishRenderingCurrentFrame();
         graphics->StartRenderingCurrentFrame();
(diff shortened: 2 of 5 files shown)


darraghjburke:
This fixes the issue for me on MacOS, but it seems like there's a minor type mismatch which is causing the Windows build to fail

Apps/Playground/iOS/LibNativeBridge.mm
```diff
@@ -98,8 +97,8 @@ - (void)render
 {
     if (graphics)
     {
-        graphics->FinishRenderingCurrentFrame();
         graphics->StartRenderingCurrentFrame();
+        graphics->FinishRenderingCurrentFrame();
```
bghgary, line 101: Note that this will change the behavior of rendering a bit. The idea with calling in this order (Finish then Start) is that Finish will block and thus we want to give the app the most amount of time to do other things. @syntheticmagus Any thoughts?

PR #898, opened 2021-09-24
Move XHR URL encoding to JavaScript

darraghjburke:
This PR removes platform-specific URI encoding in UrlLib, and replaces it with a call to encodeURI() from the XMLHttpRequest polyfill. This helps ensure consistency between platforms and allows us to remove the platform-specific encoding code for Apple and Unix.

### DIFF
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
@@ -183,7 +183,10 @@ namespace Babylon::Polyfills::Internal
             auto inputURL{info[1].As<Napi::String>()};
             auto decodedURL{info.Env().Global().Get("decodeURI").As<Napi::Function>().Call({inputURL}).As<Napi::String>()};
             auto encodedURL{inputURL};
-            if (inputURL.StrictEquals(decodedURL)) {
+            // We need to percent encode the URL
+            // If decoding the URL changes something, it's already encoded - don't change anything
+            if (inputURL.StrictEquals(decodedURL))
+            {
                 encodedURL = info.Env().Global().Get("encodeURI").As<Napi::Function>().Call({inputURL}).As<Napi::String>();
             }
             m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), encodedURL.Utf8Value());
--- a/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
+++ b/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
@@ -180,7 +180,13 @@ namespace Babylon::Polyfills::Internal
     {
         try
         {
-            m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), info[1].As<Napi::String>().Utf8Value());
+            auto inputURL{info[1].As<Napi::String>()};
+            auto decodedURL{info.Env().Global().Get("decodeURI").As<Napi::Function>().Call({inputURL}).As<Napi::String>()};
+            auto encodedURL{inputURL};
+            if (inputURL.StrictEquals(decodedURL)) {
+                encodedURL = info.Env().Global().Get("encodeURI").As<Napi::Function>().Call({inputURL}).As<Napi::String>();
+            }
+            m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), encodedURL.Utf8Value());
             SetReadyState(ReadyState::Opened);
         }
         catch (const std::exception& e)
--- a/Apps/UnitTests/Scripts/tests.js
+++ b/Apps/UnitTests/Scripts/tests.js
@@ -39,11 +39,11 @@ describe("XMLHTTPRequest", function () {
         expect(xhr.status).to.equal(200);
     })
     it("should load URLs with escaped spaces", async function () {
-        const xhr = await createRequest("GET", "https://github.com/BabylonJS/Assets/raw/master/meshes/aerobatic%20plane.glb");
+        const xhr = await createRequest("GET", "https://powermrblobtest.blob.core.windows.net/modles/Airplane%20Galaxy.glb");
         expect(xhr.status).to.equal(200);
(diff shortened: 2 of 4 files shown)


CedricGuillemet:
More or less related to this PR, Unittest can run on Linux? We should enable them on the CI. IIRC VT can run on iOS/MacOS and Android, did you test all platforms?

darraghjburke:
> More or less related to this PR, Unittest can run on Linux? We should enable them on the CI. IIRC VT can run on iOS/MacOS and Android, did you test all platforms?

Unix Unit Tests are already running on the CI :) https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=15632&view=logs&j=22373658-a8df-585e-20b1-bc76ffa2e424&t=120303ba-ca4a-5c0f-2efd-ab7bfff33769

I have not manually run the Validation Tests yet, I will do that this morning.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -180,7 +180,16 @@ namespace Babylon::Polyfills::Internal
...
-            m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), info[1].As<Napi::String>().Utf8Value());
+            auto inputURL{info[1].As<Napi::String>()};
+            auto decodedURL{info.Env().Global().Get("decodeURI").As<Napi::Function>().Call({inputURL}).As<Napi::String>()};
+            auto encodedURL{inputURL};
+            // We need to percent encode the URL
+            // If decoding the URL changes something, it's already encoded - don't change anything
+            if (inputURL.StrictEquals(decodedURL))
+            {
+                encodedURL = info.Env().Global().Get("encodeURI").As<Napi::Function>().Call({inputURL}).As<Napi::String>();
+            }
```
bghgary, line 191: I'm not sure I follow this logic. What happens if the URL has both encoded spaces and not encoded spaces?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -180,7 +180,16 @@ namespace Babylon::Polyfills::Internal
...
+            m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), encodedURL.Utf8Value());
```
bghgary, line 192: Maybe store `encodedURL` as a `Napi::Value` so that you can factor out the `.As<Napi::String>` also.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
darraghjburke: If we want to handle that case (a partially-encoded URL), we will need to write our own custom encoding function. We can do this, but then we are responsible for handling all edge cases. encodeURI is standardized.

I'm not sure how likely it is that we would get a string like this. I'd be surprised if someone encoded only part of their URL, but I guess it is possible. (Maybe they concatenate "some%20encoded%20url" + "some non-encoded string".)

If we need to support this case, I can write a function in XMLHttpRequest to do this. At least it would be cross-platform that way.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
```diff
@@ -46,3 +46,6 @@ namespace Babylon::Polyfills::Internal
         std::unordered_map<std::string, std::vector<Napi::FunctionReference>> m_eventHandlerRefs;
     };
 }
+
+bool IsHexChar(const char& c);
```
darraghjburke, line 50: Is this the right way to add little helper functions that are purely for internal use?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -201,20 +209,21 @@ namespace Babylon::Polyfills::Internal
             throw Napi::Error::New(info.Env(), "XMLHttpRequest must be opened before it can be sent");
             return;
         }
-        m_request.SendAsync().then(m_runtimeScheduler, arcana::cancellation::none(), [env{info.Env()}, this](arcana::expected<void, std::exception_ptr> result) {
```
darraghjburke, line 204: 204-226 are just the autoformatter

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -180,7 +181,14 @@ namespace Babylon::Polyfills::Internal
     {
         try
         {
-            m_request.Open(MethodType::StringToEnum(info[0].As<Napi::String>().Utf8Value()), info[1].As<Napi::String>().Utf8Value());
+            auto inputURL{info[1].As<Napi::String>()};
+            // If the input URL contains any true % characters, encode them as %25
+            auto encodedPercentURL{Napi::String::New(info.Env(), EncodePercent(inputURL.Utf8Value()))};
```
darraghjburke, line 186: I realized that the only way to safely encode a URL is to start from a fully decoded one. Otherwise, you run the risk of re-encoding "%20" as "%2520.". To implement this, I came up with this solution:
1. If there's a standalone % character in the URL, encode it as %25.
2. Use decodeURI to get a truly decoded URL. This URL will have *nothing* percent encoded. If a % exists, it's because there's a real % character.
3. Finally, use encodeURI to get an encoded URL.

This method still primarily uses decodeURI/encodeURI, which are standards. The only custom code is for the first step, to turn % into %25 when it's not followed by hex characters. I think since this function is so simple, the risk of bugs is fairly low.

I added several tests which represent pretty much the worst possible URLs that I could imagine, and this solution handles them correctly, so it seems pretty safe.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.h
bghgary: Nope :-) If you put these functions in the header file, then they are exposed to anyone who includes them. What you should do instead is put the implementation functions at the top of the top inside the anonymous namespace.

PR #899, opened 2021-09-24
Fix Android URL Requests

darraghjburke:
Fixes loading files from URLs on Android. Load from the full URL string rather than just the path.

### DIFF
--- a/Dependencies/UrlLib/Source/Android/UrlRequest.cpp
+++ b/Dependencies/UrlLib/Source/Android/UrlRequest.cpp
@@ -45,7 +45,8 @@ namespace UrlLib
         void Open(UrlMethod method, std::string url)
         {
             m_method = method;
-            Uri uri{Uri::Parse(std::move(url).data())};
+            m_url = std::move(url);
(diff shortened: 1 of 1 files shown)


Dependencies/UrlLib/Source/Android/UrlRequest.cpp
```diff
@@ -45,7 +45,8 @@ namespace UrlLib
         void Open(UrlMethod method, std::string url)
         {
             m_method = method;
-            Uri uri{Uri::Parse(std::move(url).data())};
+            m_url = std::move(url);
+            Uri uri{Uri::Parse(m_url.data())};
             m_schemeIsApp = ((std::string)uri.getScheme() == "app");
             m_urlOrAppPath = uri.getPath();
```
bghgary, line 51:
```suggestion
            if ((std::string)uri.getScheme() == "app")
            {
                m_schemeIsApp = true;
                m_urlOrAppPath = uri.getPath();
            }
            else
            {
                m_urlOrAppPath = std::move(url);
            }
```

and revert the other stuff

bghgary (approved):

PR #901, opened 2021-09-28
Fix Apple Memory Management Bug in UrlLib

darraghjburke:
~~Currently, we are storing a pointer to an NSURL object when we call Open(), and then creating the URLRequest in SendAsync(). The problem is that it's possible for the underlying object to be released in between these calls. This fix simply increments the refCount on Open() and then decrements it when the object is destroyed or when Open() is called again.~~

Makes UrlLib use Automated Reference Counting: https://en.wikipedia.org/wiki/Automatic_Reference_Counting

This way, we no longer have to manually retain/release objects.

Also, stops storing the URL from Open() as a member variable unless it's a valid URL. If it's a bad URL, we should not be retaining it.

### DIFF
--- a/Dependencies/UrlLib/Source/Apple/UrlRequest.mm
+++ b/Dependencies/UrlLib/Source/Apple/UrlRequest.mm
@@ -1,7 +1,5 @@
 #if ! __has_feature(objc_arc)
-
 #error "ARC is off"
-
 #endif
 
 #include <UrlLib/UrlLib.h>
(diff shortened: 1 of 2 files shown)


bghgary (approved):
LGTM. @ryantrem Can you check this? I'm not as familiar with Obj-C

Dependencies/UrlLib/Source/Apple/UrlRequest.mm
```diff
@@ -1,3 +1,9 @@
+#if ! __has_feature(objc_arc)
+
+#error "ARC is off"
+
+#endif
```
ryantrem, line 5:
```suggestion
#if ! __has_feature(objc_arc)
#error "ARC is off"
#endif
```

ryantrem (approved):

PR #903, opened 2021-09-30
Add console logging to the Playground App on Apple Devices

darraghjburke:
Initializes the Console polyfill on MacOS+iOS and uses NSLog to print the messages.


ryantrem (approved):

bghgary (approved):
Ah, that's what was missing.

PolygonalSun (approved):

PR #908, opened 2021-10-05
testUtils plugin

CedricGuillemet:
Moving testUtils from a single header into its own plugin.
I'll extend this plugin with bgfx statistics in an upcoming PR.


Apps/ValidationTests/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -103,7 +103,7 @@ extern "C"
 
                 Babylon::Polyfills::XMLHttpRequest::Initialize(env);
 
-                Babylon::TestUtils::CreateInstance(env, window);
+                 Babylon::Plugins::TestUtils::Initialize(env, window);
```
darraghjburke, line 106: nit: extra space

PR #909, opened 2021-10-06
Update NativeXR's PointerEvent to store the pointer type, like on browser

rickfromwork:
Now babylonNative apps can properly take advantage of the pointerType value on pointer events.

### DIFF
--- a/Plugins/NativeXr/Source/NativeXr.cpp
+++ b/Plugins/NativeXr/Source/NativeXr.cpp
@@ -814,7 +814,8 @@ namespace Babylon
                     env,
                     JS_CLASS_NAME,
                     {
-                        InstanceAccessor("pointerId", &PointerEvent::GetPointerId, nullptr)
+                        InstanceAccessor("pointerId", &PointerEvent::GetPointerId, nullptr),
+                        InstanceAccessor("pointerType", &PointerEvent::GetPointerType, nullptr)
                     });
 
                 env.Global().Set(JS_CLASS_NAME, func);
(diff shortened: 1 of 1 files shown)


ryantrem (approved):

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -814,7 +814,8 @@ namespace Babylon
                     env,
                     JS_CLASS_NAME,
                     {
-                        InstanceAccessor("pointerId", &PointerEvent::GetPointerId, nullptr)
+                        InstanceAccessor("pointerId", &PointerEvent::GetPointerId, nullptr),
+                        InstanceAccessor("pointerType", &PointerEvent::GetPointerType, nullptr)
```
rgerd, line 818: Given that this object will receive its values on instantiation, the values won't change, and the values are primarily read on the JS side, you could do something like this:

```c++
// This part is optional. 
// If you do decide to have these in the class definition, 
// the napi_writable is necessary to allow assignment in the constructor.
{
    InstanceValue("pointerId", env.Undefined(), napi_writable),
    InstanceValue("pointerType", env.Undefined(), napi_writable),
}

...

PointerEvent(const Napi::CallbackInfo& info)
    : Napi::ObjectWrap<PointerEvent>{info}
{
    auto thisObject = info.This().As<Napi::Object>();
    const auto params = info[1].As<Napi::Object>();
    
    thisObject.Set("pointerId", params.Get("pointerId"));
    m_pointerId = params.Get("pointerId").As<Napi::Number>().Int32Value();

    thisObject.Set("pointerType", params.Get("pointerType"));
    m_pointerType = params.Get("pointerType").As<Napi::String>().Utf8Value();
}
```

And then no more need for the accessor functions, and no more communication over the wire for static values.

PR #917, opened 2021-10-13
Native video

CedricGuillemet:
Native video is divided in 3 parts:
- the plugin: JS objects and link with C++
- Video decoding
- Video encoding

1. The plugin structure is done and stubs for all supported platform are in place.

1. Video decoding is only supported for Apple devices. It uses Media foundation. Sound is not supported as there is nothing in BN to playback sound. The code is not production ready. Some parts are weak and a bit more attention is needed here and there. Also, MP4 decoding is only possible for local files. It's not possible to decode streams for https for example. I think there is a way with Media Foundation to automate the process to download a file and play it back but got no time to test it. There are OS APIs to decode video files on Windows and Android. For Linux, I found nothing but install FFmpeg package and link against. FFmpeg can also be used with all other platforms. "Just fill in the blanks"

1. Video encoding works on Apple but the API and exposed JS functions do not correspond to any web standard (web codecs). It's just 3 functions: open, push_frame and close.

Note: it can be tempting to only have a FFMpeg backend. It will come at a cost of memory footprint for the mobiles (whereas using OS apis is cheaper) and performance. I believe on Apple for example, video encoding/decoding is done in HW are at least with coprocessors on M1.


CedricGuillemet:
@bghgary Do we want to keep this PR open?

bghgary:
Up to you. I'll be okay either way.

CedricGuillemet:
Closing.

Issue #916, opened 2021-10-12
[NativeCapture] Capturing a framebuffer crashes bgfx


CedricGuillemet:
Capturing a framebuffer created on the JS side crashes bgfx because texture has not been registered with GraphicsImpl.
Quick repro:

```
var renderTarget = new BABYLON.RenderTargetTexture("depth", 1024, scene, true);
...
capture = new NativeCapture(renderTarget._renderTarget._framebuffer);

```

I did a quick workaround by adding:
```
m_graphicsImpl.AddTexture(texture->Handle, width, height, generateMips, 1, format);
```
in `NativeEngine::CreateFrameBuffer`. Is it the way to go @syntheticmagus  ?

darraghjburke:
Even if I insert your workaround in line 1273 of NativeEngine, I'm still seeing:
`Assertion failed: bgfx::isTextureValid(0, false, 1, format, BGFX_TEXTURE_RT), file C:\Users\darraghburke\BabylonNative\Plugins\NativeEngine\Source\NativeEngine.cpp, line 1285`

My code is just:
```javascript
        var engine = new BABYLON.NativeEngine();
        var scene = new BABYLON.Scene(engine);
        var renderTarget = new BABYLON.RenderTargetTexture("depth", 1024, scene, true);
        var capture = new NativeCapture(renderTarget._renderTarget._framebuffer);
```

I'm on the latest version of master. Any thoughts?

PR #919, opened 2021-10-14
Add unit tests for NativeEngine + NativeCapture

darraghjburke:
Requested for #916. Tests should fail currently as there is an issue with NativeCapture. Should only be merged after that issue is resolved.

@CedricGuillemet would love feedback on what specific tests you would like for the NativeCapture object. Do you want to have a listener and then check that the data exists? I am not familiar with this class so I'm not sure what the right thing to check is.

We could also probably add some more checks to the NativeEngine test.


Apps/UnitTests/Scripts/tests.js
```diff
@@ -100,6 +100,24 @@ describe("RequestFile", function () {
...
 })
+describe("NativeEngine", function() {
+    this.timeout(0);
+    it("should create an engine and scene", function() {
+        var engine = new BABYLON.NativeEngine();
+        var scene = new BABYLON.Scene(engine);
+        expect(scene.meshes.length, "meshes.length").to.equal(0);
+    })
+})
+describe("NativeCapture", function() {
```
CedricGuillemet, line 111: will you try to do a capture and check callback values in this PR?

bghgary:
Closing for now. We can reopen this later when the feature is used more.

PR #920, opened 2021-10-14
[WIP] Refactor NativeXR for performance

rgerd:
This refactor gives an 8% improvement in frame time on HoloLens when visualizing scene understanding meshes, and significantly decreases (90%) `xrFrame.getPose` CPU time. In a performance test with 200 anchors, time spent per frame was reduced 70%.

mesh-and-anchors-perf

* Set properties directly on native-backed javascript objects as instance values rather than instance accessors to make for fast reading on the javascript side.
* Rewrite part of the XRFrame::getPoses function in javascript and use memcpy to avoid multiple expensive Set calls over-the-wire (see the babylon.js part of this PR (https://github.com/BabylonJS/Babylon.js/pull/11276) for more details).
* Return spaces instead of reference spaces for planes and anchors.
* Move the Space struct from `xr::System::Session::Frame::Space` to `xr::Space` so that (1) we can attach a space to an anchor to be consistent with the OpenXR and WebXR specs and (2) remove the implication that a space is a frame-bound construct.

Still want to do some more profiling & cleaning up, some other small to-do's:
- [x] Profile anchor tracking performance
- [x] Fix mobile compile errors
- [ ] Test & profile on mobile

### DIFF
--- a/Dependencies/xr/Include/XR.h
+++ b/Dependencies/xr/Include/XR.h
@@ -270,7 +270,7 @@ namespace xr
 
                 struct Plane
                 {
-                    using Identifier = size_t;
+                    using Identifier = uint32_t;
                     const Identifier ID{ NEXT_ID++ };
                     Pose Center{};
                     std::vector<float> Polygon{};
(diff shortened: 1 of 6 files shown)


rgerd:
Closing as I've been splitting this into smaller PR's.

Issue #821, opened 2021-07-02
Abstract and generalize the Graphics component's "update" concept


syntheticmagus:
At present, the Graphics component supports only one "update" per frame, which is tightly coupled with the rendering. This proposal is to extract, abstract, and generalize this concept to allow for arbitrary distinct "updates" to be performed per-frame, varying by the needs of the application.

This thinking is precipitated by the observation that certain components such as the Canvas polyfill have an explicit need to do work after all other JavaScript work has been completed. In the Canvas's case, it needs to do this because the underlying implementation depends on having all the commands for the current frame serialized before the final submission calls can be executed, which means all JS work must be done before the implementation can "report" this work to the Graphics component for rendering. Two options have been explored for ways to do this so far:

1. Perform the actual submission of the drawing on the _next_ frame, thus ensuring that all submissions from _this_ frame are completed.
2. Introduce a third scheduler to the Graphics component which allows code to be run on the Graphics thread _after_ the "update" has finished and _before_ `bgfx::frame` is called.

Both of these approaches have flaws. The first approach introduces an unexpected asynchronicity which certainly adds complexity and may be adding bugs. The second approach avoids these issues, but it does so by forcing the post-update work to be executed on the rendering thread, which is arbitrary, counter to our parallelism goals, and may scale poorly if work of this kind mounts. For example, consider a hypothetical app running a 4x4 grid of completely separate JavaScript contexts all rendering with Babylon Native and all using the Canvas. The current architecture should allow the core rendering work of all 16 of these engines to be done in parallel. However, if the Canvas work is forced to run on the rendering thread, all 16 contexts worth of Canvas drawing will all be executed sequentially on the least blockable thread in the application _every frame_. This is less than ideal.

[...]

PR #926, opened 2021-10-20
Multi update

syntheticmagus:
The changes adds the concept of "updates" to the Graphics component, allowing for an arbitrary number of semantically-distinctive "passes" to be run as part of the rendering process. Many game engines provide mechanisms for developers to sequence their operations (for example, `Update()` versus `LateUpdate()`, etc. in Unity) so that all operations of one semantic affiliation can be guaranteed to be completed by the time the next one begins. With this change, Babylon Native now allows this to be done generically and lazily, with various components requesting their "updates" by name while the Graphics-owning consuming app controls the sequence and timing. The most immediate need for this is in components which require all "update" work to be done before running a sort of gather step prior to rendering; the Canvas component and NativeEngine's command buffer logic, in particular, both naturally want to gather up the issued instructions in a "postUpdate" step that nevertheless happens before rendering.

To be clear, this change does not introduce this "postUpdate" update or remove the workarounds we used to avoid needing it before. In fact, the goal of this PR is to not change behavior at all, but only to switch the existing behavior to the new update-based paradigm so that it can be modified in later PRs to use the new multi-update capabilities.


syntheticmagus:
Fixes #821 although there's more to do to actually *use* the abstraction.

PR #927, opened 2021-10-21
Build improvements

bghgary:
- llhttp in v8inspector now assumes npm is already installed
- CMakeLists automatically call npm to install packages
- Make Android npm packages Android specific
- Move App node_modules to App/Dependencies
- Touch babylonjs files in after npm install to avoid incremental build issues
- Update README


darraghjburke:
Doing it this way seems like it makes more sense. What do you think @ryantrem?

Another question: right now I can manually build Babylon.js and put the output JS files into node_modules if I want to test my local changes. I'm guessing I won't be able to do this anymore, at least on Windows, since CMake will do an `npm install` every time I build. I guess an alternative would be to put the files into my build folder after the build finishes.

ryantrem:
It's weird to me. When I think about the developer workflow, I think of running CMake as a "nested loop" inside the running NPM "loop." Like I only run NPM when I need packages installed/updated, which is occasional. I run CMake when I add/remove source files, which is somewhat more frequent. So in my mind, NPM is a courser grained more outer tool, so the "dependency" seems backwards to me. I was thinking more like you do the `npm install`, then you can run `npm run <some command name>` to do the various other build related things. This pattern would match what we do in BRN as well.

bghgary:
With BRN, we know exactly the platforms we need to target and the exact flags we need to pass to CMake. If we wrapped CMake inside an npm call, we would have to be able to specify CMake arguments somehow since we don't know what the user wants to do with the CMake arguments.

Perhaps we should be splitting the concepts out a bit.

The library itself (core, plugins, polyfills, dependencies) must be CMake. There is only one thing that depends on npm and that's the `llhttp` dependency. This should most definitely be encapsulated in the CMake as CMake is how we consume dependencies.

It might make sense to use npm for the apps (e.g. the `Apps` folder), but doing it just for Apps is weird because CMake is the one generating the solution for the overall project.

ryantrem:
That's what the -- is for in npm commands. So something like "npm run myCommand -- arg1 arg2". So I think it would still work, but your call on whatever pattern you think is right!

bghgary:
Yes, I'm aware of this, but it doesn't translate directly to CMake arguments. We still have to decide what the argument after `--` actually does. It will be tricky to get right.

CedricGuillemet:
> Yes, I'm aware of this, but it doesn't translate directly to CMake arguments. We still have to decide what the argument after `--` actually does. It will be tricky to get right.

I feel like 95%, I'm using default cmake args and remaining 5%, you can run cmake after `npm ... ` with more fine grained arguments.

bghgary:
Closing for now. We can reactive later if necessary.

PR #935, opened 2021-10-25
Parse rgb() color strings in Canvas polyfill

darraghjburke:
Parses strings of the form "rgb(#,#,#)". Does not currently support rgba() strings; I'm happy to add support for those later if we need them. Needed for LensRenderingPipeline.


CedricGuillemet (approved):
LGTM

PR #940, opened 2021-10-27
Test post processes

darraghjburke:
Adds unit tests for compiling post processes. Many are currently broken, so those tests are commented out. As we fix them, we can uncomment the tests.

These new tests rely on Graphics, which meant that some additional work was needed for Ubuntu specifically. I have separated out the platform specific code.

Having these tests will be useful for shader pipeline refactoring, to make sure that nothing gets broken.


.github/jobs/linux.yml
```diff
@@ -66,6 +66,9 @@ jobs:
       condition: failed()
 
     - script: |
+        export DISPLAY=:99
+        Xvfb :99 -screen 0 1600x900x24 &
+        sleep 3
```
bghgary, line 71: what's this sleep for?

Apps/UnitTests/Scripts/tests.js
```diff
@@ -101,6 +101,144 @@ describe("RequestFile", function () {
...
+                    vertexSource: `void main() {}`,
+                    fragmentSource: `void main() {}`
+                }
+                const mat = new BABYLON.ShaderMaterial("shader", scene, shaders, {});
+                sphere.material = mat;
+            }, done);
+        }
+        createEmptyShaderMat()
+    })
+    // it("GradientMaterial should compile", function(done) {
```
bghgary, line 130: might be good to group the ones that are commented out and add a comment explaining why they are commented out

Apps/UnitTests/Scripts/tests.js
```diff
@@ -101,6 +101,144 @@ describe("RequestFile", function () {
     })
 })
 
+function createSceneAndWait(callback, done) {
+    const engine = new BABYLON.NativeEngine();
+    const scene = new BABYLON.Scene(engine);
+    scene.createDefaultCameraOrLight();
```
bghgary, line 107: Do you actually need a light? Are we actually rendering? Also, it doesn't look like camera is used in all cases. Maybe it should be created only in the cases that are needed?

Apps/UnitTests/Shared/Tests.cpp
```diff
@@ -14,6 +15,7 @@
 
 std::unique_ptr<Babylon::AppRuntime> runtime{};
 std::unique_ptr<Babylon::Graphics> graphics{};
```
bghgary, line 17: should remove this

Apps/UnitTests/Shared/Tests.cpp
```diff
@@ -28,11 +30,9 @@ void SetExitCode(const Napi::CallbackInfo& info)
     exitCode.set_value(info[0].As<Napi::Number>().Int32Value());
 }
 
-int run()
+int run(std::unique_ptr<Babylon::Graphics> graphicsObject)
 {
-    Babylon::ContextConfiguration graphicsConfig{};
-    graphics = Babylon::Graphics::CreateGraphics(graphicsConfig);
-
+    graphics = std::move(graphicsObject);
```
bghgary, line 35: Don't need this. Just use the passed in graphics?

Apps/UnitTests/Shared/Tests.cpp
```diff
@@ -28,11 +30,9 @@ void SetExitCode(const Napi::CallbackInfo& info)
     exitCode.set_value(info[0].As<Napi::Number>().Int32Value());
 }
 
-int run()
+int run(std::unique_ptr<Babylon::Graphics> graphicsObject)
```
bghgary, line 33: nit: didn't notice the first time, but it probably should be this:
```suggestion
int Run(std::unique_ptr<Babylon::Graphics> graphicsObject)
```

Apps/UnitTests/X11/App.cpp
```diff
@@ -0,0 +1,64 @@
...
+    XInternAtoms(display, (char **)&wmDeleteWindowName, 1, False, &wmDeleteWindow);
+    XSetWMProtocols(display, window, &wmDeleteWindow, 1);
+
+    XMapWindow(display, window);
+    XStoreName(display, window, s_applicationName);
+
+    Babylon::WindowConfiguration graphicsConfig{};
+    graphicsConfig.WindowPtr = (void*)(uintptr_t)window;
+    graphicsConfig.Width = static_cast<size_t>(width);
+    graphicsConfig.Height = static_cast<size_t>(height);
```
bghgary, line 62: Inconsistent usage of C-style cast and C++-style cast.

Apps/UnitTests/X11/App.cpp
```diff
@@ -0,0 +1,64 @@
...
+    memset(&attr, 0, sizeof(attr) );
+    XChangeWindowAttributes(display, window, CWBackPixel, &attr);
+
+    const char* wmDeleteWindowName = "WM_DELETE_WINDOW";
+    Atom wmDeleteWindow;
+    XInternAtoms(display, (char **)&wmDeleteWindowName, 1, False, &wmDeleteWindow);
+    XSetWMProtocols(display, window, &wmDeleteWindow, 1);
+
+    XMapWindow(display, window);
+    XStoreName(display, window, s_applicationName);
```
bghgary, line 57: This is probably a copy and paste, but this code is a bit inconsistent. There are spacing issues all over and constants defined sometimes with the code (wmDeleteWindowName) and sometimes outside (s_applicationName).

.github/jobs/linux.yml
CedricGuillemet: AFAIK, there is no way to wait for xvfb to be ready.

Apps/UnitTests/Scripts/tests.js
darraghjburke: True, we don't need a light. For testing materials we don't need a camera, but it didn't seem worth it to duplicate createDefaultCamera() across every PP test case. I guess it would make the individual tests a bit more isolated, but the benefit didn't seem worth it to me.

Apps/UnitTests/X11/App.cpp
darraghjburke: I swapped the WindowPtr line for a reinterpret_cast since the compiler wouldn't let me do a static_cast between the types. Does this seem okay?

Apps/UnitTests/X11/App.cpp
darraghjburke: Okay, I cleaned it up a bit. I can do the same for the ValidationTests/Playground if desired

.github/jobs/linux.yml
bghgary: Quick search returns this: https://unix.stackexchange.com/questions/310679/how-to-poll-for-xvfb-to-be-ready

Apps/UnitTests/Scripts/tests.js
bghgary: You could pass arguments to the function specifying whether to create them or not. Not a big deal either way.

Apps/UnitTests/Scripts/tests.js
bghgary: I would recommend one comment per block. This block still doesn't have a comment. It's hard to tell if it's a mistake or it's intentional.

Apps/UnitTests/X11/App.cpp
```diff
@@ -0,0 +1,47 @@
...
+#define XK_LATIN1
+#include <X11/Xlib.h> // will include X11 which #defines None... Don't mess with order of includes.
+#include <X11/Xutil.h>
+#undef None
+#include "../Shared/Tests.cpp"
+
+static const char* s_applicationName = "Babylon Native Unit Tests";
+static const int width = 640;
+static const int height = 480;
+static const char* wmDeleteWindowName = "WM_DELETE_WINDOW";
```
bghgary, line 11: nit: prefer anonymous namespace to static
nit: can use constexpr for these

bghgary (approved):

Apps/UnitTests/X11/App.cpp
```diff
@@ -5,11 +5,12 @@
...
-static const char* s_applicationName = "Babylon Native Unit Tests";
-static const int width = 640;
-static const int height = 480;
-static const char* wmDeleteWindowName = "WM_DELETE_WINDOW";
-
+namespace {
+    constexpr char* s_applicationName = "Babylon Native Unit Tests";
+    constexpr int width = 640;
+    constexpr int height = 480;
+    constexpr char* wmDeleteWindowName = "WM_DELETE_WINDOW";
```
bghgary, line 12: constexpr doesn't replace const (common misunderstanding). You still need the const.

bghgary: There are two constness modifiers for raw char pointer. One is for the constness of the char itself and one for the constness of the pointer. Before you didn't have a constness modifier for the pointer itself and thus you can assign the pointer to something else. Without constexpre, you can fix this by doing something like this:
```
const char* const foo = "asdf";
```

With constexpr, you don't need to put the const modifier after the * to disambiguate.

```
constexpr const char* foo = "asdf";
```

bghgary: Also, you should remove the `s_` Hungarian notation or put them on all of the variables.

```suggestion
    constexpr const char* applicationName = "Babylon Native Unit Tests";
    constexpr int width = 640;
    constexpr int height = 480;
    constexpr const char* wmDeleteWindowName = "WM_DELETE_WINDOW";
```

Apps/UnitTests/X11/App.cpp
bghgary: Yes.

Apps/UnitTests/Shared/Tests.cpp
```diff
@@ -54,10 +51,13 @@ int run()
...
     loader.LoadScript("app:///Scripts/chai.js");
     loader.LoadScript("app:///Scripts/mocha.js");
     loader.LoadScript("app:///Scripts/tests.js");
-    
+    graphics->StartRenderingCurrentFrame();
+    graphics->FinishRenderingCurrentFrame();
+
     auto code{exitCode.get_future().get()};
     runtime.reset();
     graphics.reset();
```
bghgary, line 63: nit: i think these are probably not necessary anymore

Apps/UnitTests/X11/App.cpp
```diff
@@ -0,0 +1,48 @@
...
+#undef None
+#include "../Shared/Tests.cpp"
+
+namespace {
+    constexpr const char* applicationName = "Babylon Native Unit Tests";
+    constexpr const  int width = 640;
+    constexpr const int height = 480;
+    constexpr const char* wmDeleteWindowName = "WM_DELETE_WINDOW";
+}
+int main()
```
bghgary, line 14: nit: should probably have a newline here
```suggestion
}

int main()
```

Apps/UnitTests/X11/App.cpp
```diff
@@ -0,0 +1,48 @@
+#define XK_MISCELLANY
+#define XK_LATIN1
+#include <X11/Xlib.h> // will include X11 which #defines None... Don't mess with order of includes.
+#include <X11/Xutil.h>
+#undef None
+#include "../Shared/Tests.cpp"
+
+namespace {
+    constexpr const char* applicationName = "Babylon Native Unit Tests";
```
bghgary, line 9: nit: formatting
```suggestion
namespace
{
    constexpr const char* applicationName = "Babylon Native Unit Tests";
```

bghgary (approved):

PR #947, opened 2021-11-09
Refactor Shader Traversers

### LINKED ISSUE #914: Refactor Shader Compiler to reduce preprocessor directives
The Shader Compiler code currently contains several preprocessor directives, including macros and API-specific #defines. We should try to factor these out to make the code cleaner and more maintainable.

darraghjburke:
This PR is a first pass at cleaning up the shader compiler code. The main change is to remove the preprocessor directives for VertexVaryingInTraverser, and replace them with separate subclasses classes. Since OpenGL and Metal use the same logic, they both share a subclass, while DirectX has it's own.

This approach seemed like the best way to get rid of preprocessors while still minimizing code duplication.

I left the IF_NAME_RETURN_ATTRIB macro in because there didn't seem to be much benefit to factoring it out; I could have rewritten it with a list and iterated over that, but I think the current solution is pretty readable as is.

(Also, sorry about the messy commit history; this branch diverged a bit from master and rebasing seems to have left a bunch of extraneous commits)

Resolves #914

### DIFF
--- a/Plugins/NativeEngine/Source/ShaderCompilerD3D.cpp
+++ b/Plugins/NativeEngine/Source/ShaderCompilerD3D.cpp
@@ -98,7 +98,7 @@ namespace Babylon
         ShaderCompilerTraversers::IdGenerator ids{};
         auto utstScope = ShaderCompilerTraversers::MoveNonSamplerUniformsIntoStruct(program, ids);
         std::unordered_map<std::string, std::string> vertexAttributeRenaming = {};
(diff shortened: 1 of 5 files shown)


darraghjburke:
Seems like there is an issue with uniforms on Windows, the validation tests are generating this error:
`Duplicated input semantics can't change type, size, or layout ('TEXCOORD0').`

I will investigate

bghgary (approved):

PR #953, opened 2021-11-23
Modify Playground to use JS InputManager

PolygonalSun:
This PR is a continuation of @ryantrem's [PR 610](https://github.com/BabylonJS/BabylonNative/pull/610) that adds Win32, UWP, and Linux functionality (with some slight modifications for Android).  This should not change the user input functionality (only support for single left click with drag) but will enable the use of things like `scene.onPointerObservable` (including picking).


Apps/Playground/macOS/ViewController.mm
```diff
@@ -51,7 +53,12 @@ - (void)viewDidLoad {
 }
 
 - (void)uninitialize {
-    inputBuffer.reset();
+    if (graphics)
+    {
+        graphics->FinishRenderingCurrentFrame();
+    }
+
+    nativeInput = nullptr;
```
syntheticmagus, line 61: It might be worth adding a comment on why this isn't a leak. I'm pretty sure it isn't as the JS context owns that memory, but it might be worth calling that out explicitly as seeing a raw pointer just set to null like this is a little jarring.

Apps/Playground/Scripts/experience.js
```diff
@@ -81,9 +61,8 @@ CreateBoxAsync().then(function () {
 //BABYLON.SceneLoader.AppendAsync("https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/ClearCoatTest/glTF/ClearCoatTest.gltf").then(function () {
     BABYLON.Tools.Log("Loaded");
 
-    scene.createDefaultCamera(true);
+    scene.createDefaultCamera(true, true, true);
     scene.activeCamera.alpha += Math.PI;
-    CreateInputHandling(scene);
```
syntheticmagus, line 86: So, if we're getting rid of this stuff, can we get rid of the old `InputManager` stopgap thing altogether?

Apps/Playground/Scripts/experience.js
```diff
@@ -27,7 +27,7 @@ function CreateSpheresAsync() {
     for (var i = 0; i < size; i++) {
         for (var j = 0; j < size; j++) {
             for (var k = 0; k < size; k++) {
-                var sphere = BABYLON.Mesh.CreateSphere("sphere" + i + j + k, 32, 0.9, scene);
+                var sphere = BABYLON.Mesh.CreateSphere("sphere" + i + j + k, 32, 0.9);
```
syntheticmagus, line 30: Would it not be better to pass in the scene to `CreateSpheresAsync` instead?

Apps/Playground/UWP/App.h
```diff
@@ -44,7 +44,7 @@ ref class App sealed : public Windows::ApplicationModel::Core::IFrameworkView
     std::unique_ptr<Babylon::Graphics> m_graphics{};
     std::unique_ptr<Babylon::Graphics::Update> m_update{};
     std::unique_ptr<Babylon::AppRuntime> m_runtime{};
-    std::unique_ptr<InputManager<Babylon::AppRuntime>::InputBuffer> m_inputBuffer{};
+    Babylon::Plugins::NativeInput* nativeInput{};
```
bghgary, line 47:
```suggestion
    Babylon::Plugins::NativeInput* m_nativeInput{};
```

Apps/Playground/UWP/App.cpp
```diff
@@ -129,7 +129,7 @@ void App::Uninitialize()
     }
 
     m_chromeDevTools.reset();
-    m_inputBuffer.reset();
+    nativeInput = nullptr;
```
bghgary, line 132: nit: Inconsistent way to reset pointer. Earlier it was this:
```suggestion
    nativeInput = {};
```

Apps/Playground/UWP/App.cpp
```diff
@@ -208,26 +208,28 @@ void App::OnWindowClosed(CoreWindow^ sender, CoreWindowEventArgs^ args)
 
 void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
 {
-    if (m_inputBuffer != nullptr)
+    if (nativeInput)
```
bghgary, line 211: nit: I would recommend doing this for raw pointers:
```suggestion
    if (nativeInput != nullptr)
```

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
```diff
@@ -55,7 +56,7 @@ extern "C"
         g_chromeDevTools.reset();
         g_nativeXr.reset();
         g_scriptLoader.reset();
-        g_inputBuffer.reset();
+        g_nativeInput = {};
```
bghgary, line 59: Is there a threading issue here? g_nativeInput is being assigned/constructed on the JS thread, but being reset on the main thread. Not sure if there are lurking issues. Granted, this is what we are doing with the other objects also, but we don't check and/or use these objects on the main thread.

Apps/Playground/Scripts/experience.js
bghgary: If we do that, then we should also do it to CreateBoxAsync.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
ryantrem: I think there is potentially an issue here. I think with a different change we should rework `NativeInput` to be structured more like `NativeXr` and `NativeCamera`, where there is an impl object that has a native "projection" and a JS "projection" and the impl is ref counted. I think that is a safer approach, but I think that should be fixed separately.

Apps/Playground/Scripts/experience.js
PolygonalSun: That's the plan though I'm not sure that we can get rid of the actual code from the repo though

Apps/Playground/UWP/App.h
PolygonalSun: Changed

Apps/Playground/UWP/App.cpp
PolygonalSun: Changed

Apps/Playground/UWP/App.cpp
PolygonalSun: Fixed

Apps/Playground/Scripts/experience.js
PolygonalSun: @ryantrem @bghgary Do we know if InputManager (Native ver) is used in anything outside just an internal use?  If it's not, I can just delete the file.

bghgary: I doubt it is used elsewhere. It was a stopgap that @syntheticmagus wrote because we didn't have input hooked up.

Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
PolygonalSun: Unless you think that I should address this in this PR, I think that it'd be best to address a rework in a different follow-up PR.

bghgary: Okay with me.

Plugins/NativeInput/Source/Shared/DeviceInputSystem.cpp
```diff
@@ -109,6 +110,15 @@ namespace Babylon::Plugins
         }
     }
 
+    Napi::Value NativeInput::Impl::DeviceInputSystem::IsDeviceAvailable(const Napi::CallbackInfo& info)
+    {
+        uint32_t deviceType = info[0].As<Napi::Number>().Uint32Value();
+
+        bool isAvailable =  m_nativeInput.IsDeviceAvailable(static_cast<DeviceType>(deviceType));
```
bghgary, line 117: nit:
```suggestion
        bool isAvailable = m_nativeInput.IsDeviceAvailable(static_cast<DeviceType>(deviceType));
```

Plugins/NativeInput/Source/Unix/NativeInput.cpp
```diff
@@ -0,0 +1,10 @@
+#include "../Shared/NativeInput.h"
+
+namespace Babylon::Plugins
+{
+    bool NativeInput::Impl::HasMouse()
+    {
+        // Assume a mouse is present (or find a way to check this programmatically).
```
bghgary, line 7: Is this a TODO?

Apps/Playground/Scripts/experience.js
PolygonalSun: I'll add this change to script

Plugins/NativeInput/Source/Unix/NativeInput.cpp
PolygonalSun: Adding a TODO to comment and changed wording to clarify intentions

Plugins/NativeInput/Source/Shared/DeviceInputSystem.cpp
PolygonalSun: Fixed

Apps/Playground/macOS/ViewController.mm
PolygonalSun: I added a comment

Apps/Playground/Scripts/experience.js
PolygonalSun: I removed the header file and cmake list reference.

Plugins/NativeInput/Source/Unix/NativeInput.cpp
```diff
@@ -0,0 +1,10 @@
+#include "../Shared/NativeInput.h"
+
+namespace Babylon::Plugins
+{
+    bool NativeInput::Impl::HasMouse()
+    {
+        // TODO: Currently assuming a mouse is present.  Need to evaluate possibly detecting this programmatically.
```
bghgary, line 7: I would recommend filing an issue to address this later.

bghgary (approved):

Plugins/NativeInput/Source/Unix/NativeInput.cpp
PolygonalSun: Added as task to https://github.com/BabylonJS/BabylonNative/issues/956

PR #955, opened 2021-12-03
Fix an issue with array buffer finalizers for napi/jsi implementation

bghgary:
There is an issue with using jsi host objects as prototypes on react native windows. This change works around the problem by using a side property instead of chaining the prototype. Using prototype chaining is perhaps not a good idea anyways since it introduces another layer when accessing arraybuffer members.


CedricGuillemet (approved):
Tested and approved!

PR #961, opened 2021-12-15
Use getPoseData for XRFrame

rgerd:
We now wrap new XRFrame's in NativeXRFrame instances so that we can use more performant JS+Native implementations of functions, starting with `getPose`, which has been changed to a `getPoseData` helper function on the native side because a memcpy to an ArrayBuffer is more performant than updating a JS object's members from the native side.

See https://github.com/BabylonJS/Babylon.js/pull/11655 for context.


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2055,7 +2071,10 @@ namespace Babylon
 
             static Napi::Object New(const Napi::CallbackInfo& info)
             {
-                return info.Env().Global().Get(JS_CLASS_NAME).As<Napi::Function>().New({});
+                // Instead of creating just a C++ XRFrame object, wrap it in a JS NativeXRFrame object (see nativeXRFrame.ts in Babylon.JS)
+                const auto nativeImpl = info.Env().Global().Get(JS_CLASS_NAME).As<Napi::Function>().New({});
+                const auto nativeXRFrame = JsRuntime::NativeObject::GetFromJavaScript(info.Env()).Get("NativeXRFrame").As<Napi::Function>().New({nativeImpl}).As<Napi::Object>();
```
bghgary, line 2076: Is this called often? Might consider caching the NativeXRFrame constructor somehow.

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2198,23 +2217,13 @@ namespace Babylon
...
-                {
-                    auto* xrSpace = XRReferenceSpace::Unwrap(info[0].As<Napi::Object>());
-                    assert(xrSpace != nullptr);
-                    Napi::Object napiPose = XRPose::New(info);
-                    XRPose* pose = XRPose::Unwrap(napiPose);
-                    pose->Update(xrSpace->GetTransform());
-                    return std::move(napiPose);
-                }
+                const auto& space = *info[0].As<Napi::External<xr::Space>>().Data();
+                auto vectorBuffer = info[2].As<Napi::ArrayBuffer>();
```
bghgary, line 2223: What about `info[1]`?

rgerd: We currently only support one reference space, which is stored in info[1] and is assumed to be the identity with respect to the provided space

Plugins/NativeXr/Source/NativeXr.cpp
rgerd: It's actually called only once when the session is created, and then the single frame instance is updated every frame with new frame data from the XR runtime

Plugins/NativeXr/Source/NativeXr.cpp
bghgary: Maybe assert or throw exception when it's not?

ryantrem (approved):

Apps/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.0.1",
   "private": true,
   "dependencies": {
-    "babylonjs": "5.0.0-alpha.60",
-    "babylonjs-gui": "5.0.0-alpha.60",
-    "babylonjs-loaders": "5.0.0-alpha.60",
-    "babylonjs-materials": "5.0.0-alpha.60",
+    "babylonjs": "5.0.0-alpha.65",
```
ryantrem, line 6: When I tried upgrading to alpha.64 in another project earlier today, I saw XR was not working on iOS (solid background rendered instead of the camera frame). I guess you aren't seeing this behavior?

rgerd: I'm not seeing the behavior on OpenXR devices because our OpenXR and mobile XR implementations currently expect opposite clearing behavior on the JS side. This PR brings us back to the original behavior that changed due to a change on the JS side that looks for a specific type we don't currently define:
https://github.com/BabylonJS/BabylonNative/pull/964

rgerd:
I'd like to merge #963 first so that planes and meshes aren't broken by this change.

PR #965, opened 2022-01-05
Fix xr.mm texture cleanup timing issues

### LINKED ISSUE #932: Metal Validation Error with AR/iOS
I've checked commit and it seems this commit breaks Metal Validation
https://github.com/BabylonJS/BabylonNative/commit/c79d748174cae8aa158de63f1e4f809b567105d6

Everything is ok with 
https://github.com/BabylonJS/BabylonNative/commit/b43087a5b26ac8d3dfcb7d1421249d7aed765e35

Run Playground with `ar = true;` in experience.js

image

ryantrem:
We used to block the command buffer until it completed, and then set texture resources to purgeable. To improve performance, we removed the blocking, but that left the code in a state where we end up making texture resources purgeable before the command buffer completes. I guess we got lucky that this didn't seem to cause a problem in release builds, but in debug builds with Metal API validation enabled these issues are flagged and the debugger stops app execution.

The change here is to ensure that texture resources are made purgeable only after the command buffer completes.

Fixes #932

### DIFF
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -741,6 +741,10 @@ namespace xr {
         }
 
         ~Impl() {
+            if (currentCommandBuffer != nil) {
+                [currentCommandBuffer waitUntilCompleted];
+            }
+
             if (ActiveFrameViews[0].ColorTexturePointer != nil) {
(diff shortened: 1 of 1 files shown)


CedricGuillemet (approved):
Tested OK!

PR #970, opened 2022-01-11
Handle selection node fixing repro for shader compilation bug.

### LINKED ISSUE #959: [Shader Traverser] Exception with UniformTypeChangeTraverser
When changing uniform types to vec4/mat4, there is one case that is not handled by `UniformTypeChangeTraverser` here throws an exception here: https://github.com/BabylonJS/BabylonNative/blob/9b2bad35c4e020afde0d4355d7d817ebe0466afb/Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp#L369

This can be tested with https://playground.babylonjs.com/#ycy2il#0

before changing type, there is `float _alphaTest` and a test in the fragment shader : `if (alphaTest)`. After changing type, `_alphaTest` becomes a float4 but the if expression stays the same (hence the throw).

This was not happening with D3D11 before https://github.com/BabylonJS/BabylonNative/pull/826 . It does now, and for all other Graphics API.

cc @bghgary

syntheticmagus:
Fixes #959. Tested on Windows D3D11 using the repro described in the bug.

### DIFF
--- a/Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
+++ b/Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
@@ -364,6 +364,21 @@ namespace Babylon::ShaderCompilerTraversers
                         {
                             unary->setOperand(shapeConversion);
                         }
+                        else if (auto* selection = parent->getAsSelectionNode())
+                        {
+                            if (selection->getCondition() == node)
+                            {
+                                selection->setCondition(shapeConversion);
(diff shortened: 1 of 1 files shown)


CedricGuillemet (approved):
Tested OK D3D11 and 12

CedricGuillemet:
I tried to add a sprites VT but ran into this issue: https://github.com/BabylonJS/BabylonNative/issues/971

PR #972, opened 2022-01-14
Various build and consistency fixes

bghgary:
While trying to take a new version into Babylon React Native and analyzing Android build binary sizes, I found a bunch of issues with the code and build.

- Fix Android Playground build to actually use shared C++ library.
- Clean up the options we set for all dependencies to ensure we are disabling options we don't need.
- Add option to disable building the Apps folder.
- Add dependent option for using swap chain panel for UWP.
- Rename `windowPtr` to `window` as some of the types are not pointers.


CedricGuillemet (approved):
LGTM

PR #973, opened 2022-01-17
Disable MSAA for iOS and Android

CedricGuillemet:
Move bgfx reset flags into an header per platform.
MSAA is disabled for Android and iOS. Comments/links on why it's disabled are in the code.

- tested on Android, Windows
- tested on MacOS & iOS with highlighting


Core/Graphics/Source/GraphicsImpl.cpp
```diff
@@ -30,7 +31,7 @@ namespace Babylon
 
         auto& init = m_state.Bgfx.InitState;
         init.type = s_bgfxRenderType;
-        init.resolution.reset = BGFX_RESET_VSYNC | BGFX_RESET_MSAA_X4 | BGFX_RESET_MAXANISOTROPY;
```
bghgary, line 33: I'm not sure we should split this into parts. The intention is that they are the same. Can you think of a reason why we want the reset flags to be different on purpose? IMO we can just use #ifdefs here and put a comment saying it's broken for Android/iOS like you did but in one place.

Core/Graphics/CMakeLists.txt
```diff
@@ -26,15 +26,17 @@ set(SOURCES
     "Source/GraphicsImpl_${PLATFORM_NAME}.${BABYLON_NATIVE_PLATFORM_IMPL_EXT}"
     "Source/GraphicsImpl_${GRAPHICS_API}.cpp"
     "Source/SafeTimespanGuarantor.cpp"
-    "Source/SafeTimespanGuarantor.h")
+    "Source/SafeTimespanGuarantor.h"
+    "Source/${PLATFORM_NAME}/GraphicsFlags.h")
```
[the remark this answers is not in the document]
CedricGuillemet: GraphicsPlatform.h is public. It means bgfx.h must be included and public. I wanted to keep bgfx private here.

Core/Graphics/Source/GraphicsImpl.cpp
CedricGuillemet: Using #ifdef was my first choice but you proposed to put the flags in header, per platform.

bghgary: Sorry, yes, I did, but I wasn't thinking about what the intention was.

bghgary: Let me take care of this since I broke the Android version.

PR #974, opened 2022-01-18
Root package.json

CedricGuillemet:
Following discussion here : https://github.com/BabylonJS/BabylonNative/pull/927

I'm testing having a root package.json.
Build steps would be for ios:

1. git clone https://github.com/BabylonJS/BabylonNative.git
1. cd BabylonNative
1. npm install
1. npm run cmake-ios

Which, to me, looks much simpler.
For advanced use, user can cd to the build folder and re-run cmake with according flags. For 90-95% of times, when you want to test or develop with common parameters, this is super handy.

I suggest to replace build steps in readme.md for something less verbose and move cmake commands to a document in Documentation folder.

This PR is open to discussion/improvements/ideas.

### DIFF
--- a/.github/jobs/android.yml
+++ b/.github/jobs/android.yml
@@ -18,6 +18,7 @@ jobs:
     - task: Npm@1
       inputs:
         command: 'install'
+        workingDir: './'
       displayName: 'Install Babylon.js NPM packages'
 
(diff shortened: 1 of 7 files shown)


CedricGuillemet:
Closing until we discuss more on integration.

PR #979, opened 2022-01-24
Tear down XrInput before OpenXR context is reset

rgerd:
There's an issue that shows up when tearing down the XR session with BabylonReactNative on windows where the input manager resources aren't disposed before the OpenXR context is reset.

### DIFF
--- a/Dependencies/xr/Source/OpenXR/XR.cpp
+++ b/Dependencies/xr/Source/OpenXR/XR.cpp
@@ -362,6 +362,8 @@ namespace xr
 
         ~Impl()
         {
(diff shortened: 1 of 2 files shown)


Dependencies/xr/Source/OpenXR/XR.cpp
```diff
@@ -361,6 +362,7 @@ namespace xr
 
         ~Impl()
         {
+            InputManager.reset();
```
syntheticmagus, line 365: Can you add a comment to make explicit that this has to happen before `XrRegistry::Reset()` is called?

syntheticmagus (approved):

Issue #923, opened 2021-10-19
Morph targets are broken


bghgary:
1. Open any model that uses morph targets (e.g., [MorphPrimitivesTest](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MorphPrimitivesTest))

Result: Error in JS: `Unable to get property 'TEXTURE_2D_ARRAY' of undefined or null reference`

**Expected behavior**
Model to load without error

bghgary:
Work around: put this somewhere before loading a model.

```
BABYLON.MorphTargetManager.EnableTextureStorage = false;
```

bghgary:
Once fixed, we should add validation tests so that this doesn't break again.

bghgary:
> Once fixed, we should add validation tests so that this doesn't break again.

@SergioRZMasson I think we missed doing this. It would be good to do so that we can't break it again.

bghgary:
Turns out, morph targets are still not working properly.

PR #981, opened 2022-01-24
Fixed issue #923 by implementing 2D Texture Array to be used with Morph  targets.

SergioRZMasson:
Added support for creating 2D texture arrays using the native engine. This allows morph targets to work properly. Also added length check to shader uniforms to avoid memory corruption when JS sends data with incorrect length to uniforms.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -16,7 +16,7 @@
 #include <bimg/decode.h>
 #include <bimg/encode.h>
 
-// STB_IMAGE_RESIZE_IMPLEMENTATION will define implementation in stb_image_resize.h. Many .cpp can include it 
+// STB_IMAGE_RESIZE_IMPLEMENTATION will define implementation in stb_image_resize.h. Many .cpp can include it
 // but only one can define implementation or linking errors will popup.
 #define STB_IMAGE_RESIZE_IMPLEMENTATION
 #include <stb/stb_image_resize.h>
@@ -107,8 +107,16 @@ namespace Babylon
             if (image->m_format == bimg::TextureFormat::R8 ||
                 image->m_format == bimg::TextureFormat::RG8)
             {
-                static const TransformFn UnpackR8{[](const uint8_t* src, uint8_t* dst) { dst[0] = dst[1] = dst[2] = src[0]; dst[3] = 1; }};
-                static const TransformFn UnpackRG8{[](const uint8_t* src, uint8_t* dst) { dst[0] = dst[1] = dst[2] = src[0]; dst[3] = src[1]; }};
+                static const TransformFn UnpackR8{[](const uint8_t* src, uint8_t* dst)
+                    {
+                        dst[0] = dst[1] = dst[2] = src[0];
+                        dst[3] = 1;
+                    }};
+                static const TransformFn UnpackRG8{[](const uint8_t* src, uint8_t* dst)
+                    {
+                        dst[0] = dst[1] = dst[2] = src[0];
+                        dst[3] = src[1];
+                    }};
 
                 // bimg loads grayscale textures with and without alpha as R8 and RG8 respectively.
                 // Unpack to RGB and RGBA such that RGB is the grayscale and the A is the alpha.
@@ -210,7 +218,8 @@ namespace Babylon
                     bgfx::ReleaseFn releaseFn{};
                     if (mip == image->m_numMips - 1)
                     {
-                        releaseFn = [](void*, void* userData) {
+                        releaseFn = [](void*, void* userData)
+                        {
                             bimg::imageFree(static_cast<bimg::ImageContainer*>(userData));
                         };
                     }
@@ -254,9 +263,10 @@ namespace Babylon
                     {
                         bimg::ImageContainer* image{images[(side * numMips) + mip]};
 
-                        bgfx::ReleaseFn releaseFn{[](void*, void* userData) {
-                            bimg::imageFree(static_cast<bimg::ImageContainer*>(userData));
-                        }};
+                        bgfx::ReleaseFn releaseFn{[](void*, void* userData)
+                            {
+                                bimg::imageFree(static_cast<bimg::ImageContainer*>(userData));
+                            }};
 
                         const bgfx::Memory* mem{bgfx::makeRef(image->m_data, image->m_size, releaseFn, image)};
                         bgfx::updateTextureCube(texture->Handle, 0, side, mip, 0, 0, static_cast<uint16_t>(image->m_width), static_cast<uint16_t>(image->m_height), mem);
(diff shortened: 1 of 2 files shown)


Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -73,10 +73,13 @@ namespace Babylon
         };
 
         std::unordered_map<uint16_t, UniformValue> Uniforms{};
+        std::unordered_map<uint16_t, size_t> UniformMaxElementLength{};
 
         void SetUniform(bgfx::UniformHandle handle, gsl::span<const float> data, size_t elementLength = 1)
         {
             UniformValue& value = Uniforms[handle.idx];
+            size_t& maxLenght = UniformMaxElementLength[handle.idx];
+            elementLength = std::max(static_cast<size_t>(1), std::min(maxLenght, elementLength));
```
bghgary, line 82: nit: 
```suggestion
            size_t& maxLength = UniformMaxElementLength[handle.idx];
            elementLength = std::max(static_cast<size_t>(1), std::min(maxLength, elementLength));
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1102,6 +1106,38 @@ namespace Babylon
...
+
+        if (!data.IsNull())
+        {
+            if (data.ByteLength() != bimg::imageGetSize(nullptr, width, height, 1, false, false, depth, format))
+                throw std::runtime_error{"The data size does not match width, height, depth and format"};
+
+            uint8_t* dataPtr = static_cast<uint8_t*>(data.ArrayBuffer().Data()) + data.ByteOffset();
+            uint32_t dataSize = static_cast<uint32_t>(data.ByteLength());
+            const bgfx::Memory* dataCopy = bgfx::copy(dataPtr, dataSize); // This is required since BGFX must manage the data the memory.
+            bgfx::updateTexture2D(texture->Handle, 0, 0, 0, 0, width, height, dataCopy);
```
bghgary, line 1137: > // This is required since BGFX must manage the data the memory.

I'm not sure this is true. We could use bgfx::makeRef instead and hold on to the JS memory until it is used.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1102,6 +1106,38 @@ namespace Babylon
...
+        texture->Height = height;
+        texture->CreationFlags = BGFX_TEXTURE_NONE;
+        texture->Flags = BGFX_TEXTURE_NONE | BGFX_SAMPLER_NONE | BGFX_CAPS_TEXTURE_2D_ARRAY;
+        texture->Handle = bgfx::createTexture2D(width, height, generateMips, depth, Cast(format), texture->Flags);
+        texture->OwnsHandle = true;
+
+        if (!data.IsNull())
+        {
+            if (data.ByteLength() != bimg::imageGetSize(nullptr, width, height, 1, false, false, depth, format))
+                throw std::runtime_error{"The data size does not match width, height, depth and format"};
```
bghgary, line 1132: nit:
```suggestion
            if (data.ByteLength() != bimg::imageGetSize(nullptr, width, height, 1, false, false, depth, format))
            {
                throw std::runtime_error{"The data size does not match width, height, depth and format"};
            }
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -711,19 +714,20 @@ namespace Babylon
                 bgfx::UniformInfo info{};
                 bgfx::getUniformInfo(uniforms[index], info);
                 auto itStage = uniformStages.find(info.name);
-                uniformInfos.emplace(std::make_pair<std::string, UniformInfo>(info.name, {itStage == uniformStages.end() ? uint8_t{} : itStage->second, uniforms[index]}));
+                auto& handle = uniforms[index];
+                uniformInfos.emplace(std::make_pair<std::string, UniformInfo>(info.name, {itStage == uniformStages.end() ? uint8_t{} : itStage->second, handle}));
+                uniformSizes[uniforms[index].idx] = static_cast<size_t>(info.num);
```
bghgary, line 719: I think you can store this info inside UniformInfo so that you don't need a second map?

Plugins/NativeEngine/Source/NativeEngine.h
bghgary, line 82: What is the `std::max` for?

SergioRZMasson: In the case where the unordered_map was not yet initialized for the handle.idx value it will return 0 for maxLenght. std::max prevents the length value from been set to 0.

bghgary: Can that happen? Isn't this a problem if it happens?

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: I think you would have to propagate the info to UniformValue and then it should work. Let's chat offline.

SergioRZMasson: Changed the code so that MaxElementSize is now part of UniformInfo struct.

Plugins/NativeEngine/Source/NativeEngine.h
SergioRZMasson: With the updated codebase this is no longer required. Now we check if UniformInfo exists in the unordered map, if it does it should have valid data.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -701,7 +702,9 @@ namespace Babylon
             throw Napi::Error::New(info.Env(), ex.what());
         }
 
-        static auto InitUniformInfos{[](bgfx::ShaderHandle shader, const std::unordered_map<std::string, uint8_t>& uniformStages, std::unordered_map<std::string, UniformInfo>& uniformInfos) {
+        static auto InitUniformInfos
+        {
+            [](bgfx::ShaderHandle shader, const std::unordered_map<std::string, uint8_t>& uniformStages, std::unordered_map<std::string, UniformInfo>& uniformInfos, std::unordered_map<uint16_t, size_t>& uniformSizes){
```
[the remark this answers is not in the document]
SergioRZMasson: Latest commit was formatted using clang-format

bghgary (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1124,8 +1124,12 @@ namespace Babylon
...
         const auto generateMips = info[6].As<Napi::Boolean>().Value();
         const auto invertY = info[7].As<Napi::Boolean>().Value();
-        const auto samplingMode = info[8].As<Napi::Number>().Int32Value();
-        const auto textureType = info[10].As<Napi::Number>().Int32Value();
+
+         if (generateMips)
+            throw std::runtime_error{"Texture 2D array currently do not support mipmaps."};
+
+         if (invertY)
+             throw std::runtime_error{"Texture 2D array currently do not support invert Y."};
```
bghgary, line 1132: nit: we don't typically use single line if statements, also missing a space
```suggestion
         if (generateMips)
         {
             throw std::runtime_error{"Texture 2D array currently do not support mipmaps."};
         }

         if (invertY)
         {
             throw std::runtime_error{"Texture 2D array currently do not support invert Y."};
         }
```

PR #986, opened 2022-01-27
Updated npm package to latest version of babylonjs, fixed texture flag bug on 2D Texture arrays

### LINKED ISSUE #923: Morph targets are broken

SergioRZMasson:
- Updated npm package to latest version of babylonjs, 
- Fixed texture flag bug on 2D Texture arrays. The BGFX_CAPS_TEXTURE_2D_ARRAY flag should not be set to the texture flags property, if should only be used on to call bgfx::createTexture2D.
- Fixes #923


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1137,9 +1137,8 @@ namespace Babylon
 
         texture->Width = width;
         texture->Height = height;
-        texture->CreationFlags = BGFX_TEXTURE_NONE;
-        texture->Flags = BGFX_TEXTURE_NONE | BGFX_SAMPLER_NONE | BGFX_CAPS_TEXTURE_2D_ARRAY;
-        texture->Handle = bgfx::createTexture2D(width, height, generateMips, depth, Cast(format), texture->Flags);
+        uint64_t flags{BGFX_TEXTURE_NONE | BGFX_SAMPLER_NONE | BGFX_CAPS_TEXTURE_2D_ARRAY};
```
bghgary, line 1140: I can't leave a comment for code not being changed, but you need to throw Napi::Error instead of std::runtime_error above.

SergioRZMasson: Replaced std::exception with Napi::Error in the latest commit.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1106,7 +1106,7 @@ namespace Babylon
         const auto bytes{static_cast<uint8_t*>(data.ArrayBuffer().Data()) + data.ByteOffset()};
         if (data.ByteLength() != bimg::imageGetSize(nullptr, width, height, 1, false, false, 1, format))
         {
-            throw std::runtime_error{"The data size does not match width, height, and format"};
+            Napi::Error::New(Env(), "The data size does not match width, height, and format").ThrowAsJavaScriptException();
```
bghgary, line 1109: this or you need to return to prevent the following code from running
```suggestion
            throw Napi::Error::New(Env(), "The data size does not match width, height, and format");
```

bghgary (changes requested):

SergioRZMasson (commented):
Napi::Error is been thrown in the latest commit.

bghgary (approved):

PR #990, opened 2022-02-01
Thin Instances

CedricGuillemet:
This PR needs https://github.com/BabylonJS/Babylon.js/pull/11900 that's in beta.6. package.json has been updated to reflect this change. VT need also this PR : https://github.com/BabylonJS/Babylon.js/pull/11917 this PR will be updated to reflect the change (NPM update) but it won't have any impact on Native code.
It will be difficult to fully understand changes without reading this comment. 

# Context:
bgfx instancing is quite limited: only 5 vec4 attributes can be used for instancing (world matrix + color basically). These attributes named i_data0 .. 4 and D3D semantic texcoord3..7 are bound with the instance buffer data when available.
No more than 5 vec4 special streams! see https://github.com/bkaradzic/bgfx/issues/2008#issuecomment-578819111
Most of this work would have been unnecessary if it was possible to set `instance divisor` to any stream. Like OpenGL/WebGL allows (and I guess, any other graphic API). In bjs, any stream can have a divisor==1. 
More over, streams for instancing are stored in a different buffer (with a lifetime of 1 frame). Where bjs treats any stream (instance or vertex buffer) the same.

The change is divided in 2 parts:
- make difference in recorded vertex streams so Native Engine knows if it must be stored as an instance buffer or a regular vertex buffer.
- Find and modify vertex attributes in shaders so bgfx can bind them as instanced buffer. This work is dependant on the rendering API.

# Vertex Buffer vs Instance Buffer

Now, instance divisor is forwarded to native when recording.
If the divisor is 1, there are pushed into the `m_vertexBufferInstanceRecords` map. A divisor == 0 is a regular vertex stream. When binding a vertex array for rendering, these buffer arrays are merged together into a bgfx instance buffer. This instance buffer lives for 1 frame (no destruction/storage). All the instance streams must be grouped inside this instance buffer in this pattern:
[i_dataA[0], i_dataB[0], i_dataC[0]....
i_dataA[1], i_dataB[1], i_dataC[1]...]

## Direct3D
Vec4 datas are stored in reverse order of semantics: https://github.com/bkaradzic/bgfx/blob/4581f14cd481bad1e0d6292f0dd0a6e298c2ee18/src/renderer_d3d11.cpp#L2701
Vertex buffer is stored in ascending order in `VertexArray` because `m_vertexBufferInstanceRecords` is `std::map<bgfx::Attrib::Enum, InstanceVertexBufferRecord>` so vertex buffer iteration is in reverse order in `VertexArray::SetVertexBuffers` as pointer to bgfx instance array is increased linearly. Same for OpenGL.

## Metal  and OpenGL
Order of vec4 is the same as the location. No reverse iterator here.

# Shader attribute definition

## BabylonJS
Before https://github.com/BabylonJS/Babylon.js/pull/11900 there was 1 attribute named `color`. it either contained Mesh Vertex colors OR instance Color. This was simple: depending on divisor and bound vertex buffer, the same shader code use the color to render the mesh. It was not possible to have Mesh vertex color AND Instance color at the same time but it was not a need. But as native shaders need name to properly bind the attribute, name needs to be different and `instanceColor` has been added.

## Direct3D
Quite simple: in `VertexVaryingInTraverserD3D`, add new attribute names from BabylonJS, bind them as expected semantic for bgfx and rename them as well. As bgfx binds with semantics, the new `instanceColor` has the appropriate semantic.

## OpenGL
Binding is based on name, hence the renaming (i_data0...i_data5). As long as i_dataX and location ares in increasing order, it's fine. First instance attribute, first served.

## Metal
Binding based on attribute location. The difference here is bgfx is binding the last streams (a mix between OGL and D3D). It means, if for some reason (order of attributes in bjs shaders) attributes appear like this to the shader:
- location 0: instanceColor
- location 1: position
- location 2: world0
- location 3: world1
- location 4: world2
- location 5: world3
It needs to be remapped to :
- location 0: position
- location 1: instanceColor
- location 2: world0
- location 3: world1
- location 4: world2
- location 5: world3
That's why there are 2 passes. 1st one for stream, second for instanced ones.

## Other technical possibilities

It's always possible to have a different technical path but with time allowed, this is the best I could do.

Other possibilities:
- modify bgfx to allow instance divisor per stream. It would be in sync with webGL. This tiny bit of work cover a change in API contract and also reimplementing instances for a few platforms (D3D11, D3D12, Metal, OpenGL, Vulkan) while keeping the backward compatibility. This also implies discussing the work with Branimir before.
- discard HW instancing and add loop on instance instead. As many `submit` call as mesh to render. Instance attributes need to be replaced by uniform. This means some work in Shader Traversers and also some questions: if a vertex buffer is used with a divisor ==1, how is it known when compiling shader that 1 (or more) particular attribute must be replaced by a uniform?  Then, what's the performance will be? Especially on mobile.

# Tests
All platforms tested OK!

Related : https://github.com/BabylonJS/BabylonReactNative/issues/232 #992


Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
```diff
@@ -495,89 +495,189 @@ namespace Babylon::ShaderCompilerTraversers
...
+            static void Traverse(TProgram& program, IdGenerator& ids, std::unordered_map<std::string, std::string>& replacementToOriginalName)
+            {
+                auto intermediate{program.getIntermediate(EShLangVertex)};
+                VertexVaryingInTraverserMetal traverser{};
+                intermediate->getTreeRoot()->traverse(&traverser);
+                VertexVaryingInTraverserMetal::Traverse(intermediate, ids, replacementToOriginalName, traverser);
+            }
+
+        private:
+            static void Traverse(TIntermediate* intermediate, IdGenerator& ids, std::unordered_map<std::string, std::string>& replacementToOriginalName, VertexVaryingInTraverser& traverser)
```
bghgary, line 622: I think it is easier if you made this an instance function instead of static since you are passing in the instance anyways?

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -35,44 +35,54 @@ namespace Babylon
...
+        if (divisor == 1)
         {
-            layout.add(attrib, static_cast<uint8_t>(numElements), bgfx::AttribType::Float);
-            layout.m_stride = static_cast<uint16_t>(sizeof(float) * numElements);
-            vertexBuffer->PromoteToFloats(attribType, numElements, byteOffset, byteStride);
+            // bgfx allows instancing on 5 vec4 attributes
+            if (m_vertexBufferInstanceRecords.size() < 5)
+            {
+                m_vertexBufferInstanceRecords[attrib] = {vertexBuffer, byteOffset, byteStride, static_cast<uint16_t>(sizeof(float) * numElements)};
+            }
```
bghgary, line 47: Probably should at least log a warning if more than 5?

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -85,7 +95,50 @@ namespace Babylon
 
     void VertexArray::SetVertexBuffers(bgfx::Encoder* encoder, uint32_t startVertex, uint32_t numVertices)
     {
-        for (auto& pair : m_vertexBufferRecords)
+        const bgfx::Caps* caps = bgfx::getCaps();
+
+        // Check if instancing is supported.
+        const bool instancingSupported = 0 != (BGFX_CAPS_INSTANCING & caps->supported);
```
bghgary, line 101: Doesn't Babylon.js already check this? Also, if we are dropping things, we should log a warning.

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -85,7 +95,50 @@ namespace Babylon
...
+            // copy instance datas
+            uint8_t* data{instanceDataBuffer.data};
+            uint32_t offset{};
+
+            // reverse because of bgfx also reverting : https://github.com/bkaradzic/bgfx/blob/4581f14cd481bad1e0d6292f0dd0a6e298c2ee18/src/renderer_d3d11.cpp#L2701
+#if D3D11 || D3D12 
+            for (auto iter = m_vertexBufferInstanceRecords.rbegin(); iter != m_vertexBufferInstanceRecords.rend(); ++iter)
+#else
+            for (auto iter = m_vertexBufferInstanceRecords.cbegin(); iter != m_vertexBufferInstanceRecords.cend(); ++iter)
+#endif
```
bghgary, line 128: would it be okay to always reverse?

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -85,7 +95,50 @@ namespace Babylon
...
+            }
+
+            // create instance datas. Instance Data Buffer is transient.
+            bgfx::allocInstanceDataBuffer(&instanceDataBuffer, instanceCount, instanceStride);
+
+            // copy instance datas
+            uint8_t* data{instanceDataBuffer.data};
+            uint32_t offset{};
+
+            // reverse because of bgfx also reverting : https://github.com/bkaradzic/bgfx/blob/4581f14cd481bad1e0d6292f0dd0a6e298c2ee18/src/renderer_d3d11.cpp#L2701
```
bghgary, line 123: nit:
```suggestion
            // reverse because of bgfx also reversing : https://github.com/bkaradzic/bgfx/blob/4581f14cd481bad1e0d6292f0dd0a6e298c2ee18/src/renderer_d3d11.cpp#L2701
```

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -85,7 +95,50 @@ namespace Babylon
...
+            for (auto iter = m_vertexBufferInstanceRecords.rbegin(); iter != m_vertexBufferInstanceRecords.rend(); ++iter)
+#else
+            for (auto iter = m_vertexBufferInstanceRecords.cbegin(); iter != m_vertexBufferInstanceRecords.cend(); ++iter)
+#endif
+            {
+                const auto& element{iter->second};
+                const auto* source{element.Buffer->GetBytes().data()};
+                for (uint32_t instance = 0; instance < instanceCount; instance++)
+                {
+                    memcpy(data + instance * instanceStride + offset, source + instance * element.Stride + element.Offset, element.ElementSize);
```
bghgary, line 134:
```suggestion
                    std::memcpy(data + instance * instanceStride + offset, source + instance * element.Stride + element.Offset, element.ElementSize);
```

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -85,7 +95,50 @@ namespace Babylon
...
     void VertexArray::SetVertexBuffers(bgfx::Encoder* encoder, uint32_t startVertex, uint32_t numVertices)
     {
-        for (auto& pair : m_vertexBufferRecords)
+        const bgfx::Caps* caps = bgfx::getCaps();
+
+        // Check if instancing is supported.
+        const bool instancingSupported = 0 != (BGFX_CAPS_INSTANCING & caps->supported);
+        if (!m_vertexBufferInstanceRecords.empty() && instancingSupported)
+        {
+            bgfx::InstanceDataBuffer instanceDataBuffer{};
```
bghgary, line 104: I think it would cleaner if this function (that creates a `bgfx::InstanceDataBuffer` from a span of VertexBuffer) is encapsulated in a public static function under VertexBuffer, so that you don't have to expose the bytes with `GetBytes`. It would also make this code a lot easier to read.

bghgary (commented):
Mostly nits.

Plugins/NativeEngine/Source/VertexArray.cpp
CedricGuillemet: logged with `JsConsoleLogger`

Plugins/NativeEngine/Source/VertexArray.cpp
CedricGuillemet: warning logged. We don't expose many things from bgfx caps up to bjs. there is already an issue for that.

Plugins/NativeEngine/Source/ShaderCompilerTraversers.cpp
CedricGuillemet: Done.

Plugins/NativeEngine/Source/VertexArray.cpp
CedricGuillemet: I tried to find a solution for this without success. Because of attribute types ordering and bgfx, I didn't find a solution that fixes it.

Plugins/NativeEngine/Source/VertexArray.cpp
CedricGuillemet: Done.

bghgary: This isn't exactly what I meant, but we can fix it later.

bghgary (approved):

Plugins/NativeEngine/Source/VertexArray.cpp
bghgary: Is the answer to my question no then?

CedricGuillemet: yes, it's no.

Issue #604, opened 2021-02-16
Crash when attempting to load model with large number of meshes


Alex-MSFT:
When loading a GLB with more than 4096 meshes BabylonNative causes the app to crash (see attached GLB).

The error is due to us running over the hard coded Vertex buffer limit of 4096 then in setVertexBuffer it is crashing on:
BGFX_CHECK_HANDLE("setVertexBuffer", s_ctx->m_vertexBufferHandle, _handle);

Relevant Traces:
BGFX WARNING: Failed to allocate vertex buffer handle (BGFX_CONFIG_MAX_VERTEX_BUFFERS, max: 4096).
BGFX CHECK Invalid handle. setVertexBuffer handle: 65535 (max 4096)

Callstack:
bx::debugBreak() debug.cpp:55
Babylon::BgfxCallback::fatal(char const*, unsigned short, bgfx::Fatal::Enum, char const*) BgfxCallback.cpp:33
bgfx::fatal(char const*, unsigned short, bgfx::Fatal::Enum, char const*, ...) bgfx.cpp:443
bgfx::Encoder::setVertexBuffer(unsigned char, bgfx::VertexBufferHandle, unsigned int, unsigned int, bgfx::VertexLayoutHandle) bgfx.cpp:3684
bgfx::setVertexBuffer(unsigned char, bgfx::VertexBufferHandle, unsigned int, unsigned int, bgfx::VertexLayoutHandle) bgfx.cpp:4987
auto Babylon::VertexBufferData::SetAsBgfxVertexBuffer(unsigned char, unsigned int, bgfx::VertexLayoutHandle) const::'lambda'(auto)::operator()<bgfx::VertexBufferHandle>(auto) const NativeEngine.cpp:324
void Babylon::VariantHandleHolder<bgfx::VertexBufferHandle, bgfx::DynamicVertexBufferHandle>::DoForHandleTypes<Babylon::VertexBufferData::SetAsBgfxVertexBuffer(unsigned char, unsigned int, bgfx::VertexLayoutHandle) const::'lambda'(auto) const, Babylon::VertexBufferData::SetAsBgfxVertexBuffer(unsigned char, unsigned int, bgfx::VertexLayoutHandle) const::'lambda0'(auto) const>(auto&, Babylon::VertexBufferData::SetAsBgfxVertexBuffer(unsigned char, unsigned int, bgfx::VertexLayoutHandle) const::'lambda0'(auto) const&) const NativeEngine.cpp:166
[...]

PR #991, opened 2022-02-02
Large mesh count

SergioRZMasson:
Fixes issue #604, once the number of vertex buffers goes higher than the max supported by bgfx, babylon native will log a warning message in the JS console and will not render those meshes.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -624,6 +626,11 @@ namespace Babylon
         const bool normalized = info[7].As<Napi::Boolean>().Value();
 
         vertexArray->RecordVertexBuffer(vertexBuffer, location, byteOffset, byteStride, numElements, type, normalized);
```
bghgary, line 628: I think maybe it would be better for RecordVertexBuffer to return a boolean instead of a separate function? It would mean vertexBuffer::CreateHandle will need to return a boolean also, which makes perfect sense.

Also, does index buffer need the same check?

Core/JsRuntime/Include/Babylon/JsConsoleLogger.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
```
bghgary, line 1: I'm not sure this helper class belongs in JsRuntime, but I also don't have a better idea. Maybe we can keep it in NativeEngine folder for now? @syntheticmagus Any thoughts?

Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -142,11 +147,17 @@ namespace Babylon
     {
         if (m_dynamic)
         {
-            encoder->setVertexBuffer(stream, m_dynamicHandle, startVertex, numVertices, layoutHandle);
+            if (bgfx::isValid(m_dynamicHandle))
```
bghgary, line 150: You can do this check once at the top. Yes, it's a union, but they are both identical layout, so it should be fine. We already do this the CreateHandle function.

Plugins/NativeEngine/Source/NativeEngine.cpp
SergioRZMasson: Yep, we will need to check index buffer creation as well. It can fail the same way Vertex buffer does. One car argue that vertex buffer creation will fail first, but since we will not prevent the user from creating new Index buffers once they run into an error on creating Vertex buffers, we should handle that for index buffers as well.

Plugins/NativeEngine/Source/VertexBuffer.cpp
SergioRZMasson: Updated on latest commit.

Plugins/NativeEngine/Source/NativeEngine.cpp
SergioRZMasson: Updated commit with validation for Index buffer as well. Also, removed the IsValid method from VertexBuffer and made CreateHandle return false if the handler was invalid. However, in order to completely skip binding the vertex buffers and index buffers during draw call it was required to add a IsValid() method to VertexArray, this will be set to false if any associated index of vertex buffer is invalid, this way we completely skip the draw if the VertexArray has invalid data.

Core/JsRuntime/Include/Babylon/JsConsoleLogger.h
syntheticmagus: _Hmm_, funky. It feels like it wants to be part of the `Console` Polyfill, but other native Components can't take dependencies on Polyfills.

Maybe the console functionality needs to be split in two, with the core functionality housed in a Plugin (or even Core) that other native Components can depend on, and the `Console` Polyfill that exposes the functionality to JS depends on that. Thoughts?

bghgary: Some JS engine exposes a Console object by default (like JSC) and if we are integrating with an existing environment, it may already have Console (like React Native). I don't think it makes sense to split the functionality.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1410,6 +1417,9 @@ namespace Babylon
 
         if (m_boundVertexArray != nullptr)
         {
+            if (!m_boundVertexArray->IsValid())
+                return;
```
bghgary, line 1421: It seems like you are already storing the valid boolean inside vertex array and thus you can check inside the SetIndexBuffer and SetVertexBuffers function instead?

bghgary, line 1421: What happens if DrawInternal is called without the vertex array binding the buffers?

Core/JsRuntime/Include/Babylon/JsConsoleLogger.h
syntheticmagus: In that case, is it worth encapsulating this at all? It looks like both the places it's currently being called from have access to `Napi::CallbackInfo`s, and I would expect anywhere we'd want to log to JS from should at least have _some_ way to get a `Napi::Env`. `info.Env().Global().Get("console").As<Napi::Object>().Get("log").As<Napi::Function>().Call(...)` isn't exactly concise, but it's not bespoke, requires no infrastructure, directly hooks into whatever the JS has decided to do for the console, and will only matter speed-wise if we're spewing logs, which will be expensive anyway.

syntheticmagus: I guess another option if we wanted encapsulate to avoid the verbosity would be to add something to `napi-extensions` alongside napi_pointer.h, maybe napi_console.h or something. Then we could hide the above console-getting in helper functions like `Napi::Console::Log(Napi::Env, const char*)`, `Napi::Console::Warn(Napi::Env*, const char*)`, etc.

bghgary: I don't like the napi_console.h in napi itself. Maybe a utility dependency outside of napi would be fine. This is effectively a set of helper functions. I think maybe caching the `console.*` functions in NativeEngine as members isn't too bad. Basically, any code that wants to log to the console can do something similar. @SergioRZMasson any thoughts?

SergioRZMasson: I think this is too little of a utility to add to napi, it was designed just to avoid typing the entire thing every time and to avoid other people that might use the log to forget to check if console exist. Performance wise I think this can be considered a special case, a heathy application should not be throwing's warnings at the console anyway, I didn't want to start putting extra stuff on NativeEngine's members, unless they are essential.

SergioRZMasson: I think a simple helper function into a utility header(maybe inside NativeEngine project itself?) can be considered fine?

bghgary: I don't have a preference. I was thinking you can move these files into NativeEngine in my original reply. That would be fine with me.

Plugins/NativeEngine/Source/NativeEngine.cpp
SergioRZMasson: In the cases I tested nothing bad seams to happen but I think this might be render API specific. Moreover, I think if the shader is calling this with m_boundVertexArray not been null it should be expecting it to be bind, so I didn't want to fall into undefined behavior.

bghgary: As long as it doesn't crash, I'm okay with undefined behavior here since it can't draw it correctly regardless. We already can get into a state where the vertex array isn't bound and that's normal. If we are okay with this, then you should be able to simplify the code a bit without exposing IsValid.

SergioRZMasson:
Removed IsValid logic from VertexArrays, validation is handled by VertexBuffer and IndexBuffer during Set call and DrawInternal can be called without VertexBuffer and IndexBuffers been set.

bghgary (approved):
LGTM

PR #994, opened 2022-02-07
Vertex and Index memory optimization - replaced std::vector by strong ref.

### LINKED ISSUE #726: Babylon Native uses a lot more memory than Babylon.js when loading glbs
For this 16mb model: https://raw.githubusercontent.com/microsoft/experimental-pcf-control-assets/master/bird_of_paradise_blue.glb

Loading and rendering it with Babylon.js (within a WKWebView (WebKit)) uses about 180mb of main memory.
Loading and rendering it within Babylon Native (within an iOS React Native app) uses about 500mb of main memory.

This seems like a substantial difference and can create a lot of memory pressure on memory constrained mobile devices. We should try to understand why Babylon Native seems to be so much more memory intensive.

SergioRZMasson:
## Overview

Babylon native was consuming more memory compared to browsers when loading large glbs. One major factor that was contributing to that was due to the fact that VertexBuffers and IndexBuffers were storing data into an array while it was waiting to get sent to bgfx. During the call to CreateHandle there classes will clean the array once the memory is no longer used by bgfx. However, when calling .std::vector::clear reallocation is not guaranteed to happen, so the vertex memory copy will keep using the memory event after data has already been sent to bgfx.

## Solution

By holding a strong reference to the JS object by using Napi::Reference and storying the vertex and index data into a gsl::span we can completely avoid one extra copy of data. The downside is that VertexBuffer and IndexBuffer has to also store the JsRuntimeScheduler in order to release the Napi::Reference from JS thread once bgfx is done with it. 

## Other technical possibilities

We could also make sure the std::vector releases the memory, but that would continue creating a temporary memory copy that could lead to memory usage spikes, been release once bgfx upload was done. 

## Changes
- Fixes #726 
- VertexBuffer and IndexBuffer std::vector was replaced by gsl::span and Napi::Reference
- PromoteToFloats was rewriten to use gsl::span instead of std::vector.


Plugins/NativeEngine/Source/IndexBuffer.cpp
```diff
@@ -1,11 +1,19 @@
 #include "IndexBuffer.h"
+#include <arcana/threading/task.h>
+#include <arcana/threading/task_schedulers.h>
 
 namespace Babylon
 {
-    IndexBuffer::IndexBuffer(gsl::span<uint8_t> bytes, uint16_t flags, bool dynamic)
+    IndexBuffer::IndexBuffer(gsl::span<uint8_t> bytes,
```
tymlipari, line 7: Since `bytes` is derived from `dataRefs` at the construction site, should this parameter be removed and `m_bytes` is calculated inside the constructor?

Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -1,16 +1,22 @@
...
+    void PromoteToFloats(gsl::span<uint8_t>& bytes, uint32_t numElements, uint32_t byteOffset, uint32_t byteStride)
     {
         const size_t count = bytes.size() / byteStride;
         const size_t destinationSize = count * numElements * sizeof(float);
-        if (destinationSize != bytes.size()) // ensure both vectors have different size
+        if (destinationSize != static_cast<size_t>(bytes.size())) // ensure both vectors have different size
         {
-            std::vector<uint8_t> newBytes(destinationSize);
-            float* destination = reinterpret_cast<float*>(newBytes.data());
+            uint8_t* newBytes = new uint8_t[destinationSize];
```
tymlipari, line 16: Nit: I'd consider using a `std::unique_ptr<uint8_t[]>` here rather than a raw array to get the RAII cleanup and avoid the need for manual deletion a few lines down. In practice it doesn't really matter because I don't think any of these lines below can throw and thus leak it, but it still reads somewhat cleaner.

bghgary:
There are some issues with the code, but before getting into that, did we verify that it is okay to keep a reference to the JS buffer? What does WebGL do on a browser? What happens if someone changes this buffer on the JS side while it is rendering? What does Babylon.js do with the CPU buffer? What happens if someone calls [`scene.clearCachedVertexData`](https://doc.babylonjs.com/divingDeeper/scene/reducingMemoryUsage#removing-cached-vertex-data)?

Sorry for so many questions, but I want to be sure this won't break anything. 😏

Plugins/NativeEngine/Source/VertexBuffer.cpp
syntheticmagus: I'd actually recommend continuing to use the `std::vector` as that is (from what I understand) our current best practice for this scenario. Is there a particular reason to switch away from that pattern?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -562,13 +562,16 @@ namespace Babylon
...
         const uint32_t byteOffset = info[1].As<Napi::Number>().Uint32Value();
         const uint32_t byteLength = info[2].As<Napi::Number>().Uint32Value();
         const bool is32Bits = info[3].As<Napi::Boolean>().Value();
         const bool dynamic = info[4].As<Napi::Boolean>().Value();
 
+        auto reference = Napi::Persistent(info[0]);
+
         const uint16_t flags = (is32Bits ? BGFX_BUFFER_INDEX32 : 0);
-        IndexBuffer* indexBuffer = new IndexBuffer{gsl::make_span(static_cast<uint8_t*>(bytes.Data()) + byteOffset, byteLength), flags, dynamic};
+        IndexBuffer* indexBuffer = new IndexBuffer{gsl::make_span(static_cast<uint8_t*>(bytes.Data()) + byteOffset, byteLength), reference, flags, m_runtimeScheduler, dynamic};
```
syntheticmagus, line 574: Doesn't this take a hard dependency on the location -- the raw pointer address -- of the underlying data remaining constant into the future? Is that part of the contract for all of the JS engines? Are any of them allowed to move their data?

Plugins/NativeEngine/Source/VertexBuffer.cpp
SergioRZMasson: I agree with you @syntheticmagus, using ```std::vector``` will be the best practice and will not change the final result.

Plugins/NativeEngine/Source/NativeEngine.cpp
SergioRZMasson: I think this will be related to @bghgary 's question on if it is ok to hold a reference to the JS buffer. That implies what happens when this data is overridden, release and if JS engine might move that buffer at some point. I will do some research to see if this information can be found anywhere.

SergioRZMasson:
PR was updated with second approach been used. Looking at V8's implementation for WebGL they seam to be creating a copy of vertex and index data as well. The updated PR only makes sure that the ```std::vector``` is properly released once it is no longer been used.

Plugins/NativeEngine/Source/IndexBuffer.cpp
SergioRZMasson: Approach has changed, so this is no longer valid.

Plugins/NativeEngine/Source/VertexBuffer.cpp
SergioRZMasson: Approach has changed, so this is no longer valid.

Plugins/NativeEngine/Source/IndexBuffer.cpp
```diff
@@ -3,7 +3,7 @@
 namespace Babylon
 {
     IndexBuffer::IndexBuffer(gsl::span<uint8_t> bytes, uint16_t flags, bool dynamic)
-        : m_bytes{bytes.data(), bytes.data() + bytes.size()}
+        : m_bytes{std::vector<uint8_t>{bytes.data(), bytes.data() + bytes.size()}}
```
bghgary, line 6: nit:
```suggestion
        : m_bytes{{bytes.data(), bytes.data() + bytes.size()}}
```

Plugins/NativeEngine/Source/IndexBuffer.cpp
```diff
@@ -66,11 +64,11 @@ namespace Babylon
 
         auto releaseFn = [](void*, void* userData)
         {
-            auto* bytes = reinterpret_cast<std::vector<uint8_t>*>(userData);
-            bytes->clear();
+            auto* bytes = reinterpret_cast<std::optional<std::vector<uint8_t>>*>(userData);
```
bghgary, line 67: nit:
```suggestion
            auto* bytes = reinterpret_cast<decltype(m_bytes)*>(userData);
```

bghgary (approved):

Plugins/NativeEngine/Source/IndexBuffer.cpp
syntheticmagus: I haven't tried to actually run this, but this looks like it's going to construct the `std::vector` outside the `std::optional`, then move (or copy, not sure how it'll handle the R-value-ness) that vector into the optional. Do we think it's better to do this, or is it better to construct the `std::optional` empty and `emplace` the vector inside in the body of the constructor? If this line moves instead of copies, it probably doesn't make much difference cost-wise, but thought it was worth asking.

Plugins/NativeEngine/Source/IndexBuffer.cpp
```diff
@@ -32,9 +32,7 @@ namespace Babylon
                 bgfx::destroy(m_handle);
             }
         }
-
-        m_bytes.clear();
-
+        m_bytes = {};
```
syntheticmagus, line 35: nit: `reset()`? We reset these elsewhere.
```suggestion
        m_bytes.reset();
```

syntheticmagus (approved):

Plugins/NativeEngine/Source/IndexBuffer.cpp
bghgary: It has to move unless the implementation of `std::optional` is broken. The `std::vector` is unnamed which should end up as a rvalue reference.

PR #995, opened 2022-02-10
Canvas Fix for demo

CedricGuillemet:
- Global Alpha
- Clipping
- Text measurement
- Clear rect blending

### DIFF
--- a/Polyfills/Canvas/Source/MeasureText.cpp
+++ b/Polyfills/Canvas/Source/MeasureText.cpp
@@ -11,50 +11,18 @@
 
 namespace Babylon::Polyfills::Internal
 {
-    static constexpr auto JS_CONSTRUCTOR_NAME = "TextDimension";
-
     Napi::Value MeasureText::CreateInstance(Napi::Env env, Context* context, const std::string& text)
     {
-        Napi::HandleScope scope{ env };
-        Napi::Function func = DefineClass(
-            env,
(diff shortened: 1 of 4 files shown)


Polyfills/Canvas/Source/MeasureText.cpp
```diff
@@ -20,8 +20,10 @@ namespace Babylon::Polyfills::Internal
             env,
             JS_CONSTRUCTOR_NAME,
             {
-                InstanceAccessor("width", &MeasureText::GetWidth, &MeasureText::SetWidth),
-                InstanceAccessor("height", &MeasureText::GetHeight, &MeasureText::SetHeight),
+                InstanceAccessor("width", &MeasureText::GetWidth, nullptr),
+                InstanceAccessor("height", &MeasureText::GetHeight, nullptr),
+                InstanceAccessor("actualBoundingBoxLeft", &MeasureText::ActualBoundingBoxLeft, nullptr),
+                InstanceAccessor("actualBoundingBoxRight", &MeasureText::ActualBoundingBoxRight, nullptr)
```
ryantrem, line 26: Can we just use regular properties with values instead of accessor functions? Accessor functions are way slower, and all they are doing here is returning a value, so just putting that value directly into a js property eliminates that overhead.

CedricGuillemet: Fixed.

PR #1003, opened 2022-02-16
Android Pixel Density

CedricGuillemet:
In browsers, the canvas size is divided by the pixel density.
On my iPhone 11 with a pixel density of 2, the canvas size is 414x640 instead of native resolution.
On my Android, the density is 3. The surface area is 1/9 of native res.

In Native:
On iOS, the system (via `contentScaleFactor`) handles everything. 
On Android, the view size is returned in pixels, without taking care of pixel density. This PR adds its support.
The surfaceView bounds are not changed but the OpenGL area is resized with density factor. At presentation time, 
https://github.com/bkaradzic/bgfx/blob/3d4bd88c0635b614b453aac1003cc56a11d1ccf0/src/glcontext_egl.cpp#L375
Is taking care of setting the native window size so it gets rescaled by the compositor into the surfaceView.

### DIFF
--- a/Apps/Playground/Android/app/src/main/java/BabylonNative/BabylonView.java
+++ b/Apps/Playground/Android/app/src/main/java/BabylonNative/BabylonView.java
@@ -18,7 +18,7 @@ public class BabylonView extends FrameLayout implements SurfaceHolder.Callback2,
     private Activity mCurrentActivity;
     private final SurfaceView primarySurfaceView;
     private final SurfaceView xrSurfaceView;
-
+    private final float pixelDensityScale = getResources().getDisplayMetrics().density;
     public BabylonView(Context context, ViewDelegate viewDelegate) {
         this(context, viewDelegate, (Activity)viewDelegate);
     }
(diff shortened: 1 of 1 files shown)


ryantrem:
I'm unclear how this relates to `window.devicePixelRatio` and `Engine.setHardwareScalingLevel`. We already polyfill `window.devicePixelRatio`, and the BJS [docs](https://doc.babylonjs.com/typedoc/classes/babylon.engine#sethardwarescalinglevel) say "By default the hardware scaling level is computed from the window device ratio." Additionally, I see the [constructor in nativeEngine.js](https://github.com/BabylonJS/Babylon.js/blob/master/src/Engines/nativeEngine.ts#L910-L912) sets `this._hardwareScalingLevel` and then calls `resize`, so conceptually I would have thought the scaling is already covered.

CedricGuillemet:
There 3 things:
- engine hardwareScalingLevel
- window.devicePixelRatio
- output surface resolution

Let's forget hardwareScalingLevel as it's unchanged (always 1 until you change it manually)
`window.devicePixelRatio` is properly polyfilled and returns the correct value from the system when queried.
And the output surface resolution in pixels. In Chrome for example, the output size is computed based on devicePixelRatio. For a 1920x1080 screen and a ratio of 2, the canvas size will be 960x540. The rendering is being performed in that resolution target.
For iOS, the same thing is happening, the UIView dimension is computed based on screen res and ratio.
On Android, it was not. output res is the screen res, without taking care of that ratio.

ryantrem (approved):

PR #1004, opened 2022-02-16
XRRay object Check

### LINKED ISSUE #997: iOS ARKit XRHitTest "Invalid Argument" Error In Playground Demo
**Describe the bug**
On the latest master commit, I'm having issues across multiple devices (iPhone 13 Pro / iPhone 11 / iPad Pro) enabling both AR and XRHitTest in experience.js with the playground demo.

**To Reproduce**
1. My configuration variables in experience.js look like the following:

```
var wireframe = false;
var turntable = false;
var logfps = false;
var ibl = false;
var rtt = false;
var vr = false;
var ar = true;
var xrHitTest = true;
var xrFeaturePoints = false;
var meshDetection = false;
var text = false;
var hololens = false;
var cameraTexture = false;
```

**Expected behavior**
Expect to see the cube in AR and be able to place it.

**Actual behavior**
AR scene never loads, get the following error in the log:

```
2022-02-10 12:32:32.206247-0500 Playground[382:7972] Error: Invalid argument
2022-02-10 12:32:32.206310-0500 Playground[382:7972] Invalid argument
```

Also want to note that the ARKit scene does load if only "ar" is true and "xrHitTest" is false.

**Other**
 - Tested on various iOS versions from iOS 13 - iOS 15

CedricGuillemet:
fixes #997 

It was broken for every platforms. It didn't check the napi object could be unwrapped (exception) .

### DIFF
--- a/Plugins/NativeXr/Source/NativeXr.cpp
+++ b/Plugins/NativeXr/Source/NativeXr.cpp
@@ -929,12 +929,11 @@ namespace Babylon
 
         class XRRigidTransform : public Napi::ObjectWrap<XRRigidTransform>
         {
+            static constexpr auto JS_CLASS_NAME = "XRRigidTransform";
             // static constexpr size_t VECTOR_SIZE = 4;
             static constexpr size_t MATRIX_SIZE = 16;
 
         public:
-            static constexpr auto JS_CLASS_NAME = "XRRigidTransform";
-
             static void Initialize(Napi::Env env)
             {
                 Napi::HandleScope scope{env};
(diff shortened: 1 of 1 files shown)


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -1338,9 +1339,11 @@ namespace Babylon
                 {
                     auto argumentObject{info[0].As<Napi::Object>()};
 
-                    XRRigidTransform* transform{XRRigidTransform::Unwrap(argumentObject)};
-                    if (transform != nullptr)
+                    const auto XRRigidTransformFunction = Env().Global().Get(XRRigidTransform::JS_CLASS_NAME).As<Napi::Function>();
+
+                    if (argumentObject.InstanceOf(XRRigidTransformFunction))
```
ryantrem, line 1344: Under what circumstances will this not be true?

bghgary: InstanceOf is a very slow check. Is there some property you can check instead?

bghgary (approved):

Plugins/NativeXr/Source/NativeXr.cpp
CedricGuillemet: Using property check instead.

PR #1011, opened 2022-02-25
Improve Pointer Support

PolygonalSun:
PR for Improved Pointer Support.  This PR adds supports for additional mouse buttons and the mouse wheel for Mouse implementations and multi-touch support for touch inputs.

- [x] Win32 Mouse
- [x] UWP Mouse
- [x] MacOS Mouse
- [x] iOS Touch
- [x] Android Touch
- [x] Linux Mouse

### DIFF
--- a/Apps/Playground/UWP/App.cpp
+++ b/Apps/Playground/UWP/App.cpp
@@ -213,18 +213,27 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
 {
     if (m_nativeInput != nullptr)
     {
-        const auto& point = args->CurrentPoint->RawPosition;
+        const auto& point = args->CurrentPoint;
+        const auto& position = point->RawPosition;
 
-        m_nativeInput->MouseMove(static_cast<int>(point.X), static_cast<int>(point.Y));
+        m_nativeInput->MouseMove(static_cast<int>(position.X), static_cast<int>(position.Y));
 
         /*
-         * Note: Because PointerPressed and PointerReleased don't fire for additional buttons presses/releases,
-         * we need an alternative way to track other button clicks.  In both of the below functions, we have
-         * three booleans that track each button's state and then we run these two functions to update any buttons
-         * that we may have missed.
+         * Note: Because PointerPressed only first for the first button press (and no additional ones)
+         * and PointerReleased only fires when all buttons have been released, we need an alternative way
+         * to track additional button presses.
+         * (See Definition - PointerPressed: https://docs.microsoft.com/en-us/uwp/api/windows.ui.core.corewindow.pointerpressed?view=winrt-22000)
+         * (See Remarks - PointerReleased: https://docs.microsoft.com/en-us/uwp/api/windows.ui.core.corewindow.pointerreleased?view=winrt-22000)
+         *
+         * In both of the below functions, we have three booleans that track each button's state and then we
+         * run these two functions to update any buttons that we may have missed.
          */
-        OnPointerPressed(nullptr, args);
-        OnPointerReleased(nullptr, args);
+
+        if (point->IsInContact)
+        {
+            OnPointerPressed(nullptr, args);
+            OnPointerReleased(nullptr, args);
+        }
     }
 }
 
--- a/Apps/Playground/UWP/App.cpp
+++ b/Apps/Playground/UWP/App.cpp
@@ -91,6 +91,9 @@ void App::SetWindow(CoreWindow^ window)
     window->PointerReleased +=
         ref new TypedEventHandler<CoreWindow^, PointerEventArgs^>(this, &App::OnPointerReleased);
 
+    window->PointerWheelChanged +=
+        ref new TypedEventHandler<CoreWindow^, PointerEventArgs^>(this, &App::OnPointerWheelChanged);
+
     window->KeyDown +=
         ref new TypedEventHandler<CoreWindow^, KeyEventArgs^>(this, &App::OnKeyPressed);
 }
@@ -211,7 +214,17 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
     if (m_nativeInput != nullptr)
     {
         const auto& point = args->CurrentPoint->RawPosition;
+
         m_nativeInput->MouseMove(static_cast<int>(point.X), static_cast<int>(point.Y));
+
+        /*
+         * Note: Because PointerPressed and PointerReleased don't fire for additional buttons presses/releases,
+         * we need an alternative way to track other button clicks.  In both of the below functions, we have
+         * three booleans that track each button's state and then we run these two functions to update any buttons
+         * that we may have missed.
+         */
+        OnPointerPressed(nullptr, args);
+        OnPointerReleased(nullptr, args);
     }
 }
 
@@ -220,7 +233,21 @@ void App::OnPointerPressed(CoreWindow^, PointerEventArgs^ args)
     if (m_nativeInput != nullptr)
     {
         const auto& point = args->CurrentPoint->RawPosition;
-        m_nativeInput->MouseDown(0, static_cast<int>(point.X), static_cast<int>(point.Y));
+        if (!m_leftPressed && args->CurrentPoint->Properties->IsLeftButtonPressed)
+        {
+            m_leftPressed = true;
+            m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::LEFT_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
+        if (!m_middlePressed && args->CurrentPoint->Properties->IsMiddleButtonPressed)
+        {
+            m_middlePressed = true;
+            m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::MIDDLE_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
+        if (!m_rightPressed && args->CurrentPoint->Properties->IsRightButtonPressed)
+        {
+            m_rightPressed = true;
+            m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::RIGHT_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
     }
 }
 
@@ -229,8 +256,27 @@ void App::OnPointerReleased(CoreWindow^, PointerEventArgs^ args)
     if (m_nativeInput != nullptr)
     {
         const auto& point = args->CurrentPoint->RawPosition;
-        m_nativeInput->MouseUp(0, static_cast<int>(point.X), static_cast<int>(point.Y));
+        if (m_leftPressed && !args->CurrentPoint->Properties->IsLeftButtonPressed)
+        {
+            m_leftPressed = false;
+            m_nativeInput->MouseUp(Babylon::Plugins::NativeInput::LEFT_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
+        if (m_middlePressed && !args->CurrentPoint->Properties->IsMiddleButtonPressed)
+        {
+            m_middlePressed = false;
+            m_nativeInput->MouseUp(Babylon::Plugins::NativeInput::MIDDLE_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
+        if (m_rightPressed && !args->CurrentPoint->Properties->IsRightButtonPressed)
+        {
+            m_rightPressed = false;
+            m_nativeInput->MouseUp(Babylon::Plugins::NativeInput::RIGHT_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
     }
+}
+void App::OnPointerWheelChanged(CoreWindow^, PointerEventArgs^ args)
+{
+    const auto& delta = args->CurrentPoint->Properties->MouseWheelDelta;
+    m_nativeInput->MouseWheel(Babylon::Plugins::NativeInput::MOUSEWHEEL_Y_ID, delta);
 }
 
 void App::OnKeyPressed(CoreWindow^ window, KeyEventArgs^ args)
--- a/Apps/Playground/UWP/App.h
+++ b/Apps/Playground/UWP/App.h
@@ -32,6 +32,7 @@ private:
     void OnPointerMoved(Windows::UI::Core::CoreWindow^ sender, Windows::UI::Core::PointerEventArgs^ args);
     void OnPointerPressed(Windows::UI::Core::CoreWindow^ sender, Windows::UI::Core::PointerEventArgs^ args);
     void OnPointerReleased(Windows::UI::Core::CoreWindow^ sender, Windows::UI::Core::PointerEventArgs^ args);
+    void OnPointerWheelChanged(Windows::UI::Core::CoreWindow ^ sender, Windows::UI::Core::PointerEventArgs ^ args);
     void OnKeyPressed(Windows::UI::Core::CoreWindow^ sender, Windows::UI::Core::KeyEventArgs^ args);
 
     // DisplayInformation event handlers.
@@ -50,6 +51,9 @@ private:
     Windows::Foundation::Collections::IVectorView<Windows::Storage::IStorageItem^>^ m_files;
     bool m_windowClosed;
     bool m_windowVisible;
+    bool m_leftPressed{false};
+    bool m_middlePressed{false};
+    bool m_rightPressed{false};
     float m_displayScale{ 1.f };
 };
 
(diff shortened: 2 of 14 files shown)


Plugins/NativeInput/Source/Shared/NativeInput.cpp
```diff
@@ -188,7 +212,7 @@ namespace Babylon::Plugins
                 {
                     return;
                 }
-            }
+            }*/
```
ryantrem, line 215: If we are sure block this is no longer needed, just remove it.

Plugins/NativeInput/Source/Shared/NativeInput.cpp
```diff
@@ -111,6 +130,11 @@ namespace Babylon::Plugins
         PointerMove(MOUSE_POINTER_ID, x, y, DeviceType::Mouse);
     }
 
+    void NativeInput::Impl::MouseWheel(uint32_t wheelAxis, int32_t scrollValue)
+    {
+        PointerScroll(MOUSE_POINTER_ID, wheelAxis, scrollValue, DeviceType::Mouse);
```
ryantrem, line 135: Is there such thing as "pointer scroll" for anything other than a mouse?

Plugins/NativeInput/Source/Shared/NativeInput.cpp
```diff
@@ -234,6 +258,19 @@ namespace Babylon::Plugins
...
 
+    void NativeInput::Impl::PointerScroll(uint32_t pointerId, uint32_t scrollAxis, int32_t scrollValue, DeviceType deviceType)
+    {
+        m_runtimeScheduler([pointerId, scrollAxis, scrollValue, deviceType, this]()
+            {
+                std::vector<int32_t>& deviceInputs{GetOrCreateInputMap(deviceType, pointerId, {scrollAxis})};
+                SetInputState(deviceType, pointerId, scrollAxis, scrollValue, deviceInputs, true);
+
+                m_eventDispatcher.tick(arcana::cancellation::none());
+                SetInputState(deviceType, pointerId, scrollAxis, 0, deviceInputs, true);
```
ryantrem, line 269: Why is this set back to 0?

Apps/Playground/UWP/App.cpp
```diff
@@ -91,6 +91,9 @@ void App::SetWindow(CoreWindow^ window)
     window->PointerReleased +=
         ref new TypedEventHandler<CoreWindow^, PointerEventArgs^>(this, &App::OnPointerReleased);
 
+    window->PointerWheelChanged +=
+        ref new TypedEventHandler<CoreWindow ^, PointerEventArgs ^>(this, &App::OnPointerWheelChanged);
```
ryantrem, line 95:
```suggestion
        ref new TypedEventHandler<CoreWindow^, PointerEventArgs^>(this, &App::OnPointerWheelChanged);
```

Plugins/NativeInput/Source/Shared/NativeInput.cpp
PolygonalSun: Not really.  Closest thing would be a touchpad gesture but that would still register as a mouse wheel scroll

Plugins/NativeInput/Source/Shared/NativeInput.cpp
PolygonalSun: Good catch!  I'll remove it

Plugins/NativeInput/Source/Shared/NativeInput.cpp
PolygonalSun: Mousewheel events don't have a "wheelup" event so if we don't zero it out once the frame is complete, anything that is polling or checking this value will continue to see this as a non-zero value and can keep applying it.  Because of this, we need to zero it out when the inputChanged callback is complete

Apps/Playground/UWP/App.cpp
PolygonalSun: Removed while space

Apps/Playground/UWP/App.cpp
```diff
@@ -206,12 +209,16 @@ void App::OnWindowClosed(CoreWindow^ sender, CoreWindowEventArgs^ args)
     Uninitialize();
 }
 
-void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
+void App::OnPointerMoved(CoreWindow^ sender, PointerEventArgs^ args)
```
bghgary, line 212: why add this?

```suggestion
void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
```

Apps/Playground/UWP/App.cpp
```diff
@@ -206,12 +209,16 @@ void App::OnWindowClosed(CoreWindow^ sender, CoreWindowEventArgs^ args)
...
 {
     if (m_nativeInput != nullptr)
     {
         const auto& point = args->CurrentPoint->RawPosition;
+
         m_nativeInput->MouseMove(static_cast<int>(point.X), static_cast<int>(point.Y));
+        // Once move is handled, check for button changes
+        OnPointerPressed(sender, args);
+        OnPointerReleased(sender, args);
```
bghgary, line 221: Why do you need this? Maybe an explanation in the comment would be good.

Apps/Playground/UWP/App.cpp
```diff
@@ -220,7 +227,21 @@ void App::OnPointerPressed(CoreWindow^, PointerEventArgs^ args)
...
+        if (!m_middlePressed && args->CurrentPoint->Properties->IsMiddleButtonPressed)
+        {
+            m_middlePressed = true;
+            m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::MIDDLE_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
+        if (!m_rightPressed && args->CurrentPoint->Properties->IsRightButtonPressed)
+        {
+            m_rightPressed = true;
+            m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::RIGHT_MOUSE_BUTTON_ID, static_cast<int>(point.X), static_cast<int>(point.Y));
+        }
```
bghgary, line 244: I don't understand why you need to keep track of the state manually. Are the UWP events not correct? Can you do the same as Win32?

Apps/Playground/iOS/ViewController.swift
```diff
@@ -1,6 +1,71 @@
...
+    private var _activeTouchIds: Array<Int> = [-1, -1, -1, -1, -1, -1, -1, -1, -1, -1]
+    
+    override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent) {
+        super.touchesBegan(touches, with: event)
+        
+        for touch in touches {
+            let deviceSlot = _activeTouchIds.firstIndex(of: -1)
+            
+            if (deviceSlot != nil) {
+                _activeTouchIds[deviceSlot!] = touch.hash
```
bghgary, line 33: What happens if there are more than 10 ids? Should we check this? Does swift arrays automatically check?

Apps/Playground/iOS/ViewController.swift
```diff
@@ -1,6 +1,71 @@
...
+    public var onTouchDown: (Int32, Int32, Int32)->Void = {(pointerId: Int32, inX: Int32, inY: Int32) -> Void in
+        // Do nothing
+    }
+    // Callback for touch movement events
+    public var onTouchMove: (Int32, Int32, Int32)->Void = {(pointerId: Int32, inX: Int32, inY: Int32) -> Void in
+        // Do nothing
+    }
+    // Callback for touch up events
+    public var onTouchUp: (Int32, Int32, Int32)->Void = {(pointerId: Int32, inX: Int32, inY: Int32) -> Void in
+        // Do nothing
```
bghgary, line 21: Probably makes more sense for these to be private `let` variables and initialized by the constructor (`init`)

Apps/package-lock.json
```diff
@@ -1,1854 +1,1854 @@
-{
```
bghgary, line 1: Why is this file changed?

Plugins/NativeInput/Source/Shared/NativeInput.cpp
```diff
@@ -176,20 +200,6 @@ namespace Babylon::Plugins
...
-                    index != POINTER_Y_INPUT_INDEX &&
-                    index != POINTER_DELTA_HORIZONTAL_INDEX &&
-                    index != POINTER_DELTA_VERTICAL_INDEX &&
-                    index != POINTER_MOVE_INDEX &&
-                    deviceInputs[index] > 0)
-                {
-                    return;
-                }
-            }
-
```
bghgary, line 192: Why is this not necessary anymore?

Apps/Playground/UWP/App.cpp
PolygonalSun: So this was part of a complex solution that I had to implement because UWP input doesn't work the same as Win32 input.  First of all, PointerPressed was only firing for the first button pressed so no other buttons were getting picked up.  PointerReleased had the same issue but with only firing when the last button is released.  For the button state tracking, that was implemented as the only way I've found to track when a button is pressed/released so that I can properly use the correct event without firing unnecessary up/down events.

Apps/Playground/UWP/App.cpp
PolygonalSun: Explained in previous comment

Apps/Playground/iOS/ViewController.swift
PolygonalSun: If all of the entries in the `_activeTouchIds` are not -1 (all taken) the `firstIndex` should return `nil` so it will work like web and ignore additional touches

Plugins/NativeInput/Source/Shared/NativeInput.cpp
PolygonalSun: For Mouse, we return before hitting that and for Touch, we only have one button to zero and it's already zeroed so this does nothing.

Apps/Playground/UWP/App.cpp
PolygonalSun: I can flesh out the comment explaining this and I'll also remove the sender since we don't use it.

Apps/Playground/iOS/ViewController.swift
PolygonalSun: Lemme see if I can get this changed over to initialize via the constructor

Apps/package-lock.json
PolygonalSun: Looks like it was a change included when I merged master in yesterday.  Lemme see if I can fix it

PolygonalSun: Looks like it was a change included when I merged master in yesterday.  Lemme see if I can fix it

Plugins/NativeInput/Source/Shared/NativeInput.cpp
ryantrem: What is "wheelup" (or "wheeldown")? I thought this was for an axis change (e.g. scroll) as opposed to click? I must be missing something.

Plugins/NativeInput/Source/Shared/NativeInput.cpp
ryantrem: Meh I guess either way, I was just unsure since it was parameterized but I wasn't sure if we'd ever have any other kind of pointer with scrolling.

Apps/Playground/UWP/App.cpp
bghgary: I don't think it's obvious why UWP doesn't send the events. A link to the documentation which says this would be great.

bghgary: Do we need to check `IsInContact`? The sample seems to check it, but I'm not sure we need it.

https://github.com/microsoft/Windows-universal-samples/blob/main/Samples/BasicInput/cpp/2-PointerPointProperties.xaml.cpp#L55-L60

Apps/Playground/UWP/App.h
```diff
@@ -32,6 +32,7 @@ ref class App sealed : public Windows::ApplicationModel::Core::IFrameworkView
     void OnPointerMoved(Windows::UI::Core::CoreWindow^ sender, Windows::UI::Core::PointerEventArgs^ args);
     void OnPointerPressed(Windows::UI::Core::CoreWindow^ sender, Windows::UI::Core::PointerEventArgs^ args);
     void OnPointerReleased(Windows::UI::Core::CoreWindow^ sender, Windows::UI::Core::PointerEventArgs^ args);
+    void OnPointerWheelChanged(Windows::UI::Core::CoreWindow ^ sender, Windows::UI::Core::PointerEventArgs ^ args);
```
bghgary, line 35:
```suggestion
    void OnPointerWheelChanged(Windows::UI::Core::CoreWindow ^ sender, Windows::UI::Core::PointerEventArgs^ args);
```

Apps/Playground/UWP/App.h
```diff
@@ -50,6 +51,9 @@ ref class App sealed : public Windows::ApplicationModel::Core::IFrameworkView
     Windows::Foundation::Collections::IVectorView<Windows::Storage::IStorageItem^>^ m_files;
     bool m_windowClosed;
     bool m_windowVisible;
+    bool m_leftPressed{false};
+    bool m_middlePressed{false};
+    bool m_rightPressed{false};
```
bghgary, line 56: nit: apps folder uses default VS conventions
```suggestion
    bool m_leftPressed{ false };
    bool m_middlePressed{ false };
    bool m_rightPressed{ false };
```

Apps/Playground/Win32/App.cpp
```diff
@@ -401,15 +401,60 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
             SetCapture(hWnd);
```
bghgary, line 401: Does SetCapture refcount? Does this work properly if multiple buttons are down?

bghgary (commented):
Other than package-lock.json, everything else seems fine to me.

Plugins/NativeInput/Source/Shared/NativeInput.cpp
PolygonalSun: What I mean by "wheelup" is just that when you trigger a wheelevent, there no defined end event.  Furthermore, because of how we handle input values and storing, we need a way to purge stale values.  Since these values should only be used via the input changed observable, we would just purge them at the next tick.  Basically, we turned wheel input into a sort of button.

Apps/Playground/Win32/App.cpp
PolygonalSun: AFAIK, it does not.  I have modified the code to account for this.

Apps/Playground/UWP/App.h
PolygonalSun: Changed

Apps/Playground/UWP/App.h
PolygonalSun: Fixed

Apps/Playground/X11/App.cpp
```diff
@@ -239,16 +239,46 @@ int main(int _argc, const char* const* _argv)
...
+                                    nativeInput->MouseDown(Babylon::Plugins::NativeInput::LEFT_MOUSE_BUTTON_ID, xmotion.x, xmotion.y);
+                                    break;
+                                case Button2:
+                                    nativeInput->MouseDown(Babylon::Plugins::NativeInput::MIDDLE_MOUSE_BUTTON_ID, xmotion.x, xmotion.y);
+                                    break;
+                                case Button3:
+                                    nativeInput->MouseDown(Babylon::Plugins::NativeInput::RIGHT_MOUSE_BUTTON_ID, xmotion.x, xmotion.y);
+                                    break;
+                                case Button4:
+                                    nativeInput->MouseWheel(Babylon::Plugins::NativeInput::MOUSEWHEEL_Y_ID, -120);
```
bghgary, line 256: Just curious. Is 120 a specific number that matches browser behavior?

Apps/Playground/Win32/App.cpp
```diff
@@ -398,20 +399,82 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
...
             break;
         }
         case WM_LBUTTONUP:
         {
             if (nativeInput != nullptr)
             {
-                nativeInput->MouseUp(0, GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam));
+                nativeInput->MouseUp(Babylon::Plugins::NativeInput::LEFT_MOUSE_BUTTON_ID, GET_X_LPARAM(lParam), GET_Y_LPARAM(lParam));
+            }
+            if (buttonRefCount-- == 1)
```
bghgary, line 418: It would be more symmetrical and very slightly faster if you did this. (Ditto everywhere else)
```suggestion
            if (--buttonRefCount == 0)
```

Apps/Playground/X11/App.cpp
PolygonalSun: That is the current value of deltaY when you trigger a wheel event in Chrome.

Apps/Playground/Win32/App.cpp
PolygonalSun: Changed

bghgary (approved):

Issue #987, opened 2022-01-28
Support Images in NativeEngine.createImageBitmap


ryantrem:
We need this as part of the XR image tracking work. Specifically, image tracking has a list of reference images:

https://github.com/BabylonJS/Babylon.js/blob/master/src/XR/features/WebXRImageTracking.ts#L22

These can either be strings (e.g. a url) or an `ImageBitmap` that has already been loaded. In the case of a string, the XR image tracking feature will use it as the source of an `Image`, load and decode that image, then pass the `Image` to `Engine.createImageBitmap` to create an `ImageBitmap`:

https://github.com/BabylonJS/Babylon.js/blob/master/src/XR/features/WebXRImageTracking.ts#L186-L197

Babylon Native already has a native polyfill for `Image`:

https://github.com/BabylonJS/BabylonNative/blob/master/Polyfills/Canvas/Source/Image.cpp

And this already supports loading from a string by setting the `src` property, and supports the `onload` callback, but it doesn't have the `decode` function today. That said, I believe the image decoding is done implicitly during load, so I think the first thing we need is:

- [ ] Add a `decode` function to the [native Image polyfill](https://github.com/BabylonJS/BabylonNative/blob/master/Polyfills/Canvas/Source/Image.cpp) that returns a `Promise` that is completed after `onload` is called.

Once that is done, then the XR image tracking code can [pass the resulting `Image` into the call to `createImageBitmap`](https://github.com/BabylonJS/Babylon.js/blob/master/src/XR/features/WebXRImageTracking.ts#L193). Babylon Native already has a native implementation of `createImageBitmap`:

https://github.com/BabylonJS/BabylonNative/blob/master/Plugins/NativeEngine/Source/NativeEngine.cpp#L1511-L1542

However as is it only knows how to deal with the input arg being an `ArrayBuffer`. It needs to also be able to handle an `Image` being passed in. From a WebAPI standpoint, `Image` is basically an opaque object (you can't get direct access to the underlying image data buffer), which means whoever is implementing WebAPI needs to understand the internals of that opaque type. In the case of Babylon Native, it defines the `Image` type (through the native polyfill) and therefore understands the internal type. Given this, I think the next thing we need is:

[...]

PR #1021, opened 2022-03-04
Support NativeCanvas images in createImageBitmap

Alex-MSFT:
This PR is making changes to prepare for supporting the WebXR Image tracking in BabylonNative.

The first step of this is to support createImageBitmap for native Images loaded by the NativeCanvas implementation as outlined in Issue #987.  The primary changes are persisting the bimg pointer on the CanvasImage object, and recognizing when we are passed in an Image object rather than an ArrayBuffer during createImageBitmap.

I am also adding a callback event for onerror so we can properly handle errors that occur during bitmap loading.

The matching Babylon.JS changes are being reviewed here: https://github.com/BabylonJS/Babylon.js/pull/12097


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1518,20 +1518,36 @@ namespace Babylon
...
-        if (!data.ByteLength())
+            image = ParseImage(m_allocator, gsl::make_span(static_cast<uint8_t*>(data.Data()), data.ByteLength()));
+            allocatedImage = true;
+        }
+        else if (info[0].IsObject())
         {
-            throw Napi::Error::New(env, "CreateImageBitmap array buffer is empty.");
+            // If this is an object, then check if it has the _imageContainer property defined.
+            auto imageObject = info[0].As<Napi::Object>();
+            if (imageObject.Has("_imageContainer")) {
```
ryantrem, line 1541:
```suggestion
            if (imageObject.Has("_imageContainer"))
            {
```

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -91,22 +94,51 @@ namespace Babylon::Polyfills::Internal
         request.SendAsync().then(m_runtimeScheduler, *m_cancellationSource, [env{info.Env()}, this, request{std::move(request)}](arcana::expected<void, std::exception_ptr> result) {
             if (result.has_error())
             {
+                if (!m_onerrorHandlerRef.IsEmpty())
+                {
+                    m_onerrorHandlerRef.Call({});
```
ryantrem, line 99: Why is an empty object passed here?

Alex-MSFT: Yes we should be returning thanks!

Added the error object to the callback params.

ryantrem (approved):

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -39,6 +41,7 @@ namespace Babylon::Polyfills::Internal
         : Napi::ObjectWrap<NativeCanvasImage>{info}
         , m_runtimeScheduler{JsRuntime::GetFromJavaScript(info.Env())}
         , m_cancellationSource{std::make_shared<arcana::cancellation_source>()}
+        , m_jsThis{Napi::Persistent(info.This().As<Napi::Object>())}
```
bghgary, line 44: I'm not sure this is okay to do. Won't this cause the garbage collector to never collect this object?

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -91,22 +94,56 @@ namespace Babylon::Polyfills::Internal
...
             {
-                throw Napi::Error::New(env, result.error());
+                const auto error = Napi::Error::New(env, result.error());
+                if (!m_onerrorHandlerRef.IsEmpty())
+                {
+                    m_onerrorHandlerRef.Call({error.Value()});
+                    return;
+                }
+
+                error.ThrowAsJavaScriptException();
```
bghgary, line 104: This is still not quite right. We still should interrupt the function when there is no error handler.

```suggestion
                const auto error = Napi::Error::New(env, result.error());
                if (m_onerrorHandlerRef.IsEmpty())
                {
                    error.ThrowAsJavaScriptException();
                }
                else
                {
                    m_onerrorHandlerRef.Call({error.Value()});
                }
                
                return;
```

bghgary: Maybe also put lines 97-105 in a private helper function.

bghgary (changes requested):

Polyfills/Canvas/Source/Image.cpp
Alex-MSFT: Fix this by using this->Value() instead.  I noticed at least one place where we use a similar pattern in NativeXR, will update that when I make the WebXRImageTracking PR.

Polyfills/Canvas/Source/Image.cpp
Alex-MSFT: Move this code to a helper function, but am still leaving it as ThrowAsJavaScriptException() rather than a native C++ exception, since that would crash the app since it is unhandled.  I don't think it makes sense to crash the whole app in the event a user provides a bad src string.

Dependencies/napi/napi-jsi/include/napi/napi-inl.h
```diff
@@ -2115,19 +2115,11 @@ inline void CallbackInfo::SetData(void* data) {
...
-
-  //Reference<Object>* instanceRef = instance;
-  //*instanceRef = Reference<Object>(env, ref);
-  //throw std::runtime_error{"TODO"};
+{
+  napi_env env{callbackInfo.Env()};
+  jsi::Runtime& rt{env->rt};
+  Reference<Object>* instanceRef = static_cast<Reference<Object>*>(this);
+  *instanceRef = Reference<Object>{env, {env, std::move(static_cast<const jsi::Value&>(callbackInfo.This()).asObject(rt))}};
 }
```
bghgary, line 2123: Looks perfect!

Just a super tiny nit: open brace needs to be on the previous line.

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -94,26 +94,29 @@ namespace Babylon::Polyfills::Internal
...
+                const auto error = Napi::Error::New(env, result.error());
                 if (!m_onerrorHandlerRef.IsEmpty())
                 {
-                    m_onerrorHandlerRef.Call({});
+                    m_onerrorHandlerRef.Call({error.Value()});
+                    return;
                 }
 
-                throw Napi::Error::New(env, result.error());
+                throw error;
```
bghgary: Where does this error end up if it's thrown? Should this be the same as the other ones?

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -94,26 +94,29 @@ namespace Babylon::Polyfills::Internal
...
-                    m_onerrorHandlerRef.Call({});
+                    m_onerrorHandlerRef.Call({error.Value()});
                     return;
                 }
 
-                Napi::Error::New(env, "Image with provided source returned empty response.").ThrowAsJavaScriptException();
+                error.ThrowAsJavaScriptException();
             }
 
             m_imageContainer = bimg::imageParse(&m_allocator, buffer.data(), static_cast<uint32_t>(buffer.size_bytes()));
```
bghgary, line 109: Looks like you missed one from lines 123 to 132.

Polyfills/Canvas/Source/Image.cpp
Alex-MSFT: I think you're looking at an old version of the PR?

bghgary (approved):

Polyfills/Canvas/Source/Image.cpp
bghgary: > am still leaving it as ThrowAsJavaScriptException() rather than a native C++ exception

Yup, I wasn't commenting on that part.

bghgary (approved):

PR #1026, opened 2022-03-08
Fix Metal Validation with Stencil in Canvas

### LINKED ISSUE #996: Metal validation Error with Space Demo
Related to surfaces and Stencil.

CedricGuillemet:
fixes #996 
For some 2D shapes, stencil is needed. As it was used without a rendertarget supporting it, it brought a Metal validation error.
Adding a Depth/Stencil for Canvas.

### DIFF
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -108,9 +108,18 @@ namespace Babylon::Polyfills::Internal
     {
         if (m_dirty)
         {
-            auto handle = bgfx::createFrameBuffer(static_cast<uint16_t>(m_width), static_cast<uint16_t>(m_height), bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT);
+            std::array<bgfx::TextureHandle, 2> textures{
+                bgfx::createTexture2D(static_cast<uint16_t>(m_width), static_cast<uint16_t>(m_height), false, 1, bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT),
+                bgfx::createTexture2D(static_cast<uint16_t>(m_width), static_cast<uint16_t>(m_height), false, 1, bgfx::TextureFormat::D24S8, BGFX_TEXTURE_RT)};
+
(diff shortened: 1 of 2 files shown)


Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -108,9 +108,18 @@ namespace Babylon::Polyfills::Internal
     {
         if (m_dirty)
         {
-            auto handle = bgfx::createFrameBuffer(static_cast<uint16_t>(m_width), static_cast<uint16_t>(m_height), bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT);
+            std::array<bgfx::TextureHandle, 2> textures{
+                bgfx::createTexture2D(static_cast<uint16_t>(m_width), static_cast<uint16_t>(m_height), false, 1, bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT),
+                bgfx::createTexture2D(static_cast<uint16_t>(m_width), static_cast<uint16_t>(m_height), false, 1, bgfx::TextureFormat::D24S8, BGFX_TEXTURE_RT)};
```
Alex-MSFT, line 113: Does this never need to be cleaned up in the event we create a new frame buffer, or will the lifetime be owned by the frame buffer handle?

Alex-MSFT (approved):

Polyfills/Canvas/Source/Canvas.cpp
bghgary: I believe the attachments of the framebuffer owns the handles, but we should check to be sure.

bghgary (approved):

Polyfills/Canvas/Source/Canvas.cpp
CedricGuillemet: Yes, FB attachments will be destroyed. 3rd parameter of `bgfx::createFrameBuffer` : https://bkaradzic.github.io/bgfx/bgfx.html#_CPPv4N4bgfx17createFrameBufferE7uint8_tPK13TextureHandleb

PR #975, opened 2022-01-19
External resource

syntheticmagus:
ExternalResource component, building on top of (and originally part of) the changes in #967. Uploaded here to be the basis for continuing work on external resources.

PR #967, opened 2022-01-10
WIP Graphics context

syntheticmagus:
Still working on the patterns, naming, and integration, so this is not ready for review yet.

PR #1028, opened 2022-03-11
Graphics Infrastructure rework

SergioRZMasson:
## Overview

The PR is a direct continuation of the work developed in PR #967 and #975. It is targeted to make the necessary changes to allow users to provide native textures to BabylonNative and consume them in javascript. This PR brings all changes from #967, fixes to make it work with current version from master as well as the additions of the GraphicsInterops library, that provides the component that should be used by projects that wish no interoperate with BabylonNative using the native graphics API.

## Graphics

```Babylon::Graphics::Device``` is now what was previously named ```Babylon::Graphics```, also ```Babylon::Graphics::DeviceContext``` is now what ```Babylon::GraphicsImpl``` was. For external consumers it is now possible to create a Babylon::Graphics::Device by specifying a ```Babylon::Graphics::DeviceConfiguration``` that allows a native graphics device to be provided. This graphics device will be used internally by bgfx. The main goal is that resources created using the native graphics device can be shared by the consuming application and BabylonNative. 

## ExternalTexture

This library project was created to hold all components for interoperability between BabylonNative and the native Graphics API texture objects. 

### ExternalTexture class

ExternalTexture is a class that can be constructed using a native graphics API handle and then be used to create Babylon::Graphics::TextureData and send this data to javascript environment using ExternalTexture::AddToContext(). It is important to notice that the constructor will only store the necessary to convert the native texture to the bgfx internal texture, but it will not create bgfx object. 

This class also assumes that the native texture was created using the same graphics device used to create the Babylon::Device. Therefore, it is assumed that ```Babylon::Graphics::Device``` was created by providing a ```Babylon::Graphics::DeviceConfiguration``` structure with the appropriate Device.

Only when ExternalTexture::AddToContext() is called that the following things will happen: 

-  Napi::Promise will be created, since the actual creation of the texture will happen over an entire frame. 
- During BeforeRender a bgfx texture will be created.
- During AfterRender this bgfx texture will be replaced to use the native provided texture instead. A TextureData will be created.
- In the JS thread a Napi::Pointer<TextureData> will be created and the Napi::Promise::Deferred will be resolved. (The bgfx texture will be released by the Napi::Pointer finalization callback)

With this approach the same native texture can be used to create multiple JS objects, each one will have ownership over the bgfx texture representation but not the actual native texture.

## Changes made to PR #967

In the original implementation for DeviceImpl::UpdateWindow and DeviceImpl::UpdateContext the m_state.Bgfx.InitState.platformData was been reset. This is a problem, since if the user wants to create a Device using Device::Create(const DeviceConfiguration& config) in order to provide the native graphics context, and than assign the render windows using Device::UpdateWindow(const WindowConfiguration& config) it would erase the provided graphics context.

Also, bgfx does not allow the user to reset the graphics context once it is initialized, trying to do it so will lead to a runtime error. This way, if the user wants to call Device::UpdateWindow at any time, the value in bgfx::PlatformData::context must not be reset from the one that was used to initialize bgfx.

## Final considerations

- This PR will also require a PR for BabylonReactNative.
- This PR does not require changes in the JS side.
- Users will need to handle texture creation on the JS side using the TextureData handle with code such the sample above:
```
function loadNativeTexture(nativePromise, width, height, format)
{
    nativePromise.then((nativeTexture) =>
    {
       const internalTexture = new BABYLON.InternalTexture(
            engine,
            BABYLON.InternalTextureSource.Unknown,
            true
        );
        internalTexture._hardwareTexture = new BABYLON.WebGLHardwareTexture(
            nativeTexture
        );

        internalTexture.generateMipMaps = false;
        internalTexture.baseWidth = internalTexture.width = width;
        internalTexture.baseHeight = internalTexture.height = height;
        internalTexture.isReady = true;
        internalTexture.format = format;
        inTexture = new BABYLON.ThinTexture(internalTexture);
});
```

### CODE REVIEW

File: Core/GraphicsInterops/Include/Shared/Babylon/GraphicsInterops/ExternalTexture.h [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+        uint32_t GetWight();
+        uint32_t GetHeight();
+        uint32_t GetFormat();
+
+    private:
+        struct Impl;
+        std::unique_ptr<Impl> m_impl{};
+
+        ExternalTexture(std::unique_ptr<Impl>);
```
syntheticmagus, line 29: If I remember correctly, this was here to be used by a static factory. With the static factory gone, do we need this private constructor?

Core/GraphicsInterops/Source/ExternalTexture.cpp
```diff
@@ -0,0 +1,52 @@
...
+    uint32_t ExternalTexture::GetHeight()
+    {
+        return m_impl->GetHeight();
+    }
+
+    uint32_t ExternalTexture::GetFormat() 
+    {
+        return m_impl->GetFormat();
+    }
+}
```
syntheticmagus, line 52: nit:
```suggestion
}

```
GitHub complains if there's no newline at the end of a file, so we usually try to have that, though it's not a big deal.

Core/GraphicsInterops/Source/ExternalTextureImpl.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <Babylon/Graphics/Texture.h>
+#include <Babylon/Graphics/DeviceContext.h>
+#include <Babylon/JsRuntime.h>
+#include <Babylon/JsRuntimeScheduler.h>
+
+#include "ExternalTextureImpl.h"
+
+#include <napi/env.h>
+#include <napi/napi_pointer.h>
+
```
syntheticmagus, line 11: nit:
```suggestion
```

Core/GraphicsInterops/Source/ExternalTextureImpl.h
```diff
@@ -0,0 +1,35 @@
...
+{
+    struct ExternalTexture::Impl
+    {
+        Impl(TextureType nativeTexture);
+        ~Impl();
+
+        //TODO: Change this to take a context.
+        Napi::Promise AddToContext(Napi::Env& env);
+
+        uint32_t GetWight();
```
syntheticmagus, line 20:
```suggestion
        uint32_t GetWidth();
```

Core/GraphicsInterops/Source/ExternalTextureImpl.h
```diff
@@ -0,0 +1,35 @@
...
+        uint32_t GetFormat();
+
+    private:
+        bgfx::TextureFormat::Enum m_format{};
+        uintptr_t m_nativeTexture{};
+        uint32_t m_width{};
+        uint32_t m_height{};
+        uint32_t m_mips{};
+
+        void GetNativeTextureData(TextureType nativeTexture);
```
syntheticmagus, line 33: nit: Having a void-returning function called "Get" is a little confusing, took me a moment to figure out what it was doing. `CacheNativeTextureData(...)`, maybe? Or `ReadPropertiesFromNativeTexture(...)`, or something like that?

syntheticmagus: Alternatively, if the only reason this is a member is so that it can set the local property caches, it might also be possible to have it return a struct (or even `auto`) containing the values rather than being a member. That way the name `GetNativeTextureData(...)` fits because it's fetching and returning the native texture data, the "ExternalTexturePlatform.cpp"s could become headers instead, and `ExternalTexture::Impl` would just call into them to get the values out. In fact, I think that would remove the reason to separate out `ExternalTexture::Impl` into its own file with its own header; it could go back to being an implementation detail inside "ExternalTexture.cpp," unless there's another reason that I'm missing to have it externalized.

syntheticmagus (commented):
Looks pretty good to me overall. Not sure about the `GraphicsInterop` name; in my experience, *interop* typically refers to things that are much more low-level than an external resources consumption mechanism. Would `GraphicsResources`, `GraphicsExternals`, or something like that be potential alternatives?

Core/GraphicsInterops/Source/ExternalTextureImpl.cpp
```diff
@@ -0,0 +1,91 @@
...
+            auto format = m_format;
+            auto mips = m_mips;
+
+            arcana::make_task(context.BeforeRenderScheduler(), arcana::cancellation_source::none(), [&context, &runtime, deferred{std::move(deferred)}, width, height, format, mips, nativeHandler]() 
+                { 
+                    auto textureHandle = bgfx::createTexture2D(static_cast<uint16_t>(width), static_cast<uint16_t>(height), mips != 1, 1, format, BGFX_TEXTURE_RT);
+
+                    arcana::make_task(context.AfterRenderScheduler(), arcana::cancellation_source::none(), [&context, &runtime, deferred{std::move(deferred)}, textureHandle, width, height, format, mips, nativeHandler]() 
+                    {
+                            if (bgfx::overrideInternal(textureHandle, nativeHandler) == 0)
```
syntheticmagus, line 46: nit: Formatting here, and in this file overall, is a little wonky.

Core/GraphicsInterops/Include/Shared/Babylon/GraphicsInterops/ExternalTexture.h
SergioRZMasson: You are right @syntheticmagus , the way we are currently doing things do not require this indirection. I have completely removed ExternalTextureImpl and made everything part of ExternalTexture.

Core/GraphicsInterops/Source/ExternalTextureImpl.h
SergioRZMasson: I like renaming it to ReadPropertiesFromNativeTexture, I don't like creating a new struct just for storing this values temporarily. I also removed ```ExternalTexture::Impl``` completely and moved the platform specific code to match the way it is been handled by Graphics, I think it is better organized now.

Core/AppRuntime/Source/AppRuntime.cpp
```diff
@@ -36,10 +36,12 @@ namespace Babylon
...
-    void AppRuntime::Dispatch(std::function<void(Napi::Env)> func)
+    void AppRuntime::Dispatch(Dispatchable<void(Napi::Env)> func)
     {
-        m_workQueue->Append([this, func{std::move(func)}](Napi::Env env) {
-            Execute([this, env, func{std::move(func)}] {
+        m_workQueue->Append([this, func{std::move(func)}](Napi::Env env) mutable
+            {
+                Execute([this, env, func{std::move(func)}]() mutable
+                {
                 try
```
bghgary, line 45:
```suggestion
        m_workQueue->Append([this, func{std::move(func)}](Napi::Env env) mutable {
            Execute([this, env, func{std::move(func)}]() mutable {
                try
```

Core/Graphics/CMakeLists.txt
```diff
@@ -64,22 +70,31 @@ target_compile_definitions(Graphics
...
     INTERFACE bx)
 
-target_compile_definitions(GraphicsInternal
+target_compile_definitions(GraphicsDeviceContext
     INTERFACE NOMINMAX)
-
-if(APPLE)
-    # compile as objc++
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -x objective-c++")
-endif()
```
bghgary, line 85: What happened to this?

Core/GraphicsInterops/Source/ExternalTexture_D3D11.cpp
```diff
@@ -0,0 +1,191 @@
...
+namespace Babylon::Graphics
+{
+    //For refrence see renderer_d3d11.cpp line 203.
+    bgfx::TextureFormat::Enum NativeToBGFXImageFormat(DXGI_FORMAT format)
+    {
+        switch (format)
+        {
+            case DXGI_FORMAT_BC1_UNORM:
+            case DXGI_FORMAT_BC1_UNORM_SRGB:
+                return bgfx::TextureFormat::Enum::BC1;
```
bghgary, line 19: The srgb part needs to be translated also. They are not the same.

bghgary:
> Not sure about the GraphicsInterop name; in my experience, interop typically refers to things that are much more low-level than an external resources consumption mechanism. Would GraphicsResources, GraphicsExternals, or something like that be potential alternatives?

I'm okay with either of these proposed names, except the plural part. Why plural?

Core/GraphicsInterops/Source/ExternalTexture.cpp
```diff
@@ -0,0 +1,85 @@
...
+                                auto jsObject = Napi::Pointer<TextureData>::Create(env, textureData, [textureData]
+                                    {
+                                        bgfx::destroy(textureData->Handle);
+                                        delete textureData;
+                                    });
+
+                                deferred.Resolve(jsObject);
+                            });
+                    });
+            });
```
syntheticmagus, line 66: Formatting here's still kind of weird; you might need to format manually if we don't have time to dive further into clang format right now.

Core/AppRuntime/Source/WorkQueue.h
```diff
@@ -19,16 +19,31 @@ namespace Babylon
...
         void Resume();
         void Run(Napi::Env);
 
     private:
+        template<typename CallableT>
+        void Invoke(CallableT& callable)
+        {
+            callable(m_env.value());
+        }
+
```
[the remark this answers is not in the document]
SergioRZMasson: After Justin's PR with the fix for the issues with the copy able is merged we can remove this.

Apps/Playground/CMakeLists.txt
```diff
@@ -110,14 +110,15 @@ target_link_to_dependencies(Playground
...
-    PRIVATE NativeInput
     PRIVATE NativeEngine
+    PRIVATE NativeInput
     PRIVATE NativeOptimizations
     PRIVATE Console
     PRIVATE Window
     PRIVATE ScriptLoader
     PRIVATE XMLHttpRequest
     PRIVATE Canvas
+    PRIVATE GraphicsInterops
```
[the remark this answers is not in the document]
SergioRZMasson: We don't, it was there for testing, it is removed now

Core/Graphics/CMakeLists.txt
SergioRZMasson: This was not required, it was copied from the Graphics library and is now removed.

Core/GraphicsInterops/Source/ExternalTexture.cpp
SergioRZMasson: Fixed formatting in the latest version.

Plugins/ExternalTexture/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    target_link_to_dependencies(ExternalTexture
+        PRIVATE android
+        PRIVATE AndroidExtensions)
+endif()
+
+target_link_to_dependencies(ExternalTexture
+    PRIVATE napi_extensions
+    PRIVATE JsRuntime
+    PRIVATE JsRuntimeInternal
+    PRIVATE GraphicsDevice
```
syntheticmagus, line 26: Should these be public dependencies? Your header depends directly on JsRuntime and GraphicsDevice concepts, at least.

syntheticmagus (approved):

Core/GraphicsInterops/Source/ExternalTexture_D3D11.cpp
SergioRZMasson: Added support for SRGB textures on the latest commit.

Plugins/ExternalTexture/CMakeLists.txt
SergioRZMasson: Moved JsRuntiem and GraphicsDevices to be public dependencies.

Plugins/ExternalTexture/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    PRIVATE Graphics
+    PRIVATE bgfx
+    PRIVATE bimg
+    PRIVATE bx)
+
+target_compile_definitions(ExternalTexture
+    PRIVATE NOMINMAX
+    PRIVATE ${GRAPHICS_API_UPPER})
+
+set_property(TARGET ExternalTexture PROPERTY FOLDER Core)
```
bghgary, line 37:
```suggestion
set_property(TARGET ExternalTexture PROPERTY FOLDER Plugins)
```

Plugins/ExternalTexture/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+target_include_directories(ExternalTexture
+    PUBLIC "Include/Shared"
+    PRIVATE "Source")
+
+if(WIN32 AND NOT WINDOWS_STORE)
+    target_link_to_dependencies(ExternalTexture
+        PRIVATE "shlwapi.lib")
+elseif(ANDROID)
+    target_link_to_dependencies(ExternalTexture
+        PRIVATE android
```
bghgary, line 18: hmm, what does this do?

Plugins/ExternalTexture/Include/Shared/Babylon/Plugins/ExternalTexture/ExternalTexture.h
```diff
@@ -0,0 +1,33 @@
...
+    {
+    public:
+        ExternalTexture(Babylon::Graphics::TextureType nativeTexture);
+        ~ExternalTexture();
+
+        Napi::Promise AddToContext(Napi::Env&);
+
+        uint32_t GetWidth();
+        uint32_t GetHeight();
+        uint32_t GetFormat();
```
bghgary, line 21:
```suggestion
        uint32_t GetWidth() const;
        uint32_t GetHeight() const;
        uint32_t GetFormat() const;
```

Plugins/ExternalTexture/Include/Shared/Babylon/Plugins/ExternalTexture/ExternalTexture.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Babylon::Plugins
+{
+    class ExternalTexture
+    {
+    public:
+        ExternalTexture(Babylon::Graphics::TextureType nativeTexture);
+        ~ExternalTexture();
+
+        Napi::Promise AddToContext(Napi::Env&);
```
bghgary, line 17: maybe?
```suggestion
        Napi::Promise AddToContext(Napi::Env&) const;
```

Plugins/ExternalTexture/Source/ExternalTexture_D3D11.cpp
```diff
@@ -1,29 +1,38 @@
...
 #include <Babylon/Graphics/Texture.h>
 #include <Babylon/Graphics/DeviceContext.h>
 
 #include <napi/env.h>
 #include <napi/napi_pointer.h>
 
-namespace Babylon::Graphics
+namespace Babylon::Plugins
 {
     //For refrence see renderer_d3d11.cpp line 203.
```
bghgary, line 12: Putting a specific line number here may get out of sync. Maybe put a link to a specific commit with a line number?
```suggestion
    // For reference see renderer_d3d11.cpp line 203.
```

Plugins/ExternalTexture/Source/ExternalTexture_D3D11.cpp
```diff
@@ -179,13 +191,15 @@ namespace Babylon::Graphics
         }
     }
 
-    void ExternalTexture::ReadPropertiesFromNativeTexture(TextureType nativeTexture) 
+    void ExternalTexture::ReadPropertiesFromNativeTexture(Babylon::Graphics::TextureType nativeTexture) 
     {
         D3D11_TEXTURE2D_DESC desc;
         nativeTexture->GetDesc(&desc);
-        m_format = NativeToBGFXImageFormat(desc.Format);
+        uint64_t flags = 0Ui64;
```
bghgary, line 198:
```suggestion
        uint64_t flags{};
```

Plugins/ExternalTexture/Source/ExternalTexture_D3D11.cpp
```diff
@@ -35,8 +44,11 @@ namespace Babylon::Graphics
             case DXGI_FORMAT_BC6H_SF16:
                 return bgfx::TextureFormat::Enum::BC6H; 
             
-            case DXGI_FORMAT_BC7_UNORM_SRGB:
             case DXGI_FORMAT_BC7_UNORM:
+                flags |= BGFX_TEXTURE_SRGB;
+                return bgfx::TextureFormat::Enum::BC7;
+
+            case DXGI_FORMAT_BC7_UNORM_SRGB:
                 return bgfx::TextureFormat::Enum::BC7;
```
bghgary, line 52:
```suggestion
            case DXGI_FORMAT_BC7_UNORM:
                return bgfx::TextureFormat::Enum::BC7;

            case DXGI_FORMAT_BC7_UNORM_SRGB:
                flags |= BGFX_TEXTURE_SRGB;
                return bgfx::TextureFormat::Enum::BC7;
```

Plugins/ExternalTexture/Source/ExternalTexture_D3D12.cpp
```diff
@@ -1,29 +1,38 @@
...
 #include <Babylon/Graphics/Texture.h>
 #include <Babylon/Graphics/DeviceContext.h>
 
 #include <napi/env.h>
 #include <napi/napi_pointer.h>
 
-namespace Babylon::Graphics
+namespace Babylon::Plugins
 {
     //For refrence see renderer_d3d12.cpp line 187.
```
bghgary, line 12: ditto as d3d11

Plugins/ExternalTexture/Source/ExternalTexture_D3D12.cpp
```diff
@@ -35,8 +44,11 @@ namespace Babylon::Graphics
             case DXGI_FORMAT_BC6H_SF16:
                 return bgfx::TextureFormat::Enum::BC6H; 
             
-            case DXGI_FORMAT_BC7_UNORM_SRGB:
             case DXGI_FORMAT_BC7_UNORM:
+                flags |= BGFX_TEXTURE_SRGB;
```
bghgary, line 48: ditto with d3d11

bghgary (changes requested):
Left a few comments, some are bugs.

Plugins/ExternalTexture/CMakeLists.txt
```diff
@@ -10,15 +10,6 @@ target_include_directories(ExternalTexture
...
-if(WIN32 AND NOT WINDOWS_STORE)
-    target_link_to_dependencies(ExternalTexture
-        PRIVATE "shlwapi.lib")
-elseif(ANDROID)
-    target_link_to_dependencies(ExternalTexture
-        PRIVATE android
-        PRIVATE AndroidExtensions)
-endif()
-
 target_link_to_dependencies(ExternalTexture
```
bghgary, line 13: Sorry, should have commented on this before. Are all of these required?

bghgary (approved):

SergioRZMasson:
Created [PR](https://github.com/BabylonJS/BabylonReactNative/pull/363) for BabylonReactNative to sync the changes introduced by this PR.

PR #1031, opened 2022-03-15
DeviceInputSystem: Update constructor to accept and use callbacks

PolygonalSun:
With recent changes to the DeviceSourceManager/DeviceInputSytem in Babylon.js, there was a regression introduced where mouse input wasn't working when the cursor was over the window before the DeviceSourceManager sets the value for onDeviceConnected.  This PR changes the constructor for the DeviceInputSystem to now accept all of the callbacks to prevent this issue.


Plugins/NativeInput/Source/Shared/DeviceInputSystem.cpp
```diff
@@ -31,68 +28,29 @@ namespace Babylon::Plugins
...
         , m_nativeInput{*NativeInput::GetFromJavaScript(info.Env()).m_impl}
-        , m_deviceConnectedTicket{m_nativeInput.AddDeviceConnectedCallback([this](DeviceType deviceType, int32_t deviceSlot) {
-            if (!m_onDeviceConnected.IsEmpty())
-            {
-                m_onDeviceConnected({
-                    Napi::Value::From(Env(), static_cast<uint32_t>(deviceType)),
-                    Napi::Value::From(Env(), deviceSlot)
-                });
-            }
+        , m_deviceConnectedTicket{m_nativeInput.AddDeviceConnectedCallback([this, callback = std::make_shared<Napi::FunctionReference>(Napi::Persistent(info[0].As<Napi::Function>()))](DeviceType deviceType, int32_t deviceSlot) {
```
bghgary, line 31: We talked about changing this to use the inplace_function that @syntheticmagus is working on to avoid the shared_ptr. @syntheticmagus Will this work here?

syntheticmagus: You'd need to change the `weak_table` in `NativeInput` to accept the new type, but yes, I think it should work if you did that. The new `Dispatchable` type (which maybe we should just put into Arcana as `unique_function`?) might be a better choice since you don't really need the "in-place"-ness 'cause `NativeInput` isn't on the stack. In theory, though, any non-copyable functor should do, unless I'm misunderstanding the scenario.

bghgary: Sounds good. Given that `Dispatchable` isn't merged yet. Let's go with shared_ptr for now.

syntheticmagus (approved):

Plugins/NativeInput/Source/Shared/DeviceInputSystem.cpp
PolygonalSun: Thanks for the info @syntheticmagus!

Issue #619, opened 2021-02-22
WebXRFeatureName.IMAGE_TRACKING support


saidmoya12:
Can you allow tracking with a image?

BabylonJs support WebXRFeatureName.IMAGE_TRACKING  but I didn't found this functionality here.

Thanks for your comments!

bghgary:
@RaananW Do you have any thoughts on this? Is this even supported by Babylon.js yet?

RaananW:
Yes, it is. The image tracking feature was added when released by the browsers and work on mobile browsers:

https://github.com/BabylonJS/Babylon.js/pull/9512

https://twitter.com/RaananW/status/1333437510840705024

bghgary:
Thanks. The next question is whether this feature is supported by the platforms we support (OpenXR, ARCore, ARKit). @RaananW How well supported is this feature on WebXR across devices?

RaananW:
Only supported on chrome for android. Its basic image recognizer, i would guess that both arcore and arkit support it naively, at least for marker recognition

bghgary:
Any ideas on OpenXR?

RaananW:
Not supported (yet?), AFAIK. couldn't find any reference online as well.

bghgary:
@saidmoya12 Is the feature still useful for you if it doesn't work everywhere?

HedwigAR:
I came across this issue when trying to find out why image tracking didn't work in Babylon ReactNative. Image recognition is supported in both ARCore and ARKit so I assumed this implementation would work with those as well. Are there any plans to implement image recognition for native android and ios in BabylonNative and BabylonReactNative?

bghgary:
We don't have any plans right now. Would this be useful if it only works for Android and iOS? It doesn't seem like there is support in OpenXR. Also, would you be capable/willing to contribute?

HedwigAR:
Ah, thanks for responding on that. I can't say I'm ready to contribute. I'm currently just exploring the Babylon libraries to see what the capabilities are. In previous project I've found image tracking to be a good starting point for experiences since people can't 'lose' it when it's attached to a physical marker. Or you can start the experience with some extra data based on which marker is scanned.

mmehmetAliIzci:
Hi @bghgary could you give some pointers to look in order to add this capability to the platform ?

Alex-MSFT:
Hi @mmehmetAliIzci - To add image tracking to the platform there are a three main things that would need to be implemented

[...]

PR #1032, opened 2022-03-16
WebXR Image Tracking Support for ARCore

Alex-MSFT:
This PR implements WebXR Image Tracking support for ARCore and introduces the WebXR polyfill in NativeXR.  I'm also adding stubs for iOS and OpenXR implementations, however those are not addressed in this PR.

The design for Image Tracking largely follows the design for Plane and Anchor tracking with the additional need to prepare the image database before beginning XR.  Another thing that sets this apart is that we are directly exposing the native ImageSpace to JS via a Napi::External, so I'm using a vector of std::unique_ptr<ImageTrackingResult> rather than storing the values directly in the vector to keep memory locations consistent.

https://github.com/BabylonJS/BabylonNative/issues/619

To improve performance I'm also making some small changes to the Babylon.JS contract for the nativeXR frame, which are being reviewed here:
https://github.com/BabylonJS/Babylon.js/pull/12176


Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
-                    filteredResults.push_back(hitResult);
+                    hitTestResult.NativeTrackable = reinterpret_cast<NativeTrackablePtr>(trackable);
+                    filteredResults.push_back(hitTestResult);
                     frameTrackables.push_back(trackable);
                 }
             }
         }
 
+        // Converts an image bitmap to grayscale assumes an RGB8 or RGB8A image for use with ARCore.
+        void ConvertBitmapToGrayscale(
```
Alex-MSFT, line 683: Would love if someone had a better more generic way to create grayscale bitmaps,  This example is largely based on a modified version from the ARCore sample code adapted for both RGB8 and RGB8A, but am open to ideas to make this more flexible.

Apps/Playground/Scripts/experience.js
```diff
@@ -290,6 +291,30 @@ CreateBoxAsync(scene).then(function () {
...
+                        {
+                            images: [
+                                { src: "https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/IridescentDishWithOlives/screenshot/screenshot_Large.jpg", estimatedRealWorldWidth: .2 },
+                                { src: "https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/DragonAttenuation/screenshot/screenshot_large.png", estimatedRealWorldWidth: .2 },   
+                        ]});
+
+                    webXRImageTrackingModule.onTrackedImageUpdatedObservable.add((imageObject) => {
+                        if (webXRTrackingMeshes[imageObject.id] === undefined) {
+                            webXRTrackingMeshes[imageObject.id] = BABYLON.Mesh.CreateBox("box1", 0.05, scene);
+                            var mat = new BABYLON.StandardMaterial("mat", scene);
```
ryantrem, line 309:
```suggestion
                            const mat = new BABYLON.StandardMaterial("mat", scene);
```

Apps/Playground/Scripts/experience.js
```diff
@@ -290,6 +291,30 @@ CreateBoxAsync(scene).then(function () {
...
                 }
 
+                // Test image tracking and detection.
+                if (imageTracking) {
+                    const webXRTrackingMeshes = [];
+                    const webXRImageTrackingModule = xr.baseExperience.featuresManager.enableFeature(
+                        BABYLON.WebXRFeatureName.IMAGE_TRACKING,
+                        "latest",
+                        {
+                            images: [
```
ryantrem, line 301: How do you test this? Do you pull these images up on your dev machine and point your device at it? Can you add some comments?

Dependencies/xr/Include/XR.h
```diff
@@ -297,6 +297,42 @@ namespace xr
...
                 };
+                
+                struct ImageTrackingRequest
+                {
+                    uint8_t* data;
+                    uint32_t width;
+                    uint32_t height;
+                    uint32_t depth;
+                    uint32_t stride;
+                    float measuredWidthInMeters;
```
ryantrem, line 308: Initialize all these to some reasonable default (`nullptr` and `0`?).

Dependencies/xr/Include/XR.h
```diff
@@ -297,6 +297,42 @@ namespace xr
...
+                };
+
+                struct ImageTrackingResult
+                {
+                    using Identifier = size_t;
+                    const Identifier ID{ NEXT_ID++ };
+                    Space ImageSpace;
+                    uint32_t Index;
+                    std::string TrackingState;
```
ryantrem, line 317: Initialize these too.

Apps/Playground/Scripts/experience.js
Alex-MSFT: Yes you can pull the image up on your dev machine and test it or you can print the image out and use tracking that way.  I'll add a comment

Dependencies/xr/Include/XR.h
```diff
@@ -310,20 +346,23 @@ namespace xr
                 std::vector<Plane::Identifier>RemovedPlanes;
                 std::vector<Mesh::Identifier>UpdatedMeshes;
                 std::vector<Mesh::Identifier>RemovedMeshes;
+                std::vector<ImageTrackingResult::Identifier>UpdatedImageTrackingResults;
```
ryantrem, line 349:
```suggestion
                std::vector<Plane::Identifier> RemovedPlanes;
                std::vector<Mesh::Identifier> UpdatedMeshes;
                std::vector<Mesh::Identifier> RemovedMeshes;
                std::vector<ImageTrackingResult::Identifier> UpdatedImageTrackingResults;
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+            const uint8_t* image_pixel_buffer,
+            int32_t width,
+            int32_t height,
+            int32_t stride,
+            uint8_t** out_grayscale_buffer)
+        {
+            uint8_t* grayscale_buffer = new uint8_t[width * height];
+            uint32_t pixelStride = stride / width;
```
ryantrem, line 691: I think a bunch of these other vars can be `const` too?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
             }
         }
 
+        // Converts an image bitmap to grayscale assumes an RGB8 or RGB8A image for use with ARCore.
+        void ConvertBitmapToGrayscale(
+            const uint8_t* image_pixel_buffer,
+            int32_t width,
+            int32_t height,
+            int32_t stride,
+            uint8_t** out_grayscale_buffer)
```
ryantrem, line 688: To me it would make more sense for the caller to allocate this memory and just pass in a pointer. Then they can manage the memory however they want (e.g. a `unique_ptr`).

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+                    uint8_t r = *pixel;
+                    uint8_t g = *(pixel + 1);
+                    uint8_t b = *(pixel + 2);
+                    grayscale_buffer[w + h * width] = static_cast<uint8_t>(0.213f * r + 0.715 * g + 0.072 * b);
+                }
+            }
+            *out_grayscale_buffer = grayscale_buffer;
+        }
+
+        std::vector<std::string> CreateAugmentedImageDatabase(std::vector<System::Session::Frame::ImageTrackingRequest>& requests)
```
ryantrem, line 706: Looks like the input vector can be const?
```suggestion
        std::vector<std::string> CreateAugmentedImageDatabase(const std::vector<System::Session::Frame::ImageTrackingRequest>& requests)
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+            ArAugmentedImageDatabase_create(xrContext->Session, &augmentedImageDatabase);
+            std::vector<std::string> scores{};
+            int32_t trackableImagesCount = 0;
+
+            // Loop over each image in the request, and add it to the image database.
+            for (System::Session::Frame::ImageTrackingRequest image : requests)
+            {
+                int32_t index;
+                uint8_t *grayscale_buffer;
+                ArStatus status;
```
ryantrem, line 717: Init these vars

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+        {
+            ArAugmentedImageDatabase_create(xrContext->Session, &augmentedImageDatabase);
+            std::vector<std::string> scores{};
+            int32_t trackableImagesCount = 0;
+
+            // Loop over each image in the request, and add it to the image database.
+            for (System::Session::Frame::ImageTrackingRequest image : requests)
+            {
+                int32_t index;
+                uint8_t *grayscale_buffer;
```
ryantrem, line 716: Related to my other comment, I'd allocate this as an `std::vector<uint8_t>` and then pass the `data` pointer to your helper. Then you don't need any code to manually delete the memory as it will just happen when the `std::vector` goes out of scope at the end of this function.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+            if (trackableImagesCount > 0)
+            {
+                // Create an ArConfig
+                ArConfig *arConfig{};
+                ArConfig_create(xrContext->Session, &arConfig);
+                ArSession_getConfig(xrContext->Session, arConfig);
+
+                // Configure the ArSession to include the image tracking database
+                ArConfig_setAugmentedImageDatabase(xrContext->Session, arConfig,
+                                                   augmentedImageDatabase);
```
ryantrem, line 779:
```suggestion
                ArConfig_setAugmentedImageDatabase(xrContext->Session, arConfig, augmentedImageDatabase);
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+                ArConfig_destroy(arConfig);
+            }
+
+            return scores;
+        }
+
+        void UpdateImageTrackingResults(std::vector<Frame::ImageTrackingResult::Identifier>& updatedResults)
+        {
+            // Get list of updated images in the current frame.
+            int32_t imageListSize;
```
ryantrem, line 798:
```suggestion
            int32_t imageListSize{};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+            ArFrame_getUpdatedTrackables(xrContext->Session, xrContext->Frame, AR_TRACKABLE_AUGMENTED_IMAGE, trackableList);
+            ArTrackableList_getSize(xrContext->Session, trackableList, &imageListSize);
+
+            // For each updated image, get the current status.
+            for (int i = 0; i < imageListSize; ++i) {
+                ArTrackable* trackable = nullptr;
+                ArTrackableList_acquireItem(xrContext->Session, trackableList, i, &trackable);
+                ArAugmentedImage* imageTrackable = ArAsAugmentedImage(trackable);
```
ryantrem, line 806: Uniform initializers for consistency?
```suggestion
                ArTrackable* trackable{nullptr};
                ArTrackableList_acquireItem(xrContext->Session, trackableList, i, &trackable);
                ArAugmentedImage* imageTrackable{ArAsAugmentedImage(trackable)};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+
+                int imageIndex;
```
ryantrem, line 808:
```suggestion
                int imageIndex{};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+                ArAugmentedImage_getIndex(xrContext->Session, imageTrackable, &imageIndex);
+
+                float measuredWidthInMeters;
```
ryantrem, line 811:
```suggestion
                float measuredWidthInMeters{};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,221 @@ namespace xr
...
+                        rawPose,
+                        measuredWidthInMeters,
+                        trackingState);
+                }
+            }
+        }
+
+        Frame::ImageTrackingResult& GetImageTrackingResultByID(Frame::ImageTrackingResult::Identifier resultID)
+        {
+            for (std::unique_ptr<Frame::ImageTrackingResult>& resultPtr : imageTrackingResults)
```
ryantrem, line 854: Maybe use `std::find_if`. We use it elsewhere so there should be examples.

Alex-MSFT: Did not know this existed, will also use this for planes!

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2605,6 +2656,34 @@ namespace Babylon
...
+                    for (uint32_t idx = 0; idx < napiTrackedImages.Length(); idx++)
+                    {
+                        auto napiImageRequest = napiTrackedImages.Get(idx).As<Napi::Object>();
+                        auto napiImage = napiImageRequest.Get("image").As<Napi::Object>();
+                        auto napiBuffer = napiImage.Get("data").As<Napi::Uint8Array>();
+                        uint32_t bufferSize = (uint32_t) napiBuffer.ByteLength();
+                        uint32_t imageHeight = napiImage.Get("height").ToNumber().Uint32Value();
+                        uint32_t stride = bufferSize / imageHeight;
+                        float estimatedWidth = napiImageRequest.Get("widthInMeters").ToNumber().FloatValue();
+                        session.m_imageTrackingRequests[idx] =
```
ryantrem, line 2675: Uniform initializers? I think that is still the standard being used?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,214 @@ namespace xr
...
+            int32_t trackableImagesCount{0};
+
+            // Loop over each image in the request, and add it to the image database.
+            for (System::Session::Frame::ImageTrackingRequest image : requests)
+            {
+                int32_t index{0};
+                ArStatus status{};
+                std::vector<uint8_t> grayscaleBuffer{};
+                grayscaleBuffer.reserve(image.width * image.height);
+                const std::unique_ptr<uint8_t> grayscale_buffer{new uint8_t[image.width * image.height]};
```
ryantrem, line 717: Remove

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,213 @@ namespace xr
...
+                    const uint8_t* pixel{&imagePixelBuffer[w * pixelStride + h * stride]};
+                    const uint8_t r{*pixel};
+                    const uint8_t g{*(pixel + 1)};
+                    const uint8_t b{*(pixel + 2)};
+                    grayscaleBuffer[w + h * width] = static_cast<uint8_t>(0.213f * r + 0.715 * g + 0.072 * b);
+                }
+            }
+        }
+
+        std::vector<std::string> CreateAugmentedImageDatabase(const std::vector<System::Session::Frame::ImageTrackingRequest>& requests)
```
ryantrem, line 704: It seems like this is the only place the input vector is marked as const. Should it be const everywhere?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1572,6 +1573,10 @@ void GetHitTestResultsLegacy(std::vector<HitResult>& filteredResults, xr::HitTes
         return m_impl->sessionImpl.GetPlaneByID(planeID);
     }
 
+    System::Session::Frame::ImageTrackingResult& System::Session::Frame::GetImageTrackingResultByID(System::Session::Frame::ImageTrackingResult::Identifier) const {
+        throw std::runtime_error("not implemented");
```
ryantrem, line 1577: Should these all be:
```suggestion
        throw std::runtime_error("Image tracking not supported");
```

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2605,6 +2658,34 @@ namespace Babylon
...
+                        const auto napiImage{napiImageRequest.Get("image").As<Napi::Object>()};
+                        const auto napiBuffer{napiImage.Get("data").As<Napi::Uint8Array>()};
+                        const uint32_t bufferSize{(uint32_t) napiBuffer.ByteLength()};
+                        const uint32_t imageHeight{napiImage.Get("height").ToNumber().Uint32Value()};
+                        const uint32_t stride{bufferSize / imageHeight};
+                        const float estimatedWidth{napiImageRequest.Get("widthInMeters").ToNumber().FloatValue()};
+                        session.m_imageTrackingRequests[idx] =
+                        {
+                            (uint8_t *) napiBuffer.Data(),
+                            napiImage.Get("width").ToNumber().Uint32Value(),
```
ryantrem, line 2680: I think this code would be slightly more readable if you move these out and store them in vars near `imageHeight` (even though these are not used multiple times like `imageHeight`).

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2605,6 +2658,34 @@ namespace Babylon
...
+                        const auto napiImageRequest{napiTrackedImages.Get(idx).As<Napi::Object>()};
+                        const auto napiImage{napiImageRequest.Get("image").As<Napi::Object>()};
+                        const auto napiBuffer{napiImage.Get("data").As<Napi::Uint8Array>()};
+                        const uint32_t bufferSize{(uint32_t) napiBuffer.ByteLength()};
+                        const uint32_t imageHeight{napiImage.Get("height").ToNumber().Uint32Value()};
+                        const uint32_t stride{bufferSize / imageHeight};
+                        const float estimatedWidth{napiImageRequest.Get("widthInMeters").ToNumber().FloatValue()};
+                        session.m_imageTrackingRequests[idx] =
+                        {
+                            (uint8_t *) napiBuffer.Data(),
```
ryantrem, line 2679: This cast should not be needed right? `Uint8Array::Data()` should already be of type `uint8_t*`.

Alex-MSFT: It looks like the deduced type of Uint8Array::Data is const char*  Here is the error I get in Android Studio without a cast:
Cannot initialize a variable of type 'uint8_t *' (aka 'unsigned char *') with an rvalue of type 'const unsigned char *'

Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2572,7 +2572,9 @@ namespace Babylon
 
                         auto napiResultRef{Napi::Weak(napiResult)};
                         auto imageTrackingArray{m_imageTrackingResultsArray.Value()};
-                        imageTrackingArray.Set(imageTrackingArray.Length(), napiResultRef.Value());
+                        auto index = imageTrackingArray.Length();
+                        imageTrackingArray.Set("length", Napi::Value::From(env, index + 1));
```
bghgary, line 2576: Should this be handled by NAPI implementation?

Alex-MSFT: Yes probably as this is an issue only in Hermes but I'm not sure if the issue is in NAPI or Hermes.  I'm not sure if I want to scope that to be part of this change though.

bghgary: No, probably not. But we can try to fix it separately beforehand maybe?

Alex-MSFT: Managed to get this working, and opened another PR here:
https://github.com/BabylonJS/BabylonNative/pull/1037

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,251 @@ namespace xr
...
+            const int32_t height,
+            const int32_t stride,
+            uint8_t* grayscaleBuffer)
+        {
+            const uint32_t pixelStride = stride / width;
+            for (int h = 0; h < height; ++h)
+            {
+                for (int w = 0; w < width; ++w)
+                {
+                    const uint8_t* pixel{&imagePixelBuffer[w * pixelStride + h * stride]};
```
ryantrem, line 696: It might be a bit safer and cleaner to use gsl::span here, but up to you.

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,251 @@ namespace xr
...
+                    hitTestResult.NativeTrackable = reinterpret_cast<NativeTrackablePtr>(trackable);
+                    filteredResults.push_back(hitTestResult);
                     frameTrackables.push_back(trackable);
                 }
             }
         }
 
+        // Converts an image bitmap to grayscale.
+        // Supported image formats: RBG(A)8, RGB(A)16, 16-bit grayscale.
+        void ConvertBitmapToGrayscale(
```
ryantrem, line 684: Maybe move this out to be a function in the anonymous namespace so it is clear that it is just a helper function and not accessing/mutating class state?

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -928,16 +1168,17 @@ namespace xr
 
         Frame::Plane& GetPlaneByID(Frame::Plane::Identifier planeID)
         {
-            // Loop over the plane vector and find the correct plane.
-            for (Frame::Plane& plane : Planes)
+            const auto end = Planes.end();
+            const auto it = std::find_if(Planes.begin(), end, [&](Frame::Plane& plane) { return plane.ID == planeID; });
```
ryantrem, line 1172:
```suggestion
            const auto end{Planes.end()};
            const auto it{std::find_if(Planes.begin(), end, [&](Frame::Plane& plane) { return plane.ID == planeID; })};
```

Dependencies/xr/Source/ARCore/XR.cpp
```diff
@@ -664,16 +669,251 @@ namespace xr
...
+                {
+                    UpdateImageTrackingResult(
+                        updatedResults,
+                        GetImageTrackingResultByID(resultIterator->second),
+                        rawPose,
+                        measuredWidthInMeters,
+                        trackingState);
+
+                    // Release reference to trackable, since we are already holding a ref count in the map.
+                    ArTrackable_release(reinterpret_cast<ArTrackable*>(imageTrackable));
```
ryantrem, line 858: Is there any additional cleanup for trackables needed when the session ends?

Dependencies/xr/Include/XR.h
```diff
@@ -190,6 +190,16 @@ namespace xr
             struct Impl;
 
         public:
+            struct ImageTrackingRequest
+            {
+                const uint8_t* data{nullptr};
```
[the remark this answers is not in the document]
Alex-MSFT: For now we handle only a select number of formats RGB(A) 8/16 and Grayscale 8/16 all encoded in big endian, which are the standard formats for .JPG and PNG files.

Dependencies/xr/Source/ARCore/XR.cpp
Alex-MSFT: Switched to spans

ryantrem (approved):

PR #1033, opened 2022-03-16
Dummy implementation for getImageTrackingResults

Alex-MSFT:
My BabylonJS PR here: https://github.com/BabylonJS/Babylon.js/pull/12097/files

Changed the contract for nativeXRFrame to require an implementation for getImageTrackingResults.  I thought this would not block Babylon.JS upgrade as we should only actually call getImageTrackingResults during image tracking, which is currently not supported, but the line in NativeXRFrame will run during XR startup. 

This PR just defines a dummy implementation for getImageTrackingResults to unblock XR startup, which should be effectively dead code at runtime since XRImageTrackingResult is not defined, so we should never allow initialization of WebXR Image Tracking

### DIFF
--- a/Plugins/NativeXr/Source/NativeXr.cpp
+++ b/Plugins/NativeXr/Source/NativeXr.cpp
@@ -2092,6 +2092,7 @@ namespace Babylon
                         InstanceMethod("getJointPose", &XRFrame::GetJointPose),
                         InstanceMethod("fillPoses", &XRFrame::FillPoses),
                         InstanceMethod("fillJointRadii", &XRFrame::FillJointRadii),
+                        InstanceMethod("getImageTrackingResults", &XRFrame::GetImageTrackingResults),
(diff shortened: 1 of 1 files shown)


Plugins/NativeXr/Source/NativeXr.cpp
```diff
@@ -2314,6 +2315,11 @@ namespace Babylon
                 return Napi::Value::From(info.Env(), true);
             }
 
+            Napi::Value GetImageTrackingResults (const Napi::CallbackInfo& info)
```
PolygonalSun, line 2318: nit: It might be good to put a small comment explaining why this is needed

PolygonalSun (approved):
LGTM, I only had the small nit

ryantrem (approved):

bghgary (approved):

PR #1035, opened 2022-03-23
When to use Babylon Native

syntheticmagus:
Somewhat fuzzy documentation page discussing current thinking about when Babylon Native technologies are the "right" choices for making an app. This is of course pretty speculative and will likely evolve significantly as the technologies evolve; but as a starting point, I think this is about where the conversation is right now. Let's discuss!


Documentation/WhenToUseBabylonNative.md
```diff
@@ -0,0 +1,170 @@
...
+In short, in all the ways that Babylon Native technologies are "last 
+resort" choices from a Web-first perspective, Babylon Native should be
+among the first options considered when looking at adding rendering from 
+a native-first perspective. Of course, Babylon Native will not be as fast
+or powerful as a full-native gaming solution like Unity or Unreal, so
+native usages that require that kind of top-tier power will likely not be 
+able to make do with Babylon Native. However, for native apps looking to 
+quickly and easily add feature-rich rendering, few alternatives will 
+provide as much rendering capability for as little native infrastructure 
+cost as Babylon Native can offer.
```
bghgary, line 170: The arguments for using Babylon Native may be download size. Maybe we can incorporate this into the text?

bghgary (commented):
I like it overall. I think I want to see some kind of chart or a graph summarizing the pros and cons of each choice. I would expect this to evolve. For example, we make no mention of the performance issues on iOS without JIT that probably should be mentioned as a consideration.

Documentation/WhenToUseBabylonNative.md
```diff
@@ -0,0 +1,194 @@
...
+-   [My app is a React Native app.](#babylon-react-native-is-probably-a-good-choice-for-your-project)
+-   [My app's platform uses JavaScript, but it isn't made with React Native.](#there-may-be-other-choices-better-suited-to-your-project-than-babylon-native-technologies)
+-   [My app's platform doesn't inherently use JavaScript.](#do-you-have-exceptional-rendering-requirements-such-as-aaa-quality-graphics-or-extreme-performance-needs)
+
+***
+
+## Babylon Native is probably a good choice for your project!
+
+Congratulations! If you ended up here, then we think Babylon Native may
+be a suitable choice for adding 3D to your project. Typically,
```
bghgary, line 40: We shouldn't assume 3D. Babylon Native can be used for scenarios that require access to the GPU.

Documentation/WhenToUseBabylonNative.md
```diff
@@ -0,0 +1,194 @@
...
+-   A 3D content generation tool written in C++ that wants to be able to 
+    preview its output.
+
+The main reason Babylon Native is well-suited for these scenarios is 
+because Babylon Native makes it possible to bring Babylon.js's 
+rendering capabilities into native applications with *extremely* minimal
+requirements regarding app and development infrastructure. In other words,
+Babylon Native is perhaps the simplest way to add Web-style 3D rendering
+to a native application without having to pivot that application to 
+*become* a Web app.
```
bghgary, line 59: nit: just to avoid "js`s"

```suggestion
The main reason Babylon Native is well-suited for these scenarios is 
because Babylon Native makes it possible to bring the rendering
capabilities of Babylon.js into native applications with *extremely* minimal
requirements regarding app and development infrastructure. In other words,
Babylon Native is perhaps the simplest way to add Web-style 3D rendering
to a native application without having to pivot that application to 
*become* a Web app.
```

Documentation/WhenToUseBabylonNative.md
```diff
@@ -0,0 +1,194 @@
...
+approach: it can sometimes yield better JavaScript performance (on iOS,
+WebViews are allowed to run JavaScript with JIT enabled while React 
+Native, and consequently Babylon React Native, are not) and it will 
+provide access to the very latest Babylon.js features, which may not 
+always be available in Babylon React Native. However, there are 
+disadvantages as well: a WebView will create a second JavaScript context
+distinct from React Native's main JavaScript context and you will have to
+manage communication between the contexts manually, which can be complex
+and may negatively impact memory footprint as you pay the memory cost for
+two JavaScript contexts instead of one. Furthermore, running inside a
```
bghgary, line 88: It may also be prohibitively slow to send the data across.

syntheticmagus: I'd assume that would depend on how chatty the relationship is, so I don't know how broad a claim we can make, but I'll add a callout to potential performance impacts.

bghgary (approved):
LGTM, we can iterate on this more as we go

PR #1037, opened 2022-03-25
Fix crash on hermes when calling Napi::Array::Set beyond length of array

Alex-MSFT:
Native Hermes APIs have differing behavior from from JSC and V8 where attempting to set an index value >= the length of the array will throw an unhandled exception.  This PR changes the napi-jsi implementation to account for this by manually extending the length of the array before calling setValueAtIndex.

### DIFF
--- a/Dependencies/napi/napi-jsi/include/napi/napi-inl.h
+++ b/Dependencies/napi/napi-jsi/include/napi/napi-inl.h
@@ -692,7 +692,12 @@ inline Value Object::Get(uint32_t index) const {
 template <typename ValueType>
 inline void Object::Set(uint32_t index, const ValueType& value) {
   // TODO: does this need to work for non-array objects?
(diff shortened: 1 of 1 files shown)


bghgary (approved):
Is it possible that Hermes will eventually fix this and this change will be obsolete?

Alex-MSFT:
> Is it possible that Hermes will eventually fix this and this change will be obsolete?

I believe this is intended behavior as we are actually hitting an intentional error case:
https://github.com/facebook/hermes/blob/9757378b211d6fbce5432b15995a400b3901f12a/API/hermes/hermes.cpp#L1859

I don't think it is likely to change unless there's lots of community pushback.  The reason for this is possibly because they are not backing with sparse arrays, so if someone were to set the value at index 100000 for example it would allocate an array of size 100000 instead of just one element.

bghgary:
Ouch. That will likely cause some issues for some JavaScript code.

PR #1051, opened 2022-04-01
[Canvas] Color parsing improvements

### LINKED ISSUE #1001: [Canvas] Color parsing
`rgba(r,g,b,a)` parsing not supported by regex.
Also reorder code so regex is used last for better performance.
Bonus point: check parsing in test unit.

CedricGuillemet:
fixes #1001 
- reverted experience.js changes
- rgb or rgba regex parsing
- improvement colors '#xxx' parsing


CedricGuillemet:
Switching to draft until I add test units.

Polyfills/Canvas/Source/Colors.h
```diff
@@ -162,40 +162,52 @@ namespace Babylon::Polyfills::Internal
...
+                components[i] += components[i] << bitShift;
             }
             return nvgRGBA(components[0], components[1], components[2], components[3]);
         }
         else
         {
-            // matches strings of the form rgb(#,#,#)
-            static const std::regex rgbRegex("rgb\\(\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*\\)");
+            // matches strings of the form rgb(#,#,#) or rgba(#,#,#,#)
+            static const std::regex rgbRegex("rgba?\\(\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*\\,?\\s*(:?\\d{1,3})?\\)");
```
ryantrem, line 197: I'm unclear on a few changes here:
1. Why is there an extra `\\` right before the comma? It seems like it is trying to escape a comma, which isn't needed?
2. Why is there an optional `:` before the alpha value? It almost looks like it was meant to be a non-capturing group (`?:`) except that doesn't make sense for that capture group.
3. It currently has the `,` as optional separately from the alpha value, so technically I believe the regex would match something like `rgba(1,2,3,)`, which is not idea. In this case, a non-capturing group *could* be used to group (but not capture) the `,`.
4. I think we also lost the trailing white space before the final `)`, so `rgb(1,2,3,4    )` would not match.
Given the above, I'd think it should be (untested):
```suggestion
            static const std::regex rgbRegex("rgba?\\(\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*(?:,\\s*(\\d{1,3}))?\\s*\\)");
```

Polyfills/Canvas/Source/Colors.h
```diff
@@ -162,40 +162,52 @@ namespace Babylon::Polyfills::Internal
             {"yellow", 0xffff00},
             {"yellowgreen", 0x9acd32} };
 
-        if (str[0] == '#' && str.length() == 4)
+        if (str[0] == '#')
```
ryantrem, line 165: Why don't we use regex in this case too? Also will you add a comment with example strings we are trying to match (like the later section does for `rgb(#,#,#)` or `rgba(#,#,#,#)`)? Having the example strings makes it much easier to understand what the code is trying to do.

bghgary (approved):
I think some unit tests for this will be good, but otherwise, just one comment.

Polyfills/Canvas/Source/Colors.h
bghgary: I thought the same thing, but I'm not sure regex is much better for this case. I think having unit tests will cover the example strings.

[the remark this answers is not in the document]
CedricGuillemet: Fixed.

Apps/Playground/Scripts/experience.js
```diff
@@ -3,340 +3,316 @@
...
-    scene,
-);
-const particleSystem = new BABYLON.ParticleSystem('particles', 2500, scene);
-particleSystem.particleTexture = fogTexture;
-particleSystem.emitter = new BABYLON.Vector3(0, 0.5, 0);
-particleSystem.start();
-
-engine.runRenderLoop(function () {
-    scene.render();
+var wireframe = false;
```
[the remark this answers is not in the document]
CedricGuillemet: Merged.

Polyfills/Canvas/Source/Colors.h
CedricGuillemet: Fixed.

Polyfills/Canvas/Source/Colors.h
CedricGuillemet: I'm not a fan of regex. I did a quick test to parse 1 million rgba color strings. Timings in release:
Time for RegExp : 17.5821 seconds
Time for scanf : 1.2384 seconds
Time for strtol : 0.1339 seconds
We should discuss this sometimes.

bghgary: Does the RegExp time include the time to parse the initial regex string?

Polyfills/Canvas/Source/Colors.h
```diff
@@ -162,58 +163,72 @@ namespace Babylon::Polyfills::Internal
...
+                    }
+                    else
+                    {
+                        return nvgRGB(std::stoi(rgbMatch[1]), std::stoi(rgbMatch[2]), std::stoi(rgbMatch[3]));
+                    }
+                }
             }
         }
-        return nvgRGBA(255, 0, 255, 255);
+        throw Napi::Error::New(env, std::string{"Unable to parse color : "} + str);
```
bghgary, line 232: nit:
```suggestion
        throw Napi::Error::New(env, std::string{"Unable to parse color: "} + str);
```

Polyfills/Canvas/Source/Colors.h
CedricGuillemet: No, std::regex is instanciated once. If it's instanciated for each color parsing, I get these values:

Time for RegExp all: 33.9778 seconds
Time for scanf all: 1.2475 seconds
Time for strtol all: 0.1473 seconds

Which is less than I anticipated. Basically, computing the regex object is more of less the same than parsing the color.

PR #1059, opened 2022-04-07
WebXRImageTracking support for iOS

Alex-MSFT:
This PR builds on top of the NativeXR and ARCore implementation of WebXR image tracking in #1032 by extending it to support ARKit on iOS as part of #619.

This is designed very similarly to the existing implementations for Planes and Meshes on ARKit as they are all built on top of ARKit anchors.  The primary difference here is when constructing the image database at session start validation is offloaded to a worker thread to not block frame processing as this can block for multiple frames during XR initialization.


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1336,6 +1392,110 @@ bool IsTracking() const {
...
+
+                    // If we have any images that qualified for tracking then enable image detection.
+                    if (imageCount > 0 && configuration != nil) {
+                        configuration.detectionImages = imageSet;
+                        
+                        if (@available(iOS 13.0, *)) {
+                            configuration.automaticImageScaleEstimationEnabled = true;
+                        }
+
+                        configuration.maximumNumberOfTrackedImages = imageCount > 4 ? 4 : imageCount;
```
ryantrem, line 1491: Are images beyond 4 just ignored then?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1120,7 +1129,41 @@ void UpdatePlanes(std::vector<Frame::Plane::Identifier>& updatedPlanes, std::vec
...
+                for (ARImageAnchor* updatedImage : *updatedARImages) {
+                    // Update the existing image tracking result if it exists, otherwise create a new result and add it to our list of results.
+                    const auto resultIterator{imageTrackingMap.find({[updatedImage.identifier.UUIDString UTF8String]})};
+                    if (resultIterator != imageTrackingMap.end()) {
+                        UpdateImageTrackingResult(updatedImageTrackingResults, GetImageTrackingResultByID(resultIterator->second), updatedImage);
+                    } else {
+                        // This is a new result. Create it and initialize its values.
+                        imageTrackingResults.push_back(std::make_unique<Frame::ImageTrackingResult>());
+                        auto& result{ *imageTrackingResults.back() };
+                        result.Index = [updatedImage.referenceImage.name intValue];
```
ryantrem, line 1157: This feels slightly sketchy. I guess `name` is not used for anything else and we have full control over it? I kind of would have thought just having an `unordered_map` of image to index would be cleaner, but maybe that would make memory management trickier and more prone to bugs?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1336,6 +1392,110 @@ bool IsTracking() const {
...
+            if (imageTrackingScoresValid) {
+                return &imageTrackingScores;
+            } else {
+                return nil;
+            }
+        }
+        
+        void CreateAugmentedImageDatabase(const std::vector<ImageTrackingRequest>& requests) {
+            // Create a dispatch group for the validation tasks.
+            __block dispatch_queue_t validationQueue{dispatch_queue_create("Image Validation Queue", nil)};
```
ryantrem, line 1406: We already use arcana's task system in this file for other types of asynchrony. Could we use arcana here as well instead of using all the dispatch stuff? It seems like it would be less of a maintenance burden because people only need to understand one system (arcana) rather than two systems, plus it would be pure c++ and not use a bunch of obj-c. Thoughts?

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -413,89 +431,80 @@ -(void)updateDisplayOrientedPose:(ARCamera*)camera {
...
-        [self UnlockPlanes];
-    }
-
+    if (planeDetectionEnabled || meshDetectionEnabled) {
+        [self LockAnchors];
+        @try {
+            for (ARAnchor* deletedAnchor : anchors) {
+                if (planeDetectionEnabled && [deletedAnchor isKindOfClass:[ARPlaneAnchor class]]) {
+                    deletedPlanes.push_back((ARPlaneAnchor*)deletedAnchor);
+                } else if (meshDetectionEnabled) {
```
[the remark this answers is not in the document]
Alex-MSFT: Correct, image anchors are never removed automatically.  You have to manually remove them by explicitly calling remove:
"Consider when to allow detection of each image to trigger (or repeat) AR interactions. ARKit adds an image anchor to a session exactly once for each reference image in the session configuration’s [detectionImages](https://developer.apple.com/documentation/arkit/arworldtrackingconfiguration/2941063-detectionimages) array."
https://developer.apple.com/documentation/arkit/content_anchors/detecting_images_in_an_ar_experience

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: In ARKit there are basically "high quality" tracked images which can be updated every frame which is what this max number represents.  If any of those images leaves the viewport then it will free up room to start tracking additional images.

If you leave this setting at the default of 0 then image tracking does not enforce this limit, but anchors are updated very slowly at ~1 tick per second.

https://developer.apple.com/documentation/arkit/arworldtrackingconfiguration/2968182-maximumnumberoftrackedimages

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: Name isn't used for anything and we do have full control over it. I went this route because it didn't add any extra members to the class, and since we only ever read it once when first create the image tracking result.

But I can use a map of ARReferenceImage* to index instead.

Dependencies/xr/Source/ARKit/XR.mm
Alex-MSFT: I'll give it a shot

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1402,20 +1402,20 @@ bool IsTracking() const {
...
-            __block dispatch_queue_t validationQueue{dispatch_queue_create("Image Validation Queue", nil)};
-            __block dispatch_group_t validationGroup{dispatch_group_create()};
-            
-            // Initialize values for tracking image scores.
+            if (requests.size() == 0) {
+                return;
+            }
+
+            // Create and resize vectors to hold request results.
+            std::vector<arcana::task<ARReferenceImage*, std::exception_ptr>> validationTasks;
```
ryantrem, line 1410:
```suggestion
            std::vector<arcana::task<ARReferenceImage*, std::exception_ptr>> validationTasks{};
```

Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -1472,30 +1471,33 @@ void CreateAugmentedImageDatabase(const std::vector<ImageTrackingRequest>& reque
...
-                            configuration.automaticImageScaleEstimationEnabled = true;
-                        }
-
-                        configuration.maximumNumberOfTrackedImages = imageCount > 4 ? 4 : imageCount;
-                        [SystemImpl.XrContext->Session runWithConfiguration: configuration];
-                        [sessionDelegate SetImageDetectionEnabled:true];
-                        imageTrackingScoresValid = true;
+            arcana::when_all(gsl::make_span(validationTasks))
+                .then(arcana::inline_scheduler, arcana::cancellation::none(), [this](std::vector<ARReferenceImage*> referenceImages) {
+                size_t imageCount = 0;
```
ryantrem, line 1476: Seems like this whole block of code needs one more level of indentation.

ryantrem (approved):

PR #1060, opened 2022-04-12
Add Graphics::Device::GetPlatformInfo

bghgary:


Core/Graphics/Include/Shared/Babylon/Graphics/Device.h
```diff
@@ -9,6 +9,11 @@
 
 namespace Babylon::Graphics
 {
+    struct PlatformInfo
+    {
+        DeviceType Device;
```
syntheticmagus, line 14: nit: This might be just me, but I don't really like `DeviceType` as a type alias because it sounds like an enum, such that I might expect to see something like
```
DeviceType Device{ DeviceType::D3D12 };
```
The convention I prefer is `DeviceT` for type aliases as that invokes template conventions, which I feel is more similar to what we're trying to express here than an enum is. Not a huge deal, but it confuses me ever time I encounter something like this.

Core/Graphics/Source/DeviceImpl.h
```diff
@@ -114,7 +117,7 @@ namespace Babylon::Graphics
 
         struct
         {
-            std::recursive_mutex Mutex{};
+            mutable std::recursive_mutex Mutex{};
```
syntheticmagus, line 120: Is this an intentional part of this change?

syntheticmagus (approved):
Not 100% sure about the mutability change, but the rest of it makes perfect sense to me, pre-existing naming quibbles aside.

Core/Graphics/Source/DeviceImpl.h
bghgary: Yes, I guess I didn't need to do it, but it's weird to me that Get functions are not const, especially just to get a lock.

Core/Graphics/Include/Shared/Babylon/Graphics/Device.h
bghgary: I'm okay with that, but let's save it for a different PR.

PR #1063, opened 2022-04-20
Pattern change to work around memory leak

syntheticmagus:
Changed class definition pattern for native context to avoid apparent leak in `Napi::DefineClass` on certain JavaScript engines. Pertains to https://github.com/BabylonJS/BabylonReactNative/issues/394.

### DIFF
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -235,6 +235,8 @@ namespace Babylon::Polyfills
         Internal::NativeCanvas::CreateInstance(env);
         Internal::NativeCanvasImage::CreateInstance(env);
 
+        Internal::Context::Initialize(env);
+
         return {impl};
     }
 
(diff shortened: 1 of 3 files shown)


CedricGuillemet:
I guess the same applies here :
https://github.com/BabylonJS/BabylonNative/blob/a08af6240559a98aa515184f6505499df6677606/Polyfills/Canvas/Source/Image.cpp#L19

https://github.com/BabylonJS/BabylonNative/blob/a08af6240559a98aa515184f6505499df6677606/Polyfills/Canvas/Source/ImageData.cpp#L16

syntheticmagus:
> I guess the same applies here :
> 
> https://github.com/BabylonJS/BabylonNative/blob/a08af6240559a98aa515184f6505499df6677606/Polyfills/Canvas/Source/Image.cpp#L19
> 
> https://github.com/BabylonJS/BabylonNative/blob/a08af6240559a98aa515184f6505499df6677606/Polyfills/Canvas/Source/ImageData.cpp#L16

Potentially, yes. However, I didn't see those as part of this particular repro, so treating this as a spot fix I've only targeted what I directly saw so far.

CedricGuillemet (approved):

PR #1065, opened 2022-04-21
Fix memory leak in ExternalTexture class

bghgary:
Also, update AddToContext to AddToContextAsync and move docs


Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
```diff
@@ -38,39 +38,40 @@ namespace Babylon::Plugins
         arcana::make_task(context.BeforeRenderScheduler(), arcana::cancellation_source::none(),
             [&context, &runtime, deferred = std::move(deferred), impl = m_impl]()
         {
-            auto handle = bgfx::createTexture2D(impl->Width(), impl->Height(), impl->HasMips(), 1, impl->Format(), impl->Flags());
+            // REVIEW: The bgfx texture handle probably needs to be an RAII object to make sure it gets clean up during the asynchrony.
+            //         For example, if any of the schedulers/dispatches below don't fire, then the texture handle will leak.
```
syntheticmagus, line 42: Could you address this with an `expected`-handling cleanup continuation?

syntheticmagus (approved):

Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
bghgary: No, on shutdown, the continuation may never happen right?

PR #1072, opened 2022-05-06
Multi-touch support for Windows (Win32/UWP)

PolygonalSun:
This PR adds multi-touch support for Windows, both Win32 and UWP.  There is also a slight change made UWP's pointer move logic so that PointerPressed and PointerReleased are only called inside of PointerMoved when a button change is detected.

- [x] Win32
- [x] UWP

This has been tested on a Windows Machine and the Windows Simulator.


Apps/Playground/UWP/App.cpp
```diff
@@ -215,24 +215,36 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
-
-        if (point->IsInContact)
+        if (deviceType == Windows::Devices::Input::PointerDeviceType::Mouse)
         {
-            OnPointerPressed(nullptr, args);
-            OnPointerReleased(nullptr, args);
+            m_nativeInput->MouseMove(static_cast<int>(position.X), static_cast<int>(position.Y));
+
+            /*
+             * Note: Because PointerPressed only first for the first button press (and no additional ones)
```
ryantrem, line 226: I think this is supposed to say the following?
```suggestion
             * Note: Because PointerPressed only fires for the first button press (and no additional ones)
```

Apps/Playground/UWP/App.cpp
```diff
@@ -215,24 +215,36 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
+             * and PointerReleased only fires when all buttons have been released, we need an alternative way
+             * to track additional button presses.
+             * (See Definition - PointerPressed: https://docs.microsoft.com/en-us/uwp/api/windows.ui.core.corewindow.pointerpressed?view=winrt-22000)
+             * (See Remarks - PointerReleased: https://docs.microsoft.com/en-us/uwp/api/windows.ui.core.corewindow.pointerreleased?view=winrt-22000)
+             *
+             * In both of the below functions, we have three booleans that track each button's state and then we
```
ryantrem, line 232: What functions? I just see blocks of code below this. I know these comments aren't new, but I'm reading them now to try to understand the logic.

Apps/Playground/UWP/App.cpp
```diff
@@ -215,24 +215,36 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
+             */
+
+            // Check buttons for any button changes
+            const auto& properties = args->CurrentPoint->Properties;
+
+            if ((m_leftPressed == properties->IsLeftButtonPressed) || (m_middlePressed == properties->IsMiddleButtonPressed) || (m_rightPressed && properties->IsRightButtonPressed))
+            {
+                OnPointerPressed(nullptr, args);
+                OnPointerReleased(nullptr, args);
+            }
```
ryantrem, line 243: I don't get this logic, even with the big comment above.

Apps/Playground/UWP/App.cpp
PolygonalSun: Good catch!  I'll get that fixed up

Apps/Playground/UWP/App.cpp
```diff
@@ -215,24 +215,36 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
+             * (See Remarks - PointerReleased: https://docs.microsoft.com/en-us/uwp/api/windows.ui.core.corewindow.pointerreleased?view=winrt-22000)
+             *
+             * In both of the below functions, we have three booleans that track each button's state and then we
+             * run these two functions to update any buttons that we may have missed.
+             */
+
+            // Check buttons for any button changes
+            const auto& properties = args->CurrentPoint->Properties;
+
+            if ((m_leftPressed == properties->IsLeftButtonPressed) || (m_middlePressed == properties->IsMiddleButtonPressed) || (m_rightPressed && properties->IsRightButtonPressed))
```
ryantrem, line 239: Was the previous check (`IsInContact`) just wrong? Like was it missing mouse up state changes when one mouse button remains down, or something like that?

Apps/Playground/Win32/App.cpp
```diff
@@ -391,33 +395,39 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
         }
         case WM_MOUSEMOVE:
         {
-            if (nativeInput != nullptr)
+            if (nativeInput != nullptr && !IsTouchEvent(GetMessageExtraInfo()))
```
ryantrem, line 398: Does this mean when a touch happens, we get both a `WM_MOUSEMOVE` and a `WM_POINTERDOWN`?

ryantrem (approved):

Apps/Playground/UWP/App.cpp
PolygonalSun: IsInContact just defines if at least one button is pressed.  Because this check is now checking for changes to button states, we no longer needed to check IsInContact

Apps/Playground/Win32/App.cpp
PolygonalSun: I had originally included it because I noticed an odd occurrence in the Windows Simulator where a single WM_MOUSEMOVE was being fired when the touch cursor entered the screen area but generally, WM_MOUSEMOVE events should **not** occur with touch events.

bghgary (changes requested):

Apps/Playground/Win32/App.cpp
```diff
@@ -26,6 +26,10 @@
 
 #define MAX_LOADSTRING 100
 
+#define MI_WP_SIGNATURE 0xFF515700
+#define SIGNATURE_MASK 0xFFFFFF00
+#define IsTouchEvent(dw) (((dw) & SIGNATURE_MASK) == MI_WP_SIGNATURE)
```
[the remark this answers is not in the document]
PolygonalSun: Yup, that's where I found this snippet.  I'll add a link to that in a comment

Apps/Playground/Win32/App.cpp
```diff
@@ -477,6 +487,30 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
             }
             break;
         }
+        case WM_POINTERDOWN:
```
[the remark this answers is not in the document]
PolygonalSun: A WM_POINTERDOWN event occurs when a pointer makes contact with the window.  As far as Windows understands things, a mouse cursor never actually makes contact.  It just moves around and then fires a button event when a mouse button is pressed.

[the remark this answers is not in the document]
bghgary: That is not what the documentation says... https://docs.microsoft.com/en-us/windows/win32/inputmsg/wm-pointerdown

> [IS_POINTER_FIRSTBUTTON_WPARAM](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-is_pointer_firstbutton_wparam)(wParam): a flag that indicates whether there is a primary action.
> 
> This is analogous to a mouse left button down.
> A touch pointer will have this set when it is in contact with the digitizer surface.
> A pen pointer will have this set when it is in contact with the digitizer surface with no buttons pressed.

Apps/Playground/Win32/App.cpp
PolygonalSun: I feel like it's ultimately unnecessary.  I'm gonna remove this check because I think this may be more on the simulator than how it would actually work.

Apps/Playground/UWP/App.cpp
PolygonalSun: Reverted to IsInContact, discussed offline

Apps/Playground/Win32/App.cpp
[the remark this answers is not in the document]
PolygonalSun: So after doing research on this, I am actually wrong.  What I thought was code for touch is actually just generic and I guess turning on mouse triggering for WM_POINTERDOWN is actually something you have to set.  I think that I'd actually prefer to use the WM_POINTER* for everything as I'd need to do it for pen support in the future.  Lemme update the logic.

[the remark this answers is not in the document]
bghgary: Yes, that would be my preference also.

Apps/Playground/UWP/App.cpp
```diff
@@ -213,26 +213,41 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
+                case PointerUpdateKind::RightButtonReleased:
+                    m_nativeInput->MouseUp(Babylon::Plugins::NativeInput::RIGHT_MOUSE_BUTTON_ID, static_cast<int>(position.X), static_cast<int>(position.Y));
+                    break;
+            }
+        }
+        else
         {
-            OnPointerPressed(nullptr, args);
-            OnPointerReleased(nullptr, args);
+            m_nativeInput->TouchMove(deviceSlot, static_cast<int>(position.X), static_cast<int>(position.Y));
```
bghgary, line 250: Maybe factor out `static_cast<int>(position.X)` and `static_cast<int>(position.Y)` into local variables `x` and `y` so you don't have to keep repeating them. Ditto elsewhere.

Apps/Playground/UWP/App.cpp
```diff
@@ -213,26 +213,41 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
+                case PointerUpdateKind::LeftButtonReleased:
+                    m_nativeInput->MouseUp(Babylon::Plugins::NativeInput::LEFT_MOUSE_BUTTON_ID, static_cast<int>(position.X), static_cast<int>(position.Y));
+                    break;
+                case PointerUpdateKind::MiddleButtonReleased:
+                    m_nativeInput->MouseUp(Babylon::Plugins::NativeInput::MIDDLE_MOUSE_BUTTON_ID, static_cast<int>(position.X), static_cast<int>(position.Y));
+                    break;
+                case PointerUpdateKind::RightButtonReleased:
+                    m_nativeInput->MouseUp(Babylon::Plugins::NativeInput::RIGHT_MOUSE_BUTTON_ID, static_cast<int>(position.X), static_cast<int>(position.Y));
+                    break;
+            }
```
bghgary, line 246: nit: Consider adding a default case and assert/throw

Apps/Playground/UWP/App.cpp
```diff
@@ -213,26 +213,41 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
+            {
+                case PointerUpdateKind::LeftButtonPressed:
+                    m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::LEFT_MOUSE_BUTTON_ID, static_cast<int>(position.X), static_cast<int>(position.Y));
+                    break;
+                case PointerUpdateKind::MiddleButtonPressed:
+                    m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::MIDDLE_MOUSE_BUTTON_ID, static_cast<int>(position.X), static_cast<int>(position.Y));
+                    break;
+                case PointerUpdateKind::RightButtonPressed:
+                    m_nativeInput->MouseDown(Babylon::Plugins::NativeInput::RIGHT_MOUSE_BUTTON_ID, static_cast<int>(position.X), static_cast<int>(position.Y));
+                    break;
```
bghgary, line 236: It seems like you can factor this code out into a separate function that is called by this and pointer down.

bghgary (approved):
My comments for Win32 also apply to UWP. I think we should provide a default implementation for each platform so that people don't have to implement this code over and over again.

Apps/Playground/UWP/App.cpp
PolygonalSun: x and y have been factored out into local vars to avoid multiple parse calls

Apps/Playground/UWP/App.cpp
```diff
@@ -213,26 +239,25 @@ void App::OnPointerMoved(CoreWindow^, PointerEventArgs^ args)
...
-         * run these two functions to update any buttons that we may have missed.
-         */
-
-        if (point->IsInContact)
+        const auto& position = args->CurrentPoint->RawPosition;
+        const auto& deviceType = args->CurrentPoint->PointerDevice->PointerDeviceType;
+        const auto& deviceSlot = args->CurrentPoint->PointerId;
+        const auto& updateKind = args->CurrentPoint->Properties->PointerUpdateKind;
+        const auto& x = static_cast<int>(position.X);
+        const auto& y = static_cast<int>(position.Y);
```
ryantrem, line 247: Kind of a nit, but a const ref to an int seems weird since the ref is the same size as (or maybe bigger than) the int, so might as well just copy it?
```suggestion
        const auto x = static_cast<int>(position.X);
        const auto y = static_cast<int>(position.Y);
```

ryantrem (approved):

PR #1075, opened 2022-05-11
Fix Android Morph Targets - Add Native Engine Caps

SergioRZMasson:
## Overview

Morph targets are currently failing on Android since bgfx does not support Texture2D arrays for Android API BGFX_CONFIG_RENDERER_OPENGLES lower than 31 (currently set to 30). The appropriate way of handling this type of problem would be to check bgfx capabilities and handle it appropriately on the JS side. This PR exposes all capabilities provided by bgfx so they can be accessed by the js NativeEngine and appropriately handled by Babylon.js 

## Aditions

- Exposed bgfx caps to javascript. 
- This PR requires needs to merge after [this PR](https://github.com/BabylonJS/Babylon.js/pull/12515) from Babylon.js is merged.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -310,6 +310,8 @@ namespace Babylon
         // Initialize the JavaScript side.
         Napi::HandleScope scope{env};
 
+        auto caps = bgfx::getCaps();
+
         // clang-format off
         Napi::Function func = DefineClass(
             env,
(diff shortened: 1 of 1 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -310,6 +310,8 @@ namespace Babylon
         // Initialize the JavaScript side.
         Napi::HandleScope scope{env};
 
+        auto caps = bgfx::getCaps();
```
bghgary, line 313: nit: Maybe store `limits` instead of `caps` to avoid typing?

bghgary (approved):

PR #1084, opened 2022-06-01
Less gargantuan debug libraries

### LINKED ISSUE #1062: Windows debug libs are gargantuan
**Describe the bug**
The Babylon Native Windows static libs (debug) are quite large, and in fact too large now to be packaged into an npm package for Babylon React Native and published to npmjs.org (e.g. https://github.com/BabylonJS/BabylonReactNative/runs/6043625321?check_suite_focus=true).

Some of the most notable large libs are (note each of these are built for four architectures, so multiple the package size impact by four!):
- Graphics.lib: ~48mb
- Canvas.lib: ~20mb
- MachineIndependent.lib: @25mb
- NativeEngine.lib: ~20mb
- NativeXR.lib: ~20mb
- openxr_loaderd.lib: ~20mb
- spirv-*.lib: ~15mb - ~20mb each

@bghgary did a little initial digging on Graphics.lib and found a couple of things:
1. "just including <winrt/Windows.UI.Xaml.Controls.h> without using any types causes the size to go kaboom" (e.g. ~6mb -> 44mb for his local build)
2. Somewhere between the BRN 1.0.1 release and 1.1.0 release, Graphics.lib increased from ~27mb to ~48mb (unknown reason currently)

**Expected behavior**
Libs should be small enough to publish BRN to npmjs.org.

**Other**
 - Platform: UWP

CedricGuillemet:
fixes #1062 

Problem is `cppwinrt\winrt\Windows.UI.Xaml.Controls.h` defines a lot of template that gets compiled inside .obj and concatenated in the .lib.
The more .cpp including `<Babylon/Graphics/Platform.h>` the bigger the library.
There is no define ([WIN32_LEAN_AND_MEAN like) that will help. It looks like some build parameters with implicit templates might help but it doesn't cover our use case here.

Only solution is to limit the number of .cpp including it. That's what I did in this PR with forward reference and an added method `UpdateSize` so `WindowConfiguration` is forward until last moment in a .cpp that already includes the winrt header.

This way, **we get from 40+Mb to 18**. It's possible to shave off another 7Mb by removing Plaform.h inclusion from DeviceImpl but this line will produce an error:

`Resize(config.Width, config.Height);`

I have some thoughts that I leave to your appreciation:
- have a base struct `WindowConfigurationBase` that `WindowConfiguration` will inherit from. It will contain only Width and Height. So, it's possible to cast from a `WindowConfiguration` to a `WindowConfigurationBase` reference in DeviceImpl and only use Width/Height
- Try to do an amalgamated build for this project using cmake. Basically, cmake build a new temporary file that include every .cpp. Only this file will be compiled. So, template will only be instanciated once. On a large scale, this can help reduce build time for other projects too (I experience around 40% less time in other projects)

### DIFF
--- a/Core/Graphics/Include/Shared/Babylon/Graphics/Device.h
+++ b/Core/Graphics/Include/Shared/Babylon/Graphics/Device.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <Babylon/JsRuntime.h>
-#include <Babylon/Graphics/Platform.h>
 #include <Babylon/Graphics/RendererType.h>
 
 #include <future>
@@ -9,6 +8,8 @@
 
 namespace Babylon::Graphics
 {
+    struct WindowConfiguration;
+
     struct PlatformInfo
     {
         DeviceT Device;
@@ -74,6 +75,7 @@ namespace Babylon::Graphics
 
         void UpdateWindow(const WindowConfiguration& config);
         void UpdateSize(size_t width, size_t height);
+        void UpdateSize(const WindowConfiguration& config);
 
         void AddToJavaScript(Napi::Env);
         Napi::Value CreateContext(Napi::Env);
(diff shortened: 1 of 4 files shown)


CedricGuillemet:
Closing because unity build is more promising.

PR #1088, opened 2022-06-13
Fix textureLod being flipped

bghgary:


CedricGuillemet (approved):

bghgary:
This change breaks PBR shader :(

bghgary:
Never mind, it was a typo.

PR #1091, opened 2022-06-22
MSAA parameter exposed to users

CedricGuillemet:
Default is no MSAA unless set in the `WindowConfiguration` or updated later with `UpdateMSAA`.
Desktop playground and ValidationTests app MSAA set to 4.

### DIFF
--- a/Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
+++ b/Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
@@ -11,5 +11,7 @@ namespace Babylon::Graphics
         WindowType Window{};
         size_t Width{};
         size_t Height{};
+        // MSAA sample values can be 2, 4, 8 or 16. Any other value will disable MSAA
+        unsigned char MSAASamples{};
     };
 }
--- a/Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
+++ b/Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <android/native_window.h>
+#include <stdint.h>
 
 namespace Babylon::Graphics
 {
@@ -12,6 +13,6 @@ namespace Babylon::Graphics
         size_t Width{};
         size_t Height{};
         // MSAA sample values can be 2, 4, 8 or 16. Any other value will disable MSAA
-        unsigned char MSAASamples{};
+        uint8_t MSAASamples{};
     };
 }
--- a/Core/Graphics/Source/Device.cpp
+++ b/Core/Graphics/Source/Device.cpp
@@ -31,6 +31,7 @@ namespace Babylon::Graphics
         std::unique_ptr<Device> device{new Device()};
         device->UpdateWindow(config);
         device->UpdateSize(config.Width, config.Height);
+        device->UpdateMSAA(config.MSAASamples);
         return device;
     }
 
@@ -44,6 +45,11 @@ namespace Babylon::Graphics
     void Device::UpdateSize(size_t width, size_t height)
     {
         m_impl->Resize(width, height);
+    }
+
+    void Device::UpdateMSAA(uint8_t MSAASamples)
+    {
+        m_impl->SetMSAA(MSAASamples);
     }
 
     void Device::AddToJavaScript(Napi::Env env)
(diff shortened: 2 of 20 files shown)


Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
```diff
@@ -11,5 +11,7 @@ namespace Babylon::Graphics
         WindowType Window{};
         size_t Width{};
         size_t Height{};
+        // MSAA sample values can be 2, 4, 8 or 16. Any other value will disable MSAA
+        unsigned char MSAASamples{};
```
bghgary, line 15: Probably better to use this like in DeviceImpl.h.

```suggestion
        uint8_t MSAASamples{};
```

Core/Graphics/Source/Device.cpp
```diff
@@ -46,6 +47,11 @@ namespace Babylon::Graphics
         m_impl->Resize(width, height);
     }
 
+    void Device::UpdateMSAA(uint8_t MSAASamples)
```
bghgary, line 50: nit: casing is not right. maybe just use `value` here.
```suggestion
    void Device::UpdateMSAA(uint8_t value)
```

Core/Graphics/Source/DeviceImpl.h
```diff
@@ -39,6 +39,7 @@ namespace Babylon::Graphics
         void UpdateWindow(const WindowConfiguration& config);
         void UpdateContext(const DeviceConfiguration& config);
         void Resize(size_t width, size_t height);
+        void SetMSAA(uint8_t MSAASamples);
```
bghgary, line 42: ditto:
```suggestion
        void SetMSAA(uint8_t value);
```

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -70,6 +62,29 @@ namespace Babylon::Graphics
         UpdateBgfxResolution();
     }
 
+    void DeviceImpl::SetMSAA(uint8_t MSAASamples)
```
bghgary, line 65: ditto:
```suggestion
    void DeviceImpl::SetMSAA(uint8_t value)
```

Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
bghgary: On second thought, maybe this should be an enum? What happens if you pass in something that is not allowed by bgfx?

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -70,6 +62,29 @@ namespace Babylon::Graphics
...
+            case 4:
+                init.resolution.reset |= BGFX_RESET_MSAA_X4;
+                break;
+            case 8:
+                init.resolution.reset |= BGFX_RESET_MSAA_X8;
+                break;
+            case 16:
+                init.resolution.reset |= BGFX_RESET_MSAA_X16;
+                break;
+        }
```
bghgary, line 85: Maybe we should throw here if someone passes in an invalid value? Or use an enum for this like I mentioned for the contract (in which case, there should be an assert here)?

CedricGuillemet: I had (and still have) a lot of hesitation using an enum. What bugs me with an enum is the number of definitions and switch all along the way:
With BRN Android, there would be an enum in JS, in Java and C++ and a switch/case conversion each time. Whereas a number of samples is just an integer that can be forwarded easily.

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -82,6 +86,9 @@ namespace Babylon::Graphics
             case 16:
                 init.resolution.reset |= BGFX_RESET_MSAA_X16;
                 break;
+            default:
+                m_bgfxCallback.trace(__FILE__, __LINE__,  "WARNING: Setting an incorrect value for SetMSAA (%d). Correct values are 0, 1 (disable MSAA) or 2, 4, 8, 16.", int(value));
```
bghgary, line 90: It's a bit weird to rely on bgfx for API contract warnings, but okay for now. I wonder if throwing an exception is better?

bghgary (approved):

Core/Graphics/Source/DeviceImpl.cpp
CedricGuillemet: No strong opinion here. Throwing an exception here seems a bit too much. It may not be the rendering you want but it will work. To me an exception is needed when the behavior of the app will be (heavily) affected.

PR #1092, opened 2022-06-24
Fix VertexBuffer update

CedricGuillemet:
Issue reported by one partner when doing a ThinInstances buffer partial update.
byteOffset was not taken into account and buffer was forced to be the partial span.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -645,7 +645,7 @@ namespace Babylon
         const uint32_t byteOffset = info[2].As<Napi::Number>().Uint32Value();
         const uint32_t byteLength = info[3].As<Napi::Number>().Uint32Value();
 
-        vertexBuffer->Update(info.Env(), gsl::make_span(static_cast<uint8_t*>(bytes.Data()) + byteOffset, byteLength));
+        vertexBuffer->Update(info.Env(), gsl::make_span(static_cast<uint8_t*>(bytes.Data()), byteLength), byteOffset);
     }
 
     // Change VS output coordinate system
--- a/Plugins/NativeEngine/Source/VertexBuffer.cpp
+++ b/Plugins/NativeEngine/Source/VertexBuffer.cpp
@@ -75,7 +75,7 @@ namespace Babylon
         else
         {
             // Buffer hasn't been finalized yet, all that's necessary is to swap out the bytes.
-            if (m_bytes && !(*m_bytes).empty())
+            if (m_bytes && !m_bytes->empty())
             {
                 // update a portion of the vertex buffer bytes
                 memcpy((*m_bytes).data() + byteOffset, bytes.data(), bytes.size());
--- a/Plugins/NativeEngine/Source/VertexBuffer.cpp
+++ b/Plugins/NativeEngine/Source/VertexBuffer.cpp
@@ -60,7 +60,7 @@ namespace Babylon
         m_disposed = true;
     }
 
-    void VertexBuffer::Update(Napi::Env env, gsl::span<uint8_t> bytes)
+    void VertexBuffer::Update(Napi::Env env, gsl::span<uint8_t> bytes, size_t byteOffset)
     {
         if (!m_dynamic)
         {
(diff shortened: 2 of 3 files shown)


Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -75,7 +75,22 @@ namespace Babylon
         else
         {
             // Buffer hasn't been finalized yet, all that's necessary is to swap out the bytes.
-            m_bytes = {bytes.data(), bytes.data() + bytes.size()};
+            if (m_bytes && !(*m_bytes).empty())
```
bghgary, line 78:
```suggestion
            if (m_bytes && !m_bytes->empty())
```

Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -75,7 +75,22 @@ namespace Babylon
...
+            {
+                // update a portion of the vertex buffer bytes
+                memcpy((*m_bytes).data() + byteOffset, bytes.data(), bytes.size());
```
bghgary, line 81: This seems dangerous. What happens if someone passes an out-of-bounds byteOffset?

Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -60,7 +60,7 @@ namespace Babylon
         m_disposed = true;
     }
 
-    void VertexBuffer::Update(Napi::Env env, gsl::span<uint8_t> bytes)
+    void VertexBuffer::Update(Napi::Env env, gsl::span<uint8_t> bytes, size_t byteOffset)
```
bghgary, line 63: Do we need to handle `byteOffset` for the dynamic path?

Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -75,7 +75,22 @@ namespace Babylon
...
+            if (m_bytes && !(*m_bytes).empty())
+            {
+                // update a portion of the vertex buffer bytes
+                memcpy((*m_bytes).data() + byteOffset, bytes.data(), bytes.size());
+            }
+            else
+            {
+                if (byteOffset != 0)
+                {
+                    throw Napi::Error::New(env, "Cannot update a vertex buffer that has no data yet.");
```
bghgary, line 87: This doesn't seem right. We should be able to store what needs to be updated and then update it later?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -645,7 +645,7 @@ namespace Babylon
         const uint32_t byteOffset = info[2].As<Napi::Number>().Uint32Value();
         const uint32_t byteLength = info[3].As<Napi::Number>().Uint32Value();
 
-        vertexBuffer->Update(info.Env(), gsl::make_span(static_cast<uint8_t*>(bytes.Data()) + byteOffset, byteLength));
+        vertexBuffer->Update(info.Env(), gsl::make_span(static_cast<uint8_t*>(bytes.Data()), byteLength), byteOffset);
```
bghgary, line 648: I'm confused. This is updating a dynamic buffer, but the dynamic path isn't updated in VertexBuffer?

bghgary (changes requested):
I'm rather confused by this change. 😕

Plugins/NativeEngine/Source/VertexBuffer.cpp
CedricGuillemet: True. Adding a check.

Plugins/NativeEngine/Source/VertexBuffer.cpp
CedricGuillemet: No, it's a safe guard that should never happen: You can't update a VB that has not been filled before. But, if ever, then you'll get an exception.

Plugins/NativeEngine/Source/VertexBuffer.cpp
CedricGuillemet: It's not part of this PR: only update in the context of Thin Instances.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: I still don't get it. Static vertex buffers can be updated? I thought only dynamic vertex buffers can be updated. What am I missing?

CedricGuillemet: There 3 types of vertex buffer: GPU static, GPU dynamic and CPU. The last one is only available when used with Thin Instancing. In this case, datas are moved to GPU each frame : https://github.com/BabylonJS/BabylonNative/blob/1abe1ef7f52cd1330a57c42d178ddb03a36486bd/Plugins/NativeEngine/Source/VertexBuffer.cpp#L169

This PR covers this case where a subset of the CPU mem needs to be updated.

Plugins/NativeEngine/Source/VertexBuffer.cpp
CedricGuillemet: Added an exception when using byte offset with GPU VB.

bghgary (approved):

PR #1093, opened 2022-06-24
Fixes for resizing windows in UWP for D3D11

bghgary:
- Update bgfx to include fixes for UWP
- Set max frame latency for bgfx init
- Use width and height from graphics instead of bgfx

#118

### DIFF
--- a/Core/Graphics/Source/DeviceImpl.cpp
+++ b/Core/Graphics/Source/DeviceImpl.cpp
@@ -27,6 +27,7 @@ namespace Babylon::Graphics
         auto& init = m_state.Bgfx.InitState;
         init.type = s_bgfxRenderType;
         init.resolution.reset = BGFX_RESET_VSYNC | BGFX_RESET_MAXANISOTROPY;
+        init.resolution.maxFrameLatency = 1;
         if (s_bgfxFlipAfterRender) init.resolution.reset |= BGFX_RESET_FLIP_AFTER_RENDER;
 
         init.callback = &m_bgfxCallback;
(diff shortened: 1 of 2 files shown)


Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -27,6 +27,7 @@ namespace Babylon::Graphics
         auto& init = m_state.Bgfx.InitState;
         init.type = s_bgfxRenderType;
         init.resolution.reset = BGFX_RESET_VSYNC | BGFX_RESET_MAXANISOTROPY;
+        init.resolution.maxFrameLatency = 1;
```
ryantrem, line 30: What does this mean?

ryantrem (approved):

Core/Graphics/Source/DeviceImpl.cpp
bghgary: [IDXGIDevice1::SetMaximumFrameLatency (dxgi.h) - Win32 apps | Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgidevice1-setmaximumframelatency?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(DXGI%252FIDXGIDevice1%253A%253ASetMaximumFrameLatency)%3Bk(IDXGIDevice1%253A%253ASetMaximumFrameLatency)%3Bk(SetMaximumFrameLatency)%3Bk(SolutionItemsProject)%3Bk(SolutionItemsProject)%3Bk(DevLang-C%252B%252B)%3Bk(TargetOS-Windows)%3Bk(TargetFrameworkMoniker-MoCOM-CLR)%26rd%3Dtrue)

bghgary: I don't see any noticeable difference when resizing, but it's supposed to make the input vs render lag better. It is common for games to set this.

PR #1094, opened 2022-06-24
Implement ExternalTexture for metal

bghgary:


Plugins/ExternalTexture/Source/ExternalTexture_Metal.mm
```diff
@@ -0,0 +1,200 @@
...
+
+namespace Babylon::Plugins
+{
+    class ExternalTexture::Impl
+    {
+    public:
+        Impl(Graphics::TextureT ptr)
+        {
+            m_ptr = ptr;
+            [m_ptr retain];
```
ryantrem, line 128: We've had so many memory management issues on iOS, can we just enable ARC for the ExternalTexture lib? We do this in other BN libs in CMakeLists.txt with:
```
if (APPLE)
    set_target_properties(NativeTracing PROPERTIES
        XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES
    )
endif()
```

bghgary: I think so. That mean I don't need to call retain/release anymore, right?

ryantrem: Yes that should be correct. You can verify by leaving it there, in which case you should get a compile error when ARC is enabled.

ryantrem (approved):

SergioRZMasson (approved):

PR #1098, opened 2022-07-07
Add support for updating an external texture

bghgary:


SergioRZMasson (approved):
Looks fine for me.

Plugins/ExternalTexture/Source/ExternalTexture_Base.h
```diff
@@ -0,0 +1,79 @@
...
+        static bool IsFullMipChain(T1 mipLevel, T2 width, T3 height)
+        {
+            return mipLevel == static_cast<T1>(std::floor(std::log2(std::max(static_cast<float>(width), static_cast<float>(height))) + 1));
+        }
+
+        void UpdateHandles(uintptr_t ptr)
+        {
+            std::scoped_lock lock{m_mutex};
+
+            for (auto it = m_handles.begin(); it != m_handles.end(); ++it)
```
ryantrem, line 52: Could this also be:
```c++
for (auto handle : m_handles)
```

Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
```diff
@@ -1,7 +1,34 @@
...
+        uint16_t height{};
+        bgfx::TextureFormat::Enum format{bgfx::TextureFormat::Unknown};
+        uint64_t flags{};
+        bool hasMips{};
+        GetInfo(width, height, hasMips, format, flags);
+        assert(width == m_width);
+        assert(height == m_height);
+        assert(format == m_format);
+        assert(flags == m_flags);
+        assert(hasMips == m_hasMips);
```
ryantrem, line 26: Should these really only be debug asserts and not runtime errors?

Plugins/ExternalTexture/Source/ExternalTexture_D3D12.cpp
```diff
@@ -129,74 +131,63 @@ namespace
...
+
+        uintptr_t Ptr() const
+        {
+            return reinterpret_cast<uintptr_t>(m_ptr.get());
+        }
+
+    private:
+        void Init(Graphics::TextureT ptr)
         {
             m_ptr.copy_from(ptr);
```
ryantrem, line 149: This order seems weird for the update case, because you will actually have an instance of the `ExternalTexture` and the m_ptr will be updated but it will also throw. It seems like the copy_from should happen after the throw (doesn't happen).

ryantrem (approved):

Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
bghgary: I could make it a runtime error, but I'm not sure what impact it will have on perf.

Plugins/ExternalTexture/Source/ExternalTexture_D3D12.cpp
bghgary: I agree, but then the code will be harder to write.

PR #1100, opened 2022-07-19
Alpha premultiplied backbuffer

CedricGuillemet:
This change concerns mostly UWP/swapchainpanel but is also available for other platforms as bgfx allows it.
More infos on the rendering side here : https://github.com/bkaradzic/bgfx/pull/2837

- API contract mimics the MSAA one.
- bgfx update to the latest

### DIFF
--- a/Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
+++ b/Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
@@ -14,5 +14,7 @@ namespace Babylon::Graphics
         size_t Height{};
         // MSAA sample values can be 2, 4, 8 or 16. Any other value will disable MSAA
         uint8_t MSAASamples{};
+        // When enabled, back buffer will be premultiplied with alpha value
+        bool AlphaPremultiplied{};
     };
 }
--- a/Core/Graphics/Source/DeviceImpl.cpp
+++ b/Core/Graphics/Source/DeviceImpl.cpp
@@ -93,6 +93,22 @@ namespace Babylon::Graphics
         }
     }
 
+    void DeviceImpl::SetAlphaPremultiplied(bool enabled)
+    {
+        if (bgfx::getCaps()->supported & BGFX_CAPS_TRANSPARENT_BACKBUFFER)
+        {
+            std::scoped_lock lock{m_state.Mutex};
+            m_state.Bgfx.Dirty = true;
+            auto& init = m_state.Bgfx.InitState;
+            init.resolution.reset &= ~BGFX_CAPS_TRANSPARENT_BACKBUFFER;
+            init.resolution.reset |= enabled ? BGFX_CAPS_TRANSPARENT_BACKBUFFER : 0;
+        }
+        else
+        {
+            m_bgfxCallback.trace(__FILE__, __LINE__, "WARNING: BGFX_CAPS_TRANSPARENT_BACKBUFFER unsupported on this platform.");
+        }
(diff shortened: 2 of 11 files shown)


Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -93,6 +93,15 @@ namespace Babylon::Graphics
         }
     }
 
+    void DeviceImpl::SetAlphaPremultiplied(bool enabled)
+    {
+        std::scoped_lock lock{m_state.Mutex};
+        m_state.Bgfx.Dirty = true;
+        auto& init = m_state.Bgfx.InitState;
+        init.resolution.reset &= ~BGFX_RESET_TRANSPARENT_BACKBUFFER;
+        init.resolution.reset |= enabled ? BGFX_RESET_TRANSPARENT_BACKBUFFER : 0;
```
bghgary, line 102: nit: i find this to be easier to read
```suggestion
        if (enabled)
        {
            init.resolution.reset |= BGFX_RESET_TRANSPARENT_BACKBUFFER;
        }
        else
        {        
            init.resolution.reset &= ~BGFX_RESET_TRANSPARENT_BACKBUFFER;
        }
```

Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
```diff
@@ -14,5 +14,7 @@ namespace Babylon::Graphics
         size_t Height{};
         // MSAA sample values can be 2, 4, 8 or 16. Any other value will disable MSAA
         uint8_t MSAASamples{};
+        // When enabled, back buffer will be premultiplied with alpha value
+        bool AlphaPremultiplied{};
```
bghgary, line 18: Should we rename this to `TransparentBackbuffer` like in bgfx?

CedricGuillemet: I don't think so. There is a strong link between alpha premultiplied back buffer and transparency. Alpha premul is a post process needed in a particular case of view transparency. On iOS/Android, transparency is with straight alpha. And on these 2 platforms, setting true to this boolean property will not make the view transparent.

Core/Graphics/Source/DeviceImpl.cpp
CedricGuillemet: I like compactness. Like https://github.com/BabylonJS/BabylonNative/blob/4c270c2af5b1b47ac7f7a4584f2edbcab83388f2/Plugins/NativeEngine/Source/NativeEngine.cpp#L800

bghgary: It's a nit, so no worries either way.

Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
bghgary: Ok, I don't quite understand this flag, but we can discuss offline.

bghgary: Can you add a comment on what platforms are currently supported and then remove them once they are all supported?

bghgary (approved):

Core/Graphics/Include/Platform/Android/Babylon/Graphics/Platform.h
CedricGuillemet: Done.

PR #1101, opened 2022-07-19
Manage Java JNI object lifetimes from the C++ side

CoPrez:
While enabling the NativeCamera plugin in BabylonReactNative I started running into issues where the JNI layer was throwing exceptions due to calling methods with JNI jobject and jclass objects that had stale references on the JNI side. The issue was that the C++ layer holds onto the jclass and jobject references across boundary calls. That can lead to the Java side destroying the object while the C++ side still retains a reference to it.

The fix in this PR is to update the C++ side to create global references to the jobject and jclass references. That required adding getters/setters for managing registering and deregistering the global references from the Java side. Additionally I added copy and move semantic implementations to also ensure the jobject and jclass references are getting cleaned up/moved appropriately.

### DIFF
--- a/Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
+++ b/Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
@@ -72,6 +72,9 @@ namespace java::lang
     {
         if (this != &other)
         {
+            // Clear out the jclass using the current environment before updating to the
+            // other Class's environment and jclass. The jclass should be used in conjunction
+            // with the environment it was created in.
             JClass(nullptr);
             m_env = other.m_env;
             JClass(other.JClass());
@@ -119,9 +122,9 @@ namespace java::lang
             m_env->DeleteGlobalRef(m_class);
         }
 
-        m_class = nullptr;
+        m_class = classObj;
 
-        if (classObj)
+        if (m_class)
         {
             m_class = static_cast<jclass>(m_env->NewGlobalRef(classObj));
         }
@@ -160,6 +163,9 @@ namespace java::lang
     {
         if (this != &other)
         {
+            // Clear out the jobject using the current environment before updating to the
+            // other Object's environment and jobject. The jobject should be used in conjunction
+            // with the environment it was created in.
             JObject(nullptr);
             m_env = other.m_env;
             m_class = other.m_class;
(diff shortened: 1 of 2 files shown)


Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -68,6 +108,25 @@ namespace java::lang
...
+        if (m_class) {
+            m_env->DeleteGlobalRef(m_class);
+        }
+
+        m_class = nullptr;
+
+        if (classObj)
+        {
+            m_class = static_cast<jclass>(m_env->NewGlobalRef(classObj));
+        }
```
ryantrem, line 127: It seems like this should be:
```suggestion
        m_class = classObj;

        if (m_class)
        {
            m_class = static_cast<jclass>(m_env->NewGlobalRef(classObj));
        }
```
Was there some reason we couldn't do it that way? Can't remember.

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
```diff
@@ -48,16 +48,56 @@ namespace java::lang
...
+    Class::Class(const Class& other)
+        : Class(other.m_class)
+    {
+    }
+
+    Class& Class::operator=(const Class& other)
+    {
+        if (this != &other)
+        {
+            JClass(nullptr);
```
ryantrem, line 75: Maybe worth a comment that we do this explicitly because the env can change and then the release ref would be wrong. Alternatively we could have a getter/setter for the env and if it changes, clear out the ref and the stored class, but that might be overkill.

ryantrem (approved):

bghgary (approved):

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
CoPrez: No real reason. I've updated to your proposed format for both the Class and Object implementation.

Dependencies/AndroidExtensions/Source/JavaWrappers.cpp
CoPrez: I added a comment to the Class and Object copy operators to make it clear that the ordering is important and that the jobject/jclass should be kept in sync with the JNI enviroment.

CoPrez:
I've logged an issue to follow up on this PR and look into simplifying the Class and Object JNI wrappers: #1105

PR #1102, opened 2022-07-19
NativeCamera - Don't access Graphics::DeviceContext from the constructor

CoPrez:
While adding the NativeCamera Plugin to Babylon React-Native I was hitting an issue where the plugin couldn't be initialized until after Graphics::DeviceContext had been initialized. The issue was that the plugin's constructor was saving off the current deviceContext reference to use later in the runtime. Because it was only using the deviceContext in one location I just removed it from the constructor and instead fetch it as needed later on when the plugin is used.

### DIFF
--- a/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
+++ b/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
@@ -149,7 +149,7 @@ namespace Babylon::Plugins
 #endif
 
     Camera::Impl::Impl(Napi::Env env, bool overrideCameraTexture)
-        : m_deviceContext{Graphics::DeviceContext::GetFromJavaScript(env)}
+        : m_env{env}
         , m_overrideCameraTexture{overrideCameraTexture}
     {
 #if __ANDROID_API__ < 24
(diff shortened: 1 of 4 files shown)


bghgary:
> While adding the NativeCamera Plugin to Babylon React-Native I was hitting an issue where the plugin couldn't be initialized until after Graphics::DeviceContext had been initialized.

I'm curious why this will happen. The device context should get created before plugins are initialized right?

CoPrez:
> 

@ryantrem, may be able to provide more context for why BabylonReactNative initializes most plugins before the DeviceContext. [BabylonNative.cpp](https://github.com/BabylonJS/BabylonReactNative/blob/master/Modules/%40babylonjs/react-native/shared/BabylonNative.cpp) in BabylonReactNative are where the existing plugins are being initialized. Most are initialized before the React component has been rendered, but the DeviceContext isn't registered until after the component receives its first update.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -239,7 +239,7 @@ namespace Babylon::Plugins
                 auto id = GetCameraId(frontCamera);
                 ACameraManager_openCamera(m_cameraManager, id.c_str(), &cameraDeviceCallbacks, &m_cameraDevice);
 
-                m_textureWindow = ANativeWindow_fromSurface(GetEnvForCurrentThread(), surface);
+                m_textureWindow = ANativeWindow_fromSurface(m_env, surface);
```
CoPrez, line 242: I am reverting this change. I was mixing up the Napi env and the JNI env.

bghgary (approved):
LGTM

CedricGuillemet (approved):

PR #1104, opened 2022-07-19
Native readPixels implementation

ryantrem:
This change provides a native implementation for reading a texture to obtain an rbga8 buffer. It does not currently support cube maps.

Notable changes:
- Added more properties to Graphics::Texture so we have enough info to read pixels from an arbitrary texture.
- Added a ReadTexture function to NativeEngine that is called by the JS side.

See the corresponding Babylon.js change here: https://github.com/BabylonJS/Babylon.js/pull/12768


Apps/Playground/Scripts/experience.js
```diff
@@ -83,6 +84,30 @@ CreateBoxAsync(scene).then(function () {
...
+                return buffer;
+            })
+            .then(buffer => {
+                // This read reuses the existing buffer.
+                texture.readPixels(undefined, mip, buffer, undefined, undefined, x, y, width, height).then((buffer) => {
+                    console.log(`Read ${buffer.byteLength} pixel bytes.`);
+                });
+            });
+        });
+    }
```
bghgary, line 109: I don't like this pattern, but it's okay for now. We need to move these into jest tests once jest is ready to use.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1314,6 +1318,106 @@ namespace Babylon
...
+            // Allocate a buffer to store the source pixel data.
+            std::vector<uint8_t> textureBuffer(sourceTextureInfo.storageSize);
+
+            // Read the source texture.
+            m_graphicsContext.ReadTextureAsync(sourceTextureHandle, textureBuffer, mipLevel).then(arcana::inline_scheduler, *m_cancellationSource, [this, textureBuffer{std::move(textureBuffer)}, sourceTextureInfo, targetTextureInfo]() mutable {
+                // If the source texture format does not match the target texture format, convert it.
+                if (targetTextureInfo.format != sourceTextureInfo.format)
+                {
+                    std::vector<uint8_t> convertedTextureBuffer(targetTextureInfo.storageSize);
+                    bimg::imageConvert(&m_allocator, convertedTextureBuffer.data(), bimg::TextureFormat::Enum(targetTextureInfo.format), textureBuffer.data(), bimg::TextureFormat::Enum(sourceTextureInfo.format), sourceTextureInfo.width, sourceTextureInfo.height, /*depth*/ 1);
```
bghgary, line 1390: Should probably check the result and throw exception if it failed.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1314,6 +1318,106 @@ namespace Babylon
...
+                // Flip the image vertically if needed.
+                if (bgfx::getCaps()->originBottomLeft)
+                {
+                    FlipImage(textureBuffer, targetTextureInfo.height);
+                }
+
+                return textureBuffer;
+            }).then(m_runtimeScheduler, *m_cancellationSource, [bufferRef{Napi::Persistent(buffer)}, deferred](std::vector<uint8_t> textureBuffer) {
+                // Copy the pixel data into the JS ArrayBuffer.
+                std::memcpy(bufferRef.Value().Data(), textureBuffer.data(), textureBuffer.size());
```
bghgary, line 1403: Probably should assert that bufferRef.Value().ByteLength() is big enough to hold the data. You have a check earlier, but this is just in case.

bghgary, line 1403: You need to take `bufferOffset` into account?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1314,6 +1318,106 @@ namespace Babylon
...
+                deferred.Resolve(bufferRef.Value());
+            }).then(m_runtimeScheduler, arcana::cancellation::none(), [env, deferred, tempTexture, sourceTextureHandle](const arcana::expected<void, std::exception_ptr>& result) {
+                if (tempTexture)
+                {
+                    bgfx::destroy(sourceTextureHandle);
+                }
```
bghgary, line 1409: This technically can still leak right now as we don't wait for continuations to complete before shutting down, but we can deal with this later after the pinnable cancellation tokens are merged.

Plugins/NativeEngine/Source/NativeEngine.cpp
ryantrem: Updated. I'm throwing a regular C++ exception which I believe would be properly converted in the final continuation.

Plugins/NativeEngine/Source/NativeEngine.cpp
ryantrem: I realized that I was only checking that the offset/length passed in were valid for the buffer passed in, but I wasn't actually checking that the buffer length was sufficient for the output pixel data, so I added that check as well. With that, I believe the only way things could be wrong on this line is due to a programming error, so I also added an assert ensuring the final textureBuffer size matches the pre-computed storageSize of the output (which is what is validated earlier in the function).

ryantrem: Yep, oops.

bghgary (approved):

PR #1109, opened 2022-07-21
Fix crash on smooth shading for meshes with no index buffers.

### LINKED ISSUE #1108: Smooth shading is not working in Babylon Native
Reported from user on the forum: https://forum.babylonjs.com/t/smooth-shading-is-not-working-in-babylon-react-native/32295

**Describe the bug**
When enabling forceSharedVertices on a mesh that has no index buffers (that will be rendered by lines or other types of primitives), causes crash on Babylon Native and Babylon React Native applications. 

**Expected behavior**
Smooth shading behavior for Babylon Native should match the one in Babylon.js

SergioRZMasson:
## Overview 

When forceSharedVertices is called on a mesh that does not have a index buffer, the forceSharedVertices will create empty arrays for its positions and normals vertex attributes. This will result in a NativeEngine::RecordVertexBuffer been called with undefined arguments been passed as VertexBuffer* argument. 

A similar behavior happens on the Web engine; however, the web engine creates vertex attribute arrays of size zero, which results in no rendering for those meshes. By simply returning from NativeEngine::RecordVertexBuffer we get the same behavior on the Native engine.  

## Issues
- Fixes #1108

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -608,6 +608,12 @@ namespace Babylon
     void NativeEngine::RecordVertexBuffer(const Napi::CallbackInfo& info)
     {
         VertexArray* vertexArray = info[0].As<Napi::Pointer<VertexArray>>().Get();
+        
+        auto vertexBufferValue = info[1];
+        
+        if (vertexBufferValue.IsUndefined())
+            return;
+
         VertexBuffer* vertexBuffer = info[1].As<Napi::Pointer<VertexBuffer>>().Get();
         const uint32_t location = info[2].As<Napi::Number>().Uint32Value();
         const uint32_t byteOffset = info[3].As<Napi::Number>().Uint32Value();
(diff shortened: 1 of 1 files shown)


CedricGuillemet (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -608,7 +608,13 @@ namespace Babylon
     void NativeEngine::RecordVertexBuffer(const Napi::CallbackInfo& info)
     {
         VertexArray* vertexArray = info[0].As<Napi::Pointer<VertexArray>>().Get();
-        VertexBuffer* vertexBuffer = info[1].As<Napi::Pointer<VertexBuffer>>().Get();
+        
+        auto vertexBufferValue = info[1];
+        
+        if (vertexBufferValue.IsUndefined())
+            return;
```
bghgary, line 615: I would recommend doing this check on the JS side.

bghgary (changes requested):

SergioRZMasson:
This fix was moved to the JS side and should be solved with PR https://github.com/BabylonJS/Babylon.js/pull/12781

Issue #1107, opened 2022-07-20
Fix various issues with NativeCamera/VideoTexture


ryantrem:
An initial implementation of `NativeCamera` and `VideoTexture` was merged mid 2021 (for Android and iOS), but these implementations have some issues that need to be addressed for these components be ready for production usage:

### Android
- [x] Resolution selection - currently, it seems NativeCamera on Android gets the highest resolution camera feed, creates a texture that is exactly sized to the min width/height requested, and then uses a shader to draw the full res camera texture into the newly created texture. Instead, I believe we need to be querying the camera to find a resolution that most closely satisfies the constraints (min width/height), and create a second (renderable) texture that is the same size, and then draw the source texture into the target texture. NOTE: It seems like there might be a useful example for resolution/frame rate enumeration for [Camera2](https://developer.android.com/training/camera2) here: https://github.com/android/camera-samples/tree/main/Camera2Video. It is a Java example, but should map the the [native API](https://developer.android.com/ndk/reference/group/camera) closely.
  - It looks to me like we can enumerate resolutions like this:
        ```
        ACameraMetadata_const_entry configInfo{};
        ACameraMetadata_getConstEntry(metadataObj, ACAMERA_SCALER_AVAILABLE_STREAM_CONFIGURATIONS, &configInfo);
        for (int32_t i = 0; i < configInfo.count; i += 4)
        {
            const int32_t width{entry.data.i32[i + 1]};
            const int32_t height{entry.data.i32[i + 2]};
        }
        ```
  - From https://developer.android.com/ndk/reference/group/camera#acameradevice_createcapturesession, it looks to me like we can select a resolution by calling `m_surfaceTexture.setDefaultBufferSize` (needs to be added to JavaWrappers) and setting the resolution *before* creating the `Surface` from the `SurfaceTexture`.
- [x] Orientation - from some quick testing, it seems like the VideoTexture orientation/rotation is not the same as in the browser. We should take a closer look and see if the shader should be doing some additional rotation or something.
- [x] API Level - the NDK Camera2 API is used, which requires API level 24. The code currently does a compile time check, but really we want to do a runtime check and return an error or something so that the package as a whole can still be used with lower API levels (and NativeCamera will fail gracefully).
  - #1112
[...]

PR #1112, opened 2022-07-29
NativeCamera Fixes and Improvements

CoPrez:
This PR addresses a couple issues with the NativeCamera implementation. The first issue was that the `NativeCamera::Open` method returned a promise in an already completed state. For iOS this doesn't seem to be a problem. On Android it works so long as the permissions to use the camera were previously granted leading to the function executing synchronously. On Android if the permission prompt was showing the call to update the texture would be called while the Open method was still initializing leading to an error thrown trying to access the graphic context member variable. The fix was to have the `NativeCamera::Open` method return a promise in an uncompleted state and only complete it after the internal async work has finished.

The second issue this PR address was that the NativeCameraImpl for Android was using if defs to access the NDK camera API which requires Android API 24 or higher. That means that the consumer can only use NativeCamera if they define their minimum supported API as 24 or higher. This poses problems for BabylonReactNative where the BabylonNative library is pre-compiled against API 21. Another curveball is that on Android all symbols for a library must be fulfilled at the time of loading. That means a library statically linked against a specific API often cannot be loaded on a device running an older API ([more info](https://developer.android.com/ndk/guides/sdk-versions#compilesdkversion)).

This PR changes from using pre-processor arguments to access the newer API's to using dynamic linking (`dlopen` and `dlsym`) to access the API at runtime. This means that the functionality requiring newer API support can still be accessed if the device OS supports it. Older devices still gracefully handle the scenario by using a runtime check to access the current devices API level support.

### DIFF
--- a/Plugins/NativeCamera/Source/Android/CameraWrappers.h
+++ b/Plugins/NativeCamera/Source/Android/CameraWrappers.h
@@ -13,7 +13,7 @@
 #undef __ANDROID_API__
 #define __ANDROID_API__ 24
 namespace API24 {
-#include <camera/NdkCameraManager.h>
+    #include <camera/NdkCameraManager.h>
     #include <camera/NdkCameraCaptureSession.h>
     #include <camera/NdkCameraDevice.h>
     #include <camera/NdkCameraError.h>
@@ -32,18 +32,18 @@ namespace API24 {
 
 namespace
 {
-    static const int API_LEVEL{ android_get_device_api_level() };
+    const int API_LEVEL{ android_get_device_api_level() };
 
     // Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
     // on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
     // to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
-    static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW): nullptr };
+    void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW): nullptr };
 
-    static std::map<std::string, void*> CameraDynamicFunctions{};
+    std::map<std::string, void*> CameraDynamicFunctions{};
 }
 
 namespace Babylon::Plugins {
-    static inline void* GetCameraDynamicFunction(const char* functionName) {
+    inline void* GetCameraDynamicFunction(const char* functionName) {
         if (!libCamera2NDK)
         {
             return nullptr;
--- a/Plugins/NativeCamera/Source/Android/CameraWrappers.h
+++ b/Plugins/NativeCamera/Source/Android/CameraWrappers.h
@@ -37,7 +37,7 @@ namespace
     // Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
     // on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
     // to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
-    static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW | RTLD_LOCAL): nullptr };
+    static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW): nullptr };
 
     static std::map<std::string, void*> CameraDynamicFunctions{};
 }
--- a/Plugins/NativeCamera/Source/Android/CameraWrappers.h
+++ b/Plugins/NativeCamera/Source/Android/CameraWrappers.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <android/api-level.h>
-#include <arcana/macros.h>
 #include <dlfcn.h>
 #include <map>
 
@@ -29,7 +28,7 @@ namespace API24 {
 #undef __ORIG_ANDROID_API__
 
 // Helper macro to make calling camera NDK api's more type safe
-#define GET_CAMERA_FUNCTION(function) reinterpret_cast<decltype(&API24::function)>(GetCameraDynamicFunction(#function))
+#define GET_CAMERA_FUNCTION(function) reinterpret_cast<decltype(&API24::function)>(Babylon::Plugins::GetCameraDynamicFunction(#function))
 
 static const int API_LEVEL{ android_get_device_api_level() };
 
@@ -40,17 +39,18 @@ static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NO
 
 static std::map<std::string, void*> CameraDynamicFunctions{};
 
-static inline void* GetCameraDynamicFunction(const char* functionName)
-{
-    if (!libCamera2NDK)
-    {
-        return nullptr;
-    }
-    auto cameraDynamicFunctionIt{CameraDynamicFunctions.find(functionName)};
-    if (cameraDynamicFunctionIt == CameraDynamicFunctions.end())
-    {
-        cameraDynamicFunctionIt = CameraDynamicFunctions.emplace(functionName, dlsym(libCamera2NDK, functionName)).first;
-    }
+namespace Babylon::Plugins {
+    static inline void* GetCameraDynamicFunction(const char* functionName) {
+        if (!libCamera2NDK)
+        {
+            return nullptr;
+        }
+        auto cameraDynamicFunctionIt{CameraDynamicFunctions.find(functionName)};
+        if (cameraDynamicFunctionIt == CameraDynamicFunctions.end())
+        {
+            cameraDynamicFunctionIt = CameraDynamicFunctions.emplace(functionName, dlsym(libCamera2NDK, functionName)).first;
+        }
 
-    return cameraDynamicFunctionIt->second;
+        return cameraDynamicFunctionIt->second;
+    }
 }
--- a/Plugins/NativeCamera/Source/Android/CameraWrappers.h
+++ b/Plugins/NativeCamera/Source/Android/CameraWrappers.h
@@ -30,14 +30,17 @@ namespace API24 {
 // Helper macro to make calling camera NDK api's more type safe
 #define GET_CAMERA_FUNCTION(function) reinterpret_cast<decltype(&API24::function)>(Babylon::Plugins::GetCameraDynamicFunction(#function))
 
-static const int API_LEVEL{ android_get_device_api_level() };
+namespace
+{
+    static const int API_LEVEL{ android_get_device_api_level() };
 
-// Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
-// on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
-// to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
-static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW | RTLD_LOCAL): nullptr };
+    // Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
+    // on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
+    // to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
+    static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW | RTLD_LOCAL): nullptr };
 
-static std::map<std::string, void*> CameraDynamicFunctions{};
+    static std::map<std::string, void*> CameraDynamicFunctions{};
+}
 
 namespace Babylon::Plugins {
     static inline void* GetCameraDynamicFunction(const char* functionName) {
--- a/Plugins/NativeCamera/Source/Android/CameraWrappers.h
+++ b/Plugins/NativeCamera/Source/Android/CameraWrappers.h
@@ -1,0 +1,56 @@
+#pragma once
+
+#include <android/api-level.h>
+#include <arcana/macros.h>
+#include <dlfcn.h>
+#include <map>
+
+// The camera API's in the NDK are only supported in API24 and higher. We
+// use dlopen and dlsym to link the camera lib at runtime only if the current
+// device supports it. In order to avoid redefining the type information the
+// below macro's forcefully import the camera header files as if the min supported
+// SDK was set to 24.
+#define __ORIG_ANDROID_API__ __ANDROID_API__
+#undef __ANDROID_API__
+#define __ANDROID_API__ 24
+namespace API24 {
+#include <camera/NdkCameraManager.h>
+    #include <camera/NdkCameraCaptureSession.h>
+    #include <camera/NdkCameraDevice.h>
+    #include <camera/NdkCameraError.h>
+    #include <camera/NdkCameraManager.h>
+    #include <camera/NdkCameraMetadata.h>
+    #include <camera/NdkCameraMetadataTags.h>
+    #include <camera/NdkCameraWindowType.h>
+    #include <camera/NdkCaptureRequest.h>
+}
+#undef __ANDROID_API__
+#define __ANDROID_API__ __ORIG_ANDROID_API__
+#undef __ORIG_ANDROID_API__
+
+// Helper macro to make calling camera NDK api's more type safe
+#define GET_CAMERA_FUNCTION(function) reinterpret_cast<decltype(&API24::function)>(GetCameraDynamicFunction(#function))
+
+static const int API_LEVEL{ android_get_device_api_level() };
+
+// Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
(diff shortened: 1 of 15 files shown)


Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -123,6 +123,9 @@ - (id)init:(std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData>)implData;
             [m_implData->avCaptureSession commitConfiguration];
             [m_implData->avCaptureSession startRunning];
         });
+        
+        // This should capture an actual task and return it... but we'll deal with that later
```
CoPrez, line 127: I intend to address the iOS implementation so that the task completion properly awaits the async work before I complete the PR.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -54,88 +65,103 @@ namespace Babylon::Plugins
...
                 break;
             }
         }
 
-        ACameraManager_deleteCameraIdList(cameraIds);
+        GET_CAMERA_FUNCTION(ACameraManager_deleteCameraIdList)(cameraIds);
         return cameraId;
     }
 
+    void* Camera::Impl::GetCameraDynamicFunction(const char* functionName)
```
ryantrem, line 97: I'd be more inclined to just make this a global function and put it up by `libCamera2NDK`. I think storing the map globally is fine as well. Any specific reason to have it scoped to an instance of NativeCameraImpl?

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -54,88 +65,103 @@ namespace Babylon::Plugins
...
+        {
+            return m_cameraDynamicFunctions.at(functionName);
+        }
+
+        if (libCamera2NDK)
+        {
+            return dlsym(libCamera2NDK, functionName);
+        }
+
+        return nullptr;
```
ryantrem, line 111: I would do something like:
```suggestion
        if (!libCamera2NDK)
        {
            return nullptr;
        }

        auto cameraDynamicFunctionIt{m_cameraDynamicFunctions.find(functionName)};
        if (cameraDynamicFunction == m_cameraDynamicFunctions.end())
        {
            cameraDynamicFunctionIt = m_cameraDynamicFunctions.insert(functionName, dlsym(libCamera2NDK, functionName));
        }
        
        return cameraDynamicFunctionIt->second;
```

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
CoPrez: The latest iteration now runs asynchronously and returns a proper task to await.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: In the latest iteration I've moved all of the dynamic invocation logic into a CameraWrappers.h file to better consolidate it.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: In the latest iteration iteration I've applied your suggestion.

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,56 @@
...
+#define GET_CAMERA_FUNCTION(function) reinterpret_cast<decltype(&API24::function)>(GetCameraDynamicFunction(#function))
+
+static const int API_LEVEL{ android_get_device_api_level() };
+
+// Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
+// on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
+// to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
+static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW | RTLD_LOCAL): nullptr };
+
+static std::map<std::string, void*> CameraDynamicFunctions{};
```
ryantrem, line 41: Put these globals in an anonymous namespace to make sure they don't conflict with other global symbols:
```suggestion
namespace
{
    static const int API_LEVEL{ android_get_device_api_level() };

    // Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
    // on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
    // to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
    static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW | RTLD_LOCAL): nullptr };

    static std::map<std::string, void*> CameraDynamicFunctions{};
}
```

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,56 @@
...
+        return nullptr;
+    }
+    auto cameraDynamicFunctionIt{CameraDynamicFunctions.find(functionName)};
+    if (cameraDynamicFunctionIt == CameraDynamicFunctions.end())
+    {
+        cameraDynamicFunctionIt = CameraDynamicFunctions.emplace(functionName, dlsym(libCamera2NDK, functionName)).first;
+    }
+
+    return cameraDynamicFunctionIt->second;
+}
```
ryantrem, line 56: Put this in a Babylon::Plugins namespace to make sure it doesn't conflict with any global symbols (and update the macro to reference it from the namespace, e.g. `Babylon::Plugins::GetCameraDynamicFunction`):
```suggestion
namespace Babylon::Plugins
{
    static inline void* GetCameraDynamicFunction(const char* functionName)
    {
        if (!libCamera2NDK)
        {
            return nullptr;
        }
        auto cameraDynamicFunctionIt{CameraDynamicFunctions.find(functionName)};
        if (cameraDynamicFunctionIt == CameraDynamicFunctions.end())
        {
            cameraDynamicFunctionIt = CameraDynamicFunctions.emplace(functionName, dlsym(libCamera2NDK, functionName)).first;
        }

        return cameraDynamicFunctionIt->second;
    }
}
```

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,56 @@
...
+
+// Helper macro to make calling camera NDK api's more type safe
+#define GET_CAMERA_FUNCTION(function) reinterpret_cast<decltype(&API24::function)>(GetCameraDynamicFunction(#function))
+
+static const int API_LEVEL{ android_get_device_api_level() };
+
+// Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
+// on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
+// to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
+static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW | RTLD_LOCAL): nullptr };
```
ryantrem, line 39: What happens if this library has already been implicitly loaded because the consuming application is API level 24+ and it explicitly uses the Camera2 NDK APIs itself? Does it return the handle to the already loaded library? Or does it return null? Or throw?

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: Added

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: They're now in a namespace.

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: I tested this out by re-enabling the CMAKE file that causes the consumer to link against the libcamera2ndk and then ran it in the Playground app (with the min API set to 24). Everything still worked correctly when running the app on device with a camera feed being rendered in the scene.

I'm not sure if the pointer we get from `dlopen` is the same one that would be used internally for the statically linked library, but it does at least work as intended still.

ryantrem (approved):

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,59 @@
...
+namespace
+{
+    static const int API_LEVEL{ android_get_device_api_level() };
+
+    // Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
+    // on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
+    // to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
+    static void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW): nullptr };
+
+    static std::map<std::string, void*> CameraDynamicFunctions{};
```
bghgary, line 42: nit: static is extraneous on an anonymous namespace

```suggestion
    const int API_LEVEL{ android_get_device_api_level() };

    // Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
    // on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
    // to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
    void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW): nullptr };

    std::map<std::string, void*> CameraDynamicFunctions{};
```

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,59 @@
...
+}
+
+namespace Babylon::Plugins {
+    static inline void* GetCameraDynamicFunction(const char* functionName) {
```
bghgary, line 46: [`static`](https://stackoverflow.com/questions/2271902/static-vs-global) doesn't make much sense here.
```suggestion
    inline void* GetCameraDynamicFunction(const char* functionName) {
```

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,59 @@
...
+        if (!libCamera2NDK)
+        {
+            return nullptr;
+        }
+        auto cameraDynamicFunctionIt{CameraDynamicFunctions.find(functionName)};
+        if (cameraDynamicFunctionIt == CameraDynamicFunctions.end())
+        {
+            cameraDynamicFunctionIt = CameraDynamicFunctions.emplace(functionName, dlsym(libCamera2NDK, functionName)).first;
+        }
```
bghgary, line 55: Do we know for sure if this is faster than just calling `dlsym` every time? Is this function slow?

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,59 @@
...
+    #include <camera/NdkCameraMetadataTags.h>
+    #include <camera/NdkCameraWindowType.h>
+    #include <camera/NdkCaptureRequest.h>
+}
+#undef __ANDROID_API__
+#define __ANDROID_API__ __ORIG_ANDROID_API__
+#undef __ORIG_ANDROID_API__
+
+// Helper macro to make calling camera NDK api's more type safe
+#define GET_CAMERA_FUNCTION(function) reinterpret_cast<decltype(&API24::function)>(Babylon::Plugins::GetCameraDynamicFunction(#function))
```
bghgary, line 31: This is an interesting way to do it. I'm not sure I like it. The more typical way is to store each function the code will call in an explicit list of function pointers.

bghgary (approved):
No need to block the PR, but I have some comments.

Plugins/NativeCamera/Source/Android/CameraWrappers.h
ryantrem: No we don't actually know. If we just store the individual function pointers (like you suggest below), then we'd only call dlsym once per function anyway. We could maybe test that out, or save it for a future change. 🤷🏼‍♂️

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,59 @@
...
+// The camera API's in the NDK are only supported in API24 and higher. We
+// use dlopen and dlsym to link the camera lib at runtime only if the current
+// device supports it. In order to avoid redefining the type information the
+// below macro's forcefully import the camera header files as if the min supported
+// SDK was set to 24.
+#define __ORIG_ANDROID_API__ __ANDROID_API__
+#undef __ANDROID_API__
+#define __ANDROID_API__ 24
+namespace API24 {
+#include <camera/NdkCameraManager.h>
```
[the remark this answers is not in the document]
CoPrez: Fixed

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: Removed

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: Removed

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: I didn't test it out so take this with a grain of salt. The [docs for dlsym](https://man7.org/linux/man-pages/man3/dlsym.3.html) call out that it will do a breadth first search for the symbol first through the library passed in and then through any other shared/global libraries loaded with `dlopen`. It doesn't indicate any caching is in place, so I would say it's a good call to do some form of caching on our own side.

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: There are two main reasons I went with this route of loading the functions on demand:
1. If you don't end up using the functions you don't pay any cost of finding them in the library. It's probably a negligible amount of time... but I liked knowing they wouldn't get loaded if the NativeCamera wasn't used.
2. It reduces the boilerplate needed to call a new function out of the NDK Camera API. I'm hoping to do some follow up PR's to make the NativeCamera implementation a bit more robust (such as the issues called out in #1107). In doing so I'd like to be able to call other functions in the Camera API as easily as possible. With this current route the macro makes that very easy. Pre-defining the functions isn't much extra work... but it is a little.

As long as you're okay with it in it's current form I'd like to leave it that way and then change it later if it becomes an issue.

bghgary: > If you don't end up using the functions you don't pay any cost of finding them in the library. It's probably a negligible amount of time... but I liked knowing they wouldn't get loaded if the NativeCamera wasn't used.

This is both a pro and a con. The con part is that you won't know about the missing function until you hit the code path that uses it which can cause bugs to slip through. It's certainly not a deal breaker.

Plugins/NativeCamera/Source/Android/CameraWrappers.h
bghgary: Either way, it would be interesting to know, but it's totally fine to merge without addressing this.

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,59 @@
...
+
+// The camera API's in the NDK are only supported in API24 and higher. We
+// use dlopen and dlsym to link the camera lib at runtime only if the current
+// device supports it. In order to avoid redefining the type information the
+// below macro's forcefully import the camera header files as if the min supported
+// SDK was set to 24.
+#define __ORIG_ANDROID_API__ __ANDROID_API__
+#undef __ANDROID_API__
+#define __ANDROID_API__ 24
+namespace API24 {
```
bghgary, line 15: nit:
```suggestion
namespace API24
{
```

Plugins/NativeCamera/Source/Android/CameraWrappers.h
```diff
@@ -0,0 +1,59 @@
...
+    // Load the NDK camera lib dynamically. When running on OS 6.0 and below this will return nullptr,
+    // on OS 7.0 and up this should return a pointer to access the library using dlsym. It is technically fine
+    // to call dlopen when the API_LEVEL is below 24, but it's wasted effort on those devices.
+    void* libCamera2NDK{ API_LEVEL >= 24 ? dlopen("libcamera2ndk.so", RTLD_NOW): nullptr };
+
+    std::map<std::string, void*> CameraDynamicFunctions{};
+}
+
+namespace Babylon::Plugins {
+    inline void* GetCameraDynamicFunction(const char* functionName) {
```
bghgary, line 46: nit:
```suggestion
    inline void* GetCameraDynamicFunction(const char* functionName)
    {
```

bghgary (commented):
Sorry, a few more nits I just noticed.

Plugins/NativeCamera/Source/Android/CameraWrappers.h
CoPrez: I think this was Android Studio auto formatting my file on changes... or I just made mistakes :)

PR #1117, opened 2022-08-11
Add CMake setting for iOS product bundle identifiers

docEdub:
The iOS product bundle identifiers need to be changed when building locally for development purposes but are hard-coded as "com.BabylonNative.Playground.iOS" and "com.BabylonNative.ValidationTests.iOS".

This change adds CMake settings for the iOS product bundle identifiers so they can be changed without modifying the CMake files directly.


ryantrem:
> The iOS product bundle identifiers need to be changed when building locally for development purposes

Why? What problem is this solving? Not too familiar with product bundle ids.

docEdub:
> > The iOS product bundle identifiers need to be changed when building locally for development purposes
> 
> Why? What problem is this solving? Not too familiar with product bundle ids.

On my machine I get a bundle id error when building for iOS devices:
```
error: Failed to register bundle identifier: The app identifier "com.BabylonNative.Playground.iOS" cannot be registered to your development team because it is not available. Change your bundle identifier to a unique string to try again.
```
Changing the id gets the build working.

bghgary:
That sounds like something else is wrong. I don't think we need this change.

PR #1119, opened 2022-08-17
Use CMake `DEPLOYMENT_TARGET` variable to set iPhoneOS deployment target

docEdub:
There is a mismatch in the `XCODE_ATTRIBUTE_IPHONEOS_DEPLOYMENT_TARGET` CMake variables used for libraries and apps. Libraries set the deployment target using the `DEPLOYMENT_TARGET` CMake variable. Apps have the deployment target hard-coded to `9.0`, which causes build warnings when the `DEPLOYMENT_TARGET` CMake variable is not set to `9.0`.

This change avoids the build warnings by setting the deployment target to the `DEPLOYMENT_TARGET` CMake variable in the playground and validation test apps.


bghgary (approved):
Nice. The warnings have been annoying. :-)

PR #1121, opened 2022-08-23
Push camera dimensions back to NativeVideo texture

Alex-MSFT:
This PR pushes the actual used camera dimension back to the NativeVideo texture so that readPixels and getSize return the proper dimensions of the object rather than defaulting to whatever is passed in for maxHeight and maxWidth as outlined in #1107.  With this change we will properly report the correct resolution on both iOS and Android.  Android requires additional follow up work to select the correct camera configuration to most closely match the target resolution.

### DIFF
--- a/Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
+++ b/Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
@@ -7,11 +7,6 @@ namespace Babylon::Plugins
     class Camera final
     {
     public:
-        struct CameraDimensions {
-            uint32_t width;
-            uint32_t height;
-        };
-
         class Impl;
 
         Camera(const Camera& other) = default;
--- a/Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
+++ b/Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
@@ -7,6 +7,11 @@ namespace Babylon::Plugins
     class Camera final
     {
     public:
+        struct CameraDimensions {
+            uint32_t width;
+            uint32_t height;
+        };
+
         class Impl;
 
         Camera(const Camera& other) = default;
--- a/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
+++ b/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
@@ -160,7 +160,7 @@ namespace Babylon::Plugins
     {
     }
 
-    arcana::task<Camera::CameraDimensions, std::exception_ptr> Camera::Impl::Open(uint32_t maxWidth, uint32_t maxHeight, bool frontCamera)
+    arcana::task<Camera::Impl::CameraDimensions, std::exception_ptr> Camera::Impl::Open(uint32_t maxWidth, uint32_t maxHeight, bool frontCamera)
     {
         if (!m_deviceContext){
             m_deviceContext = &Graphics::DeviceContext::GetFromJavaScript(m_env);
--- a/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
+++ b/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
@@ -160,7 +160,7 @@ namespace Babylon::Plugins
     {
     }
 
-    arcana::task<Camera::CameraDimensions*, std::exception_ptr> Camera::Impl::Open(uint32_t maxWidth, uint32_t maxHeight, bool frontCamera)
+    arcana::task<const Camera::CameraDimensions*, std::exception_ptr> Camera::Impl::Open(uint32_t maxWidth, uint32_t maxHeight, bool frontCamera)
     {
         if (!m_deviceContext){
             m_deviceContext = &Graphics::DeviceContext::GetFromJavaScript(m_env);
@@ -275,7 +275,7 @@ namespace Babylon::Plugins
                 throw std::runtime_error{"Unable to restore GL context for camera texture init."};
             }
 
-            return &m_cameraDimensions;
+            return static_cast<const Camera::CameraDimensions*>(&m_cameraDimensions);
         });
     }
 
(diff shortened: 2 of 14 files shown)


CoPrez (approved):

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.h
```diff
@@ -15,7 +15,7 @@ namespace Babylon::Plugins
         
         Impl(Napi::Env env, bool overrideCameraTexture);
         ~Impl();
-        arcana::task<void, std::exception_ptr> Open(uint32_t width, uint32_t height, bool frontCamera);
+        arcana::task<const Camera::CameraDimensions*, std::exception_ptr> Open(uint32_t maxWidth, uint32_t maxHeight, bool frontCamera);
```
ryantrem, line 18: Why pass this as a pointer? It's a tiny struct, and there is no intention for the caller to be able to observe mutations to this struct right? Seems like it should just be returned by value.

Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
```diff
@@ -7,6 +7,11 @@ namespace Babylon::Plugins
     class Camera final
     {
     public:
+        struct CameraDimensions {
+            uint32_t width;
+            uint32_t height;
+        };
```
ryantrem, line 13: Shouldn't this be in NativeCameraImpl.h? This is not part of the public contract, just the internal one, yes?

Alex-MSFT: I put it here to not redefine it for every .h file, but I can move it.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -274,6 +274,8 @@ namespace Babylon::Plugins
             {
                 throw std::runtime_error{"Unable to restore GL context for camera texture init."};
             }
+
+            return static_cast<const Camera::CameraDimensions*>(&m_cameraDimensions);
```
ryantrem, line 278: Why is this cast here?

Plugins/NativeCamera/Source/NativeVideo.cpp
```diff
@@ -149,13 +149,18 @@ namespace Babylon::Plugins
...
             m_IsPlaying = true;
-            NativeCameraImpl->Open(m_width, m_height, m_frontCamera).then(m_runtimeScheduler, arcana::cancellation::none(), [this, env, deferred](const arcana::expected<void, std::exception_ptr>& result) {
+            NativeCameraImpl->Open(m_width, m_height, m_frontCamera).then(m_runtimeScheduler, arcana::cancellation::none(), [this, env, deferred](const arcana::expected<const Camera::CameraDimensions*, std::exception_ptr>& result) {
                 if (result.has_error())
                 {
                     deferred.Reject(Napi::Error::New(env, result.error()).Value());
                 }
                 else
                 {
+                    if (result.has_value()) {
```
ryantrem, line 159: Is this check actually needed? If it doesn't have an error, it must have a value, right?

Plugins/NativeCamera/Source/NativeVideo.cpp
```diff
@@ -149,13 +149,18 @@ namespace Babylon::Plugins
...
+                        auto cameraDimensions{result.value()};
+                        this->m_width = cameraDimensions->width;
+                        this->m_height = cameraDimensions->height;
```
ryantrem, line 162: So now we *update* the width/height after the open call returns, but we still initially store the max width/height as the width/height. Do you know how this compares to what happens on the web? I kind of would expect the width/height to be undefined until open or something.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.h
```diff
@@ -29,8 +29,7 @@ namespace Babylon::Plugins
 
         bool m_overrideCameraTexture;
 
-        uint32_t m_width{};
-        uint32_t m_height{};
+        Camera::CameraDimensions m_cameraDimensions{};
```
ryantrem, line 32: So basically we are using this to store the max width/height, then overriding it with the actual width/height after the open call. I wonder if it would make core sense to store the original camera constraints, and just return the CameraDimensions. With the WebAPI, I think you can dynamically change some constraints and restart the video or something, in which case it would be better to just store the constraints. I think we could change this later though too.

ryantrem:
Dang we have too much duplication between the different NativeCameraImpls. I don't think each platform should re-define the header. We can fix this in a later change, but if we find ourselves continuing to duplicate things across platforms in future changes, we should consider reworking this a bit.

Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
ryantrem: I would move it if it is not intended to be part of the public API. As mentioned in my other comment, we need to rework the impl headers and stuff to reduce duplication, but we could do that later.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.h
Alex-MSFT: I think we would want to have a constraints object that is stored separately (or as part of the CameraDimensions), but as it is right now we don't have any way to update the constraints during the session.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
Alex-MSFT: If you don't cast it as const then compile tries to create an invalid arcana::expected type. I'm going to change to returning the object by value, so this will go away.

Plugins/NativeCamera/Source/NativeVideo.cpp
Alex-MSFT: I think this effectively does not matter, because we won't attempt to read these properties until after Open is called, but I will separate out maxWidth and width properties and return undefined until we call open (confirmed that matches the browser implementation.)

ryantrem (approved):

PR #1123, opened 2022-08-25
[iOS native camera] - Swap height and width camera dimensions & video texture size based on initial orientation

Alex-MSFT:
Camera format resolutions are always in landscape. To the web implementation we need to swap height and width of the video format when reporting the size of the video texture when a session is started in portrait mode.

### DIFF
--- a/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
+++ b/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
@@ -114,18 +114,7 @@ namespace Babylon::Plugins
                     AVCaptureDeviceTypeBuiltInWideAngleCamera
                 ];
             }
-            
-            // In portrait mode swap height and width when selecting the best format.
-            uint32_t targetFormatHeight{maxHeight};
-            uint32_t targetFormatWidth{maxWidth};
-            
-            if (m_implData->cameraTextureDelegate->videoOrientation == AVCaptureVideoOrientationPortrait
-                ||  m_implData->cameraTextureDelegate->videoOrientation == AVCaptureVideoOrientationPortraitUpsideDown)
-            {
-                std::swap(targetFormatHeight, targetFormatWidth);
-            }
 
-            
             AVCaptureDeviceDiscoverySession* discoverySession{[AVCaptureDeviceDiscoverySession
                discoverySessionWithDeviceTypes:deviceTypes
                mediaType:AVMediaTypeVideo position:frontCamera ? AVCaptureDevicePositionFront: AVCaptureDevicePositionBack]};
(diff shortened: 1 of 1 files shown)


Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -114,6 +115,18 @@ - (id)init:(std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData>)implData;
...
             
+            // In portrait mode swap height and width when selecting the best format.
+            uint32_t targetFormatHeight{maxHeight};
+            uint32_t targetFormatWidth{maxWidth};
+            
+            if (m_implData->cameraTextureDelegate->videoOrientation == AVCaptureVideoOrientationPortrait
+                ||  m_implData->cameraTextureDelegate->videoOrientation == AVCaptureVideoOrientationPortraitUpsideDown)
+            {
+                targetFormatHeight = maxWidth;
+                targetFormatWidth = maxHeight;
```
ryantrem, line 126: Any reason not to just swap maxWidth and maxHeight rather than introduce a new set of width/height vars?

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -190,8 +203,20 @@ - (id)init:(std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData>)implData;
...
+                ||  m_implData->cameraTextureDelegate->videoOrientation == AVCaptureVideoOrientationPortraitUpsideDown)
+            {
+                cameraDimensions.height = dimensions.width;
+                cameraDimensions.width = dimensions.height;
+            }
+            else
+            {
+                cameraDimensions.width = dimensions.width;
+                cameraDimensions.height = dimensions.height;
+            }
```
ryantrem, line 219: Maybe simpler to just do:
```suggestion
            cameraDimensions.width = dimensions.width;
            cameraDimensions.height = dimensions.height;

            // For portrait orientations swap the height and width of the video format dimensions.
            Camera::Impl::CameraDimensions cameraDimensions{};
            if (m_implData->cameraTextureDelegate->videoOrientation == AVCaptureVideoOrientationPortrait
                ||  m_implData->cameraTextureDelegate->videoOrientation == AVCaptureVideoOrientationPortraitUpsideDown)
            {
                std::swap(cameraDimensions.width, cameraDimensions.height);
            }
```

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
Alex-MSFT: You can't make parameters __block variables in Objective C when marshalling to another thread, so they become immutable :/

Alex-MSFT: Reverted this change, as the web API is actually already expecting width and height to correspond to the sensor oriented dimensions rather than the device oriented dimensions.

ryantrem (approved):

PR #1124, opened 2022-08-25
Android native camera resolution and orientation

CoPrez:
This change addresses the android NativeCamera plugin implementation so that it picks a camera resolution based on the passed in maxWidth and maxHeight. Additionally this change fixes the rotation and orientation of the returned camera texture. More information about those two issues can be found in issue #1107.

### DIFF
--- a/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
+++ b/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
@@ -108,10 +108,10 @@ namespace Babylon::Plugins
             // 3: input?
 
             for (uint32_t j = 0; j < streamConfigurations.count; j += 4) {
-                int32_t format = streamConfigurations.data.i32[j + 0];
-                int32_t width = streamConfigurations.data.i32[j + 1];
-                int32_t height = streamConfigurations.data.i32[j + 2];
-                int32_t input = streamConfigurations.data.i32[j + 3];
+                int32_t format{streamConfigurations.data.i32[j + 0]};
+                int32_t width{streamConfigurations.data.i32[j + 1]};
+                int32_t height{streamConfigurations.data.i32[j + 2]};
+                int32_t input{streamConfigurations.data.i32[j + 3]};
 
                 if (input || format != AIMAGE_FORMAT_YUV_420_888 || static_cast<uint32_t>(width) > maxWidth || static_cast<uint32_t>(height) > maxHeight)
                 {
@@ -262,8 +262,8 @@ namespace Babylon::Plugins
             // The sensor rotation dictates the orientation of the camera when the phone is in it's default orientation
             // Subtracting the phone's rotation from the camera's rotation will give us the current orientation
             // of the sensor. Then add 360 and modulus 360 to ensure we're always talking about positive degrees.
-            int sensorRotationDiff = (bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360;
-            bool sensorIsPortrait = sensorRotationDiff == 90 || sensorRotationDiff == 270;
+            int sensorRotationDiff{(bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360};
+            bool sensorIsPortrait{sensorRotationDiff == 90 || sensorRotationDiff == 270};
             if (frontCamera && !sensorIsPortrait && !m_overrideCameraTexture)
             {
                 // Compensate for the front facing camera being naturally mirrored. In the portrait orientation
--- a/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
+++ b/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
@@ -270,10 +270,12 @@ namespace Babylon::Plugins
             // Subtracting the phone's rotation from the camera's rotation will give us the current orientation
             // of the sensor. Then add 360 and modulus 360 to ensure we're always talking about positive degrees.
             int sensorRotationDiff = (bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360;
-            bool flipAxis = sensorRotationDiff == 90 || sensorRotationDiff == 270;
+            bool sensorIsPortrait = sensorRotationDiff == 90 || sensorRotationDiff == 270;
 
-            m_cameraDimensions.width = !flipAxis ? bestCameraConfiguration.width : bestCameraConfiguration.height;
-            m_cameraDimensions.height = !flipAxis ? bestCameraConfiguration.height : bestCameraConfiguration.width;
+            // To match the web implementation if the sensor is rotated into a portrait orientation then the width and height
+            // of the video should be swapped
+            m_cameraDimensions.width = !sensorIsPortrait ? bestCameraConfiguration.width : bestCameraConfiguration.height;
+            m_cameraDimensions.height = !sensorIsPortrait ? bestCameraConfiguration.height : bestCameraConfiguration.width;
 
             // Check if there is an already available context for this thread
             EGLContext currentContext = eglGetCurrentContext();
--- a/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
+++ b/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
@@ -271,6 +271,13 @@ namespace Babylon::Plugins
             // of the sensor. Then add 360 and modulus 360 to ensure we're always talking about positive degrees.
             int sensorRotationDiff = (bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360;
             bool sensorIsPortrait = sensorRotationDiff == 90 || sensorRotationDiff == 270;
+            if (frontCamera && !sensorIsPortrait && !m_overrideCameraTexture)
+            {
+                // Compensate for the front facing camera being naturally mirrored. In the portrait orientation
+                // the mirrored behavior matches the browser, but in landscape it would result in the image rendering
+                // upside down. Rotate the image by 180 to compensate.
+                sensorRotationDiff = (sensorRotationDiff + 180) % 360;
+            }
 
             // To match the web implementation if the sensor is rotated into a portrait orientation then the width and height
             // of the video should be swapped
--- a/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
+++ b/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
@@ -30,32 +30,25 @@ namespace Babylon::Plugins
     )"};
 
 
-
-    static const std::string CAMERA_FRAG_SHADER_BASE{R"(#version 300 es
+    static const std::string CAMERA_FRAG_SHADER_HEADER{R"(#version 300 es
         #extension GL_OES_EGL_image_external_essl3 : require
+    )"};
+
+    static const std::string CAMERA_FRAG_SHADER{R"(
         precision mediump float;
         in vec2 cameraFrameUV;
         uniform samplerExternalOES cameraTexture;
         layout(location = 0) out vec4 oFragColor;
-        void main() {)"};
-    // The camera sensor is aligned with the phone orientation, parse the camera texture 1 to 1
-    static const std::string  CAMERA_FRAG_SHADER_0_DEGREE{R"(
-            oFragColor = texture(cameraTexture, cameraFrameUV);
-        }
-    )"};
-    // The camera sensor is 90 degrees out of alignment with the phone, swap the x and y and then flip the x axis
-    static const std::string  CAMERA_FRAG_SHADER_90_DEGREE{R"(
+        void main() {
+#if (SENSOR_ROTATION == 90)
             oFragColor = texture(cameraTexture, vec2(cameraFrameUV.y, 1.0 - cameraFrameUV.x));
-        }
-    )"};
-    // The camera sensor is upside down compared to the phone orientation, parse the camera texture flip the y coordinates
-    static const std::string  CAMERA_FRAG_SHADER_180_DEGREE{R"(
+#elif (SENSOR_ROTATION == 180)
             oFragColor = texture(cameraTexture, vec2(1.0 - cameraFrameUV.x, 1.0 - cameraFrameUV.y));
-        }
-    )"};
-    // The camera sensor is 270 degrees out of alignment with the phone, swap the x and y
-    static const std::string  CAMERA_FRAG_SHADER_270_DEGREE{R"(
+#elif (SENSOR_ROTATION == 270)
             oFragColor = texture(cameraTexture, vec2(1.0 -  cameraFrameUV.y, cameraFrameUV.x));
+#else
+            oFragColor = texture(cameraTexture, cameraFrameUV);
+#endif
         }
     )"};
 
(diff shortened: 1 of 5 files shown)


Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -29,16 +29,35 @@ namespace Babylon::Plugins
         }
     )"};
 
-    static constexpr char CAMERA_FRAG_SHADER[]{R"(#version 300 es
+
```
Alex-MSFT, line 32:
```suggestion
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -51,35 +70,92 @@ namespace Babylon::Plugins
...
+            GET_CAMERA_FUNCTION(ACameraMetadata_getConstEntry)(metadataObj, API24::ACAMERA_SCALER_AVAILABLE_STREAM_CONFIGURATIONS, &streamConfigurations);
+            // format of the data:
+            // 0: format
+            // 1: width
+            // 2: height
+            // 3: input?
+
+            for (uint32_t j = 0; j < streamConfigurations.count; j += 4) {
+                int32_t format = streamConfigurations.data.i32[j + 0];
+                int32_t width = streamConfigurations.data.i32[j + 1];
```
Alex-MSFT, line 119: On iOS this value always comes back in landscape orientation, so we had to swap width and height when the device was in portrait mode to properly apply the constraints.  Do you know if the same applies here?

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -166,10 +242,40 @@ namespace Babylon::Plugins
...
+                {
+                    throw std::runtime_error{"ConstraintError: Unable to match constraints to a supported camera configuration."};
+                }
+            }
+
+            // The sensor rotation dictates the orientation of the camera when the phone is in it's default orientation
+            // Subtracting the phone's rotation from the camera's rotation will give us the current orientation
+            // of the sensor. Then add 360 and modulus 360 to ensure we're always talking about positive degrees.
+            int sensorRotationDiff = (bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360;
+            bool sensorIsPortrait = sensorRotationDiff == 90 || sensorRotationDiff == 270;
```
Alex-MSFT, line 273:
```suggestion
            bool sensorIsPortrait{sensorRotationDiff == 90 || sensorRotationDiff == 270};
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -166,10 +242,40 @@ namespace Babylon::Plugins
...
+                if (bestCameraConfiguration.cameraID.empty())
+                {
+                    throw std::runtime_error{"ConstraintError: Unable to match constraints to a supported camera configuration."};
+                }
+            }
+
+            // The sensor rotation dictates the orientation of the camera when the phone is in it's default orientation
+            // Subtracting the phone's rotation from the camera's rotation will give us the current orientation
+            // of the sensor. Then add 360 and modulus 360 to ensure we're always talking about positive degrees.
+            int sensorRotationDiff = (bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360;
```
Alex-MSFT, line 272:
```suggestion
            int sensorRotationDiff{(bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360};
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -223,19 +329,24 @@ namespace Babylon::Plugins
 
             glBindFramebuffer(GL_FRAMEBUFFER, 0);
 
-            m_cameraShaderProgramId = android::OpenGLHelpers::CreateShaderProgram(CAMERA_VERT_SHADER, CAMERA_FRAG_SHADER);
+            const std::string fragShader =
+                    sensorRotationDiff == 90 ? CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_90_DEGREE :
+                    sensorRotationDiff == 180 ? CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_180_DEGREE :
+                    sensorRotationDiff == 270 ? CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_270_DEGREE :
+                    CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_0_DEGREE;
```
Alex-MSFT, line 336:
```suggestion
            const std::string fragShader{
                    sensorRotationDiff == 90 ? CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_90_DEGREE :
                    sensorRotationDiff == 180 ? CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_180_DEGREE :
                    sensorRotationDiff == 270 ? CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_270_DEGREE :
                    CAMERA_FRAG_SHADER_BASE + CAMERA_FRAG_SHADER_0_DEGREE};
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -51,35 +70,92 @@ namespace Babylon::Plugins
...
+        uint32_t bestWidth{0};
+        uint32_t bestHeight{0};
+        int32_t bestSensorOrientation{0};
+        bool foundExactMatch{false};
 
+        // Iterate over all of the cameras and find the one that has the best stream configuration
         for (int i = 0; i < cameraIds->numCameras; ++i)
         {
-            const char *id = cameraIds->cameraIds[i];
+            const char* id = cameraIds->cameraIds[i];
```
Alex-MSFT, line 90:
```suggestion
            const char* id{cameraIds->cameraIds[i]};
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -51,35 +70,92 @@ namespace Babylon::Plugins
...
+            // 0: format
+            // 1: width
+            // 2: height
+            // 3: input?
+
+            for (uint32_t j = 0; j < streamConfigurations.count; j += 4) {
+                int32_t format = streamConfigurations.data.i32[j + 0];
+                int32_t width = streamConfigurations.data.i32[j + 1];
+                int32_t height = streamConfigurations.data.i32[j + 2];
+                int32_t input = streamConfigurations.data.i32[j + 3];
```
Alex-MSFT, line 121:
```suggestion
                int32_t format{streamConfigurations.data.i32[j + 0]};
                int32_t width{streamConfigurations.data.i32[j + 1]};
                int32_t height{streamConfigurations.data.i32[j + 2]};
                int32_t input{streamConfigurations.data.i32[j + 3]};
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
ryantrem, line 273: Uniform initialization for these and other vars:
```suggestion
            int sensorRotationDiff{(bestCameraConfiguration.sensorRotation - phoneRotation + 360) % 360};
            bool sensorIsPortrait{sensorRotationDiff == 90 || sensorRotationDiff == 270};
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
bghgary: I have been following [this guideline](https://stackoverflow.com/questions/9976927/when-to-use-the-brace-enclosed-initializer#:~:text=If%20the%20values%20you%20are%20initializing%20with%20are,complex%20number%29%2C%20use%20curly%20braces%20initialization%20if%20available.) recently for brace initialization. I don't think we should just blindly use brace initialization.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -51,35 +70,92 @@ namespace Babylon::Plugins
...
+                    // Also ignore the configuration if either the width or height is beyond the max allowed
+                    continue;
+                }
 
-            // Found a corresponding facing camera?
-            if (facing == (frontCamera ? API24::ACAMERA_LENS_FACING_FRONT : API24::ACAMERA_LENS_FACING_BACK))
+                // Calculate pixel count and dimension differential and take the best qualifying one.
+                uint32_t pixelCount{static_cast<uint32_t>(width * height)};
+                uint32_t dimDiff{(maxWidth - width) + (maxHeight - height)};
+                if (bestCameraId == nullptr || pixelCount > bestPixelCount || (pixelCount == bestPixelCount && dimDiff < bestDimDiff))
```
ryantrem, line 133: This is the same heuristic as iOS, right? Is it worth sharing code for this between the platforms? Like given a set of constraints and a set of supported resolutions, the behavior should be the same regardless of platform right? Right now the heuristic is very simple, so there isn't that much duplication, but if we were to support more of the constraints, then we would have more and more duplication. I guess we can refactor this later if we add support for more constraints.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.h
```diff
@@ -37,7 +44,7 @@ namespace Babylon::Plugins
         CameraDimensions m_cameraDimensions{};
 
         GLuint GenerateOESTexture();
-        std::string GetCameraId(bool frontCamera);
+        CameraConfiguration GetCameraId(uint32_t maxWidth, uint32_t maxHeight, bool frontCamera);
```
ryantrem, line 47: `GetCameraConfiguration`?

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
ryantrem: 😫

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: My understanding is that the width/height from the constraints should be interpreted as if they were given in the sensors natural orientation. The API on Android always gives you back a width that correlates with the longest side of the sensor (in its natural orientation in other words).

On the web if you do a max width of 1280 and a max height of 720 you may get back a stream that is 720 wide by 1280 height. Which at first glance seems like it's breaking the height constraints, but with my above assumption it makes sense. My logic here is doing the same thing as the web.

Alex-MSFT: Double checked and you're right, will update my PR to match this behavior.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: It was definitely something I was considering as I was implementing the Android specific side. I do think in the long run it would be ideal to have some common interface to talk about the available camera streams and their capabilities in order to compare them in a common implementation... but I wasn't planning to tackle that with this PR.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: Since I was only changing the whitespace on this line anyway I'll leave it as the `=` notation

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -20,23 +20,37 @@ using namespace android::global;
 
 namespace Babylon::Plugins
 {
+    // Vertex positions for the camera texture
+    constexpr size_t CAMERA_VERTEX_COUNT{4 };
+    constexpr GLfloat CAMERA_VERTEX_POSITIONS[CAMERA_VERTEX_COUNT * 2]{-1.0f, -1.0f, 1.0f, -1.0f, -1.0f, 1.0f, 1.0f, 1.0f };
```
ryantrem, line 25:
```suggestion
    constexpr size_t CAMERA_VERTEX_COUNT{ 4 };
    constexpr GLfloat CAMERA_VERTEX_POSITIONS[CAMERA_VERTEX_COUNT * 2]{ -1.0f, -1.0f, 1.0f, -1.0f, -1.0f, 1.0f, 1.0f, 1.0f };
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -29,16 +29,35 @@ namespace Babylon::Plugins
...
+    static const std::string  CAMERA_FRAG_SHADER_0_DEGREE{R"(
             oFragColor = texture(cameraTexture, cameraFrameUV);
         }
     )"};
+    // The camera sensor is 90 degrees out of alignment with the phone, swap the x and y and then flip the x axis
+    static const std::string  CAMERA_FRAG_SHADER_90_DEGREE{R"(
+            oFragColor = texture(cameraTexture, vec2(cameraFrameUV.y, 1.0 - cameraFrameUV.x));
+        }
+    )"};
+    // The camera sensor is upside down compared to the phone orientation, parse the camera texture flip the y coordinates
```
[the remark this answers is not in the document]
CoPrez: I switched up how the shaders work and the comment is no longer in the code.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -29,16 +29,35 @@ namespace Babylon::Plugins
...
+    static const std::string  CAMERA_FRAG_SHADER_180_DEGREE{R"(
+            oFragColor = texture(cameraTexture, vec2(1.0 - cameraFrameUV.x, 1.0 - cameraFrameUV.y));
+        }
+    )"};
+    // The camera sensor is 270 degrees out of alignment with the phone, swap the x and y
```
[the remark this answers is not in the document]
CoPrez: Comment has been removed in newer iterations.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.h
CoPrez: Agreed. I've renamed it now.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -51,35 +65,92 @@ namespace Babylon::Plugins
...
+
             auto facing = static_cast<API24::acamera_metadata_enum_android_lens_facing_t>(lensInfo.data.u8[0]);
+            if (facing != (frontCamera ? API24::ACAMERA_LENS_FACING_FRONT : API24::ACAMERA_LENS_FACING_BACK))
+            {
+                // Ignore cameras facing the wrong direction
+                continue;
+            }
+
+            // Get all available stream configurations supported by the camera
+            API24::ACameraMetadata_const_entry streamConfigurations;
```
ryantrem, line 104: Should this match the others initialization?
```suggestion
            API24::ACameraMetadata_const_entry streamConfigurations = {};
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.h
```diff
@@ -28,6 +28,13 @@ namespace Babylon::Plugins
         void Close();
 
     private:
+        struct CameraConfiguration
+            {
+                std::string cameraID;
+                uint32_t width;
+                uint32_t height;
+                int32_t sensorRotation;
+            };
```
ryantrem, line 37:
```suggestion
        struct CameraConfiguration
        {
            std::string cameraID;
            uint32_t width;
            uint32_t height;
            int32_t sensorRotation;
        };
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.h
```diff
@@ -54,6 +61,7 @@ namespace Babylon::Plugins
         GLuint m_cameraRGBATextureId{};
         GLuint m_cameraShaderProgramId{};
         GLuint m_frameBufferId{};
+        const GLfloat* m_cameraUVs;
```
ryantrem, line 64:
```suggestion
        const GLfloat* m_cameraUVs{};
```

ryantrem (approved):

PR #1125, opened 2022-08-26
Fix iOS NativeCamera texture pixel format

docEdub:
The BJS side creates the texture and specifies the format (as RGBA8, in NativeEngine.createDynamicTexture), and then the native side overrides the texture handle in the Graphics::Texture, but the other properties (including format) are not updated, so the format (BGRA8) is incorrect, and operations like readPixels won't work.

This change fixes the issue by using a shader to convert the camera's YpCbCr format into RGBA, similar to how it's done in [XR.mm](https://github.com/docEdub/BabylonNative/blob/97da91efb19cde88775a83fede8e5296ae234682/Dependencies/xr/Source/ARKit/XR.mm#L554-L636).

## Related issue
https://github.com/babylonjs/babylonnative/issues/1107

### DIFF
--- a/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
+++ b/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
@@ -1,3 +1,7 @@
+#if ! __has_feature(objc_arc)
+#error "ARC is off"
+#endif
+
 #import <MetalKit/MetalKit.h>
 #include <bgfx/bgfx.h>
 #include <bgfx/platform.h>
--- a/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
+++ b/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
@@ -43,6 +43,19 @@ namespace {
         return false;
     }
 
+    typedef struct {
+        vector_float2 position;
+        vector_float2 uv;
+    } XRVertex;
+
+    static XRVertex vertices[] = {
+        // 2D positions, UV
+        { { -1, -1 },   { 0, 1 } },
+        { { -1, 1 },    { 0, 0 } },
+        { { 1, -1 },    { 1, 1 } },
+        { { 1, 1 },     { 1, 0 } },
+    };
+
     constexpr char shaderSource[] = R"(
         #include <metal_stdlib>
         #include <simd/simd.h>
@@ -346,9 +359,12 @@ namespace Babylon::Plugins
     void Camera::Impl::UpdateCameraTexture(bgfx::TextureHandle textureHandle)
     {
         arcana::make_task(m_deviceContext->BeforeRenderScheduler(), arcana::cancellation::none(), [this, textureHandle] {
-            if (m_implData->textureRGBA)
-            {
-                bgfx::overrideInternal(textureHandle, reinterpret_cast<uintptr_t>(m_implData->textureRGBA));
+            @synchronized(m_implData->cameraTextureDelegate) {
+                if (m_implData->overrideBgfxTexture && m_implData->textureRGBA != nil)
+                {
+                    bgfx::overrideInternal(textureHandle, reinterpret_cast<uintptr_t>(m_implData->textureRGBA));
+                    m_implData->overrideBgfxTexture = false;
+                }
             }
         });
     }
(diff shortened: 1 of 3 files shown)


docEdub:
Converted this to a draft because it's not up to date with Alex's latest changes. I'm not 100% happy with it yet but I'd like to get some eyes on it to make sure I'm not doing something dumb.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -327,25 +456,107 @@ - (void)captureOutput:(AVCaptureOutput *)__unused captureOutput didOutputSampleB
...
+
+    // Update both metal textures used by the renderer to display the camera image.
+    id<MTLTexture> textureY = [self getCameraTexture:pixelBuffer plane:0];
+    id<MTLTexture> textureCbCr = [self getCameraTexture:pixelBuffer plane:1];
+
+    if(textureY != nil && textureCbCr != nil)
     {
-        textureBGRA = CVMetalTextureGetTexture(texture);
-        CFRelease(texture);
+        @synchronized(self) {
```
ryantrem, line 468: Seems like all this drawing needs to happen in `UpdateCameraTexture`. The camera is constantly capturing frames, but Babylon's texture is only supposed to be updated when `UpdateCameraTexture` is called (from the call in the JS side to `VideoTexture.update()`). Otherwise, this code is updating Babylon's texture regardless of the call to `VideoTexture.update()`, and when rendering with it it would be updated unexpectedly.

It seems like `UpdateCameraTexture` should recreate the RBGATexture as needed (e.g. if the resolution changes) and do the `overrideInteral` directly, and then do the actual drawing operation. This way, we aren't even doing much work unless/until the BJS side makes a call to `VideoTexture.update()`.

ryantrem: This is also really similar to what happens on the Android side.

docEdub: Ya, this seemed like the wrong place to me, too, but when it's in `UpdateCameraTexture` the Metal command buffer errors out after a few minutes with:
```
2022-08-29 14:42:39.689979-0400 Playground[40466:1388338] Execution of the command buffer was aborted due to an error during execution. Caused GPU Timeout Error (00000002:kIOGPUCommandBufferCallbackErrorTimeout)
```
...but it works fine if it's in `didOutputSampleBuffer`.

docEdub: I've got some ideas on what the issue is. Trying a fix now.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -48,7 +137,14 @@ - (id)init:(std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData>)implData;
...
         AVCaptureSession* avCaptureSession{};
         CVMetalTextureCacheRef textureCache{};
-        id <MTLTexture> textureBGRA{};
+        id <MTLTexture> textureRGBA{};
+        id<MTLRenderPipelineState> cameraPipelineState{};
+        id<MTLDevice> metalDevice{};
+        id<MTLCommandQueue> commandQueue{};
+        id<MTLCommandBuffer> currentCommandBuffer{};
+        size_t width{0};
+        size_t height{0};
```
Alex-MSFT, line 146: This should probably be kept in a Camera::Impl::CameraDimensions struct to match the Android side

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -241,9 +365,62 @@ - (id)init:(std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData>)implData;
     void Camera::Impl::UpdateCameraTexture(bgfx::TextureHandle textureHandle)
     {
         arcana::make_task(m_deviceContext->BeforeRenderScheduler(), arcana::cancellation::none(), [this, textureHandle] {
-            if (m_implData->textureBGRA)
+            id<MTLTexture> textureY{};
+            id<MTLTexture> textureCbCr{};
+            @synchronized(m_implData->cameraTextureDelegate) {
+                textureY = [m_implData->cameraTextureDelegate getCameraTextureY];
+                textureCbCr = [m_implData->cameraTextureDelegate getCameraTextureCbCr];
```
Alex-MSFT, line 372: Do we need to explicitly increment the retain count via CVBufferRetain here or is there an implicit retain/release from taking a reference to this ID?  It looks like we explicitly release the buffer in cleanupTextures.

This blog post seems to suggest that CV Buffers need to be manually ref counted:
https://blog.krybot.com/a?ID=01650-fca75502-f8ee-4678-a905-a50e0de0b1ad

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -26,18 +30,103 @@ @interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBuffe
...
 @end
 
+namespace {
+    static bool isPixelFormatSupported(uint32_t pixelFormat)
+    {
+        switch (pixelFormat) {
+            case kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange:
+            case kCVPixelFormatType_420YpCbCr8BiPlanarFullRange:
+                return true;
+        }
```
Alex-MSFT, line 45: Should we add preferring full range over video range for the pixel format to the heuristic to select the best format?

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
docEdub: The docs at https://developer.apple.com/documentation/corevideo/1456754-cvmetaltexturecachecreatetexture?language=objc say the underlying `CVPixelBuffer` is retained by the texture, so the pixel buffer is retained for as long as we hold a strong reference to the texture. The Xcode iOS AR app template does not explicitly retain the pixel buffer, and I'm not seeing any memory leaks when profiling this change, so I'm fairly certain this is being done correctly here.

Thanks for this. While researching it I found a potential issue in [getCameraTexture](https://github.com/docEdub/BabylonNative/blob/28b4c5c9ac2e28c18f80a083c47f8d6a32da0f59/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm#L583-L587). In the Xcode iOS AR app template the texture ref is explicitly released on error before returning nil but we're not doing that here or in [XR.mm](https://github.com/docEdub/BabylonNative/blob/28b4c5c9ac2e28c18f80a083c47f8d6a32da0f59/Dependencies/xr/Source/ARKit/XR.mm#L335-L339), which is what this PR is based on. I will update NativeCameraImpl.mm in this PR to plug that potential leak, but changing XR.mm is out of scope for this PR so I'll do another PR for it later if someone else doesn't get to it first.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
docEdub: It's `private` on the Android side, so I don't know why do we need this? For consistency maybe?

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
docEdub: According to the docs here: https://developer.apple.com/documentation/avfoundation/capture_setup/choosing_a_capture_device?language=objc, `AVCaptureDeviceDiscoverySession` lists the best formats first and Apple puts video range before full range in the returned arrays. I'm not sure how Apple decides what's best but we should probably honor their decision instead of overriding it.

docEdub:
This is now working for macOS again and is ready for merge if there are no more changes needed.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -26,18 +30,103 @@ @interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBuffe
...
+                float4 ycbcr = float4(cameraSampleY.r, cameraSampleCbCr.rg, 1.0);
+                float4 cameraSample = ycbcrToRGBTransform * ycbcr;
+                cameraSample.a = 1.0;
+                return cameraSample;
+            }
+            else
+            {
+                return 0;
+            }
+        }
```
ryantrem, line 110: Maybe improve the formatting here with some line breaks:
```suggestion
        #include <metal_stdlib>
        #include <simd/simd.h>

        using namespace metal;

        #include <simd/simd.h>

        typedef struct
        {
            vector_float2 position;
            vector_float2 uv;
        } Vertex;

        typedef struct
        {
            float4 position [[position]];
            float2 uv;
        } RasterizerData;

        vertex RasterizerData
        vertexShader(uint vertexID [[vertex_id]],
                        constant Vertex *vertices [[buffer(0)]])
        {
            RasterizerData out;
            out.position = vector_float4(vertices[vertexID].position.xy, 0.0, 1.0);
            out.uv = vertices[vertexID].uv;
            return out;
        }

        fragment float4 fragmentShader(RasterizerData in [[stage_in]],
            texture2d<float, access::sample> cameraTextureY [[ texture(1) ]],
            texture2d<float, access::sample> cameraTextureCbCr [[ texture(2) ]])
        {
            constexpr sampler linearSampler(mip_filter::linear, mag_filter::linear, min_filter::linear);

            if (!is_null_texture(cameraTextureY) && !is_null_texture(cameraTextureCbCr))
            {
                const float4 cameraSampleY = cameraTextureY.sample(linearSampler, in.uv);
                const float4 cameraSampleCbCr = cameraTextureCbCr.sample(linearSampler, in.uv);

                const float4x4 ycbcrToRGBTransform = float4x4(
                    float4(+1.0000f, +1.0000f, +1.0000f, +0.0000f),
                    float4(+0.0000f, -0.3441f, +1.7720f, +0.0000f),
                    float4(+1.4020f, -0.7141f, +0.0000f, +0.0000f),
                    float4(-0.7010f, +0.5291f, -0.8860f, +1.0000f)
                );

                float4 ycbcr = float4(cameraSampleY.r, cameraSampleCbCr.rg, 1.0);
                float4 cameraSample = ycbcrToRGBTransform * ycbcr;
                cameraSample.a = 1.0;

                return cameraSample;
            }
            else
            {
                return 0;
            }
        }
```

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -335,26 +546,69 @@ - (void)captureOutput:(AVCaptureOutput *)__unused captureOutput didOutputSampleB
...
+        cameraTextureCbCr = textureCbCr;
+
+        if (implData->width != width || implData->height != height) {
+            implData->width = width;
+            implData->height = height;
+            MTLTextureDescriptor *textureDescriptor = [MTLTextureDescriptor texture2DDescriptorWithPixelFormat:MTLPixelFormatRGBA8Unorm width:width height:height mipmapped:NO];
+            textureDescriptor.usage = MTLTextureUsageRenderTarget | MTLTextureUsageShaderRead;
+            implData->textureRGBA = [implData->metalDevice newTextureWithDescriptor:textureDescriptor];
+            implData->overrideBgfxTexture = true;
+        }
```
ryantrem, line 568: Why not do this in UpdateTexture as well?

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
Alex-MSFT: Good to know it's already covered by the ordering in AVCaptureDeviceDiscoverySession

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
Alex-MSFT: I don't think this is a needed change for correctness, but there are two main reasons I think it might be preferable to keep it in a struct.

1.) The primary reason is to make it slightly easier to address the line item in #1107 to handle resizing of the camera textures when shifting orientation.  By keeping it in a CameraDimensions object we can pass it up via const pointer to monitor changes or reference without having to push it into a struct and do some casting every time.
2.) Keeping the way each implementation tracks its state as consistent as possible makes sense to me.  I think we will eventually want to get rid of the redundancy with the split NativeCameraImpl.h, and move the actual shared API into the shared NativeCamera.h with platform specific implementation details only being implemented in their specific .mm/cpp files. In this case we may want to pull CameraDimensions into the public contract since it should be the same across platforms.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -215,14 +328,32 @@ - (id)init:(std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData>)implData;
...
+            // Create a pipeline state for converting the camera output to RGBA.
+            id<MTLLibrary> lib = CompileShader(m_implData->metalDevice, shaderSource);
+            id<MTLFunction> vertexFunction = [lib newFunctionWithName:@"vertexShader"];
+            id<MTLFunction> fragmentFunction = [lib newFunctionWithName:@"fragmentShader"];
+
+            MTLRenderPipelineDescriptor *pipelineStateDescriptor = [[MTLRenderPipelineDescriptor alloc] init];
+            pipelineStateDescriptor.label = @"Native Camera YCbCr to RGBA Pipeline";
+            pipelineStateDescriptor.vertexFunction = vertexFunction;
+            pipelineStateDescriptor.fragmentFunction = fragmentFunction;
+            pipelineStateDescriptor.colorAttachments[0].pixelFormat = MTLPixelFormatBGRA8Unorm;
```
Alex-MSFT, line 348: Should this be MTLPixelFormatRGBA8Unorm?

docEdub: Yes, thank you! Interesting that it still works with the wrong format being set.

Alex-MSFT (approved):

PR #1126, opened 2022-08-26
Support external render target textures

bghgary:
This change makes it possible to use external textures with render target textures. It fixes setting the bgfx flag when constructing an external texture as well as make it such that creating a render target from an existing texture works in native engine.


SergioRZMasson (approved):
LGTM

PR #1127, opened 2022-08-29
Fix playground scene input

CoPrez:
With a previous change to improve the NativeCamera Plugin the default playground scene was modified in such a way that inputs were broken and the camera feature was enabled by default.

I've restored the previous camera instantiation and default to having the camera feature off. I then modified the setup for when the camera feature is on so it still functions as intended without interfering with when the feature is off.

I tested the changes on an Android device.

### DIFF
--- a/Apps/Playground/Scripts/experience.js
+++ b/Apps/Playground/Scripts/experience.js
@@ -15,7 +15,7 @@ var xrFeaturePoints = false;
 var meshDetection = false;
 var text = false;
 var hololens = false;
-var cameraTexture = true;
+var cameraTexture = false;
 var imageTracking = false;
 const readPixels = false;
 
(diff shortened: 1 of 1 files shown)


Apps/Playground/Scripts/experience.js
```diff
@@ -77,6 +75,9 @@ CreateBoxAsync(scene).then(function () {
     }
 
     if (cameraTexture) {
+        scene.activeCamera.position = new BABYLON.Vector3(0, 1, -10);
```
bghgary, line 78:
```suggestion
        scene.activeCamera.position.set(0, 1, -10);
```

ryantrem (approved):

bghgary (approved):

PR #1129, opened 2022-09-06
Allow negative rgb and rgba values and clamp them to range 0-255

docEdub:
When converting rgb and rgba strings to colors, browsers tolerate numbers out of the acceptable color range but BabylonNative does not.

This change updates the color string regex to accept negative numbers and clamps them to the acceptable color range of 0 to 255.

See https://forum.babylonjs.com/t/react-native-ssao-render-error/33489.
Related PR https://github.com/BabylonJS/Babylon.js/pull/12914.

### DIFF
--- a/Polyfills/Canvas/Source/Colors.h
+++ b/Polyfills/Canvas/Source/Colors.h
@@ -212,7 +212,7 @@ namespace Babylon::Polyfills::Internal
             }
             
             // matches strings of the form rgb(#,#,#) or rgba(#,#,#,#)
-            static const std::regex rgbRegex("rgba?\\(\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*(?:,\\s*(\\d{1,3}))?\\s*\\)");
+            static const std::regex rgbRegex("rgba?\\(\\s*(-?\\d{1,3})\\s*,\\s*(-?\\d{1,3})\\s*,\\s*(-?\\d{1,3})\\s*(?:,\\s*(-?\\d{1,3}))?\\s*\\)");
             std::smatch rgbMatch;
             if (std::regex_match(str, rgbMatch, rgbRegex))
             {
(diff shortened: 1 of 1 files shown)


Polyfills/Canvas/Source/Colors.h
```diff
@@ -212,19 +212,19 @@ namespace Babylon::Polyfills::Internal
...
+            static const std::regex rgbRegex("rgba?\\(\\s*(-?\\d{1,3})\\s*,\\s*(-?\\d{1,3})\\s*,\\s*(-?\\d{1,3})\\s*(?:,\\s*(-?\\d{1,3}))?\\s*\\)");
             std::smatch rgbMatch;
             if (std::regex_match(str, rgbMatch, rgbRegex))
             {
                 if (rgbMatch.size() == 5)
                 { 
                     if (rgbMatch[4].matched)
                     {
-                        return nvgRGBA(std::stoi(rgbMatch[1]), std::stoi(rgbMatch[2]), std::stoi(rgbMatch[3]), std::stoi(rgbMatch[4]));
+                        return nvgRGBA(std::clamp(std::stoi(rgbMatch[1]), 0, 255), std::clamp(std::stoi(rgbMatch[2]), 0, 255), std::clamp(std::stoi(rgbMatch[3]), 0, 255), std::clamp(std::stoi(rgbMatch[4]), 0, 255));
```
ryantrem, line 223: Kind of tempted to put the limits (0 and 255) in constants to prevent any types, but then that's also kind of the same as just using `std::numeric_limits<uint8_t>::min()` and `std::numeric_limits<uint8_t>::max()`. Meh up to you.

ryantrem (approved):

bghgary (approved):

Polyfills/Canvas/Source/Colors.h
docEdub: Ya I wondered about this, too, and decided 0 and 255 was more readable.

Issue #1131, opened 2022-09-07
Properly handle readTextureAsync cancellation in Native Capture


Alex-MSFT:
In NativeCapture, we currently do not handle the possibility that readTextureAsync can fail.  It is possible that if NativeCapture is running at the end of the engine lifetime that we will not be able to hand back results to NativeCapture, and improperly handle the return value of the task.

We should add some level of exception handling or pre-emptively end the NativeCapture session and drain all remaining readTextures requests during engine shutdown.

See the changes in #1132 for more info.

bghgary:
NativeCapture is not the long term solution, so we will not likely invest in this feature.

PR #1132, opened 2022-09-07
Fix a few issues around readTexture and iOS camera

Alex-MSFT:
This PR fixes a few issues related to the native camera implementation on iOS and readPixels
1. The current implementation for readTextureAsync assumes that the frame number will always increment by 1 when DeviceImpl::Frame is called.  This would be true if we only ever call bgfx::frame in DeviceImpl::Frame, but we have at least two other call sites for bgfx::frame, so this is not a true assumption.  In general this implementation was prone to the queue getting blocked, so I'm making it a bit more lenient.  (See #1131 for a related issue).
2. NativeCameraImpl was not creating a new ImplData object when calling reset() which meant that the unique pointer was invalid on subesquent NativeCamera open calls, and leading to an AV.
3. On Apple devices when initializing a video texture in landscape mode on we were never creating the textureRGBA object, because the resolution from the camera already matched the initial camera dimensions, so I am explicitly setting the textureRGBA value to null when calling Open and validating that we have a valid texture at the start of updateCameraTexture.
4. iOS was using dispatch_sync when creating video textures, this can cause deadlocks on iOS if already running on the main thread or if the main thread is blocked on the current thread.  Changed this to dispatch_async instead.

### DIFF
--- a/Core/Graphics/Source/DeviceImpl.cpp
+++ b/Core/Graphics/Source/DeviceImpl.cpp
@@ -176,10 +176,7 @@ namespace Babylon::Graphics
             // Drain readTextures queue, completing them in an error state.
             while (!m_readTextureRequests.empty())
             {
-                auto error = arcana::make_unexpected(
-                    std::make_exception_ptr(
-                        std::system_error(std::make_error_code(std::errc::operation_canceled))));
-
+                auto error = arcana::make_unexpected(std::make_exception_ptr(std::system_error(std::make_error_code(std::errc::operation_canceled))));
                 m_readTextureRequests.front().second.complete(error);
                 m_readTextureRequests.pop();
             }
@@ -189,7 +186,7 @@ namespace Babylon::Graphics
             FinishRenderingCurrentFrame();
 
             m_cancellationSource->cancel();
-            
+
             bgfx::shutdown();
             m_state.Bgfx.Initialized = false;
 
--- a/Core/Graphics/Source/DeviceImpl.cpp
+++ b/Core/Graphics/Source/DeviceImpl.cpp
@@ -173,12 +173,23 @@ namespace Babylon::Graphics
 
         if (m_state.Bgfx.Initialized)
         {
+            // Drain readTextures queue, completing them in an error state.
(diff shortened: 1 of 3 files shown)


Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -390,8 +398,7 @@ namespace Babylon::Graphics
         uint32_t frameNumber{bgfx::frame()};
 
         // Process read texture requests.
-        assert(m_readTextureRequests.empty() || m_readTextureRequests.front().first >= frameNumber);
-        while (!m_readTextureRequests.empty() && m_readTextureRequests.front().first == frameNumber)
+        while (!m_readTextureRequests.empty() && m_readTextureRequests.front().first <= frameNumber)
```
CoPrez, line 401: Are the readPixels pending promises in this situation still valid? Or should they just be error'd out like you do above?

CoPrez (approved):

Core/Graphics/Source/DeviceImpl.cpp
Alex-MSFT: As far as I can tell the results are still valid.  I'll set up a test to explicitly delay the sending of results to double check.

Alex-MSFT: The results are still valid up to at least 1000 frames later.  I don't think we'll ever hit a scenario where this goes longer than 1-2 frames over.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -457,7 +463,7 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
 
     void Camera::Impl::Close()
     {
-        m_implData.reset();
+        m_implData.reset(new ImplData);
```
ryantrem, line 466: Why does this have to be destroyed? And if it does have to be destroyed, can we instead do `m_implData = std::make_unique<ImplData>()`?

PR #1133, opened 2022-09-07
Console with static methods

CedricGuillemet:
Recent changes with logger assume console.log/warn/err are static methods.

https://github.com/BabylonJS/Babylon.js/blob/ddb79112d18e74de7a98826c7d4a85f857314783/packages/dev/core/src/Misc/logger.ts#L103

This PR changes console polyfill to use static methods as well.

### DIFF
--- a/Polyfills/Console/Source/Console.cpp
+++ b/Polyfills/Console/Source/Console.cpp
@@ -32,7 +32,7 @@ namespace Babylon::Polyfills::Internal
         const auto existingConsole = env.Global().Get(JS_INSTANCE_NAME);
         if (!existingConsole.IsUndefined())
         {
-            s_engineConsole = Napi::Persistent(existingConsole.As<Napi::Object>());
+            s_engineConsole = std::make_shared<Napi::ObjectReference>(Napi::Persistent(existingConsole.As<Napi::Object>()));
         }
         env.Global().Set(JS_INSTANCE_NAME, func);
         s_callback = callback;
@@ -41,6 +41,11 @@ namespace Babylon::Polyfills::Internal
     Console::Console(const Napi::CallbackInfo& info)
         : ParentT{info}
     {
+    }
+    
+    Console::~Console()
+    {
+    
     }
 
     void Console::Log(const Napi::CallbackInfo& info)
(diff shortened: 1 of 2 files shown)


Polyfills/Console/Source/Console.cpp
```diff
@@ -16,32 +16,27 @@ namespace
...
             });
 
-        auto console = func.New({});
-        Console::Unwrap(console)->m_callback = std::move(callback);
-        const auto existingConsole = env.Global().Get(JS_INSTANCE_NAME);
-        if (!existingConsole.IsUndefined())
-        {
-            Console::Unwrap(console)->m_engineConsole = std::make_unique<Napi::ObjectReference>(Napi::Persistent(existingConsole.As<Napi::Object>()));
-        }
-        env.Global().Set(JS_INSTANCE_NAME, console);
```
docEdub, line 39: Why did the `existingConsole` logic get removed?

Also note that this makes `JS_INSTANCE_NAME` in Console.h unused.

CedricGuillemet: JS_INSTANCE_NAME removed. Changed from an instance to static def.

bghgary: What about `existingConsole`? This is here because JSC automatically provides a `console` object.

Polyfills/Console/Source/Console.h
```diff
@@ -16,13 +16,13 @@ namespace Babylon::Polyfills::Internal
...
-        void InvokeEngineCallback(const std::string functionName, const Napi::CallbackInfo& info);
+        static void Log(const Napi::CallbackInfo& info);
+        static void Warn(const Napi::CallbackInfo& info);
+        static void Error(const Napi::CallbackInfo& info);
+        static void InvokeCallback(const Napi::CallbackInfo& info, Babylon::Polyfills::Console::LogLevel logLevel);
+        static void InvokeEngineCallback(const std::string functionName, const Napi::CallbackInfo& info);
 
-        std::unique_ptr<Napi::ObjectReference> m_engineConsole{};
-        Babylon::Polyfills::Console::CallbackT m_callback{};
+        static inline std::unique_ptr<Napi::ObjectReference> s_engineConsole{};
```
bghgary, line 25: I think this will cause an issue with shutdown? Since this is static, it will never be released right? I wonder if we can do something with the prototype chain instead of this.

CedricGuillemet: `SuppressDestruct` is no-op for JSI. Is it smth that can be fixed?
Other solution is to use a raw pointer allocation and never delete it.

bghgary: Prototype chain doesn't work because the functions are static🤦. But, I still managed to [implement a version](https://github.com/BabylonJS/BabylonNative/pull/1140) that doesn't use static globals. Thoughts?

CedricGuillemet:
Closing. Replacement : https://github.com/BabylonJS/BabylonNative/pull/1140

PR #1135, opened 2022-09-13
Fix iOS Camera Texture memory leak on dispose

Alex-MSFT:
The current implementation of the cameraTextureDelegate was taking a strong reference to the implData object, which was preventing the destructor from being run on camera close.  This PR adds a small cleanup protocol to cleanly handle deallocation, and removes the circular reference.

Before:
Screen Shot 2022-09-13 at 10 39 43 AM

After:
Screen Shot 2022-09-13 at 11 50 13 AM

### DIFF
--- a/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
+++ b/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
@@ -603,7 +603,7 @@ namespace Babylon::Plugins
 /**
  Updates the captured texture with the current pixel buffer.
 */
-- (CVMetalTextureRef)getCameraTexture:(CVPixelBufferRef)pixelBuffer plane:(int)planeIndex {    
+- (CVMetalTextureRef)getCameraTexture:(CVPixelBufferRef)pixelBuffer plane:(int)planeIndex {
     CVReturn ret = CVPixelBufferLockBaseAddress(pixelBuffer, kCVPixelBufferLock_ReadOnly);
     if (ret != kCVReturnSuccess) {
         return {};
--- a/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
+++ b/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
@@ -25,11 +25,11 @@
 {
     @public AVCaptureVideoOrientation videoOrientation;
 
-    std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData> implData;
+    CVMetalTextureCacheRef textureCache;
     bool orientationUpdated;
 }
 
-- (id)init:(std::shared_ptr<Babylon::Plugins::Camera::Impl::ImplData>)implData;
+- (id)init:(CVMetalTextureCacheRef)textureCache;
 - (id<MTLTexture>)getCameraTextureY;
 - (id<MTLTexture>)getCameraTextureCbCr;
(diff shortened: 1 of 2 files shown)


CoPrez (approved):

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -590,7 +603,7 @@ - (void)captureOutput:(AVCaptureOutput *)__unused captureOutput didOutputSampleB
 /**
  Updates the captured texture with the current pixel buffer.
 */
-- (CVMetalTextureRef)getCameraTexture:(CVPixelBufferRef)pixelBuffer plane:(int)planeIndex {
+- (CVMetalTextureRef)getCameraTexture:(CVPixelBufferRef)pixelBuffer plane:(int)planeIndex {
```
docEdub, line 606: Extra whitespace here

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -463,7 +468,9 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
 
     void Camera::Impl::Close()
     {
-        m_implData.reset(new ImplData);
+        [m_implData->cameraTextureDelegate reset];
+        m_implData.reset();
+        m_implData = std::make_unique<ImplData>();
```
docEdub, line 473: We should probably wait to make this until it's needed

docEdub: I'll leave it up to you. It's fine by me either way.

Alex-MSFT: I think I'll just leave it since the Open call assumes we have a valid pointer

docEdub, line 473: We should probably change `m_implData` back to `std::unique_ptr` in the header now that it's not shared anymore.

Alex-MSFT: Good call!

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -347,7 +350,9 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
             // Actually start the camera session.
             [m_implData->avCaptureSession addOutput:dataOutput];
             [m_implData->avCaptureSession commitConfiguration];
-            [m_implData->avCaptureSession startRunning];
+            dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
ryantrem, line 353: Does `dispatch_get_global_queue` have a specific thread affinity? I took a quick look at the docs and couldn't tell. Overall what is this `dispatch_async` call trying to do?

Alex-MSFT: The goal is to run this on any background thread, as running it on the main thread will hold up app responsiveness.  In general the important one to offload is stopRunning as that will lock up the app for ~3 seconds if run on the main thread, but a warning is generated when running this on the main thread as well.

PR #1137, opened 2022-09-19
iOS NativeCamera session start/stop synchronization

Alex-MSFT:
# Description
This PR fixes a few issues around NativeCamera start/stop synchronization.  

- The primary change is switching from using dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0) to using an arcana background scheduler for starting/stopping the AR camera session.  This is necessary as calls to  avCaptureSession stopRunning and avCaptureSession startRunning must be serialized or it will result in a crash at runtime.
- The second major change is changing CameraImpl to not be disposed on reset, but to instead handle managing its resources, and re-use durable resources across sessions.
- Removed unnecessary dispatch to the main thread in Open which could cause the camera to hang
- Texture creation promise will not resolve until the native camera session is actually ready

I recommend viewing this PR with [whitespace ignored](https://github.com/BabylonJS/BabylonNative/pull/1137/files?diff=split&w=1), as the removal of the synchronization caused a lot of spacing changes.

# Validation
Ran in the Babylon Native Playground, as well as in our test application that consumes this code through Babylon React Native and heavily exercises this code path.

### DIFF
--- a/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
+++ b/Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
@@ -339,36 +339,36 @@ namespace Babylon::Plugins
         }
 
         // Kick off camera session on a background thread.
-        return arcana::make_task(m_cameraSessionDispatcher, arcana::cancellation::none(), [this, input, devicePixelFormat, cameraDimensions]() mutable {
-            if (m_implData->avCaptureSession == nil) {
-                m_implData->avCaptureSession = [[AVCaptureSession alloc] init];
+        return arcana::make_task(m_cameraSessionDispatcher, arcana::cancellation::none(), [implObj = shared_from_this(), input, devicePixelFormat, cameraDimensions]() mutable {
+            if (implObj->m_implData->avCaptureSession == nil) {
+                implObj->m_implData->avCaptureSession = [[AVCaptureSession alloc] init];
             } else {
-                for (AVCaptureInput* input in [m_implData->avCaptureSession inputs]) {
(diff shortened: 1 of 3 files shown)


Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -185,198 +173,204 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
+        }
+        
+        // Check for failed initialisation.
+        if (!input)
+        {
+            return arcana::task_from_error<CameraDimensions>(std::make_exception_ptr(std::runtime_error{"Error Getting Camera Input"}));
+        }
+
+        // Kick off camera session on a background thread.
+        return arcana::make_task(m_cameraSessionDispatcher, arcana::cancellation::none(), [this, input, devicePixelFormat, cameraDimensions]() mutable {
```
ryantrem, line 342: Is it possible for this class to be destroyed while an async operation is still in progress? Instead of capturing `this`, do we need to capture a shared_ptr (since this class already inherits from `enable_shared_from_this`)?

ryantrem (approved):

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
Alex-MSFT: I don't think it's possible in the current integration in BRN, but it could be possible for other integrations, so this is a good suggestion for safety, will look into it.

CoPrez (approved):

PR #1140, opened 2022-09-26
Better console polyfill implementation

bghgary:
Better console polyfill implementation that uses static functions. Shim into the existing `log`, `warn`, and `error` functions if a `console` object is already present.

Testing
- [x] Win32
- [x] macOS
- [x] Against latest Babylon.js which [depends on console functions being static](https://github.com/BabylonJS/Babylon.js/blob/ddb79112d18e74de7a98826c7d4a85f857314783/packages/dev/core/src/Misc/logger.ts#L103)
- [x] napi-jsi


Polyfills/Console/Source/Console.cpp
```diff
@@ -1,100 +1,106 @@
...
     {
-        Internal::Console::CreateInstance(env, std::move(callback));
+        Napi::HandleScope scope{env};
+
+        auto console = env.Global().Get(JS_INSTANCE_NAME).As<Napi::Object>();
+        if (console.IsUndefined())
+        {
+            console = Napi::Object::New(env);
+
+            console.Set("log", Napi::Function::New(env, [callback](const Napi::CallbackInfo& info)
```
CedricGuillemet, line 65: I would refactor these calls to console.set inside a new function to avoid copy/paste and improve readability.

Polyfills/Console/Source/Console.cpp
```diff
@@ -1,100 +1,106 @@
...
+            console.Set("error", Napi::Function::New(env, [callback](const Napi::CallbackInfo& info)
+            {
+                InvokeCallback(callback, info, LogLevel::Error);
+            }, "error"));
+
+            env.Global().Set(JS_INSTANCE_NAME, console);
+        }
+        else
+        {
+            auto existingLog = console.Get("log").As<Napi::Function>();
```
CedricGuillemet, line 84: Same here, a new function and 3 calls instead of copy/pasted code.

CedricGuillemet (approved):

PR #1141, opened 2022-09-29
WIP: Fixes for shutting down during async operations

bghgary:


Core/JsRuntime/Include/Babylon/JsRuntimeScheduler.h
```diff
@@ -1,30 +1,146 @@
...
+    // for a N-API object to use with arcana tasks.  It is different than the typical scheduler as this class itself
+    // is not a scheduler directly, but instead hands out scheduler via its `Get()` function.  It provides special
+    // handling for when the JsRuntime begins shutting down, i.e., when JsRuntime::NotifyDisposing is called.
+    //   1. Calling `Rundown` blocks execution until all outstanding schedulers are invoked on the JavaScript thread.
+    //   2. Once the JsRuntime begins shutting down, all schedulers will reroute its dispatch calls from the
+    //      JsRuntime to a separate dispatcher owned by the JsRuntimeScheduler itself.  This class will then be able
+    //      to pump this dispatcher in its destructor to prevent deadlocks.
+    //
+    // The typical pattern for an arcana task will look something like this:
+    //   class MyClass
```
bghgary, line 21: `: ObjectWrap`

Core/AppRuntime/Source/WorkQueue.cpp
```diff
@@ -53,5 +54,10 @@ namespace Babylon
 
         // There should no longer be any outstanding work once the queue is drained.
         assert(m_dispatcher.empty());
+
+        // Clear the shutdown queue to make sure the callables are destroyed on this thread.
+        m_shutdownQueue.clear();
```
bghgary, line 59: check destruction element order

Core/AppRuntime/Source/AppRuntime_Chakra.cpp
```diff
@@ -22,31 +22,23 @@ namespace Babylon
...
-            };
-
         JsRuntimeHandle jsRuntime;
         ThrowIfFailed(JsCreateRuntime(JsRuntimeAttributeNone, nullptr, &jsRuntime));
         JsContextRef context;
         ThrowIfFailed(JsCreateContext(jsRuntime, &context));
         ThrowIfFailed(JsSetCurrentContext(context));
         ThrowIfFailed(JsSetPromiseContinuationCallback(
             [](JsValueRef task, void* callbackState) {
+                auto* pThis = reinterpret_cast<AppRuntime*>(callbackState);
```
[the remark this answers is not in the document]
bghgary: JsSetPromiseContinuationCallback takes a static C function. Only no capture lambdas are convertible to static C function.

Core/JsRuntime/InternalInclude/Babylon/JsRuntimeScheduler.h
```diff
@@ -1,30 +1,146 @@
...
+    //           arcana::make_task(arcana::threadpool_scheduler, m_cancellationSource, []() {
+    //               // do some asynchronous work
+    //           }).then(m_runtimeScheduler.Get(), m_cancellationSource, [thisRef = Napi::Persistent(info.This()), callback = Napi::Persistent(callback)]() {
+    //               callback.Call({});
+    //           });
+    //       }
+    //
+    //   private:
+    //       arcana::cancellation_source m_cancellationSource;
+    //       JsRuntimeScheduler m_runtimeScheduler;
```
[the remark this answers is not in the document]
bghgary: Yes, that's good idea.

Plugins/NativeCamera/Source/MediaDevices.cpp
```diff
@@ -38,8 +38,7 @@ namespace Babylon::Plugins::Internal
                 }
             }
 
-            auto runtimeScheduler{std::make_unique<JsRuntimeScheduler>(JsRuntime::GetFromJavaScript(env))};
-            MediaStream::NewAsync(env, videoConstraints).then(*runtimeScheduler, arcana::cancellation::none(), [runtimeScheduler = std::move(runtimeScheduler), env, deferred](const arcana::expected<Napi::Object, std::exception_ptr>& result) {
+            MediaStream::NewAsync(env, videoConstraints).then(arcana::inline_scheduler, arcana::cancellation::none(), [env, deferred](const arcana::expected<Napi::Object, std::exception_ptr>& result) {
```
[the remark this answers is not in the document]
bghgary: I am enforcing that `NewAsync` must return on the JS thread, which it already does.

Core/JsRuntime/Include/Babylon/JsRuntime.h
```diff
@@ -22,30 +23,48 @@ namespace Babylon
...
         static JsRuntime& GetFromJavaScript(Napi::Env);
-        void Dispatch(std::function<void(Napi::Env)>);
 
-    protected:
-        JsRuntime(const JsRuntime&) = delete;
-        JsRuntime& operator=(const JsRuntime&) = delete;
+        struct IDisposingCallback
+        {
+            virtual void Disposing() = 0;
+        };
```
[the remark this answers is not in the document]
bghgary: I could, but then it would be more complex. We either have to return an RAII registration token that auto unregisters or force consumers to store the std::function in order to unregister. Either way, the consumer will then need to keep a member around. It's not a big deal, but this is simpler.

Core/JsRuntime/InternalInclude/Babylon/JsRuntimeScheduler.h
```diff
@@ -1,30 +1,146 @@
...
+
+        // Wait until all of the schedulers are invoked.
+        void Rundown()
         {
-            m_runtime.Dispatch([callable{std::forward<CallableT>(callable)}](Napi::Env) {
-                callable();
-            });
+            while (m_count > 0)
+            {
+                m_disposingDispatcher.blocking_tick(arcana::cancellation::none());
```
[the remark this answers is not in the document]
bghgary: I don't think it matters. The disposing dispatcher queue should never be empty in this loop, but I guess it is cheaper to use `tick`.

Core/JsRuntime/Source/JsRuntime.cpp
```diff
@@ -40,6 +41,32 @@ namespace Babylon
...
+        auto callbacks = std::move(runtime.m_disposingCallbacks);
+        for (auto* callback : callbacks)
+        {
+            callback->Disposing();
+        }
+    }
+
+    void JsRuntime::RegisterDisposing(JsRuntime& runtime, IDisposingCallback* callback)
+    {
+        auto& callbacks = runtime.m_disposingCallbacks;
```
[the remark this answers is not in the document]
bghgary: It shouldn't happen assuming correct usage patterns with cancellation. But if it does happen, then callbacks registered after `NotifyDisposing` will not fire. All of these `*Disposing` functions must be called on the JavaScript thread.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -70,7 +53,7 @@ namespace Babylon::Polyfills::Internal
 
         if (time <= earliestTime)
         {
-            m_runtime.Dispatch([this](Napi::Env) {
+            m_runtimeScheduler.Get()([this]() {
```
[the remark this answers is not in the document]
bghgary: I need to be able to call `Rundown` on the pending ones similar to other async operations to guard against class member access after destruction.

Core/JsRuntime/InternalInclude/Babylon/JsRuntimeScheduler.h
```diff
@@ -1,30 +1,146 @@
...
+            --m_count;
+        }
+
+        mutable std::mutex m_mutex{};
+        JsRuntime* m_runtime{};
+        std::function<void()> m_disposingCallback{};
+
+        SchedulerImpl m_scheduler;
+        std::atomic<int> m_count{0};
+        arcana::manual_dispatcher<128> m_disposingDispatcher{};
```
ryantrem, line 144: If each `JsRuntimeScheduler` instance has its own dispatch queue, and it is only pumped by a call to `Rundown` in the destructor, then if we ever have a situation where one object is trying to call into another object (as part of the async code executing during destruction), then the separate dispatch queue in the separate object will be in a state where it is getting pumped and we will basically deadlock. It seems like this is fixable by having a shared dispatch queue across all objects, and a given object's destructor's call to Rundown is pumping for all queued work. I don't think this breaks anything, but I think it would prevent this potential deadlock situation. I'm not sure where a shared dispatch queue would exist... maybe we'd need to push some of this logic up to JSRuntime.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -14,46 +14,29 @@ namespace Babylon::Polyfills::Internal
...
         }
 
-        m_shutdown = true;
+        m_cancellationSource.cancel();
         m_condVariable.notify_one();
         m_thread.join();
+
+        // Wait for async operations to complete.
+        m_runtimeScheduler.Rundown();
     }
```
bghgary, line 37: Is this actually necessary?

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -58,13 +58,21 @@ namespace Babylon::Plugins
 
     MediaStream::~MediaStream()
     {
+        m_cancellationSource.cancel();
+
+        // HACK: This is a hack to make sure the camera device is destroyed on the JS thread.
+        // The napi-jsi adapter currently calls the destructors of JS objects possibly on the wrong thread.
+        // Once this is fixed, this hack will no longer be needed.
```
PolygonalSun, line 65: Do we have/need an issue created to address this hack in the future?

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -58,13 +58,21 @@ namespace Babylon::Plugins
...
         if (m_cameraDevice != nullptr)
         {
             // The cameraDevice should be destroyed on the JS thread as it may need to access main thread resources
             // move ownership of the cameraDevice to a lambda and dispatch it with the runtimeScheduler so the destructor
```
PolygonalSun, line 69: Nit: "and move..."

Core/JsRuntime/InternalInclude/Babylon/JsRuntimeScheduler.h
```diff
@@ -1,30 +1,146 @@
...
+    //       JsRuntimeScheduler m_runtimeScheduler;
+    //   };
+    //
+    // **IMPORTANT**:
+    //   1. To prevent continuations from accessing freed memory, the destructor of the N-API class is expected to call
+    //      `Rundown()` which blocks execution until all of its schedulers are invoked. Failing to do so will result in
+    //      an assert if there are outstanding schedulers not yet invoked.
+    //   2. The last continuation that accesses members of the N-API object, including the cancellation associated with
+    //      the continuation, must capture a persistent reference to the N-API object itself to prevent the GC from
+    //      collecting the N-API object during the asynchronous operation. Failing to do so will result in a hang
```
bghgary, line 54: Add comment about lifetime of members during GC.

Core/AppRuntime/Source/AppRuntime.cpp
```diff
@@ -1,65 +1,36 @@
...
     AppRuntime::AppRuntime()
-        : AppRuntime{DefaultUnhandledExceptionHandler}
+        : m_impl{std::make_unique<AppRuntimeImpl>()}
     {
     }
 
     AppRuntime::AppRuntime(std::function<void(const std::exception&)> unhandledExceptionHandler)
-        : m_workQueue{std::make_unique<WorkQueue>([this] { RunPlatformTier(); })}
-        , m_unhandledExceptionHandler{unhandledExceptionHandler}
+        : m_impl{std::make_unique<AppRuntimeImpl>(unhandledExceptionHandler)}
```
ryantrem, line 12: This this be:
`: m_impl{std::make_unique<AppRuntimeImpl>(std::move(unhandledExceptionHandler))}

Core/JsRuntime/InternalInclude/Babylon/JsRuntimeScheduler.h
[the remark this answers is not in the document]
bghgary: Nope, I'm wrong. The disposing dispatcher may be empty. There can be outstanding asynchronous work happening on a worker thread and hasn't tried to schedule a continuation yet. In these cases, we want to block the JS thread until the work is complete.

Plugins/NativeCamera/Source/MediaStream.cpp
bghgary: I didn't find one. I will create one.

PR #1144, opened 2022-10-04
XMLHttpRequest headers

CedricGuillemet:
Fixing Nightly with these changes:

- Nightly bjs build download done with a node script instead of CI script (can be run locally)
- Fix syntax for objects created on the stack to be consistent with the rest of the code base
- add `getResponseHeader` and `getAllResponseHeaders` to XMLHttpRequest as it's now used in bjs

Note: impl added for Win32 and missing for Linux, Apple and Android.

### DIFF
--- a/Apps/scripts/getNightly.js
+++ b/Apps/scripts/getNightly.js
@@ -1,0 +1,21 @@
+var https = require('https');
+var fs = require('fs');
+
+function download(filename, url) {
+  var file = fs.createWriteStream(filename);
+  var request = https.get(url, function(response) {
+    response.pipe(file);
+  });
+}
+
+console.log('Downloading babylon.js nightly');
+download('node_modules/babylonjs/babylon.max.js', 'https://preview.babylonjs.com/babylon.max.js');
+download('node_modules/babylonjs/babylon.max.js.map', 'https://preview.babylonjs.com/babylon.max.js.map');
+download('node_modules/babylonjs-materials/babylonjs.materials.js', 'https://preview.babylonjs.com/materialsLibrary/babylonjs.materials.js');
+download('node_modules/babylonjs-materials/babylonjs.materials.js.map', 'https://preview.babylonjs.com/materialsLibrary/babylonjs.materials.js.map');
+download('node_modules/babylonjs-loaders/babylonjs.loaders.js', 'https://preview.babylonjs.com/loaders/babylonjs.loaders.js');
+download('node_modules/babylonjs-loaders/babylonjs.loaders.js.map', 'https://preview.babylonjs.com/loaders/babylonjs.loaders.js.map');
+download('node_modules/babylonjs-gui/babylon.gui.js', 'https://preview.babylonjs.com/gui/babylon.gui.js');
+download('node_modules/babylonjs-gui/babylon.gui.js.map', 'https://preview.babylonjs.com/gui/babylon.gui.js.map');
+download('node_modules/chai/chai.js', 'https://unpkg.com/chai/chai.js');
+download('node_modules/mocha/mocha.js', 'https://unpkg.com/mocha/mocha.js');
--- a/Apps/scripts/getNightly.js
+++ b/Apps/scripts/getNightly.js
@@ -17,5 +17,3 @@ download('node_modules/babylonjs-loaders/babylonjs.loaders.js', 'https://preview
 download('node_modules/babylonjs-loaders/babylonjs.loaders.js.map', 'https://preview.babylonjs.com/loaders/babylonjs.loaders.js.map');
 download('node_modules/babylonjs-gui/babylon.gui.js', 'https://preview.babylonjs.com/gui/babylon.gui.js');
 download('node_modules/babylonjs-gui/babylon.gui.js.map', 'https://preview.babylonjs.com/gui/babylon.gui.js.map');
-//download('node_modules/chai/chai.js', 'https://unpkg.com/chai/chai.js');
-//download('node_modules/mocha/mocha.js', 'https://unpkg.com/mocha/mocha.js');
--- a/Apps/scripts/getNightly.js
+++ b/Apps/scripts/getNightly.js
@@ -17,5 +17,5 @@ download('node_modules/babylonjs-loaders/babylonjs.loaders.js', 'https://preview
 download('node_modules/babylonjs-loaders/babylonjs.loaders.js.map', 'https://preview.babylonjs.com/loaders/babylonjs.loaders.js.map');
 download('node_modules/babylonjs-gui/babylon.gui.js', 'https://preview.babylonjs.com/gui/babylon.gui.js');
 download('node_modules/babylonjs-gui/babylon.gui.js.map', 'https://preview.babylonjs.com/gui/babylon.gui.js.map');
-download('node_modules/chai/chai.js', 'https://unpkg.com/chai/chai.js');
-download('node_modules/mocha/mocha.js', 'https://unpkg.com/mocha/mocha.js');
+//download('node_modules/chai/chai.js', 'https://unpkg.com/chai/chai.js');
+//download('node_modules/mocha/mocha.js', 'https://unpkg.com/mocha/mocha.js');
--- a/Dependencies/UrlLib/Include/UrlLib/UrlLib.h
+++ b/Dependencies/UrlLib/Include/UrlLib/UrlLib.h
@@ -39,7 +39,7 @@ namespace UrlLib
 
         void Abort();
 
-        void Open(UrlMethod method, std::string url);
+        void Open(UrlMethod method, const std::string& url);
 
         UrlResponseType ResponseType() const;
 
@@ -53,9 +53,7 @@ namespace UrlLib
 
         std::string_view ResponseString() const;
 
-        std::string ResponseHeader(const std::string& headerName) const;
-
-        std::string ResponseHeaders() const;
+        std::optional<std::string> GetResponseHeader(const std::string& headerName) const;
 
         gsl::span<const std::byte> ResponseBuffer() const;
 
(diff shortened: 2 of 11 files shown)


Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -161,10 +174,10 @@ namespace Babylon::Polyfills::Internal
 
     void XMLHttpRequest::AddEventListener(const Napi::CallbackInfo& info)
     {
-        std::string eventType = info[0].As<Napi::String>().Utf8Value();
-        Napi::Function eventHandler = info[1].As<Napi::Function>();
+        const std::string eventType{info[0].As<Napi::String>().Utf8Value()};
```
bghgary, line 177: I've been following this guideline for brace initialization. We don't just want to blanket replace all with brace initialization. https://stackoverflow.com/a/9977536. So, in this case, it should not be a brace initialization and also pretty much everywhere changed in this file.

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -149,6 +151,17 @@ namespace Babylon::Polyfills::Internal
         m_request.ResponseType(ResponseType::StringToEnum(value.As<Napi::String>().Utf8Value()));
     }
 
+    Napi::Value XMLHttpRequest::GetResponseHeader(const Napi::CallbackInfo& info)
+    {
+        const auto headerName{info[0].As<Napi::String>().Utf8Value()};
+        return Napi::Value::From(Env(), m_request.ResponseHeader(headerName));
```
bghgary, line 157: This implementation is incorrect.

[Return value](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getResponseHeader#return_value)

> A string representing the header's text value, or null if either the response has not yet been received or the header doesn't exist in the response.

We have to return null in some cases.

Dependencies/UrlLib/Include/UrlLib/UrlLib.h
```diff
@@ -53,6 +53,10 @@ namespace UrlLib
 
         std::string_view ResponseString() const;
 
+        std::string ResponseHeader(const std::string& headerName) const;
```
bghgary, line 56: This contract is a bit off.

```suggestion
        std::optional<std::string> GetResponseHeader(const std::string& headerName) const;
```

If the response headers are not available yet or if the `headerName` doesn't exist, we need to be able to distinguish from a header that returned an empty string.

This is a function call and not a property access function, so it should have a `Get` prefix.

Also, in the `Open` method, I used `std::string` as input which is inconsistent with this. I'm always annoyed that C++ doesn't have a standard way of passing in strings. I keep flipflopping between what is best here. It seems like `const std::string&` reduces copying, so let's stick with this. Can you update `Open` to use `const std::string&`?

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
```diff
@@ -149,6 +151,17 @@ namespace Babylon::Polyfills::Internal
...
+    }
+
+    Napi::Value XMLHttpRequest::GetAllResponseHeaders(const Napi::CallbackInfo&)
+    {
+        return Napi::Value::From(Env(), m_request.ResponseHeaders());
```
bghgary, line 162: Same for this:

[Return value](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getAllResponseHeaders#return_value)

> A string representing all of the response's headers (except those whose field name is Set-Cookie) separated by [CRLF](https://developer.mozilla.org/en-US/docs/Glossary/CRLF), or null if no response has been received. If a network error happened, an empty string is returned.

Dependencies/UrlLib/Include/UrlLib/UrlLib.h
```diff
@@ -53,6 +53,10 @@ namespace UrlLib
...
+
+        std::string ResponseHeaders() const;
```
bghgary, line 58: Similar for this:

```suggestion
        std::optional<std::string> ResponseHeaders() const;
```

Dependencies/UrlLib/Source/Windows/UrlRequest.cpp
```diff
@@ -164,6 +170,29 @@ namespace UrlLib
...
+        {
+            std::string str{};
+            for (auto& iter : m_headers)
+            {
+                str += iter.first;
+                str += ":";
+                str += iter.second;
+                str += "\n";
+            }
+            return str;
```
bghgary, line 193: Should use `std::stringstream` to be more efficient.

Also, I noted below the contract for this API must be able to return `null` and filter out `Set-Cookie`.

bghgary: I think maybe we should hold off on implementing this function since Babylon.js doesn't use it. If we want to add it, we should at least add tests.

Apps/scripts/getNightly.js
```diff
@@ -0,0 +1,21 @@
...
+download('node_modules/babylonjs/babylon.max.js', 'https://preview.babylonjs.com/babylon.max.js');
+download('node_modules/babylonjs/babylon.max.js.map', 'https://preview.babylonjs.com/babylon.max.js.map');
+download('node_modules/babylonjs-materials/babylonjs.materials.js', 'https://preview.babylonjs.com/materialsLibrary/babylonjs.materials.js');
+download('node_modules/babylonjs-materials/babylonjs.materials.js.map', 'https://preview.babylonjs.com/materialsLibrary/babylonjs.materials.js.map');
+download('node_modules/babylonjs-loaders/babylonjs.loaders.js', 'https://preview.babylonjs.com/loaders/babylonjs.loaders.js');
+download('node_modules/babylonjs-loaders/babylonjs.loaders.js.map', 'https://preview.babylonjs.com/loaders/babylonjs.loaders.js.map');
+download('node_modules/babylonjs-gui/babylon.gui.js', 'https://preview.babylonjs.com/gui/babylon.gui.js');
+download('node_modules/babylonjs-gui/babylon.gui.js.map', 'https://preview.babylonjs.com/gui/babylon.gui.js.map');
+download('node_modules/chai/chai.js', 'https://unpkg.com/chai/chai.js');
+download('node_modules/mocha/mocha.js', 'https://unpkg.com/mocha/mocha.js');
```
bghgary, line 21: I know this is just moving code, but why are we doing this? Doesn't `npm install` already bring down the right files?

Dependencies/UrlLib/Source/Windows/UrlRequest.cpp
```diff
@@ -114,6 +115,11 @@ namespace UrlLib
 
                             m_responseUrl = winrt::to_string(responseMessage.RequestMessage().RequestUri().RawUri());
 
+                            for (auto&& iter : responseMessage.Headers())
+                            {
+                                m_headers.insert(std::make_pair(winrt::to_string(iter.Key()), winrt::to_string(iter.Value())));
+                            }
```
bghgary, line 121: Would it be better to store the `Web::Http::Headers::HttpResponseHeaderCollection` directly to avoid copying the strings if no one ever needs them?

bghgary (changes requested):

Dependencies/UrlLib/Source/Windows/UrlRequest.cpp
CedricGuillemet: I'm removing the method.

Dependencies/UrlLib/Include/UrlLib/UrlLib.h
CedricGuillemet: Fixed

Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp
CedricGuillemet: Fixed

Apps/scripts/getNightly.js
CedricGuillemet: version on `unpkg.com` are more recent than npm. It looks like Darragh did it because of an issue with one of these NPM packages : https://github.com/BabylonJS/BabylonNative/pull/897

Dependencies/UrlLib/Source/Windows/UrlRequest.cpp
CedricGuillemet: The header collection becomes invalid outside of the task. What's the good practice to keep a reference on it? If it needs a copy, then doing it with a std::map is more consistent with the rest of the class IMO.

Apps/scripts/getNightly.js
CedricGuillemet: CI is OK when download is removed. I guess it was a bug with a specific NPM package that's been fixed since.

bghgary: Interesting. We don't do this locally. Was this a CI only issue?

Dependencies/UrlLib/Source/Windows/UrlRequest.cpp
bghgary: > The header collection becomes invalid outside of the task

It's not ref counted like normal WinRT object? This isn't that big a deal either way.

Apps/scripts/getNightly.js
CedricGuillemet: Darragh must have been the only one testing unittests locally. Or the mocha/chai issue has been fixed not long after this has been put in the CI.

bghgary: I have run them a few times since this was added. Never seen any issues. Anyways, sounds like we can remove this now.

CedricGuillemet: It's been removed.

bghgary (approved):

Issue #893, opened 2021-09-23
Implement clearTimeout in Window polyfill


darraghjburke:
We should implement the clearTimeout function. This would enable the use of timeouts in Mocha.js tests. This should probably be done once #843 is finished, since that work will change the implementation of setTimeout.

bghgary:
Related: https://github.com/BabylonJS/BabylonNative/pull/844

docEdub:
Implemented in https://github.com/BabylonJS/BabylonNative/pull/1146.

Issue #253, opened 2020-05-07
Improve setTimeout implementation


bghgary:
Currently setTimeout polls until the time elapses.

One suggestion from @syntheticmagus:
> You could have a single thread managing a priority queue of them.
> It wakes up either when it's time to send one off or when it's time to insert a new one.
> That might make sense as a feature of the work queue: basically a side pot of extra work that will insert itself into the work queue at a given time in the future.

bghgary:
Related: #843

PR #1146, opened 2022-10-05
Make `setTimeout` asynchronous and add `clearTimeout`

docEdub:
## Overview
Improves `setTimeout` performance by making it asynchronous, and implements `clearTimeout`.

## Details
The current implementation of `setTimeout` dispatches repeatedly until the given delay is reached. This causes performance issues.

This PR fixes the issue by building on the initial investigative work done in PR https://github.com/BabylonJS/BabylonNative/issues/843 to implement `setTimeout` asynchronously on a separate thread.

The `clearTimeout` function is currently not implemented and since it is closely related to `setTimeout`, it makes sense to implement it in this change, too.

## Related tickets
https://github.com/BabylonJS/BabylonNative/issues/253
https://github.com/BabylonJS/BabylonNative/issues/893


Polyfills/Window/Source/Window.cpp
```diff
@@ -8,20 +8,28 @@
 
 namespace
 {
-    struct TimeoutFunction
+    using TimeoutId = Babylon::Polyfills::Internal::TimeoutId;
+    using Milliseconds = std::chrono::milliseconds::rep;
+    using TimePoint = std::chrono::time_point<std::chrono::steady_clock, std::chrono::milliseconds>;
```
bghgary, line 13: I'm not sure I understand what is going on with this. It's a bit confusing to use `Milliseconds` as a `rep` which somewhat defeats the purpose of using `std::chrono::duration`.

Polyfills/Window/Source/Window.cpp
```diff
@@ -8,20 +8,28 @@
...
+        return std::chrono::time_point_cast<std::chrono::milliseconds, std::chrono::steady_clock>(std::chrono::steady_clock::now());
+    }
 
-        TimeoutFunction(std::shared_ptr<Napi::FunctionReference> func, std::chrono::time_point<std::chrono::system_clock> time)
-            : Function{std::move(func)}
-            , TimePoint{time}
+    struct Timeout
+    {
+        TimeoutId id;
+        std::shared_ptr<Napi::FunctionReference> function;
```
bghgary, line 23: Does this really need to be shared_ptr?

Polyfills/Window/Source/Window.cpp
```diff
@@ -32,46 +40,85 @@ namespace
...
-        void Dispatch(std::shared_ptr<Napi::FunctionReference> func, std::chrono::system_clock::time_point time) 
-        {   
-            if (time <= std::chrono::system_clock::now())
+        TimeoutId Dispatch(std::shared_ptr<Napi::FunctionReference> func, Milliseconds time)
+        {
+            if (time <= 0)
             {
                 CallFunction(std::move(func));
-                return;
+                return 0;
```
bghgary, line 48: Is this the correct behavior? `setTimeout(func, 0)` is supposed to call the `func` on the next frame. See the `Late timeouts` section from https://developer.mozilla.org/en-US/docs/Web/API/setTimeout.

Polyfills/Window/Source/Window.cpp
```diff
@@ -32,46 +40,85 @@ namespace
...
 
+            const auto now = Now();
             std::unique_lock<std::mutex> lk(m_mutex);
             const auto soonestTimePoint =
-                m_queue.empty()
-                    ? time + std::chrono::milliseconds{1}
-                    : m_queue.top().TimePoint;
+                m_timeMap.empty()
+                ? now + std::chrono::milliseconds{ 1 }
+            : (*m_timeMap.cbegin()).second.front().timePoint;
```
bghgary, line 56: Formatting issues

bghgary, line 56: Generally, all over the file, you can use `iterator->` syntax instead of `(*iterator).`

Polyfills/Window/Source/Window.cpp
bghgary: Also, setTimeout must return a positive number according to the spec.

Polyfills/Window/Source/Window.cpp
bghgary, line 56: I don't understand this variable. What does it mean to be the soonest time point?

Polyfills/Window/Source/Window.cpp
```diff
@@ -103,11 +148,31 @@ namespace
...
+        }
+
         Babylon::JsRuntime& m_runtime;
         std::thread m_thread;
         std::mutex m_mutex{};
         std::condition_variable m_condVariable{};
-        std::priority_queue<TimeoutFunction, std::vector<TimeoutFunction>, decltype(&TimeoutFunction::Compare)> m_queue{TimeoutFunction::Compare};
+        TimeoutId m_lastTimeoutId = 0;
+        std::map<TimeoutId, Timeout> m_idMap;
+        std::map<Milliseconds, std::list<Timeout>> m_timeMap;
```
bghgary, line 175: These maps are weird to me. There should only be one owner of the `Timeout` struct. I think it would be better if the id map is the owner of the `Timeout` struct and the second container is a sorted list of pointers to the Timeout struct in the id map. I suppose vice versa should also work (i.e., sorted list is the owner and id map points to the entry in the sorted list).

Polyfills/Window/Source/Window.h
```diff
@@ -22,12 +24,11 @@ namespace Babylon::Polyfills::Internal
         JsRuntime& m_runtime;
         std::unique_ptr<TimeoutDispatcher> m_timeoutDispatcher;
```
bghgary, line 25: This probably could be an `std::optional` instead.

Polyfills/Window/Source/Window.cpp
```diff
@@ -88,10 +135,8 @@ namespace
...
             {
                 std::unique_lock<std::mutex> lk(m_mutex);
-                while (!m_queue.empty())
-                {
-                    m_queue.pop();
-                }
+                m_idMap.clear();
+                m_timeMap.clear();
                 m_shutdown = true;
                 m_condVariable.notify_one();
```
bghgary, line 141: Probably need to call `m_thread.join()` to make sure the thread is done using member variables before the destructor completes.

Polyfills/Window/Source/Window.cpp
```diff
@@ -88,10 +135,8 @@ namespace
         {
             {
                 std::unique_lock<std::mutex> lk(m_mutex);
-                while (!m_queue.empty())
-                {
-                    m_queue.pop();
-                }
+                m_idMap.clear();
+                m_timeMap.clear();
                 m_shutdown = true;
```
bghgary, line 140: `m_shutdown` isn't protected inside the mutex lock in the thread proc. It is probably okay on most platforms, but maybe we should use `std::atomic<bool>` or `arcana::cancellation` instead.

bghgary (commented):
I just did a first pass. I will do a second pass after some of the feedback is addressed.

Polyfills/Window/Source/Window.cpp
docEdub: Thread is joined further down in the destructor:

https://github.com/docEdub/BabylonNative/blob/3a67ac6606074ac01554aed372747b215041a565/Polyfills/Window/Source/Window.cpp#L144-L147

bghgary: Ahh, yes, GitHub hid it because I'm only looking at the commits you made.

Polyfills/Window/Source/Window.cpp
docEdub: Would you use bare pointers for this? ...or is there an STL pointer type that would be better?

bghgary: Raw pointers would be fine for this.

Polyfills/Window/Source/Window.cpp
docEdub: This is Robert's name for it. It's the time point that will be reached next. It's being used to tell the thread how long to sleep, i.e., "wake up the thread when the soonest time point is reached".

docEdub: Maybe change it to "earliestTimePoint"?

bghgary: Why is it 1 millisecond from now when the time map is empty?

docEdub: Should probably be `std::chrono::millisecond::max`?

docEdub: Oh wait, I think the way Robert set this up is the 1 millisecond setting wakes up the thread and the thread sets process function sets the next wake up time. This can probably be done better.

docEdub: Ya, good catch. I tested it with a single 1 second timeout and it never fires. Changing it to `TimePoint::max()` fixes it.

Polyfills/Window/Source/Window.cpp
```diff
@@ -32,46 +40,85 @@ namespace
             , m_thread{&TimeoutDispatcher::WaitThenCallProc, this}
         { }
 
-        void Dispatch(std::shared_ptr<Napi::FunctionReference> func, std::chrono::system_clock::time_point time) 
-        {   
-            if (time <= std::chrono::system_clock::now())
+        TimeoutId Dispatch(std::shared_ptr<Napi::FunctionReference> func, Milliseconds time)
+        {
+            if (time <= 0)
```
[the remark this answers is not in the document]
docEdub: Calls to `setTimeout` can have a negative duration.

Polyfills/Window/Source/Window.cpp
docEdub: For late timeouts, how do we make sure they don't get dispatched until after all the currently executing code has been run?

Polyfills/Window/Source/Window.cpp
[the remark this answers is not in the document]
bghgary: What should it do when that happens?

Polyfills/Window/Source/Window.cpp
bghgary: I think we have to use the WorkQueue.

Polyfills/Window/Source/Window.cpp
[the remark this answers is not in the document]
docEdub: https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html bullet point 4 says to set the delay to zero if it's less than zero.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
+#include "TimeoutDispatcher.h"
+
+#include <assert.h>
```
bghgary, line 3: https://stackoverflow.com/questions/43011737/is-it-better-to-include-cassert-or-assert-h
```suggestion
#include <cassert>
```

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+namespace Babylon::Polyfills::Internal
+{
+    using TimePoint = std::chrono::time_point<std::chrono::steady_clock, std::chrono::milliseconds>;
+
+    TimePoint Now()
+    {
+        return std::chrono::time_point_cast<std::chrono::milliseconds, std::chrono::steady_clock>(std::chrono::steady_clock::now());
+    }
```
bghgary, line 12: nit: this should probably be in an anonymous namespace

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    struct Timeout
```
bghgary, line 14: I would recommend adding code to delete the copy constructors.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+    {
+        {
+            std::unique_lock<std::mutex> lk(m_mutex);
+            m_idMap.clear();
+            m_timeMap.clear();
+            m_shutdown = true;
+            m_condVariable.notify_one();
+        }
+
+        if (m_thread.joinable())
```
bghgary, line 47: I don't think it's necessary to call this. If we can't join the thread, it's programming error the way we are using it.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    TimeoutId TimeoutDispatcher::Dispatch(Napi::Function function, std::chrono::milliseconds delay)
+    {
+        if (delay.count() <= 0)
+        {
+            delay = std::chrono::milliseconds{0};
+        }
+
+        std::unique_lock<std::mutex> lk(m_mutex);
```
bghgary, line 60: nit: multiple places
```suggestion
        std::unique_lock<std::mutex> lk{m_mutex};
```

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+        const auto earliestTime = m_timeMap.empty() ? TimePoint::max()
+                                                    : m_timeMap.cbegin()->second->time;
+        const auto id = NextTimeoutId();
+        const auto time = Now() + delay;
+        m_idMap.insert({id, std::make_unique<Timeout>(id, std::move(function), time)});
+        m_timeMap.insert({time, m_idMap[id].get()});
```
bghgary, line 67: `insert` returns the iterator where it inserted. You can use that instead of accessing the map again. Something like this:

```suggestion
        const auto it = m_idMap.insert({id, std::make_unique<Timeout>(id, std::move(function), time)});
        m_timeMap.insert({time, it->second.get()});
```

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+            m_condVariable.notify_one();
+        }
+
+        return id;
+    }
+
+    void TimeoutDispatcher::Clear(TimeoutId id)
+    {
+        std::unique_lock<std::mutex> lk(m_mutex);
+        const auto idIterator = m_idMap.find(id);
```
bghgary, line 80: nit: maybe use this convention `it<name of map>`
```suggestion
        const auto itId = m_idMap.find(id);
```

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+            assert(m_timeMap.lower_bound(timeout->time) != m_timeMap.end() && "m_idMap and m_timeMap are out of sync");
+
+            for (auto i = m_timeMap.lower_bound(timeout->time); i->first == timeout->time; i++)
+            {
+                if (i->second->id == id)
+                {
+                    m_timeMap.erase(i);
+                    break;
+                }
+            }
```
bghgary, line 94:
```suggestion
            for (auto itTime = m_timeMap.lower_bound(timeout->time); itTime->first == timeout->time; ++itTime)
            {
                if (itTime->second->id == id)
                {
                    m_timeMap.erase(itTime);
                    break;
                }
            }
```

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+    {
+        std::unique_lock<std::mutex> lk(m_mutex);
+        const auto idIterator = m_idMap.find(id);
+        if (idIterator != m_idMap.end())
+        {
+            const auto& timeout = idIterator->second;
+
+            assert(m_timeMap.lower_bound(timeout->time) != m_timeMap.end() && "m_idMap and m_timeMap are out of sync");
+
+            for (auto i = m_timeMap.lower_bound(timeout->time); i->first == timeout->time; i++)
```
bghgary, line 87: Maybe use [equal_range](https://en.cppreference.com/w/cpp/container/multimap/equal_range) instead? See the example. You can use the result for the assert also.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+                }
+            }
+
+            m_idMap.erase(idIterator);
+        }
+    }
+
+    void TimeoutDispatcher::ThreadFunction()
+    {
+        TimePoint nextTimePoint{};
```
bghgary, line 102: This should probably be moved inside the while loop right the first inner while loop? It's not used outside.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+        while (!m_shutdown)
+        {
+            std::unique_lock<std::mutex> lk(m_mutex);
+
+            while (!m_timeMap.empty() && Now() < (nextTimePoint = m_timeMap.begin()->second->time))
+            {
+                m_condVariable.wait_until(lk, nextTimePoint);
+            }
```
bghgary, line 110: nit: Inline assignments are hard to read for me. I would prefer this:
```suggestion
            while (!m_timeMap.empty())
            {
                nextTimePoint = m_timeMap.begin()->second->time;
                if (Now() >= nextTimePoint)
                {
                    break;
                }
            
                m_condVariable.wait_until(lk, nextTimePoint);
            }
```

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+            for (auto i = m_timeMap.begin(); i != m_timeMap.end() && i->second->time == nextTimePoint; i = m_timeMap.begin())
```
bghgary, line 112: This is confusing. It's not really a for loop at all. Why not use a while loop?

```suggestion
            while (!m_timeMap.empty() && m_timeMap.begin()->second->time == nextTimePoint)
```

Polyfills/Window/Source/TimeoutDispatcher.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include <Babylon/JsRuntime.h>
+
+#include <map>
```
bghgary, line 5:
```suggestion
#include <unordered_map>
#include <map>
```

Polyfills/Window/Source/TimeoutDispatcher.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Babylon::Polyfills::Internal
+{
+    using TimeoutId = int32_t;
+
+    struct Timeout;
```
bghgary, line 11: nit: I would recommend moving this inside the TimeoutDispatcher

Polyfills/Window/Source/Window.cpp
```diff
@@ -38,6 +37,11 @@ namespace Babylon::Polyfills::Internal
             global.Set(JS_SET_TIMEOUT_NAME, Napi::Function::New(env, &Window::SetTimeout, JS_SET_TIMEOUT_NAME, Window::Unwrap(jsWindow)));
         }
 
+        if (global.Get(JS_CLEAR_TIMEOUT_NAME).IsUndefined())
```
bghgary, line 40: We should probably check for both `setTimeout` and `clearTimeout` together. If there is an exisitng implementation of `setTimeout` but not `clearTimeout` or vice versa, this code is going to polyfill one of the functions and that won't work at all.

Polyfills/Window/Source/Window.h
```diff
@@ -1,9 +1,15 @@
...
+#include "TimeoutDispatcher.h"
+
 #include <Babylon/JsRuntime.h>
 
+#include <optional>
+
 namespace Babylon::Polyfills::Internal
 {
+    using TimeoutId = int32_t;
+
```
bghgary, line 12: remove?

bghgary (commented):
Looking good. Mostly minor comments.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,150 @@
...
+            {
+                const auto *timeout = i->second;
```
bghgary, line 114: nit: 
```suggestion
                const auto* timeout = i->second;
```

bghgary:
It might also be good to add some unit tests.

CedricGuillemet:
@docEdub Don't forget to add unit tests here: https://github.com/BabylonJS/BabylonNative/blob/master/Apps/UnitTests/Scripts/tests.js
It's handy because it runs on Mac and Linux as well.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+        Timeout(const Timeout&) = delete;
+        Timeout(Timeout&&) = delete;
+    };
+
+    TimeoutDispatcher::TimeoutDispatcher(Babylon::JsRuntime& runtime)
+        : m_runtime(runtime)
+    {
+        m_thread = std::thread{&TimeoutDispatcher::ThreadFunction, this};
+    }
```
docEdub, line 41: Note that I moved the initialization of `m_thread` into the body of the constructor to make sure `m_mutex` is initialized before the thread start. While debugging on Ubuntu, the thread kept starting up before `m_mutex` was ready.

Let me know if there is a better way to do this.

Apps/UnitTests/Scripts/tests.js
```diff
@@ -327,6 +327,110 @@ describe("PostProcesses", function() {
...
+            catch (e) {
+                done(e);
+                return;
+            }
+            done();
+         }, undefined);
+        trailingCodeExecuted = true
+    });
+    it("should call the given functions in the correct order", function (done) {
+        let called = []
```
bghgary, line 400:
```suggestion
        const called = [];
```

Apps/UnitTests/Scripts/tests.js
```diff
@@ -327,6 +327,110 @@ describe("PostProcesses", function() {
...
+    this.timeout(0);
+    it("should stop the timeout matching the given timeout id", function (done) {
+        const id = setTimeout(() => {
+            done(new Error("Timeout was not cleared"));
+        }, 0);
+        clearTimeout(id);
+        setTimeout(done, 100);
+    });
+    it("should do nothing if the given timeout id is undefined", function (done) {
+        setTimeout(() => { done(); }, 0);
```
bghgary, line 429: nit: is this necessary? can you just call `done()` after `clearTimeout`

Polyfills/Window/Source/TimeoutDispatcher.cpp
bghgary: Variable initialization should be in order of declaration. See [this](https://stackoverflow.com/questions/6308915/member-fields-order-of-construction#:~:text=Member%20variables%20are%20constructed%20in%20the%20order%20in,variables%20appear%20in%20the%20initializer%20list%20is%20irrelevant.).

You just need to move `m_mutex` before `m_thread` in the header. It would have been an uninitialized access otherwise.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,160 @@
...
+            , time{time}
+        {
+        }
+
+        Timeout(const Timeout&) = delete;
+        Timeout(Timeout&&) = delete;
+    };
+
+    TimeoutDispatcher::TimeoutDispatcher(Babylon::JsRuntime& runtime)
+        : m_runtime(runtime)
```
bghgary, line 38: nit:
```suggestion
        : m_runtime{runtime}
```

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,160 @@
...
+        const auto earliestTime = m_timeMap.empty() ? TimePoint::max()
+                                                    : m_timeMap.cbegin()->second->time;
+        const auto id = NextTimeoutId();
+        const auto time = Now() + delay;
+        const auto result = m_idMap.insert({id, std::make_unique<Timeout>(id, std::move(function), time)});
+        m_timeMap.insert({time, result.first->second.get()});
+
+        if (time <= earliestTime)
+        {
+            m_runtime.Dispatch([this](Napi::Env){ m_condVariable.notify_one(); });
```
bghgary, line 74: I think there is a problem here. `m_condVariable.notify_one();` is not under the `std::unique_lock` scope anymore, so this is not correct usage of condition variables.

bghgary: I'm also not exactly sure what this mutex is supposed to protect. All of the members or just the condition variable?

bghgary: Maybe all of this code under the lock should be on the JS thread?

Polyfills/Window/Source/TimeoutDispatcher.h
```diff
@@ -0,0 +1,45 @@
...
+        using TimePoint = std::chrono::time_point<std::chrono::steady_clock, std::chrono::microseconds>;
+
+        void ThreadFunction();
+        void CallFunction(std::shared_ptr<Napi::FunctionReference> function);
+
+        Babylon::JsRuntime& m_runtime;
+        std::thread m_thread;
+        std::mutex m_mutex{};
+        std::condition_variable m_condVariable{};
+        TimeoutId m_lastTimeoutId = 0;
```
bghgary, line 40: nit: 
```suggestion
        TimeoutId m_lastTimeoutId{0};
```

Polyfills/Window/Source/Window.cpp
```diff
@@ -72,17 +72,50 @@ namespace Babylon::Polyfills::Internal
...
-        auto function = Napi::Persistent(info[0].As<Napi::Function>());
-        auto milliseconds = std::chrono::milliseconds{info[1].As<Napi::Number>().Int32Value()};
-
         auto& window = *static_cast<Window*>(info.Data());
 
-        window.RecursiveWaitOrCall(std::make_shared<Napi::FunctionReference>(std::move(function)), std::chrono::system_clock::now() + milliseconds);
+        const auto arg0 = info[0];
+        if (!arg0.IsFunction())
+        {
+            return Napi::Value::From(info.Env(), window.m_timeoutDispatcher->NextTimeoutId());
```
bghgary, line 86: It's highly unlikely, but this doesn't guarantee that a subsequent setTimeout doesn't use the same id if the id wraps around. Are you trying to return a positive number?

Polyfills/Window/Source/Window.cpp
```diff
@@ -72,17 +72,50 @@ namespace Babylon::Polyfills::Internal
...
+                delay = std::chrono::milliseconds{std::stoi(delayString)};
+            }
+            catch (std::invalid_argument&)
+            {
+            }
+        }
+        else if (info[1].IsNumber())
+        {
+            delay = std::chrono::milliseconds{info[1].ToNumber().Int32Value()};
+        }
```
bghgary, line 105: N-API has a method called `ToNumber()` which coerces the type to be a number. Does that work?

docEdub: Ya, I was using `ToNumber()` earlier but it wasn't working with the JSI engine.

Apps/UnitTests/Scripts/tests.js
docEdub: `done()` is called in the `setTimeout` callback to make sure `clearTimeout` has not cleared it.

Polyfills/Window/Source/Window.cpp
bghgary: ToNumber isn't completely implemented in napi-jsi, but we can probably make it work.

Apps/UnitTests/Scripts/tests.js
bghgary: that seems arbitrary, but okay.

Polyfills/Window/Source/Window.cpp
docEdub: We could return `undefined` instead, so that a call to `clearTimeout` using the value returned here does not do anything.

bghgary: You could also reserve a number (say max value) for this specific case. What do browsers do?

docEdub: Chrome returns the next id.

Polyfills/Window/Source/TimeoutDispatcher.cpp
docEdub: Oof! Good catch. Ya, I'll try moving it all to the JS thread. Can remove the mutex and make `m_shutdown` a `bool` again instead of a `std::atomic<bool>` if we do it that way?

Polyfills/Window/Source/Window.cpp
bghgary: Maybe it also adds an entry into the map so that it can collide with other ones?

Apps/UnitTests/Scripts/tests.js
```diff
@@ -342,25 +342,25 @@ describe("setTimeout", function () {
         setTimeout(() => {
             try {
                 expect(new Date().getTime() - startTime).to.be.at.least(10);
+                done();
             }
             catch (e) {
                 done(e);
                 return;
```
bghgary, line 349: remove this also

Polyfills/Window/Source/TimeoutDispatcher.cpp
bghgary: I don't think so. `m_shutdown` is still accessed from the Timeout thread.

Apps/UnitTests/Scripts/tests.js
docEdub: Ya, it's not very useful but it's not taking up much time beyond making sure `clearTimeout(undefined)` doesn't crash.

Polyfills/Window/Source/TimeoutDispatcher.cpp
docEdub: Like this?
```C++
    TimeoutDispatcher::TimeoutId TimeoutDispatcher::Dispatch(Napi::Function function, std::chrono::milliseconds delay)
    {
        if (delay.count() < 0)
        {
            delay = std::chrono::milliseconds{0};
        }

        const auto id = NextTimeoutId();

        m_runtime.Dispatch([this, function, delay, id](Napi::Env){
            std::unique_lock<std::mutex> lk{m_mutex};

            const auto earliestTime = m_timeMap.empty() ? TimePoint::max()
                                                        : m_timeMap.cbegin()->second->time;
            const auto time = Now() + delay;
            const auto result = m_idMap.insert({id, std::make_unique<Timeout>(id, std::move(function), time)});
            m_timeMap.insert({time, result.first->second.get()});

            if (time <= earliestTime)
            {
                m_condVariable.notify_one();
            }
        });

        return id;
    }

    void TimeoutDispatcher::Clear(TimeoutId id)
    {
        m_runtime.Dispatch([this, id](Napi::Env){
            std::unique_lock<std::mutex> lk{m_mutex};
            const auto itId = m_idMap.find(id);
            if (itId != m_idMap.end())
            {
                const auto& timeout = itId->second;
                const auto timeRange = m_timeMap.equal_range(timeout->time);

                assert(timeRange.first != m_timeMap.end() && "m_idMap and m_timeMap are out of sync");

                for (auto itTime = timeRange.first; itTime != timeRange.second; itTime++)
                {
                    if (itTime->second->id == id)
                    {
                        m_timeMap.erase(itTime);
                        break;
                    }
                }

                m_idMap.erase(itId);
            }
        });
    }
```

Polyfills/Window/Source/Window.cpp
docEdub: Probably. It's hard to tell for certain looking at the Chromium source code.

Polyfills/Window/Source/Window.cpp
docEdub: https://github.com/BabylonJS/BabylonNative/pull/1146/commits/e83cca81c7de01aa051e2d3b453b637ba53a3698

Polyfills/Window/Source/Window.cpp
docEdub: Change in commit https://github.com/BabylonJS/BabylonNative/pull/1146/commits/6e605ccbe7648530237002cde4c03f829855264e  to use a function that does nothing.

Polyfills/Window/Source/TimeoutDispatcher.cpp
bghgary: Seems right.

Dependencies/napi/napi-jsi/include/napi/napi-inl.h
```diff
@@ -247,7 +247,24 @@ inline Boolean Value::ToBoolean() const {
...
+    NAPI_TRY()
+    try
+    {
+      return Number::New(_env, std::stod(_value.asString(_env->rt).utf8(_env->rt)));
+    }
+    catch (std::invalid_argument&)
+    {
+      return Number::New(_env, 0);
+    }
+    NAPI_CATCH()
```
bghgary, line 266: This is pretty strange. It might be better to not use the macro here.

```suggestion
    try
    {
      return Number::New(_env, std::stod(_value.asString(_env->rt).utf8(_env->rt)));
    }
    catch (std::invalid_argument&)
    {
      return Number::New(_env, 0);
    }
    catch (const std::exception& exception) {
      throw Error::New(_env, exception);
    }
```

Apps/UnitTests/Scripts/tests.js
bghgary: I still see the extraneous `return;` statements.

bghgary (approved):

Apps/UnitTests/Scripts/tests.js
docEdub: I've got this fixed locally, queued for the next push.

Polyfills/Window/Source/TimeoutDispatcher.cpp
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    TimeoutDispatcher::TimeoutId TimeoutDispatcher::Dispatch(Napi::Function function, std::chrono::milliseconds delay)
+    {
+        if (delay.count() < 0)
+        {
+            delay = std::chrono::milliseconds{0};
+        }
+
+        const auto id = NextTimeoutId();
```
bghgary, line 63: This also uses the map to check if the id exists and thus is a race condition.

bghgary: It might be prudent to use two different mutexes for the condition variable and the maps.

docEdub: I'm going to need help with separate mutexes. I gave it a try [in this commit](https://github.com/docEdub/BabylonNative/commit/d6af3410da3866344996d493fe21cad37aa2c340) but it made most of the `setTimeout` unit tests time out.

PR #1152, opened 2022-10-10
Fix breaking change from bgfx update

bghgary:
that broke morph targets

See https://github.com/bkaradzic/bgfx/pull/2885#discussion_r991464619


CedricGuillemet (approved):
LGTM

PR #1157, opened 2022-10-26
Fix unity build CI

CedricGuillemet:
Context:
Unity build(UB) is enabled in BN and BRN because of BRN binary size (more here : https://github.com/BabylonJS/BabylonReactNative/pull/416 )
UB was not properly tested in BN because of CI vars forwarding.
Once UB is enabled in BRN, some issues appear:
- There are some #define conflicts that do not allow Win32 DPI functions to be declared for core/graphics lib 
- with win32/x86, a method that throw is discarded by the compiler and an 'optimized' code path detect a lambda to not be reachable (NativeVideo). It only happens with MSVC/x86

UB is a patch, the correct way to fix the root issue is to improve BRN build/packaging so compressed binary does not take more than 200Mb.


Plugins/NativeCamera/Source/UWP/NativeCameraImpl.cpp
```diff
@@ -14,7 +14,7 @@ namespace Babylon::Plugins
 
     arcana::task<Camera::Impl::CameraDimensions, std::exception_ptr> Camera::Impl::Open(uint32_t /*width*/, uint32_t /*height*/, bool /*frontCamera*/)
     {
-        throw std::runtime_error{ "HW Camera not implemented for this platform." };
+        return {};
```
bghgary, line 17: I'm assuming this comment matches this code?
> with win32/x86, a method that throw is discarded by the compiler and an 'optimized' code path detect a lambda to not be reachable (NativeVideo). It only happens with MSVC/x86

I don't get it. This is technically not correct?

Plugins/NativeEngine/CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ add_library(NativeEngine ${SOURCES})
 
 target_include_directories(NativeEngine PUBLIC "Include")
 
+if(WIN32 AND NOT WINDOWS_STORE)
+    set_property(TARGET NativeEngine PROPERTY UNITY_BUILD false)
```
bghgary, line 34: Is this temporary? Will this cause the size to be big again?

Plugins/NativeCamera/Source/UWP/NativeCameraImpl.cpp
CedricGuillemet: Yes! it's correct. code can be discarded and unreachable code warning should(may) happen. but it's only the case for x86 MSVC...other compiler do not optimize it...or do not have a warning in that case.

Plugins/NativeEngine/CMakeLists.txt
CedricGuillemet: BRN is UWP. WINDOW_STORE == true in that case.

bghgary: Ah, okay.

bghgary: Do we even need UNITY_BUILD overall if it's not UWP?

Plugins/NativeCamera/Source/UWP/NativeCameraImpl.cpp
bghgary: I don't understand. This function is not implemented. We can't return an empty task. That's not correct behavior?

Plugins/NativeEngine/CMakeLists.txt
CedricGuillemet: I'm curious to see the impact on overall build time. But yes, otherwise, only usefull for UWP.

Plugins/NativeCamera/Source/UWP/NativeCameraImpl.cpp
CedricGuillemet: Yes, not sure about this one. I prefer to revert the empty task change and disable UB for this project.

PR #1158, opened 2022-10-26
Update NativeCamera to support MediaStreamTrack API

CoPrez:
The main goal of this PR is to update the NativeCamera implementation to fulfill more of the [MediaStreamTrack WebAPI](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack) and in particular the [capabilities and constraints](https://developer.mozilla.org/en-US/docs/Web/API/Media_Capture_and_Streams_API/Constraints) portion of the spec. The capabilities and constraints API allows the client (Babylon Native in this case) to declare which capabilities the camera device supports (torch, zoom, focus mode, etc.). At the same time the consumer can pass in constraints, both at instantiation and during playback, to update the configuration of the camera.

This is a snippet of how the consumer would use this code:
``` typescript
const mediaStream = await navigator.mediaDevices.getUserMedia({
  video:{
    torch: true // Ideally pick a device that supports torch (flashlight)
    width: { ideal: 1280, max: 1920 } // Pick a stream that is closest to 1280 pixels wide, but no more than 1920
    height: { ideal: 720, max: 1080 } // Pick a stream that is closest to 720 pixels heigh, but no more than 1080
  }
}

// Save the MediaStreamTrack for later
const mediaTrack = mediaStream.getVideoTracks()[0];

// Crate the VideoTextureto use in the scene
const videoTexture = await BABYLON.VideoTexture.CreateFromStreamAsync(scene, mediaStream, {});

...

// After some action, check if the device supports the torch capability and toggle it
if (mediaTrack.getCapabilities().torch)
{
  // Get the current set of applied constraints
  let currentConstraints = mediaTrack.getConstraints();
  // Toggle the torch constraint
  currentConstraints.torch = !mediaTrack.getSettings().torch;
  // Then apply the modified constraints
  await mediaTrack.applyConstraints(currentConstraints);  
}
```

The major changes I've made to achieve the above flow are:
- Introduced a native polyfill for the MediaStream and MediaStreamTrack objects which are constructed when `navigator.mediaDevices.getUserMedia()` is called.
  - Calling getUserMedia() is what now opens the camera stream and prompts for camera permission (which matches the web API). Previously we were opening the camera stream when constructing the native video element which is why VideoTexture.CreateFromStreamAsync takes in a constraints object that was only used by NativeEngine.
  - The Video element native polyfill no longer declares the `isNative=true` property which causes Babylon.JS to assign the result of `navigator.mediaDevices.getUserMedia()` to the Video elements `.src` property.
  - The native polyfill of the MediaStream object is now the one to maintain the lifecycle of the actual camera stream (which matches the web API as well)
- Introduced a CameraCapabilities object which fulfills the capabilities and constraints spec
  - The platform specific NativeCameraImpl classes now iterate over available camera devices on the platform and translate them into platform agnostic CameraDevice objects with registered capabilities
  - Choosing which camera device to open a stream for is now handled in the platform agnostic code as the logic to choose the device is a lot more complicated now to match the capabilities and constraints spec of the web api
 - The NativeCameraImpl classes now all share a common header file so the interface is ensured to be consistent across all platform implementations


Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -106,24 +97,57 @@ namespace Babylon
...
+            }, "then");
+
+            Napi::Object constraints{ info[0].As<Napi::Object>() };
+            auto videoConstraints{ constraints.Get("video").As<Napi::Object>() };
+            if (videoConstraints.IsUndefined())
+            {
+                videoConstraints = Napi::Object::New(env);
+            }
+
+            // For backwards compatibility support top level properties that aren't part of the official constraint spec,
```
CoPrez, line 131: Remove this

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -106,24 +97,57 @@ namespace Babylon
...
+            if((videoConstraints.Has("minHeight") || videoConstraints.Has("maxHeight")) && !videoConstraints.Has("height"))
+            {
+                auto heightObject = Napi::Object::New(env);
+                heightObject.Set("min", videoConstraints.Get("minHeight"));
+                heightObject.Set("max", videoConstraints.Get("maxHeight"));
+                videoConstraints.Set("height", heightObject);
+            }
+            
+            auto applyPromise = mediaStream->ApplyConstraints(env, videoConstraints).As<Napi::Promise>();
+            return applyPromise.Get("then").As<Napi::Function>().Call(applyPromise, {callback});
```
CoPrez, line 150: Stay in javascript or stay in native

Plugins/NativeCamera/Source/NativeVideo.cpp
```diff
@@ -79,7 +83,10 @@ namespace Babylon::Plugins
 
     void NativeVideo::UpdateTexture(bgfx::TextureHandle textureHandle)
     {
-        NativeCameraImpl->UpdateCameraTexture(textureHandle);
+        if (m_IsPlaying && !m_streamObject.Value().IsNull() && !m_streamObject.Value().IsUndefined())
+        {
+            MediaStream::Unwrap(m_streamObject.Value())->UpdateTexture(textureHandle);
```
CoPrez, line 88: Don't unwrap again, use the member variable

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -217,56 +178,42 @@ namespace Babylon::Plugins
...
+        : m_implData{std::make_unique<ImplData>(env, overrideCameraTexture)}
     {
         if (API_LEVEL < 24 && !overrideCameraTexture)
         {
             throw std::runtime_error{"Android Platform level < 24. Only camera texture override is available."};
         }
+
+        if (!overrideCameraTexture)
+        {
+            m_implData->cameraManager = GET_CAMERA_FUNCTION(ACameraManager_create)();
```
Alex-MSFT, line 190: Any reason to do this here instead of in the implData constructor?

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -309,129 +256,248 @@ namespace Babylon::Plugins
...
+                int32_t height{streamConfigurations.data.i32[j + 2]};
+                int32_t input{streamConfigurations.data.i32[j + 3]};
+
+                if (input || format != AIMAGE_FORMAT_YUV_420_888)
+                {
+                    // Ignore the configuration if it is either an input type or not a preview format
+                    continue;
+                }
+
+                auto cameraResolution{ std::make_shared<CameraTrack>() };
```
Alex-MSFT, line 378: I don't think you want a shared_ptr here. My intuition think that the supportedResolutions vector should own the struct, so this should be a plain std::vector<CameraTrack>. You would construct the camera track by using supportedResolutions.emplace_back() which would create the object, then pull it off the back and edit the values.

Take a look how something similar is done in AR.
https://github.com/BabylonJS/BabylonNative/blob/581b4b4e16a46e31b555204472886eb3fbf77418/Dependencies/xr/Source/ARCore/XR.cpp#L1113-L1116

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -309,129 +256,248 @@ namespace Babylon::Plugins
...
+            float minZoomRatio{ metaDataEntry.data.f[0] };
+            float maxZoomRatio{ metaDataEntry.data.f[1] };
+
+            // update the cameraDevice information
+            cameraDevice->implData = std::make_unique<CameraDevice::ImplData>();
+            cameraDevice->implData->cameraID = id;
+            cameraDevice->implData->sensorRotation = sensorOrientation;
+            cameraDevice->implData->facingUser = facing == API24::ACAMERA_LENS_FACING_FRONT;
+
+            cameraDevice->capabilities.push_back(std::make_shared<CameraCapabilityTemplate<std::string>>(
```
Alex-MSFT, line 409: Same for all these places, I don't think you want to use make_shared but rather emplace_back a value owned by each vector.

Plugins/NativeCamera/Source/NativeCameraImpl.h
```diff
@@ -17,24 +39,18 @@ namespace Babylon::Plugins
...
 
         struct ImplData;
-        
+
         Impl(Napi::Env env, bool overrideCameraTexture);
         ~Impl();
-        arcana::task<CameraDimensions, std::exception_ptr> Open(uint32_t maxWidth, uint32_t maxHeight, bool frontCamera);
+        arcana::task<CameraDimensions, std::exception_ptr> Open(std::shared_ptr<CameraDevice> cameraDevice, std::shared_ptr<CameraTrack> track);
+
+        std::vector<std::shared_ptr<CameraDevice>> GetCameraDevices();
```
Alex-MSFT, line 47: Following along the same vein as the other comments should ultimately be std::vector<CameraDevice>&

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -208,113 +370,24 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
 
         // Capture the format dimensions.
-        CMVideoFormatDescriptionRef videoFormatRef{static_cast<CMVideoFormatDescriptionRef>(bestFormat.formatDescription)};
+        CMVideoFormatDescriptionRef videoFormatRef{static_cast<CMVideoFormatDescriptionRef>(resolution->implData->avDeviceFormat.formatDescription)};
         CMVideoDimensions dimensions{CMVideoFormatDescriptionGetDimensions(videoFormatRef)};
 #else
-        UNUSED(maxWidth);
-        UNUSED(maxHeight);
-        UNUSED(frontCamera);
         AVCaptureDevice* captureDevice{[AVCaptureDevice defaultDeviceWithMediaType:AVMediaTypeVideo]};
```
Alex-MSFT, line 388: Is iOS vs Mac OS branch still needed? It seems like maybe we can have these work the same on iOS and Mac OS since the device selection has been moved up to GetCameraDevices.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -208,113 +370,24 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
         AVCaptureDeviceInput *input{[AVCaptureDeviceInput deviceInputWithDevice:captureDevice error:&error]};
-        CMVideoFormatDescriptionRef videoFormatRef{static_cast<CMVideoFormatDescriptionRef>(captureDevice.activeFormat.formatDescription)};
+        CMVideoFormatDescriptionRef videoFormatRef{static_cast<CMVideoFormatDescriptionRef>(resolution->implData->avDeviceFormat.formatDescription)};
```
Alex-MSFT, line 390: Is iOS vs Mac OS branch still needed? It seems like maybe we can have these work the same on iOS and Mac OS since the device selection has been moved up to GetCameraDevices.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -149,35 +175,171 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
+            {
+                // Newer iOS devices might have a telephoto that has a starting zoom higher, but so far
+                // as I can tell there is no API to determine the default real world zoom of the telephoto lens
+                zoomFactorScale = 2.0;
+            }
+            
+            cameraDevice->capabilities.push_back(std::make_shared<CameraCapabilityTemplate<double>>
+            (
+                CameraCapability::Capability::Zoom,
+                1.0 * zoomFactorScale, // The device always opens at 1.0 which needs to be adjusted for the zoomFactorScale
```
Alex-MSFT, line 303: Why do this multiplication?  It looks like this can only be 3 values 1.0, 0.5, or 2.0.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -149,35 +175,171 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
+                    [m_implData->avCaptureDevice unlockForConfiguration];
+                    if (error != nil)
+                    {
+                        return false;
+                    }
+                    return true;
+                }
+            ));
+#endif
+
```
Alex-MSFT, line 320:
```suggestion
```

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
```diff
@@ -149,35 +175,171 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
+            if (foundExactMatch)
+            {
+                break;
+            }
         }
+        
+        return cameraDevices;
+    }
+
```
Alex-MSFT, line 332:
```suggestion
```

Plugins/NativeCamera/Source/CameraCapability.cpp
```diff
@@ -0,0 +1,352 @@
...
+        switch(getConstraintType())
+        {
+            case ConstraintType::Sequence:
+            {
+                auto capability{ Napi::Array::New(env, m_acceptedValues.size()) };
+                for (uint32_t i = 0; i < m_acceptedValues.size(); i++) {
+                    capability.Set(i, Constraint::asNapiValue<T>(env, m_acceptedValues[i]));
+                }
+
+                target.Set(getName(),capability);
```
Alex-MSFT, line 66:
```suggestion
                target.Set(getName(), capability);
```

Plugins/NativeCamera/Source/CameraCapability.cpp
```diff
@@ -0,0 +1,352 @@
...
+                return;
+            }
+            case ConstraintType::Range:
+            {
+                auto capability{ Napi::Object::New(env) };
+                capability.Set("min", Constraint::asNapiValue<T>(env, m_acceptedValues[0]));
+                capability.Set("max", Constraint::asNapiValue<T>(env, m_acceptedValues[1]));
+
+                target.Set(getName(),capability);
```
Alex-MSFT, line 75:
```suggestion
                target.Set(getName(), capability);
```

Plugins/NativeCamera/Source/CameraCapability.cpp
```diff
@@ -0,0 +1,352 @@
...
+                }
+                
+                T minAccepted = m_acceptedValues[0];
+                T maxAccepted = m_acceptedValues[1];
+
+                if (constraint.exact.has_value())
+                {
+                    // The exact value has highest priority the constraint is either fully satisfied or unsatisfied based on it
+                    return constraint.exact.value() >= minAccepted && constraint.exact.value() <= maxAccepted ? CameraCapability::MeetsConstraint::FullySatisfied : CameraCapability::MeetsConstraint::Unsatisfied;
+
```
Alex-MSFT, line 137:
```suggestion
```

Plugins/NativeCamera/Source/CameraCapability.cpp
```diff
@@ -0,0 +1,352 @@
...
+        
+        if (constraintValue.IsEmpty() || constraintValue.IsUndefined() || constraintValue.IsNull())
+        {
+            // The constraints don't contain any restrictions for this capability.
+            return CameraCapability::MeetsConstraint::Unconstrained;
+        }
+
+        Constraint::Constraint<T> constraint{ Constraint::parseConstraint<T>(constraintValue) };
+
+        switch(getConstraintType())
```
Alex-MSFT, line 103:
```suggestion
        switch (getConstraintType())
```

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,335 @@
...
+        // To simply the implementation (because only a single track is supported by the native implementation) the same
+        // ObjectWrap fullfills the MediaStream and MediaStreamTrack interfaces
+        auto tracks = Napi::Array::New(info.Env(), 1);
+        tracks.Set(uint32_t{0}, info.This());
+        return std::move(tracks);
+    }
+
+    Napi::Value MediaStream::GetAudioTracks(const Napi::CallbackInfo& info)
+    {
+        // Only video tracks are currently supported
```
Alex-MSFT, line 55: Does this need to be wrapped in std::move?

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,335 @@
...
+        return Napi::Array::New(info.Env(), 0);
```
Alex-MSFT, line 56: Wrap in std::move?

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,335 @@
...
+    Napi::Value MediaStream::ApplyConstraints(const Napi::CallbackInfo& info)
+    {
+        auto env = info.Env();
+        auto constraints = info[0].As<Napi::Object>();
+        
+        auto deferred{Napi::Promise::Deferred::New(env)};
+        auto promise{deferred.Promise()};
+
+        // Apply the constraints to the existing device
+        if(!UpdateConstraints(env, constraints))
```
Alex-MSFT, line 68:
```suggestion
        if (!UpdateConstraints(env, constraints))
```

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,335 @@
...
+            {
+                // Re-throw the error from opening the camera
+                throw result.error();
+            }
+            
+            auto cameraDimensions{result.value()};
+            this->Width = cameraDimensions.width;
+            this->Height = cameraDimensions.height;
+            
+            if(!UpdateConstraints(env, constraintsRef->Value()))
```
Alex-MSFT, line 107:
```suggestion
            if (!UpdateConstraints(env, constraintsRef->Value()))
```

Alex-MSFT: How do we get into a scenario where we fail to apply constraints after the camera is open?  Shouldn't FindBestCameraStream guarantee we meet the constraints?

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,335 @@
...
+        
+        if (m_cameraDevice == nullptr)
+        {
+            // We don't have a cameraDevice selected yet.
+            return Napi::Object::New(env);
+        }
+        
+        auto capabilities = Napi::Object::New(env);
+        
+        for (uint32_t i=0; i < m_cameraDevice->capabilities.size(); i++)
```
Alex-MSFT, line 129:
```suggestion
        for (uint32_t i = 0; i < m_cameraDevice->capabilities.size(); i++)
```

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: It needs to be guarded by the API check... but I can do that in the constructor as well. I'll move it up there.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
CoPrez: I'm updating the comment to clarify this.

This line of code is setting the initial zoom factor which when opening the device it will always start at 1.0 according to the scale Apple provides. That scale is dumb though and the value always goes from 1.0+ even if you're using the wide angle lens where 1.0 really means a zoom ratio of 0.5.

I have the zoomFactorScale in place to translate from Apples scale that always starts at 1.0 (regardless of the lens's intrinsic zoom ratio) to our desired scale where the zoom ratio equates to how zoomed in the output image is.

Plugins/NativeCamera/Source/MediaStream.cpp
CoPrez: I don't think that's needed here because I'm directly returning the new object without storing it as a local variable first.

Plugins/NativeCamera/Source/NativeCameraImpl.h
```diff
@@ -0,0 +1,74 @@
...
+    class CameraDevice final
+    {
+    public:
+        ~CameraDevice();
+        
+        // Move semantics
+        CameraDevice(CameraDevice&&) noexcept;
+        CameraDevice& operator=(CameraDevice&&) noexcept;
+                
+        const std::vector<CameraTrack>& SupportedResolutions() const;
```
CoPrez, line 40: @bghgary, this is the route I landed on for giving the consumer read access to the `std::vector<CameraTrack>` that the `CameraDevice` still owns the lifetime for.

I can try to work out the `std::vector<std::reference_wrapper<CameraTrack>>` but I think that route might require allocating a second vector.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: I updated the CameraTrack class to now properly support the pimpl patter and the move operator so that the vector can contain just the class itself and not a shared or unique pointer to it.

Plugins/NativeCamera/Source/Apple/NativeCameraImpl.mm
CoPrez: I got back around to testing this and you are indeed correct. I was able to remove the main chunk of differing code between iOS and MacOS. There are still a few sporadic chunks that only work on iOS due to the camera API's being more robust on iOS. For the most part though they are now aligned and use a single implementation.

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
CoPrez: I have patched things up now so that I use emplace_back and capture the value if needed from that call. For the most part I have gotten rid of all shared_ptrs with only a few exceptions of pre-existing code that relies on it (NativeCameraImpl for instance).

Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp
```diff
@@ -184,36 +191,32 @@ namespace Babylon::Plugins
...
     }
 
     Camera::Impl::~Impl()
     {
         GET_CAMERA_FUNCTION(ACameraManager_delete)(m_implData->cameraManager);
     }
 
-    arcana::task<Camera::Impl::CameraDimensions, std::exception_ptr> Camera::Impl::Open(
-            std::shared_ptr<CameraDevice> cameraDevice, std::shared_ptr<CameraTrack> track)
+    arcana::task<Camera::Impl::CameraDimensions, std::exception_ptr> Camera::Impl::Open(CameraDevice cameraDevice, const CameraTrack* track)
```
Alex-MSFT, line 201: Does this need to be CameraDevice&?  I think this would call the copy constructor no?

Alex-MSFT (approved):

Plugins/NativeCamera/Source/CameraCapability.h
```diff
@@ -0,0 +1,109 @@
...
+        template <>
+        Constraint<bool> parseConstraint(Napi::Value value);
+        template <>
+        Napi::Value asNapiValue(Napi::Env env, bool value);
+
+        // String
+        template <>
+        Constraint<std::string> parseConstraint(Napi::Value value);
+        template <>
+        Napi::Value asNapiValue(Napi::Env env, std::string value);
```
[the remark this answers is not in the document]
bghgary: This class probably would be good in a separate file also.

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,339 @@
...
+        
+        if (cameraDevice == nullptr)
+        {
+            // We don't have a cameraDevice selected yet.
+            return false;
+        }
+        
+        for(uint32_t i=0; i < cameraDevice->Capabilities().size(); i++)
+        {
+            auto& capability{ cameraDevice->Capabilities()[i] };
```
bghgary, line 306:
```suggestion
        for (auto& capability : cameraDevice->Capabilities())
        {
```

bghgary (commented):
Lots of little things and stuff I don't understand, but overall, it looks fine. I probably need a second pass.

Plugins/NativeCamera/Source/Capability.h
```diff
@@ -0,0 +1,69 @@
...
+
+        Capability(Feature capability);
+        virtual ~Capability() = default;
+
+        std::string getName();
+
+        virtual MeetsConstraint meetsConstraints(Napi::Object constraints)=0;
+        virtual bool applyConstraints(Napi::Object constraints)=0;
+        virtual void addAsCapability(Napi::Object target)=0;
+        virtual void addAsSetting(Napi::Object target)=0;
```
bghgary, line 39: nit:
```suggestion
        virtual MeetsConstraint meetsConstraints(Napi::Object constraints) = 0;
        virtual bool applyConstraints(Napi::Object constraints) = 0;
        virtual void addAsCapability(Napi::Object target) = 0;
        virtual void addAsSetting(Napi::Object target) = 0;
```

Plugins/NativeCamera/Source/Capability.h
```diff
@@ -0,0 +1,69 @@
...
+    class CameraCapabilityTemplate : public Capability
+    {
+    public:
+        CameraCapabilityTemplate(Feature feature,
+                                 T currentValue,
+                                 T defaultValue,
+                                 std::vector<T> acceptedValues,
+                                 std::function<bool(T)> setterFunction = [](T){return true;});
+
+        MeetsConstraint meetsConstraints(Napi::Object constraints);
```
bghgary, line 55: Functions should be UpperCamelCase, everywhere

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+#include "CameraDevice.h"
+#include "Constraint.h"
+#include <assert.h>
+
+namespace Babylon::Plugins
+{
+    arcana::task<Napi::Object, std::exception_ptr> MediaStream::NewAsync(Napi::Env env, Napi::Object constraints)
+    {
+        
+
```
bghgary, line 11: nit: 
```suggestion
```

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+    arcana::task<void, std::exception_ptr> MediaStream::ApplyInitialConstraintsAsync(Napi::Env env, Napi::Object constraints)
+    {
+        // A camera device hasn't been selected yet. Find the best device and open the stream
+        auto bestCamera = FindBestCameraStream(env, constraints);
+
+        // If m_cameraDevice is still null that means a camera device could not be found that meets the constraints
+        if (!bestCamera.has_value())
+        {
+            // If no device could be fount to satisfy the constraints throw a "ConstraintError" to match the browser implementation
+            return arcana::task_from_error<void>(std::make_exception_ptr(std::runtime_error{"ConstraintError: Unable to match constraints to a supported camera configuration."}));
```
bghgary, line 99:
```suggestion
            throw std::runtime_error{"ConstraintError: Unable to match constraints to a supported camera configuration."};
```

Plugins/NativeCamera/Source/MediaStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+            this->Width = cameraDimensions.width;
+            this->Height = cameraDimensions.height;
+            
+            if (!UpdateConstraints(env, constraintsRef.Value()))
+            {
+                // Setting the constraint to the capability failed
+                throw std::runtime_error{"ConstraintError: Unable to match constraints to a supported camera configuration."};
+            }
+            
+            return;
```
bghgary, line 118:
```suggestion
```

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -1,129 +1,54 @@
...
-                Napi::Function func = NativeCamera::DefineClass(
+            Napi::Function func = DefineClass(
                     env,
-                    JS_CLASS_NAME,
+                    "_NativeCamera",
                     {
-                        NativeCamera::InstanceMethod("createVideo", &NativeCamera::CreateVideo),
-                        NativeCamera::InstanceMethod("updateVideoTexture", &NativeCamera::UpdateVideoTexture),
+                            InstanceMethod("createVideo", &NativeCamera::CreateVideo),
+                            InstanceMethod("updateVideoTexture", &NativeCamera::UpdateVideoTexture),
```
bghgary, line 21: nit:
```suggestion
                        InstanceMethod("createVideo", &NativeCamera::CreateVideo),
                        InstanceMethod("updateVideoTexture", &NativeCamera::UpdateVideoTexture),
```

bghgary (approved):
Bunch of minor stuff, but overall looks good.

Plugins/NativeCamera/Source/NativeCamera.cpp
```diff
@@ -106,24 +97,49 @@ namespace Babylon
...
+            if (info.Length() > 0 && info[0].IsObject())
+            {
+                Napi::Object constraints{ info[0].As<Napi::Object>() };
+                if (constraints.Get("video").IsObject())
+                {
+                    videoConstraints = constraints.Get("video").As<Napi::Object>();
+                }
+            }
+
+            auto runtimeScheduler{ std::make_shared<JsRuntimeScheduler>(JsRuntime::GetFromJavaScript(env)) };
```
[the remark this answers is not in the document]
CoPrez: Per our offline conversation due to how the arcana task continuations work I am now storing the runtime as a unique_ptr and capturing it in the lambda via a move operator.

Plugins/NativeCamera/Source/MediaStream.h
```diff
@@ -0,0 +1,48 @@
...
+        
+        int Width{0};
+        int Height{0};
+        
+    private:
+        arcana::task<void, std::exception_ptr> ApplyInitialConstraints(Napi::Env env, Napi::Object constraints);
+        std::optional<std::pair<CameraDevice, const CameraTrack*>> FindBestCameraStream(Napi::Object constraints);
+        bool UpdateConstraints(Napi::Env env, Napi::Object constraints);
+        
+        std::shared_ptr<Plugins::Camera::Impl> m_cameraImpl;
```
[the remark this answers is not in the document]
CoPrez: I've added a comment for why it's a shared_ptr

Plugins/NativeCamera/Include/Babylon/Plugins/NativeCamera.h
```diff
@@ -18,12 +18,11 @@ namespace Babylon::Plugins
         // Initialization with overrideCameraTexture set to true means the caller
         // is expected to override the camera source texture with a native texture.
         // In that case, the HW camera will not be open.
-        static Camera Initialize(Napi::Env env, bool overrideCameraTexture = false);
+        static Camera Initialize(Napi::Env env);
```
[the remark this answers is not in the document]
CoPrez: Per our conversation since the Native API is still experimental I'm renaming it to NativeCmara to match the pattern used by other plugins

Plugins/NativeCamera/Source/Capability.h
CoPrez: Fixed all of the functions I could find.

Plugins/NativeCamera/Source/Android/CameraDevice.cpp
```diff
@@ -0,0 +1,515 @@
...
+                    GET_CAMERA_FUNCTION(ACameraCaptureSession_setRepeatingRequest)(impl->textureSession, &captureCallbacks, 1, &impl->request, nullptr);
+                    return true;
+                }
+            ));
+
+            cameraDeviceImpl->capabilities.emplace_back(std::make_unique<CameraCapabilityTemplate<double>>
+            (
+                    Capability::Feature::Zoom,
+                    zoomRatio,
+                    1.0, // Set the default target zoom to 1.0 (no zoom)
```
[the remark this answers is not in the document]
CoPrez: Grr Android Studio really likes to muck with the file formatting. I went through in VS code and cleaned up a few different whitespace issues I spotted.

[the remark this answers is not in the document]
bghgary: Yeah, it's annoying.

PR #1173, opened 2022-11-30
Render target texture improvements

bghgary:
See corresponding changes on the JS side: https://github.com/BabylonJS/Babylon.js/pull/13275

Note that there is PROTOCOL_VERSION change in both PRs.


SergioRZMasson (approved):
LGTM

PR #1174, opened 2022-12-01
Add dispatch function to ScriptLoader

bghgary:
This is useful if the native side needs to do work on the JS thread after loading scripts.


SergioRZMasson (approved):
LGTM

PR #1177, opened 2022-12-06
Remove extraneous bgfx flag

bghgary:
I verified that `BGFX_CONFIG_MEMORY_TRACKING` is still set in the code.


CedricGuillemet (approved):
LGTM! Memory is still checked as I can see this leak  `BGFX WARN 0x1ace24c RefCount is 4 (expected 0).` in the CI test logs.

PR #1178, opened 2022-12-06
Add assert for Texture::Disown to guard against incorrect usage.

bghgary:
Follow up to #1173 to guard against calling Disown when texture is already not owned.


SergioRZMasson (approved):
LGTM

PR #1183, opened 2022-12-21
Fixed issue when setting ViewPort multiple times per frame. 

### LINKED ISSUE #1182: Changing Viewport multiple times during rendering does not work properly
## Overview

The current viewport update logic does not work when viewport needs to be changed multiple times during a single frame. One example when that might happen is when multiple camera might be used to render to different parts of the screen. 

Since viewports get automatically set in native but rendering commands get issued by a command queue, viewport assignments gets out of sync during the rendering loop, resulting in only the last viewport set been applied. 

[From forum issue](https://forum.babylonjs.com/t/viewport-ignored-for-mutiple-cameras-in-react-native/36709/3)

**Other**
 - Platform: All

SergioRZMasson:
## Overview

This PR fixes #1182, it moves viewport update from been done immediately to be done using the command queue. The way this is currently been done on master, if multiple viewport updates are done in a single frame (like when you can more than one camera been rendered using different viewports) the viewport updates actually end up overriding each other and get lost by the time the command queue starts been processed. 

## Solution

In order to fix this, I'm proposing moving the setViewPort operation to be performed in the command queue. This way, this operation is going to happen in sync with the other rendering commands like  "FrameBuffer::Clear" (which also generates a ```bgfx::View```). I also had to modify ```BabylonNative::FrameBuffer``` since in some situations the ```FrameBuffer::SetViewPort``` gets called before ```FrameBuffer::Clear``` and sometimes it gets called after. In the cases where it gets called after, it is required that we store the clear information in the FrameBuffer object so we can call ```bgfx::setViewClear``` and that information does not get lost if a new ```bgfx::View``` gets created. 

This will also require a PR in Babylon.js to change how NativeEngine calls setViewPort. [Babylon.js PR](https://github.com/BabylonJS/Babylon.js/pull/13367)

## Extra Information

This functionality started as a request in the forum:
[Forum issue](https://forum.babylonjs.com/t/viewport-ignored-for-mutiple-cameras-in-react-native/36709/4)
[Target web behaviour](https://playground.babylonjs.com/#H8VW0B)


Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -7,10 +7,10 @@ namespace Babylon::Graphics
 {
     namespace
     {
-        void setDefaultClearMode(bgfx::ViewId viewId, bgfx::FrameBufferHandle handle)
+        void setDefaultClearMode(bgfx::ViewId viewId, bgfx::FrameBufferHandle handle, uint16_t flags, uint32_t rgba, float depth, uint8_t stencil)
```
bghgary, line 10: nit: this isn't your change, but the style is wrong.. same for `setViewPort`

```suggestion
        void SetDefaultClearMode(bgfx::ViewId viewId, bgfx::FrameBufferHandle handle, uint16_t flags, uint32_t rgba, float depth, uint8_t stencil)
```

Apps/Playground/Scripts/experience.js
```diff
@@ -24,6 +24,47 @@ function CreateBoxAsync(scene) {
     return Promise.resolve();
 }
 
+function CreateMultiCameraView(scene) {
```
bghgary, line 27: Is this change intended? Maybe we can create a playground for this and use validation tests?

bghgary (approved):
LGTM. We should test XR and other scenarios are still working.

Apps/Playground/Scripts/experience.js
SergioRZMasson: No, changed to experience.js were not intended. Just reverted it to match master.

SergioRZMasson:
Working fine on IOS XR.

Issue #1184, opened 2022-12-26
Dynamic Texture context.drawImage does not work with context.clip


SergioRZMasson:
## Overview

When using context.drawImage after a context.clip the image is not properly drawn using the curve previously set.  This bug was reported on the [forum](https://forum.babylonjs.com/t/context-clip-not-working-on-babylon-react-native-dynamic-texture-context/36825). 

**To Reproduce**
Using a dynamic texture construct a path than call context.clip and context.drawImage:

```
//Create path
textureContext.beginPath();
textureContext.moveTo(75*2, 25*2);
textureContext.quadraticCurveTo(25*2, 25*2, 25*2, 62.5*2);
textureContext.quadraticCurveTo(25*2, 100*2, 50*2, 100*2);
textureContext.quadraticCurveTo(50*2, 120*2, 30*2, 125*2);
textureContext.quadraticCurveTo(60*2, 120*2, 65*2, 100*2);
textureContext.quadraticCurveTo(125*2, 100*2, 125*2, 62.5*2);
textureContext.quadraticCurveTo(125*2, 25*2, 75*2, 25*2);
textureContext.strokeStyle = "white";
textureContext.stroke();

//Clip and draw image.
textureContext.clip();
textureContext.drawImage(this, 0, 0);
```

**Expected behavior**
Expected behavior should match the following playground: https://playground.babylonjs.com/#FU0ES5#25

thomlucc:
fixed

PR #1185, opened 2022-12-26
Fixed interaction between dynamic texture and context.clip()

SergioRZMasson:
## Overview 
Allows Dynamic Texture context ```context.drawImage``` ```context.clearRect``` and ```context.fillRect``` to properly work with ```context.clip```.

## Implementation

When calling ```Babylon::Polyfills::Internal::Context::DrawImage``` the ```nvgBeginPath``` is been called every time. This will discard any path that was previously set by the user. The correct behavior should be to not call  ```nvgBeginPath``` if ```Babylon::Polyfills::Internal::Context::Clip``` was previously called. 

Also, in ```Babylon::Polyfills::Internal::Context::Clip``` the ```m_rectangleClipping``` is been used to set the Scissor values even when  ```m_rectangleClipping``` was never set before (with default values 0, 0, 0, 0). This will make some of the following operations to produce no result. The correct behavior should be to only use ```m_rectangleClipping``` if the were set, otherwise use the width and height of the render target. 

Finally, when creating Babylon::Polyfills::Internal::NativeCanvasImage the resulting data might be in RGB instead of RGBA. However, during ```NativeCanvasImage::CreateNVGImageForContext``` NVG always assumes the format to be RGBA, which might cause a memory out of bounds read if the image was originally RGB. This PR forces ```bimg::imageParse``` to always convert the image to RGBA. 

## More information
Allows the following code to work in Babylon Native: https://playground.babylonjs.com/#FU0ES5#27
Fixes issue #1184


Apps/ValidationTests/Scripts/config.json
```diff
@@ -1,383 +1,389 @@
...
-            "replace": "__folder__, Texture_Sampler, __page__, 1, __disableSRGBBuffers__, 0",
-            "referenceImage": "gltfTextureSampler1.png"
-        },
-        {
-            "title": "GLTF Alien",
-            "playgroundId": "#XN37SR#5",
-            "referenceImage": "gltfAlien.png"
-        }
-    ]
+  "root": "https://cdn.babylonjs.com",
```
CedricGuillemet, line 2: Can you change the tab so it's easier to see the change?

azure-pipelines.yml
```diff
@@ -92,8 +92,8 @@ jobs:
     parameters:
       name: Ubuntu_Clang9_V8
```
CedricGuillemet, line 93: Don't forget to change the job name

Apps/ValidationTests/Scripts/config.json
```diff
@@ -1,383 +1,389 @@
...
+  "tests": [
+    {
+      "title": "Light Projection Texture",
```
CedricGuillemet, line 5: "Light Projection Texture" It's in the config.json 2 times. is it expected?

Apps/ValidationTests/Scripts/config.json
```diff
@@ -1,383 +1,389 @@
...
-            "referenceImage": "gltfTextureSampler1.png"
-        },
-        {
-            "title": "GLTF Alien",
-            "playgroundId": "#XN37SR#5",
-            "referenceImage": "gltfAlien.png"
-        }
-    ]
+  "root": "https://cdn.babylonjs.com",
+  "tests": [
```
CedricGuillemet, line 3: Can you please add a PG test for this feature? I'm not sure this Canvas feature is tested on bjs.

CedricGuillemet (changes requested):

azure-pipelines.yml
SergioRZMasson: CI changes were not intended for this PR, just reverted them.

Apps/ValidationTests/Scripts/config.json
SergioRZMasson: Reverted changes.

Apps/ValidationTests/Scripts/config.json
SergioRZMasson: @CedricGuillemet , just added a validation test to test clip() from canvas.

CedricGuillemet (approved):
LGTM!

PR #1187, opened 2022-12-27
Allow Dynamic Texture update method to consider the invertY argument. 

### LINKED ISSUE #1186: Dynamic Texture update does not consider invertY argument values
## Overview

When calling ```DynamicTexture.update``` with values given to the invertY argument nothing happens in Babylon Native. The invertY arguments get completely ignored by the NativeEngine. 

From forum issue: https://forum.babylonjs.com/t/react-native-dynamic-texture-update-inverty-parameter-is-not-working-as-expected/36916

**Describe the bug**
When calling ```DynamicTexture.update``` we always get the same results no matter if we give true or false to the ```invertY``` argument. 

**To Reproduce**
The following Playground does not render properly in Babylon Native: https://playground.babylonjs.com/#FU0ES5#32

**Expected behavior**
Texture should be flipped on the Y axis when invertY is set to true.

SergioRZMasson:
## Overview 

When calling DynamicTexture.update with values given to the invertY argument nothing happens in Babylon Native. The invertY arguments get completely ignored by the NativeEngine.

See #1186

## Implementation

I'm adding a code path where the CPU flips the texture in the Y axis. Maybe there is a better way to do this completely on the GPU but this implementation should be good enough for users to have the expected behavior while we improve it with a more performant version. 

Fixes #1186


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1092,13 +1092,59 @@ namespace Babylon
...
+
+        arcana::make_task(m_update.Scheduler(), *m_cancellationSource, [&context, this, flipY, textureDestination, textureSource, cancellationSource = m_cancellationSource]()
         {
-            return arcana::make_task(m_runtimeScheduler, *m_cancellationSource, [this, textureDestination, textureSource, updateToken = m_update.GetUpdateToken(), cancellationSource = m_cancellationSource]()
+            return arcana::make_task(m_runtimeScheduler, *m_cancellationSource, [&context, this, flipY, textureDestination, textureSource, updateToken = m_update.GetUpdateToken(), cancellationSource = m_cancellationSource]()
             {
                 bgfx::Encoder* encoder = m_update.GetUpdateToken().GetEncoder();
-                GetBoundFrameBuffer(*encoder).Blit(*encoder, textureDestination->Handle(), 0, 0, textureSource->Handle());
+
+                // Flip texture in memory (maybe there is a way for bgfx to allow flipping to happen on the GPU?)
```
bghgary, line 1105: Going to CPU seems like it's unnecessary. At worst case, we should be able to flip using a shader?

bghgary:
Closing for now. We can reopen once we are ready to work on the issue again.

Issue #1190, opened 2023-01-04
Building with V8 on Linux with curl on Ubuntu 22.04.1 cause curl to error with SSL errors


bghgary:
Repro:
1. Build and run `UnitTests` on Ubuntu 22.04.1 using V8.

Result:
UnitTests fail for many URL tests because curl easy apis are returning `CURLE_SSL_CACERT_BADFILE`. Same tests do not fail if targeting JavaScriptCore instead.

Expect:
UnitTests to succeed.

PR #1191, opened 2023-01-04
Fix CI after Ubuntu updated from 20.04 to 22.04

bghgary:
- Update Linux to use latest Clang/GCC compilers
- Use JavaScriptCore instead of V8 to work around #1190


CedricGuillemet (approved):

SergioRZMasson (commented):
LGTM

PR #1192, opened 2023-01-05
Task chain for texture copy/destruction

### LINKED ISSUE #1193: Texture destroyed while being used
Broken nightly:

a texture copy is asked from JS : https://github.com/BabylonJS/BabylonNative/blob/9c69a24801edae3996418ef771a09f39538a23a7/Plugins/NativeEngine/Source/NativeEngine.cpp#L1093
but between that call and this one (where the copy is actually performed) https://github.com/BabylonJS/BabylonNative/blob/9c69a24801edae3996418ef771a09f39538a23a7/Plugins/NativeEngine/Source/NativeEngine.cpp#L1101

the destination texture is destroyed. I won't discussed why the texture is destroyed, if it's necessary or not. the fact that we can run process related to textures and it can be destroyed before is problematic

I think it's the same as https://github.com/BabylonJS/BabylonNative/issues/810
Depending on the work and time it takes, resource is destroyed too early.

CedricGuillemet:
- Taskchain for texture copy/destruction so they are executed in order
- removed some canvas context methods that set it as dirty
- upgrade bjs version

fixes #1193

### DIFF
--- a/Apps/package-lock.json
+++ b/Apps/package-lock.json
@@ -8,10 +8,10 @@
       "name": "BabylonNative",
       "version": "0.0.1",
       "dependencies": {
-        "babylonjs": "^5.35.1",
-        "babylonjs-gui": "^5.35.1",
-        "babylonjs-loaders": "^5.35.1",
-        "babylonjs-materials": "^5.35.1",
+        "babylonjs": "^5.41.0",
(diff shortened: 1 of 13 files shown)


CedricGuillemet:
Closing this PR as it's not an issue for now and it needs more work.

PR #1197, opened 2023-01-12
Xcode/ios deployment update

CedricGuillemet:
xcode/ios deployment version update because of new VM images.
I tried to keep a broad enough range of ios/xcode version while not having to update it next week.


SergioRZMasson (approved):
LGTM

PR #1198, opened 2023-01-18
NativeCamera - Fix iOS crash and expose enumerateDevices

CoPrez:
This PR addresses two scenarios I observed where the NativeCamera implementation would crash on iOS.

 The first was if the camera was created and then quickly destroyed. That would lead to an exception being thrown inside of the UpdateCameraTexture function due to the object being destroyed while the async task was still pending for BeforeRenderScheduler(). The fix for that scenario was to add an arcana cancelation source that gets cancelled when the stream is closed and the async task captures it as a shared reference so it persists after the parent object is destroyed.

The second crash happened during engine shutdown occasionally. The issue is that CameraDevice uses ARC to manage the lifetime of the textureRGBA member variable. It then passes that texture handler to bgfx::overrideInternal which maintains a reference to the texture but doesn't increment the reference count on it. That can lead to a crash if the CameraDevice is disposed during or before a render loop and then bgfx tries to access the now destroyed texture handle. The fix for that scenario is when closing the CameraDevice it now keeps the reference to the textureRGBA member variable alive until AfterRenderScheduler.

Additionally this PR also adds a polyfill for the [MediaDevices::enumerateDevices()  web API](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices). That function can then be used to determine if the current device doesn't have any available camera's before attempting to open a stream. It could be updated in the future to also expose more info about the devices (such as a proper deviceID and label).

### DIFF
--- a/Plugins/NativeCamera/Source/Apple/CameraDevice.mm
+++ b/Plugins/NativeCamera/Source/Apple/CameraDevice.mm
@@ -424,12 +424,9 @@ namespace Babylon::Plugins
 
     void CameraDevice::UpdateCameraTexture(bgfx::TextureHandle textureHandle)
     {
+        // Hook into the BeforeRender loop to copy over the texture. Capture the cancellation token so that the shared pointer is kept alive when
+        // arcana checks internally for cancellation.
         arcana::make_task(m_impl->deviceContext->BeforeRenderScheduler(), *m_impl->cancellationSource, [this, textureHandle, cancellationSource{m_impl->cancellationSource}] {
-            if (cancellationSource->cancelled()) {
-                // The stream was closed while we were waiting to render.
-                return;
-            }
-
             id<MTLTexture> textureY{};
             id<MTLTexture> textureCbCr{};
             int64_t width{0};
--- a/Plugins/NativeCamera/Source/Apple/CameraDevice.mm
+++ b/Plugins/NativeCamera/Source/Apple/CameraDevice.mm
@@ -626,7 +626,7 @@ namespace Babylon::Plugins
     [self updateOrientation];
 #else
     // Orientation not supported on non-iOS devices. LandscapeLeft assumes the video is already in the correct orientation.
-    self->UIOrientation = UIInterfaceOrientationLandscapeLeft;
+    self->VideoOrientation = AVCaptureVideoOrientationLandscapeLeft;
 #endif
 
     return self;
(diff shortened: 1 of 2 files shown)


Alex-MSFT (approved):

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -423,7 +424,12 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
 
     void CameraDevice::UpdateCameraTexture(bgfx::TextureHandle textureHandle)
     {
-        arcana::make_task(m_impl->deviceContext->BeforeRenderScheduler(), arcana::cancellation::none(), [this, textureHandle] {
+        arcana::make_task(m_impl->deviceContext->BeforeRenderScheduler(), *m_impl->cancellationSource, [this, textureHandle, cancellationSource{m_impl->cancellationSource}] {
+            if (cancellationSource->cancelled()) {
+                // The stream was closed while we were waiting to render.
+                return;
+            }
```
ryantrem, line 431: Do we need this? @bghgary you looked at this part of the arcana code recently... do you remember if the cancellation token state is checked before invoking the make_task callback *on the actual dispatcher thread*?

CoPrez: I put a breakpoint inside the conditional and I couldn't get it to hit. So I'm inclined to believe it's indeed not required to check for cancelation at the beginning of the async work... but I also don't have a reliable way to ensure I'm actually reproducing the scenario that leads to it being canceled while waiting to render.

bghgary: I'm pretty sure this is already checked in the task continuation code. You should be able to see the check in the call stack if you put a breakpoint at the beginning of the continuation.

CoPrez: Arcana does indeed check the cancellation status internally. I've removed my own check, but I've left the capture for the cancellationToken to ensure it's kept alive if the host object is destroyed while the async work is still pending.

ryantrem (approved):

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
bghgary: Yes, this is a stopgap until my other changes set the coding pattern.

Alex-MSFT (approved):

PR #1203, opened 2023-02-07
NativeCamera - Add support for HTMLVideoElement resize event polyfill

CoPrez:
This PR updates the implementation of the NativeCamera plugin to support the HTMLVideoElement resize event. See [this doc](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/videoHeight#examples) for an example of how the web API works.

The main scenario where the video dimensions would change for the NativeCamera is when the device orientation changes. This issue is mentioned in more detail in issue #1107 

Exposing the resize event will allow the Babylon.JS VideoTexture to respond to the underlying video size changes and update the internal texture accordingly. This enables rotating the device while maintaining the same VideoTexture object and maintaining the same NativeCamera in conjunction with it. This becomes more important if you use the move advanced camera capabilities (such as zoom and torch).

### DIFF
--- a/Plugins/NativeCamera/Source/Android/CameraDevice.cpp
+++ b/Plugins/NativeCamera/Source/Android/CameraDevice.cpp
@@ -46,6 +46,26 @@ namespace Babylon::Plugins
             return oesTexture;
         }
 
+        int GetCurrentSensorRotationDiff() {
+            // Get the phone's current rotation so we can determine if the camera image needs to be rotated based on the sensor's natural orientation
+            int32_t phoneRotation{GetAppContext().getSystemService<android::view::WindowManager>().getDefaultDisplay().getRotation() * 90};
+
+            // The sensor rotation dictates the orientation of the camera when the phone is in it's default orientation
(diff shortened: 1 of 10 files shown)


ryantrem:
> Exposing the resize event will allow the Babylon.JS VideoTexture to respond to the underlying video size changes and update the internal texture accordingly.

That requires a separate change in Babylon.js, correct? This change makes is so it is possible to add that code to BJS and have it work in both the browser and native, yes?

CoPrez:
> > Exposing the resize event will allow the Babylon.JS VideoTexture to respond to the underlying video size changes and update the internal texture accordingly.
> 
> That requires a separate change in Babylon.js, correct? This change makes is so it is possible to add that code to BJS and have it work in both the browser and native, yes?

That is correct. This change doesn't functionally change any code that BJS uses today. I'm intending to send out a PR in BJS shortly to make VideoTexture respond to the resize event (that is already fired by the browser implementation today). The change in VideoTexture essentially involves disposing and re-creating the internal texture when resized.

ryantrem (approved):

PR #1206, opened 2023-02-28
Take hardware scaling level into account

bghgary:
When `engine.getHardwareScalingLevel()` is not 1, the frame buffer that is created and the renderWidth and renderHeight returned are the wrong size.


SergioRZMasson (approved):
LGTM

CedricGuillemet (approved):

PR #1209, opened 2023-03-03
Asynchronous shader compilation

### LINKED ISSUE #402: Enable asynchronous shader compilation
Per comments by @CedricGuillemet and @bghgary on #370, the potential for increasingly complex work required to traverse and transpile shaders under certain circumstances will likely justify moving shader transpilation to a thread pool to be done asynchronously. This work should be a relatively simple matter of combining patterns already established for mipping with the shader transpilation pipeline, but some adaptation may be required to ensure rendering calls are resilient to potential delays in the completion of asynchronous compilations.

glangstonb:
This code allows making shader compilation occur in a separate thread so that it doesn't block the Javascript thread. It's intended to be reverse-compatible with BabylonJS as of commit [52d30c49fffd2d9bd446668eb0b61337f7c167aa](https://github.com/BabylonJS/Babylon.js/commit/52d30c49fffd2d9bd446668eb0b61337f7c167aa) on master, but shaders will not compile asynchronously with that version; I will make another PR into BabylonJS with the necessary Typescript code to support asynchronous shader compilation using this code, and the shader compilation will be asynchronous only when both updates are used together.

Fixes #402

The PR for the BabylonJS side has now been opened:
https://github.com/BabylonJS/Babylon.js/pull/13587

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -686,7 +686,7 @@ namespace Babylon
         return vertexSource;
     }
 
-    void NativeEngine::CreateProgramInternal(const std::string vertexSource, const std::string fragmentSource, ProgramData& program)
+    bgfx::ProgramHandle NativeEngine::CreateProgramInternal(const std::string vertexSource, const std::string fragmentSource, ProgramData& program)
     {
         ShaderCompiler::BgfxShaderInfo shaderInfo = m_shaderCompiler.Compile(ProcessShaderCoordinates(vertexSource), ProcessSamplerFlip(fragmentSource));
 
@@ -715,7 +715,7 @@ namespace Babylon
         auto fragmentShader = bgfx::createShader(bgfx::copy(shaderInfo.FragmentBytes.data(), static_cast<uint32_t>(shaderInfo.FragmentBytes.size())));
         InitUniformInfos(fragmentShader, shaderInfo.UniformStages, program.UniformInfos, program.UniformNameToIndex);
 
-        program.Handle = bgfx::createProgram(vertexShader, fragmentShader, true);
+        return bgfx::createProgram(vertexShader, fragmentShader, true);
     }
 
     Napi::Value NativeEngine::CreateProgram(const Napi::CallbackInfo& info)
@@ -725,7 +725,7 @@ namespace Babylon
         ProgramData* program = new ProgramData{};
         try
         {
-            CreateProgramInternal(vertexSource, fragmentSource, *program);
+            program->Handle = CreateProgramInternal(vertexSource, fragmentSource, *program);
         }
         catch (const std::exception& ex)
         {
@@ -742,11 +742,11 @@ namespace Babylon
         const Napi::Function onError = info[3].As<Napi::Function>();
         ProgramData* program = new ProgramData{};
         arcana::make_task(arcana::threadpool_scheduler, *m_cancellationSource,
-            [this, vertexSource, fragmentSource, program, cancellationSource{m_cancellationSource}]()
+            [this, vertexSource, fragmentSource, program, cancellationSource{m_cancellationSource}]() -> bgfx::ProgramHandle
             {
-                CreateProgramInternal(vertexSource, fragmentSource, *program);
+                return CreateProgramInternal(vertexSource, fragmentSource, *program);
             })
-            .then(m_runtimeScheduler, *m_cancellationSource, [onSuccessRef{Napi::Persistent(onSuccess)}, onErrorRef{Napi::Persistent(onError)}, cancellationSource{m_cancellationSource}](arcana::expected<void, std::exception_ptr> result)
+            .then(m_runtimeScheduler, *m_cancellationSource, [program, onSuccessRef{Napi::Persistent(onSuccess)}, onErrorRef{Napi::Persistent(onError)}, cancellationSource{m_cancellationSource}](arcana::expected<bgfx::ProgramHandle, std::exception_ptr> result)
                 {
                 if (result.has_error())
                 {
@@ -754,6 +754,7 @@ namespace Babylon
                 }
                 else
                 {
+                    program->Handle = result.value();
                     onSuccessRef.Call({});
                 } });
         return Napi::Pointer<ProgramData>::Create(info.Env(), program, Napi::NapiPointerDeleter(program));
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -686,11 +686,11 @@ namespace Babylon
         return vertexSource;
     }
 
-    ProgramData NativeEngine::CreateProgramInternal(const std::string vertexSource, const std::string fragmentSource)
+    std::unique_ptr<ProgramData> NativeEngine::CreateProgramInternal(const std::string vertexSource, const std::string fragmentSource)
     {
         ShaderCompiler::BgfxShaderInfo shaderInfo = m_shaderCompiler.Compile(ProcessShaderCoordinates(vertexSource), ProcessSamplerFlip(fragmentSource));
 
-        ProgramData program;
+        std::unique_ptr<ProgramData> program = std::make_unique<ProgramData>();
 
         static auto InitUniformInfos{
             [](bgfx::ShaderHandle shader, const std::unordered_map<std::string, uint8_t>& uniformStages, std::unordered_map<uint16_t, UniformInfo>& uniformInfos, std::unordered_map<std::string, uint16_t>& uniformNameToIndex) {
@@ -711,13 +711,13 @@ namespace Babylon
             }};
 
         auto vertexShader = bgfx::createShader(bgfx::copy(shaderInfo.VertexBytes.data(), static_cast<uint32_t>(shaderInfo.VertexBytes.size())));
-        InitUniformInfos(vertexShader, shaderInfo.UniformStages, program.UniformInfos, program.UniformNameToIndex);
+        InitUniformInfos(vertexShader, shaderInfo.UniformStages, program->UniformInfos, program->UniformNameToIndex);
 
         auto fragmentShader = bgfx::createShader(bgfx::copy(shaderInfo.FragmentBytes.data(), static_cast<uint32_t>(shaderInfo.FragmentBytes.size())));
-        InitUniformInfos(fragmentShader, shaderInfo.UniformStages, program.UniformInfos, program.UniformNameToIndex);
+        InitUniformInfos(fragmentShader, shaderInfo.UniformStages, program->UniformInfos, program->UniformNameToIndex);
 
-        program.Handle = bgfx::createProgram(vertexShader, fragmentShader, true);
-        program.VertexAttributeLocations = std::move(shaderInfo.VertexAttributeLocations);
+        program->Handle = bgfx::createProgram(vertexShader, fragmentShader, true);
+        program->VertexAttributeLocations = std::move(shaderInfo.VertexAttributeLocations);
 
         return program;
     }
@@ -729,7 +729,7 @@ namespace Babylon
         ProgramData* program = new ProgramData{};
         try
         {
-            *program = CreateProgramInternal(vertexSource, fragmentSource);
+            *program = std::move(*CreateProgramInternal(vertexSource, fragmentSource).release());
         }
         catch (const std::exception& ex)
         {
@@ -750,7 +750,7 @@ namespace Babylon
         Napi::Value jsProgram = Napi::Pointer<ProgramData>::Create(info.Env(), program, Napi::NapiPointerDeleter(program));
 
         arcana::make_task(arcana::threadpool_scheduler, *m_cancellationSource,
-            [this, vertexSource, fragmentSource, program, cancellationSource{m_cancellationSource}]() -> ProgramData
+            [this, vertexSource, fragmentSource, program, cancellationSource{m_cancellationSource}]() -> std::unique_ptr<ProgramData>
             {
                 return CreateProgramInternal(vertexSource, fragmentSource);
             })
@@ -759,7 +759,7 @@ namespace Babylon
                 jsProgramRef{Napi::Persistent(jsProgram)},
                 onSuccessRef{Napi::Persistent(onSuccess)},
                 onErrorRef{Napi::Persistent(onError)},
-                cancellationSource{m_cancellationSource}](arcana::expected<ProgramData, std::exception_ptr> result)
+                cancellationSource{m_cancellationSource}](const arcana::expected<std::unique_ptr<ProgramData>, std::exception_ptr>& result)
                 {
                     if (result.has_error())
                     {
@@ -767,7 +767,7 @@ namespace Babylon
                     }
                     else
                     {
-                        *program = std::move(result.value());
+                        *program = std::move(*result.value());
                         onSuccessRef.Call({});
                     }
                 });
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -729,7 +729,7 @@ namespace Babylon
         ProgramData* program = new ProgramData{};
         try
         {
-            *program = std::move(*CreateProgramInternal(vertexSource, fragmentSource).release());
+            *program = std::move(*CreateProgramInternal(vertexSource, fragmentSource));
         }
         catch (const std::exception& ex)
         {
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -756,10 +756,10 @@ namespace Babylon
             })
             .then(m_runtimeScheduler, *m_cancellationSource,
                 [program,
-                jsProgramRef{Napi::Persistent(jsProgram)},
-                onSuccessRef{Napi::Persistent(onSuccess)},
-                onErrorRef{Napi::Persistent(onError)},
-                cancellationSource{m_cancellationSource}](const arcana::expected<std::unique_ptr<ProgramData>, std::exception_ptr>& result)
+                    jsProgramRef{Napi::Persistent(jsProgram)},
+                    onSuccessRef{Napi::Persistent(onSuccess)},
+                    onErrorRef{Napi::Persistent(onError)},
+                    cancellationSource{m_cancellationSource}](const arcana::expected<std::unique_ptr<ProgramData>, std::exception_ptr>& result)
                 {
                     if (result.has_error())
                     {
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -727,16 +727,16 @@ namespace Babylon
         const std::string vertexSource = info[0].As<Napi::String>().Utf8Value();
         const std::string fragmentSource = info[1].As<Napi::String>().Utf8Value();
         ProgramData* program = new ProgramData{};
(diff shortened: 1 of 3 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -721,13 +709,54 @@ namespace Babylon
             }};
 
         auto vertexShader = bgfx::createShader(bgfx::copy(shaderInfo.VertexBytes.data(), static_cast<uint32_t>(shaderInfo.VertexBytes.size())));
-        InitUniformInfos(vertexShader, shaderInfo.UniformStages, program->UniformInfos, program->UniformNameToIndex);
-        program->VertexAttributeLocations = std::move(shaderInfo.VertexAttributeLocations);
+        InitUniformInfos(vertexShader, shaderInfo.UniformStages, program.UniformInfos, program.UniformNameToIndex);
+        program.VertexAttributeLocations = std::move(shaderInfo.VertexAttributeLocations);
```
bghgary, line 713: To do this properly, ProgramData should not be passed in since we only want ProgramData to be accessed from the JS thread. This function should return both the bgfx::ProgramHandle and a VertexAttributeLocations map.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -721,13 +709,54 @@ namespace Babylon
...
+
+        return bgfx::createProgram(vertexShader, fragmentShader, true);
+    }
 
-        program->Handle = bgfx::createProgram(vertexShader, fragmentShader, true);
+    Napi::Value NativeEngine::CreateProgram(const Napi::CallbackInfo& info)
+    {
+        const std::string vertexSource = info[0].As<Napi::String>().Utf8Value();
+        const std::string fragmentSource = info[1].As<Napi::String>().Utf8Value();
+        ProgramData* program = new ProgramData{};
```
bghgary, line 725: This will leak memory if CreateProgramInternal throws. If we make it such that CreateProgramInternal doesn't know about ProgramData, then we can create the ProgramData right before the Napi::Pointer which should fix it.

bghgary (changes requested):
Looks perfect except for the ProgramData VertexAttributeLocations issue.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: Creating the `ProgramData` later means `CreateProgramInternal` wouldn't be able to access it. Instead, I've set it to delete `program` before throwing in `CreateProgram`. Hopefully in `CreateProgramAsync` no delete is needed because it does actually return even on a compile error. Please let me know if this is wrong.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: I've changed this, the test seems to work but please let me know if there's something wrong with the new code.

bghgary: Sorry, the same issue for UniformInfos and UniformNameToIndex. Basically, we want to avoid accessing things in ProgramData unless we own ProgramData. Another option is to make ProgramData movable so that CreateProgramInternal returns a ProgramData and then the outer function can move the ProgramData into the pointer version.

This is more for cleanliness than correctness. It should also address the other issue.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -742,21 +744,31 @@ namespace Babylon
...
+                    if (result.has_error())
+                    {
+                        try
+                        {
+                            std::rethrow_exception(result.error());
+                        }
+                        catch (const std::exception& ex)
+                        {
+                            onErrorRef.Call({Napi::Error::New(onErrorRef.Env(), ex.what()).Value()});
+                        }
```
bghgary, line 762: `Error:New` has an overload that takes `exception_ptr`. 
```suggestion
                        onErrorRef.Call({Napi::Error::New(onErrorRef.Env(), result.error()).Value()});
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -742,21 +744,31 @@ namespace Babylon
...
+                    }
+                    else
+                    {
+                        std::pair<bgfx::ProgramHandle, std::unordered_map<std::string, uint32_t>> programInfo = result.value();
```
bghgary, line 766: This will result in a copy.

```suggestion
                        auto& programInfo = result.value();
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -721,13 +709,66 @@ namespace Babylon
...
+                        program->Handle = programInfo.first;
+                        program->VertexAttributeLocations = std::move(programInfo.second);
```
glangstonb, line 768: @bghgary, I don't think you mentioned this one, but I'm a little worried, is it dangerous to do this with the `program` pointer? Given that this can happen after `CreateProgramAsync` returns, what happens if the `NapiPointerDeleter` deletes it before this runs? (Sorry if I'm misunderstanding the functionality of these APIs.)

bghgary: This is a good callout. Without the correct pattern for shutdown, it can indeed get called after the JS object is deleted which in turn deletes the program. If you want to find out more about how the new pattern works, see [this PR](https://github.com/BabylonJS/BabylonNative/pull/1141), specifically [this comment](https://github.com/BabylonJS/BabylonNative/pull/1141/files#diff-a4dbd5b4fd890908e7876413e36816689a92f3a30cd8260147d840604dba29b5R10-R55).

bghgary: Also, note that NapiPointerDeleter only deletes the object when garbage collected. It will not be deleted until the JS side lets go of the JS object, which shouldn't happen until `DeleteProgram` is called. This NapiPointerDeleter is a safe guard in case the JS engine goes down before the engine is disposed.

glangstonb: I briefly read the file you linked to above, but it's not clear to me how it applies to what I'm doing here. What I'm doing still feels unsafe (regardless of the behavior of the BGFX resource, just dereferencing the deleted pointer at all sounds bad), unless you're saying I can rely on the `NapiPointerDeleter` not to run on this pointer until the Javascript environment shuts down, but that's actually not something I would want to rely on even if I could, because of memory leaks.

Just now I looked at the actual code behind `NapiPointerDeleter` and it's simpler than I thought. Would it be okay to replace that with a different callback that only conditionally deletes the pointer? I might be able to make it safe that way, but it sounds a bit convoluted and maybe I'm missing some API already in here that is meant for this sort of thing.

bghgary: > unless you're saying I can rely on the NapiPointerDeleter not to run on this pointer until the Javascript environment shuts down

You can rely on the JS side holding onto the resulting JS object until `DeleteProgram` is called. If `DeleteProgram` doesn't get called, then the NapiPointerDeleter is here to make sure it is still deleted.

glangstonb: That makes sense, but I'm concerned about what happens if it goes the other way around: While the shader is compiling (which might take a few frames) the Javascript side stops using that object and decides to GC it, `NapiPointerDeleter` gets called (or at least I assume that's what it's there for), and *then* the C++ side comes back with a successful compilation and tries to assign `program->Handle` etc onto (I assume) a pointer to invalid memory.

bghgary: Right, the coding pattern I mentioned above will prevent the JavaScript object from being destroyed because the continuation will capture a strong ref to the object. This can't be done until my PR is merged, so you don't have to worry about it here.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: I pushed the changes we discussed (making `ProgramData` movable) and hopefully that addresses this issue.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: I pushed the changes we discussed (creating the data inside `CreateProgramInternal` and putting it into a `unique_ptr`). I gather that the `unique_ptr` will be destructed if `CreateProgramInternal` hits an exception before returning, so hopefully there's no more memory leak with this.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: Sorry, I had no idea that API was there! 😰 I've pushed the change to do it this way.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: I pushed the changes we discussed (using `std::move` out of the result value) and hopefully that addresses this issue.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: I've pushed the changes we discussed, please let me know if there's anything wrong with the new version.

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -43,11 +43,26 @@ namespace Babylon
...
         ProgramData() = default;
-        ProgramData(ProgramData&& other) = delete;
         ProgramData(const ProgramData&) = delete;
-        ProgramData& operator=(ProgramData&& other) = delete;
         ProgramData& operator=(const ProgramData& other) = delete;
 
+        ProgramData(ProgramData&& other) noexcept
+        {
+            (*this) = std::move(other);
+        }
```
bghgary, line 52: Not using constructor initialization will cause the fields to be default constructed for no reason.

```suggestion
        ProgramData(ProgramData&& other) noexcept :
            Handle{other.Handle},
            Uniforms{std::move(other.Uniforms)},
            UniformToIndex{std::move(other.UniformToIndex)},
            UniformInfos{std::move(other.UniformInfos)},
            VertexAttributeLocations{std::move(other.VertexAttributeLocations)},
        {
            other.Handle = BGFX_INVALID_HANDLE;
        }
```

I also think we should remove the Disposed flag. It's the same as Handle being invalid.

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -43,11 +43,26 @@ namespace Babylon
...
+        ProgramData& operator=(ProgramData&& other) noexcept
+        {
+            Handle = std::move(other.Handle);
+            other.Handle = bgfx::ProgramHandle{bgfx::kInvalidHandle};
+            Disposed = std::move(other.Disposed);
+            Uniforms = std::move(other.Uniforms);
+            UniformNameToIndex = std::move(other.UniformNameToIndex);
+            UniformInfos = std::move(other.UniformInfos);
+            VertexAttributeLocations = std::move(other.VertexAttributeLocations);
+            return *this;
```
bghgary, line 63: Assuming we remove `Disposed`:
```suggestion
            Handle = other.Handle;
            other.Handle = BGFX_INVALID_HANDLE;
            Uniforms = std::move(other.Uniforms);
            UniformNameToIndex = std::move(other.UniformNameToIndex);
            UniformInfos = std::move(other.UniformInfos);
            VertexAttributeLocations = std::move(other.VertexAttributeLocations);
            return *this;
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -724,9 +729,7 @@ namespace Babylon
         ProgramData* program = new ProgramData{};
         try
         {
-            std::pair<bgfx::ProgramHandle, std::unordered_map<std::string, uint32_t>> programInfo = CreateProgramInternal(vertexSource, fragmentSource, *program);
-            program->Handle = programInfo.first;
-            program->VertexAttributeLocations = std::move(programInfo.second);
+            *program = std::move(*CreateProgramInternal(vertexSource, fragmentSource).release());
```
bghgary, line 732: If you use the same pattern as the async version, you shouldn't have to call `release`.

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -43,11 +43,26 @@ namespace Babylon
     struct ProgramData final
     {
         ProgramData() = default;
-        ProgramData(ProgramData&& other) = delete;
         ProgramData(const ProgramData&) = delete;
-        ProgramData& operator=(ProgramData&& other) = delete;
         ProgramData& operator=(const ProgramData& other) = delete;
```
bghgary, line 47: Just to be consistent

```suggestion
        ProgramData& operator=(const ProgramData&) = delete;
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -742,34 +745,34 @@ namespace Babylon
...
-                return CreateProgramInternal(vertexSource, fragmentSource, *program);
+                return CreateProgramInternal(vertexSource, fragmentSource);
             })
-            .then(m_runtimeScheduler, *m_cancellationSource, [program, onSuccessRef{Napi::Persistent(onSuccess)}, onErrorRef{Napi::Persistent(onError)}, cancellationSource{m_cancellationSource}](arcana::expected<std::pair<bgfx::ProgramHandle, std::unordered_map<std::string, uint32_t>>, std::exception_ptr> result)
+            .then(m_runtimeScheduler, *m_cancellationSource,
+                [program,
+                jsProgramRef{Napi::Persistent(jsProgram)},
+                onSuccessRef{Napi::Persistent(onSuccess)},
+                onErrorRef{Napi::Persistent(onError)},
+                cancellationSource{m_cancellationSource}](const arcana::expected<std::unique_ptr<ProgramData>, std::exception_ptr>& result)
```
bghgary, line 762: We use clang-format. VS should use the `.clang-format` file at the root of the repo if configured properly. Please format accordingly.

bghgary (commented):
Really close now.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -722,15 +712,69 @@ namespace Babylon
...
+        const std::string vertexSource = info[0].As<Napi::String>().Utf8Value();
+        const std::string fragmentSource = info[1].As<Napi::String>().Utf8Value();
+        const Napi::Function onSuccess = info[2].As<Napi::Function>();
+        const Napi::Function onError = info[3].As<Napi::Function>();
+
+        ProgramData* program = new ProgramData{};
+        Napi::Value jsProgram = Napi::Pointer<ProgramData>::Create(info.Env(), program, Napi::NapiPointerDeleter(program));
+
+        arcana::make_task(arcana::threadpool_scheduler, *m_cancellationSource,
+            [this, vertexSource, fragmentSource, program, cancellationSource{m_cancellationSource}]() -> std::unique_ptr<ProgramData>
```
bghgary, line 753: clang caught this in the build. `program` is not used.

```suggestion
            [this, vertexSource, fragmentSource, cancellationSource{m_cancellationSource}]() -> std::unique_ptr<ProgramData>
```

Plugins/NativeEngine/Source/NativeEngine.h
glangstonb: Currently, `Dispose` checks whether `Handle` is valid, but doesn't actually change it when disposing the BGFX resource. I assume that leaves it in a state where `Disposed` is `true` but `Handle` is still valid. Should `Dispose` set `Handle` to `bgfx::kInvalidHandle` when successfully disposing the BGFX resource? Or might there be a reason why we want to remember which program handle was used even after it is disposed?

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: So can I just remove `.release()` from this code? The pattern already looks basically the same except for that.

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: Good catch. Where do you see a warning about this when building? I checked my build logs and didn't see anything about this.

Plugins/NativeEngine/Source/NativeEngine.h
bghgary: I don't think we need to keep the Handle with the previous value. It might be useful for debugging at some point, but I've never heard of anyone trying to use it before.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: Not exactly. Because the try/catch below can throw which will bypass the js object owning the pointer. We need to move the creation of the js object right after `new ProgramData`, just like the async version.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: If you are building on Windows, it uses MSVC which doesn't catch this. clang is used for other platforms.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: Make sure your VS formatting settings are set to use clang-format.

image

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: Sorry, I missed the delete call in the catch. What you have should work, but the `delete` would be unnecessary if we use the pattern similar to the async code.

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -44,18 +44,22 @@ namespace Babylon
...
+
+        ProgramData(ProgramData&& other) noexcept:
+            Handle{other.Handle},
+            Uniforms{std::move(other.Uniforms)},
+            UniformNameToIndex{std::move(other.UniformNameToIndex)},
+            UniformInfos{std::move(other.UniformInfos)},
+            VertexAttributeLocations{std::move(other.VertexAttributeLocations)}
         {
-            (*this) = std::move(other);
+            other.Handle = bgfx::ProgramHandle{bgfx::kInvalidHandle};
```
bghgary, line 56: nit:
```suggestion
            other.Handle = BGFX_INVALID_HANDLE;
```

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -44,18 +44,22 @@ namespace Babylon
...
         }
 
         ProgramData& operator=(ProgramData&& other) noexcept
         {
             Handle = std::move(other.Handle);
             other.Handle = bgfx::ProgramHandle{bgfx::kInvalidHandle};
```
bghgary, line 62: nit:
```suggestion
            other.Handle = BGFX_INVALID_HANDLE;
```

Plugins/NativeEngine/Source/NativeEngine.cpp
glangstonb: It was already turned on. For whatever reason it wasn't changing this code on save. No idea why, but I hope it's the way you want it now.

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -74,15 +74,14 @@ namespace Babylon
 
         void Dispose()
         {
-            if (!Disposed && bgfx::isValid(Handle))
+            if (bgfx::isValid(Handle))
             {
                 bgfx::destroy(Handle);
+                Handle = bgfx::ProgramHandle{bgfx::kInvalidHandle};
```
bghgary, line 80:
```suggestion
                Handle = BGFX_INVALID_HANDLE;
```

bghgary (approved):

PR #1218, opened 2023-03-24
Fixed viewport issue when renderning with multiple render targets.

SergioRZMasson:
There was a difference in behavior between Babylon Native and the web regarding the handling of viewports and render targets. 

The Bind method of the FrameBuffer native class was resetting the view state, which is incorrect. However, that was introduced to fix a difference in the way WebGL clears the render target. 

In webGL the viewport has no effect when we clear a render target, WebGL assumes that we always want to clear the entire render target no matter what the viewport values are. I've changed the Clear method on the native FrameBuffer class to clear the entire screen so we can have the correct behavior for clear and we can remove the state reset from Bind. 

I've also added a couple of tests for validating that those work.

### DIFF
--- a/Core/Graphics/Source/FrameBuffer.cpp
+++ b/Core/Graphics/Source/FrameBuffer.cpp
@@ -53,15 +53,15 @@ namespace Babylon::Graphics
 
     void FrameBuffer::Clear(bgfx::Encoder& encoder, uint16_t flags, uint32_t rgba, float depth, uint8_t stencil)
     {
-        //BGFX requires us to create a new viewID, this will ensure that the view gets cleaned. 
+        // BGFX requires us to create a new viewID, this will ensure that the view gets cleaned.
         m_viewId = m_context.AcquireNewViewId(encoder);
 
         bgfx::setViewMode(m_viewId.value(), bgfx::ViewMode::Sequential);
         bgfx::setViewClear(m_viewId.value(), flags, rgba, depth, stencil);
         bgfx::setViewFrameBuffer(m_viewId.value(), m_handle);
 
-        //BGFX will consider the viewport when cleaning the screen, but WebGL always cleans the entire screen. 
-        //That's why we always set the viewport to {0, 0, 1, 1} when cleaning.
+        // BGFX will consider the viewport when cleaning the screen, but WebGL always cleans the entire screen.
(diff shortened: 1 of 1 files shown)


Apps/ValidationTests/Scripts/config.json
```diff
@@ -1,6 +1,6 @@
 {
     "root": "https://cdn.babylonjs.com",
-    "tests": [
+    "tests": [
```
docEdub, line 3: extra white space

docEdub: 👍 LGTM other than that

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -159,6 +103,27 @@ namespace Babylon::Graphics
...
+        if (m_viewId.has_value() && viewState.Equals(m_bgfxViewPort))
+        {
+            return;
+        }
+
+        m_viewId = m_context.AcquireNewViewId(encoder);
+
+        m_bgfxViewPort = viewState;
+
+        bgfx::setViewMode(m_viewId.value(), bgfx::ViewMode::Sequential);
```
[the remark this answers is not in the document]
SergioRZMasson: We haven't changed this, this is the way it Babylon Native has always set views to behave.

Apps/ValidationTests/Scripts/config.json
SergioRZMasson: Removed the white spaces.

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -82,61 +53,34 @@ namespace Babylon::Graphics
 
     void FrameBuffer::Clear(bgfx::Encoder& encoder, uint16_t flags, uint32_t rgba, float depth, uint8_t stencil)
     {
-        m_flags = flags;
-        m_rgba = rgba;
-        m_depth = depth;
-        m_stencil = stencil;
+        m_viewId = m_context.AcquireNewViewId(encoder);
```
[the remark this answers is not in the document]
SergioRZMasson: Added  comments explaining the clean logic.

bghgary (approved):

PR #1221, opened 2023-03-31
Update the graphics contract to support passing in a custom device

bghgary:
This change updates the graphics contract such that the graphics configuration is merged into one.

I found out recently that bgfx still creates a graphics device when a window is passed in. I think everyone thought that bgfx was somehow getting the graphics device from the window, but it seems this isn't possible, at least with XamlSwapChains. If someone wanted to use their own graphics device with a window (e.g., create a multi-threaded D3D device), the current contract doesn't allow this. This is the main reason for this change.

Another reason is that the MSAA and premultiplied-alpha flags apply for both when there is a window and when there is not (at least for D3D11). The only time that these flags don't apply is when there is a given back buffer, but we don't currently support this.

### DIFF
--- a/Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
+++ b/Apps/Playground/Android/app/src/main/cpp/BabylonNativeJNI.cpp
@@ -151,7 +151,7 @@ extern "C"
         if (runtime)
         {
             ANativeWindow* window = ANativeWindow_fromSurface(env, surface);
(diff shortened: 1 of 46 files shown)


ryantrem (approved):

SergioRZMasson (approved):
LGTM

bghgary:
Not merging until I confirm this works with BabylonReactNative.

bghgary:
I've confirmed it's working in BabylonReactNative.

PR #1226, opened 2023-04-20
Build and code cleanup

bghgary:
- Separate `JsRuntimeDispatcher` into its own target `JsRuntimeInternal` as it will depend on arcana in a future PR.
- Make the build style more consistent across different CMakeLists.txt


CedricGuillemet (approved):
LGTM

PR #1228, opened 2023-05-11
Fix ninja build when using gcc or clang

TomFalkner:
Fixes issue where running ninja after building with either gcc or clang on Ubuntu would throw the following error:

‘class std::basic_ostream<char>’ has no member named ‘str’

Related discussion:
https://stackoverflow.com/questions/59473325/no-member-named-str-in-stdbasic-ostreamchar-with-gcc-and-clang-but-no-p

Tested with
gcc 9.4.0
clang 10.0.0-4ubuntu1

### DIFF
--- a/Dependencies/UrlLib/Source/UrlRequest_Unix.cpp
+++ b/Dependencies/UrlLib/Source/UrlRequest_Unix.cpp
@@ -12,7 +12,9 @@ namespace
     {
         if (code != CURLE_OK)
         {
-            throw std::runtime_error{(std::stringstream{} << "CURL call failed with code (" << code << ")").str()};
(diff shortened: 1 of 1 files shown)


TomFalkner:
@microsoft-github-policy-service agree company="Spectrum"

bghgary:
Thanks @TomFalkner for the PR. I've noticed this code doesn't build in older versions of the compiler when I made these changes originally. If we use newer versions, it will compile just fine. We have tests that build with various compilers. Do you have a requirement to build with an old version of GCC / Clang?

bghgary:
Closing due to no activity. Please reopen if you still need this change.

PR #1231, opened 2023-05-24
5 more GLTF Tests

CedricGuillemet:


Apps/ValidationTests/ReferenceImages/gltfTextureLinearInterpolation.png
docEdub, line 1: Should the sphere on the right be darker?

This is how it looks at https://playground.babylonjs.com/#WGZLGJ#3009 ...
image

CedricGuillemet: Ho! good catch @docEdub !!!

bghgary: They should look the same.

bghgary: https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/TextureLinearInterpolationTest/README.md

docEdub: So, it's wrong in the JS playground then? Interesting.

CedricGuillemet: After talking with Alexis: rendering is correct with Native (and WebGPU) but is not with webgl.

bghgary: Is this expected or is there an issue with webgl?

CedricGuillemet: issue with webgl because of compatibility with some browsers. there is also a flag to force srgb in that situation: https://github.com/BabylonJS/Babylon.js/blob/9c1a1d7d464bcd100b188a4b098653bc68259162/packages/dev/core/src/Engines/thinEngine.ts#L193

bghgary: Yes, I'm aware of this, but the test should force this flag on. There is no reason for the test otherwise.

bghgary (approved):
Thanks for updating! Tests are failing though :-(

PR #1234, opened 2023-05-29
Test Install

### LINKED ISSUE #1067: Add CI tests for the install target
Currently we have an install target that bundles the include and libs into a nice clean folder, but it is very easy to do this wrong (e.g., missing libs, etc.) This can be mitigated by adding some tests in CI that consume the output of the install and try to build.

CedricGuillemet:
fixes #1067

- Installation is a bit weird with openxr_loader (.lib copied 2 times)
- lib directory contains suffixed library with 'd' for some libraries, not all. It means, it's impossible to have debug and release. 

What platforms should be tested?

### DIFF
--- a/.github/jobs/win32.yml
+++ b/.github/jobs/win32.yml
@@ -69,7 +69,7 @@ jobs:
       - script: |
           cmake --build build${{ variables.solutionName }} --target INSTALL --config RelWithDebInfo
           cd Install/Test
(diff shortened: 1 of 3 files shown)


CedricGuillemet:
Back to draft. After discussing with @bghgary , I think it should be tested with cpp unittests instead of simple .cpp. Will get back to this PR in the coming weeks.

CedricGuillemet:
Better version in https://github.com/BabylonJS/BabylonNative/pull/1260 Closing.

PR #1237, opened 2023-06-05
Merge PG and VT

CedricGuillemet:


Apps/Playground/Scripts/ReferenceImages/Billboard.png
bghgary, line 1: Seems like you added the ReferenceImages twice?

bghgary (approved):
I guess this is fine for a first step. We should think about how to organize the files inside the Playground app.

Apps/Playground/Scripts/ReferenceImages/Billboard.png
CedricGuillemet: Indeed! Fixed now

CedricGuillemet:
> I guess this is fine for a first step. We should think about how to organize the files inside the Playground app.

Yes, I'm not happy with the way ref images are handled. It's part of PG and it should be a separate option.

PR #1243, opened 2023-06-14
Implement scissor

docEdub:
Implements scissors. See https://github.com/BabylonJS/Babylon.js/pull/13960 for changes required on the JS side.


Apps/Playground/Scripts/config.json
```diff
@@ -206,6 +206,17 @@
...
+            "title": "Multi camera rendering with scissors",
+            "playgroundId": "#1LK70I#35",
+            "referenceImage": "multiCameraRenderingWithScissors.png"
+        },
+        {
+            "title": "Multi cameras and output render target with scissors",
+            "renderCount": 2,
+            "playgroundId": "#BCYE7J#45",
+            "referenceImage": "multiCamerasOutputRenderTargetWithScissors.png"
+        },
```
bghgary, line 219: Are these native specific? I don't think we have any visual tests that are native specific. Ideally, there shouldn't be any.

Apps/Playground/Win32/App.cpp
```diff
@@ -131,6 +131,8 @@ namespace
 
             Babylon::Polyfills::Console::Initialize(env, [](const char* message, auto) {
                 OutputDebugStringA(message);
+                printf("%s", message);
+                fflush(stdout);
```
bghgary, line 135: This is needed because of the merge with validation tests? Was the CI missing output then? cc @CedricGuillemet

Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
```diff
@@ -7,14 +7,14 @@ namespace Babylon::Graphics
 {
     class DeviceContext;
 
-    struct ViewPort
+    struct Rect
     {
         float X{0.0f};
         float Y{0.0f};
         float Width{1.0f};
         float Height{1.0f};
```
bghgary, line 15: nit: I would recommend changing the defaults if the name is changing. Must change the defaults below also.
```suggestion
        float X{};
        float Y{};
        float Width{};
        float Height{};
```

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -53,37 +53,69 @@ namespace Babylon::Graphics
...
+        // Note that the view rect and view scissor are reset to the desired dimensions before the encoder is submitted.
+        if (m_desiredScissor.Width != 0.0 || m_desiredScissor.Height != 0.0)
+        {
+            const auto flippedScissor = GetFlippedScissor();
+            bgfx::setViewRect(
+                m_viewId.value(),
+                static_cast<uint16_t>(flippedScissor.X),
+                static_cast<uint16_t>(flippedScissor.Y),
+                static_cast<uint16_t>(flippedScissor.Width),
+                static_cast<uint16_t>(flippedScissor.Height));
```
bghgary, line 79: This is currently truncated. Should it round?

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -53,37 +53,69 @@ namespace Babylon::Graphics
...
-        bgfx::setViewRect(m_viewId.value(), 0, 0, Width(), Height());
+        // If a scissor is set, we need to set the BGFX view rect to match it before clearing.
+        // We set the view rect instead of the view scissor because BGFX clears after the view rect is set and before
+        // the view scissor is set.
+        // 
+        // If no scissor is set, WebGL cleans the entire screen, so set the view rect to cover the entire screen before
+        // clearing to match WebGL's behavior; otherwise BGFX will only clear the view rect.
+        //
+        // Note that the view rect and view scissor are reset to the desired dimensions before the encoder is submitted.
+        if (m_desiredScissor.Width != 0.0 || m_desiredScissor.Height != 0.0)
```
bghgary, line 71:
```suggestion
        if (m_desiredScissor.Width != 0.0f || m_desiredScissor.Height != 0.0f)
```

Also, float comparison with exactly zero is not typical. Should this be a epsilon check?

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -107,16 +139,35 @@ namespace Babylon::Graphics
...
+    /// Note that this function sets the y coord to 0 if the desired scissor's height is 0 because BGFX requires the
+    /// x, y, width, and height to all be zero to disable the scissor. Without this extra check this function will set
+    /// the y coord to the framebuffer's height, which will make BGFX set an invalid scissor rect (the top and bottom
+    /// of the scissor rect will have the same value), and may crash the renderer in some cases.
+    ///
+    Rect FrameBuffer::GetFlippedScissor() const
     {
-        if (m_viewId.has_value() && viewPort.Equals(m_bgfxViewPort))
+        Rect scissor = m_desiredScissor;
+        scissor.Y = m_desiredScissor.Height == 0.0f ? 0.0f : std::max(0.0f, (Height() - m_desiredScissor.Y) - m_desiredScissor.Height);
```
bghgary, line 156: We are clamping the Y. Should we also clamp the X?

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -53,37 +53,69 @@ namespace Babylon::Graphics
...
     void FrameBuffer::SetViewPort(bgfx::Encoder& encoder, float x, float y, float width, float height)
     {
         m_desiredViewPort = {x, y, width, height};
-        SetBgfxViewPort(encoder, m_desiredViewPort);
+        SetBgfxViewPortAndScissor(encoder, m_desiredViewPort, m_desiredScissor);
+    }
+
+    void FrameBuffer::SetScissor(bgfx::Encoder& encoder, float x, float y, float width, float height)
+    {
+        m_desiredScissor = {x, y, width, height};
```
bghgary, line 105: Is this in a different unit than view port. View port is normalized 0 to 1. Is scissor not normalized? If not normalized, how does hardware scaling affect this?

Plugins/NativeEngine/Source/NativeEngine.h
```diff
@@ -201,6 +201,7 @@ namespace Babylon
         void GetFrameBufferData(const Napi::CallbackInfo& info);
         void SetStencil(NativeDataStream::Reader& data);
         void SetViewPort(NativeDataStream::Reader& data);
+        void SetScissor(NativeDataStream::Reader& data);
```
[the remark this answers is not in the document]
bghgary: I had a similar thought when reviewing the JS code. From contract/performance standpoint, it probably is better to have a separate command for disabling scissor, but it's not a big deal. The fact that bgfx uses specific values to disable is an implementation detail.

Apps/Playground/Scripts/config.json
docEdub: Ok. Babylon.js has one vis test for scissor we can bring over. Should I remove the two I added here or put them in the Babylon.js vis test list?

Core/Graphics/Source/FrameBuffer.cpp
docEdub: Hmm, maybe we should pass these from the JS side as ints instead of floats? WebGL uses ints for scissors but Babylon.js `Engine.enableScissor` uses floats.

docEdub: If we stay with floats then I'm not sure if we should round or not. I think since WebGL uses ints then it's ok to truncate.

Core/Graphics/Source/FrameBuffer.cpp
docEdub: Ya, and should we decide what an invalid scissor rect is and how invalid rects should be made valid?

Core/Graphics/Source/FrameBuffer.cpp
docEdub: Scissor is not normalized and is not affected by hardware scaling when I change it with `Engine.setHardwareScalingLevel`.

Plugins/NativeEngine/Source/NativeEngine.h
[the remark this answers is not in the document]
docEdub: Ok, thanks. I'll leave it as-is.

Apps/Playground/Win32/App.cpp
docEdub: Ya, the validation tests for this change should fail for this PR until the Babylon.js change gets deployed. In the first CI runs for this change, the validation test process hung as expected but was not considered to be failed. Cedric made some changes to fix that and added them to this PR.

Apps/Playground/Scripts/config.json
bghgary: It depends if the coverage is good or not. Is the Babylon.js one covering the same things?

Core/Graphics/Source/FrameBuffer.cpp
docEdub: This is to check if the scissor is being disabled, in which case the width and height are explicitly set to 0.0. I don't think an epsilon is needed for this unless we want to start checking for invalid rects.

Core/Graphics/Source/FrameBuffer.cpp
bghgary: There is no such thing as pass as ints since JS doesn't have ints. JS only has doubles (well, a form of a double).

> I think since WebGL uses ints then it's ok to truncate.

I'm not sure I understand this statement. `GLint` is just an alias for `Number`. We need to find out what the underlying behavior is supposed to do. Maybe we can test it and see?

Core/Graphics/Source/FrameBuffer.cpp
bghgary: It seems the code is using `Width()` and `Height()` to do some calculations relating to scissor rects. `Width()` and `Height()` do take hardware scaling into account. Is that an issue?

bghgary: Maybe we can add a test with non-default hardware scaling with scissor rects also?

Core/Graphics/Source/FrameBuffer.cpp
bghgary: What does WebGL do?

Core/Graphics/Source/FrameBuffer.cpp
bghgary: Ok, since we are checking for exactly zero only for width and height, why do we need an additional check for y in the flip code?

Core/Graphics/Source/FrameBuffer.cpp
docEdub: Maybe I'm confused. What's the test for this? I already tested this by setting `engine.setHardwareScalingLevel` to `0.5` and `0.25`, and the scissor rect took up the same amount of space on the screen for each level. Am I missing something?

bghgary: If that's the case, then hardware scaling is being taken into account.

> Scissor is not normalized and is not affected by hardware scaling when I change it with `Engine.setHardwareScalingLevel`

Maybe I'm not interpreting this statement correctly.

Core/Graphics/Source/FrameBuffer.cpp
docEdub: Ya, and I tried to document this in the comment, maybe unsuccessfully.

BGFX doesn't consider the rect to be empty if x or y is non-zero. In the case of y being non-zero, bgfx sets the top and bottom of the D3D scissor rect to y, which is what was crashing the playground on my machine yesterday.

Apps/Playground/Scripts/config.json
docEdub: The Babylon.js vis test covers the same things as the 2nd test I added here named "Multi cameras and output render target with scissors". I'll replace it with the JS test.

I think we should keep the 1st test I added here, though, since it has viewports and multiple cameras, too, which can create some unexpected interactions if we're not careful. Should I add this 1st test back to the Babylon.js tests, too? ...or do we call this a native-specific test and leave it out of the JS tests?

Core/Graphics/Source/FrameBuffer.cpp
docEdub: I tried out some fractional values in Babylon.js and Spector shows they're being rounded, not truncated. Good call.

Apps/Playground/Scripts/config.json
bghgary: > Should I add this 1st test back to the Babylon.js tests, too?

I think so. I don't see a reason why there should be a difference in testing. If it makes sense to test in native, it should also make sense to test in JS.

Core/Graphics/Source/FrameBuffer.cpp
bghgary: I think I understood your comment, maybe. Looking at this code, I don't see how scissor rect will be used if the width and height is zero. If I'm right, then bgfx will never be passed the non-zero x/y scissor values. Am I missing something?

Core/Graphics/Source/FrameBuffer.cpp
bghgary: Just to clarify, we are sure that this code works correctly if we set hardware scaling on the engine?

docEdub: Ya, the code works correctly with hardware scaling. I checked it manually. There is no visualization test for this, yet.

Core/Graphics/Source/FrameBuffer.cpp
docEdub: WebGL is fairly permissive with this. It allows negative x and y values and it allow the width and height to be set larger than the framebuffer. I updated this PR to do the same.

Apps/Playground/Scripts/config.json
docEdub: I think some of the tests I added here will only test browser implementations in JS, not Babylon.js itself. I don't think it makes sense to add all of them to the JS tests.

For example, the out of bounds tests I added today don't make much sense to add to the JS tests since Babylon.js just passes the scissor parameters directly to the WebGL implementation.

Core/Graphics/Source/FrameBuffer.cpp
docEdub: Tests added. Note that bgfx screenshots adjust to the hardware scaling level so I updated validation_native.js to match it. This could lead to large output images when the hardware scaling level is small (a 0.25 scaling level will result in a ~15MB .png). I used a 0.9 scaling level to keep the image size down in this PR but maybe there's a better way to keep the .png files small?

Apps/Playground/Scripts/config.json
bghgary: If that's the case, we need to come up with a mechanism for Babylon Native specific tests. Otherwise, it will get more difficult to maintain what tests go where.

docEdub: So the current config.json would be just for tests that apply to both native and JS, and then add another config file for native-only tests, named something like config-native.json?

bghgary: I don't know. Let's discuss offline.

docEdub: I asked Seb if he wants the new tests in Babylon.js and he's cool with it, so there's no need to have separate that are specific to Babylon Native, yet.

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -53,37 +53,72 @@ namespace Babylon::Graphics
...
-        encoder.touch(m_viewId.value());
+        // If a scissor is not set, WebGL clears the entire screen, so set the view rect to cover the entire screen
+        // before clearing to match WebGL's behavior; otherwise BGFX will only clear the view rect.
+        //
+        // If a scissor is set, we need to set the BGFX view rect to match it before clearing.
+        // We set the view rect instead of the view scissor because BGFX clears after the view rect is set and before
+        // the view scissor is set.
+        //
+        // Note that the view rect and view scissor are reset to the desired dimensions before the encoder is submitted.
+        if (m_desiredScissor.X == 0.0f && m_desiredScissor.Y == 0.0f && m_desiredScissor.Width == 0.0f && m_desiredScissor.Height == 0.0f)
```
CedricGuillemet, line 71: float comparison is a bit dangerous. use a fabsf with an epsylon instead. This said, x,y,width and height are in pixels, so why not keep uint16 for these values all along? And if needed, use floats at the very end.

docEdub: These values are all explicitly set to 0.0 when the scissor is disabled, so I don't think an epsilon is needed here.

CedricGuillemet (approved):

Apps/Playground/Scripts/config.json
```diff
@@ -237,6 +237,42 @@
...
+            "title": "Multi camera rendering with scissors",
+            "playgroundId": "#1LK70I#35",
+            "referenceImage": "multiCameraRenderingWithScissors.png"
+        },
+        {
+            "title": "Scissor test",
+            "renderCount": 10,
+            "playgroundId": "#W7E7CF#29",
+            "referenceImage": "scissor-test.png",
+            "comments": "Reset playgroundId to #W7E7CF#12 when d3d12 mip-map issue is fixed"
```
bghgary, line 250: Is there an issue filed for this?

docEdub: There is now :)

https://github.com/BabylonJS/BabylonNative/issues/1265

PR #1249, opened 2023-06-30
iOS17

### LINKED ISSUE #1248: NativeCamera AVCaptureVideoOrientation Deprecated with iOS 17
iOS17 suspected to  be released in September 2023

CedricGuillemet:
- Add job with xcode15/ios17
- `AVCaptureVideoOrientation` is deprecated in newer targets. Replaced with an enum when targeting iOS17+ or macOS14+.

Fixes #1248


.github/jobs/ios.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         vmImage: ${{ parameters.vmImage }}
 
     - script: |
+        ls /Applications
```
bghgary, line 20: What's this for?

Apps/Playground/CMakeLists.txt
```diff
@@ -175,6 +175,11 @@ if(APPLE)
             XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES
             RESOURCE "${RESOURCE_FILES}"
             FOLDER "Playground")
+        # The NSDocumentRevisionsDebugMode argument is typically added automatically by Xcode when running a macOS application in debug mode.
```
bghgary, line 178: nit: newline before this would be good for readability

bghgary: Maybe also add comment on why we are disabling?

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -20,10 +20,15 @@
 
 @class CameraTextureDelegate;
 
-@interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBufferDelegate>
+@interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBufferDelegate, NSObject>
 {
+#if (__MAC_OS_X_VERSION_MIN_REQUIRED >= 140000 || __IPHONE_OS_VERSION_MIN_REQUIRED >= 170000)
+    // angle in half PI (0 = 0, 1 = 0.5PI, 2 = PI, 3 = 1.5PI)
+    // Might change in the future for a float.
+    @public int videoRotationAngle;
```
bghgary, line 28: Should be UpperCamelCase since it's public? I also don't understand the purpose of this int. If Apple is deprecating the enum, can we introduce our own? Why int?

.github/jobs/ios.yml
CedricGuillemet: Removed

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -666,11 +720,11 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
 
                     [m_impl->currentCommandBuffer addCompletedHandler:^(id<MTLCommandBuffer>) {
                         if (textureY != nil) {
-                            [textureY setPurgeableState:MTLPurgeableStateEmpty];
+                            [textureY setPurgeableState:MTLPurgeableStateVolatile];
                         }
 
                         if (textureCbCr != nil) {
-                            [textureCbCr setPurgeableState:MTLPurgeableStateEmpty];
+                            [textureCbCr setPurgeableState:MTLPurgeableStateVolatile];
```
docEdub, line 727: It might be good to get @ryantrem's input on this.

Apps/Playground/CMakeLists.txt
CedricGuillemet: Fixed

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
CedricGuillemet: Replaced with an enum

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -614,18 +648,38 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
                     renderPassDescriptor.colorAttachments[0].texture = m_impl->textureRGBA;
                     renderPassDescriptor.colorAttachments[0].loadAction = MTLLoadActionDontCare;
                     renderPassDescriptor.colorAttachments[0].storeAction = MTLStoreActionStore;
-
+
```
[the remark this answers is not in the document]
CedricGuillemet: Fixed

.github/jobs/ios.yml
```diff
@@ -28,10 +28,12 @@ jobs:
 
     - task: Xcode@5
       inputs:
+        actions: 'build'
         xcWorkspacePath: 'buildiOS/BabylonNative.xcodeproj'
         scheme: 'Playground'
         sdk: 'iphoneos'
         useXcpretty: false
         configuration: RelWithDebInfo
+        destinationTypeOption: simulators
```
ryantrem, line 37: Just curious, what does this mean? It seems like it conflicts with `sdk: 'iphoneos'`.

Apps/Playground/CMakeLists.txt
```diff
@@ -175,6 +175,15 @@ if(APPLE)
             XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES
             RESOURCE "${RESOURCE_FILES}"
             FOLDER "Playground")
+
+        # The NSDocumentRevisionsDebugMode argument is typically added automatically by Xcode
```
ryantrem, line 179: In the PR description, you said "It looks like an old issue (10+ years) and never seen before with BN." Are there some links you could add to these comments that have more context?

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -20,10 +20,24 @@
...
+    enum Enum
+    {
+        LandscapeRight,
+        LandscapeLeft,
+        Portrait,
+        PortraitUpsideDown
+    };
+};
 
+@interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBufferDelegate, NSObject>
```
ryantrem, line 34: Why is the extra `, NSObject` needed? Isn't that supposed to be a list of protocols / interfaces being implemented?

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -20,10 +20,24 @@
 
 @class CameraTextureDelegate;
 
-@interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBufferDelegate>
+struct VideoRotationAngle
 {
-    @public AVCaptureVideoOrientation VideoOrientation;
+    enum Enum
```
ryantrem, line 25: Can we just use `enum class` here?

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -679,13 +719,16 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
         });
-
+#if (__MAC_OS_X_VERSION_MIN_REQUIRED >= 140000 || __IPHONE_OS_VERSION_MIN_REQUIRED >= 170000)
+        return CameraDimensions{m_impl->cameraDimensions.width, m_impl->cameraDimensions.height};
+#else
         // To match the web implementation if the sensor is rotated into a portrait orientation then the width and height
         // of the video should be swapped
         return m_impl->cameraTextureDelegate->VideoOrientation == AVCaptureVideoOrientationLandscapeLeft ||
             m_impl->cameraTextureDelegate->VideoOrientation == AVCaptureVideoOrientationLandscapeRight ?
             CameraDimensions{m_impl->cameraDimensions.width, m_impl->cameraDimensions.height} :
```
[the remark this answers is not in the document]
ryantrem: I was wondering this too... we're using this code in product today, and we support all device orientations, so it seems like the current behavior is correct as far as I can tell. If that's the case, then it also seems like we can remove this condition. But also want to see of Colton remembers.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -20,10 +20,24 @@
...
+        PortraitUpsideDown
+    };
+};
 
+@interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBufferDelegate, NSObject>
+{
+#if (__MAC_OS_X_VERSION_MIN_REQUIRED >= 140000 || __IPHONE_OS_VERSION_MIN_REQUIRED >= 170000)
+    @public VideoRotationAngle::Enum VideoRotation;
+#else
+    @public AVCaptureVideoOrientation VideoOrientation;
```
ryantrem, line 39: Can we just always use the new independent VideoRotationAngle enum, and for the older code path just convert to this type? Then it seems like we don't have to duplicate nearly as much code.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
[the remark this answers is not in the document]
CoPrez: It's been a while since I've looked at this code, but from what I recall MacOS doesn't support the notion of orientation changes. For MacOS devices the VideoOrientation stays at the default of LandscapeLeft which is the "normal" orientation.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
ryantrem: Dunno, the error message Cedric included in the PR description combined with the Apple docs are very confusing...

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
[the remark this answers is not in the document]
CoPrez: Sorry. I didn't notice that the inline if was returning the same value. Yes, that does seem like a copy & paste error. The dimensions should be swapped when in a portrait orientation... I'm surprised that isn't causing issues somewhere else down the line.

My previous comment was saying that I don't think you need to fork the logic for MacOS here as it should just always be considered to be in landscape orientation.

Apps/Playground/CMakeLists.txt
CedricGuillemet: Not really. All threads I've read are the same: arguments are added automatically and it breaks command line arguments parsing. I still have no explanation on why it only happens now.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
CedricGuillemet: leftover. removed.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
CedricGuillemet: yes, fixed.

.github/jobs/ios.yml
CedricGuillemet: leftover for some tests. removed.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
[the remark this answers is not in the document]
CedricGuillemet: Thanks! I've added the swap for portrait.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
CedricGuillemet: I got the validation error with Volatile as well, som definitely not a valid fix.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
CedricGuillemet: Indeed, much better. Fixed.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -533,13 +540,12 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
-
             // To match the web implementation if the sensor is rotated into a portrait orientation then the width and height
             // of the video should be swapped
-            return implObj->m_impl->cameraTextureDelegate->VideoOrientation == AVCaptureVideoOrientationLandscapeLeft ||
-                implObj->m_impl->cameraTextureDelegate->VideoOrientation == AVCaptureVideoOrientationLandscapeRight ?
+            return implObj->m_impl->cameraTextureDelegate->VideoOrientation == VideoOrientationAngle::LandscapeLeft ||
+                implObj->m_impl->cameraTextureDelegate->VideoOrientation == VideoOrientationAngle::LandscapeRight ?
                 CameraDimensions{implObj->m_impl->cameraDimensions.width, implObj->m_impl->cameraDimensions.height} :
-                CameraDimensions{implObj->m_impl->cameraDimensions.width, implObj->m_impl->cameraDimensions.height};
+                CameraDimensions{implObj->m_impl->cameraDimensions.height, implObj->m_impl->cameraDimensions.width};
```
ryantrem, line 548: This change makes me nervous since this code is used in some shipped products. How can we be sure this isn't going to break something, and that even after this change the behavior still "matches the web implementation" as mentioned in the comment? Were you able to verify this somehow?

CedricGuillemet: I'm wondering if it should be 'fixed' as it's not broken. I'm in favor of leaving as is and add a comment.

ryantrem: Seems reasonable.

CedricGuillemet: I've reverted the fix.

bghgary (approved):

PR #1250, opened 2023-07-03
Fetch content

CedricGuillemet:
Follow up to https://github.com/BabylonJS/JsRuntimeHost/pull/39 
- Disabling ARC by default here because of PG and bgfx calls retain/release
- clean up PG gradle
- JSE names to be same as JSR
- update 2 VT PGs
- remove `npm i` for android ci
- removed `npm i` in readme

in progress: install update


Apps/Playground/Scripts/config.json
```diff
@@ -183,7 +183,7 @@
         },
         {
             "title": "LightFalloff",
-            "playgroundId": "#20OAV9#201",
+            "playgroundId": "#20OAV9#11218",
             "referenceImage": "lightFalloff.png"
         },
```
bghgary, line 188: This doesn't seem to match what's in Babylon.js anymore. Should we update it?

CMakeLists.txt
```diff
@@ -2,21 +2,32 @@
...
-FetchContent_MakeAvailable(jsruntimehost)
+
+set(CONTENT_TO_FETCH cmake-extensions jsruntimehost)
+
+if(IOS)
+    FetchContent_MakeAvailable(ios-cmake)
+    set(CMAKE_TOOLCHAIN_FILE "${ios-cmake_SOURCE_DIR}/ios.toolchain.cmake" CACHE PATH "")
+    set(PLATFORM "OS64COMBINED" CACHE STRING "")
+    set(DEPLOYMENT_TARGET "12" CACHE STRING "")
+    set(ENABLE_ARC OFF CACHE STRING "Enables or disables ARC support.")
```
bghgary, line 29: Why turn this off by default? If bgfx needs this to be off, I think we should only set it off for bgfx.

CedricGuillemet:
Duplicate with https://github.com/BabylonJS/BabylonNative/pull/1260

Apps/Playground/Scripts/config.json
CedricGuillemet: A new pass on tests should be done. Some are updated in bjs and not here. some are replaced,...

CMakeLists.txt
CedricGuillemet: I tried without success. if I force the flag for bgfx, then the brige is failing to compile. in the doc, the command line disables arc : https://github.com/BabylonJS/BabylonNative/blob/master/BUILDING.md#building-on-macos-targeting-ios

PR #1276, opened 2023-08-07
Win32 Installation test

CedricGuillemet:
Build the UnitTests app using installation lib/headers. If the build fails, something is missing in the installation process.
Some questions:
- Do we want to have Debug/Release testing?
- PDBs?
- Do we want to package the installation and test that package instead (npm? nuget?)
- Do we want to package for another platform ?
- Do we want to test the unittest build in this process instead ?

### DIFF
--- a/.github/jobs/win32.yml
+++ b/.github/jobs/win32.yml
@@ -73,7 +73,8 @@ jobs:
       - script: |
           cmake --build build${{ variables.solutionName }} --target INSTALL --config RelWithDebInfo
           cd Install/Test
-          cmake . -DBINARY_DIR=../../build${{ variables.solutionName }} ${{ variables.jsEngineDefine }} && cmake --build . --config RelWithDebInfo
+          cmake . -DBINARY_DIR=../../build${{ variables.solutionName }} ${{ variables.jsEngineDefine }}
+          cmake --build . --config RelWithDebInfo
         displayName: 'Install'
 
       - task: PublishBuildArtifacts@1
--- a/.github/jobs/win32.yml
+++ b/.github/jobs/win32.yml
@@ -70,6 +70,12 @@ jobs:
           Playground app:///Scripts/validation_native.js 
         displayName: 'Validation Tests'
 
+      - script: |
+          cmake --build build${{ variables.solutionName }} --target INSTALL --config RelWithDebInfo
+          cd Install/Test
+          cmake . -DBINARY_DIR=../../build${{ variables.solutionName }} ${{ variables.jsEngineDefine }} && cmake --build . --config RelWithDebInfo
+        displayName: 'Install'
+
       - task: PublishBuildArtifacts@1
         inputs:
           artifactName: '${{ variables.solutionName }} - ${{ parameters.graphics_api }} Rendered Pictures'
--- a/Install/Test/CMakeLists.txt
+++ b/Install/Test/CMakeLists.txt
@@ -16,7 +16,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # __cplusplus value is not in sync with language version used. MVSC needs this flag to update it accordingly
 # https://gitlab.kitware.com/cmake/cmake/-/issues/18837
-if (MSVC)
+if(MSVC)
     add_compile_options(/Zc:__cplusplus)
 endif()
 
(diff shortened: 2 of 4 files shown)


docEdub (approved):

PolygonalSun (approved):
LGTM

.github/jobs/win32.yml
```diff
@@ -70,6 +70,12 @@ jobs:
           Playground app:///Scripts/validation_native.js 
         displayName: 'Validation Tests'
 
+      - script: |
+          cmake --build build${{ variables.solutionName }} --target INSTALL --config RelWithDebInfo
+          cd Install/Test
+          cmake . -DBINARY_DIR=../../build${{ variables.solutionName }} ${{ variables.jsEngineDefine }} && cmake --build . --config RelWithDebInfo
```
bghgary, line 76: nit: Is it necessary for these two cmake commands to be on one line?

Install/Test/CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+    GIT_REPOSITORY https://github.com/BabylonJS/CMakeExtensions.git
+    GIT_TAG 366dc4a84fb20f4060d97e89948c343e74c51fc3)
+FetchContent_MakeAvailable(cmake-extensions)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# __cplusplus value is not in sync with language version used. MVSC needs this flag to update it accordingly
+# https://gitlab.kitware.com/cmake/cmake/-/issues/18837
+if (MSVC)
```
bghgary, line 19: nit
```suggestion
if(MSVC)
```

Install/Test/CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+set(HEADERS "${UNITTESTS_DIR}/Shared/Tests.h")
+
+if(WIN32)
+    set(TEST_INSTALL_APP "${UNITTESTS_DIR}/Win32/App.cpp")
+else()
+	message(FATAL_ERROR "Unrecognized platform: ${CMAKE_SYSTEM_NAME}")
+endif()
+
+# Nuget/JSI
+# TODO: move it to generic cmake
```
bghgary, line 51: What's with this TODO?

Install/Test/CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+# for napi
+target_compile_definitions(TestInstall PUBLIC NODE_ADDON_API_DISABLE_DEPRECATED)
+target_compile_definitions(TestInstall PUBLIC NODE_ADDON_API_DISABLE_NODE_SPECIFIC)
+
+set(INSTALL_LIBS_DIR "${BINARY_DIR}/install/lib")
+target_include_directories(TestInstall PRIVATE "${BINARY_DIR}/install/include")
+target_link_directories(TestInstall PRIVATE ${INSTALL_LIBS_DIR})
+
+
+target_link_directories(TestInstall PRIVATE "${V8JSI_LIB_PATH_RELEASE}")
```
bghgary, line 126: Collapse into one?

Install/Test/CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+
+if(NAPI_JAVASCRIPT_ENGINE STREQUAL "JSI")
+    set(EXTRA_LIBS "v8jsi.dll.lib")
+endif()
```
bghgary, line 130: Should this be inside the `if` above?

Install/Test/CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+
+# Nuget/JSI
+# TODO: move it to generic cmake
+set(NUGET_PATH "${BINARY_DIR}/NuGet")
+set(V8JSI_VERSION "0.64.33")
+set(CPU_ARCH "x64")
+set(PLATFORM_FOLDER "win32")
+set(V8JSI_PACKAGE_PATH "${NUGET_PATH}/packages/ReactNative.V8Jsi.Windows.${V8JSI_VERSION}")
+set(V8JSI_LIB_PATH_RELEASE "${V8JSI_PACKAGE_PATH}/lib/${PLATFORM_FOLDER}/Release/${CPU_ARCH}/")
```
bghgary, line 57: Maybe move these properties into the scope of where it is used to avoid confusing? I'm actually a bit confused if some of these properties are necessary?

Install/Test/CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+    squish
+    tinyexr
+    UrlLib
+    Window
+    XMLHttpRequest
+    )
+
+target_link_libraries(TestInstall
+    ${STD_LIBS}
+    ${EXTRA_LIBS}
```
bghgary, line 177: We typically follow a pattern of using `ADDITIONAL_LINK_LIBRARIES` as the variable to add more libs. What does `STD_LIBS` mean here?

Install/Test/CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+set(NUGET_PATH "${BINARY_DIR}/NuGet")
+set(V8JSI_VERSION "0.64.33")
+set(CPU_ARCH "x64")
+set(PLATFORM_FOLDER "win32")
+set(V8JSI_PACKAGE_PATH "${NUGET_PATH}/packages/ReactNative.V8Jsi.Windows.${V8JSI_VERSION}")
+set(V8JSI_LIB_PATH_RELEASE "${V8JSI_PACKAGE_PATH}/lib/${PLATFORM_FOLDER}/Release/${CPU_ARCH}/")
+
+if(NAPI_JAVASCRIPT_ENGINE STREQUAL "JSI")
+    set(JSI_CPP "${V8JSI_PACKAGE_PATH}/build/native/jsi/jsi/jsi.cpp")
+elseif(NAPI_JAVASCRIPT_ENGINE STREQUAL "V8")
```
bghgary, line 61: I wonder if it makes sense for the install to include these v8 specific files so that consumers don't have to include them again? We already have the files during the build.

.github/jobs/win32.yml
CedricGuillemet: nop. changed to 2 lines for clarity

Install/Test/CMakeLists.txt
CedricGuillemet: Fixed

Install/Test/CMakeLists.txt
CedricGuillemet: Random thoughts. Removed.

Install/Test/CMakeLists.txt
CedricGuillemet: `STD_LIBS` removed and `ADDITIONAL_LIBRARIES` name used

Install/Test/CMakeLists.txt
CedricGuillemet: moved inside the if codeblock above

Install/Test/CMakeLists.txt
CedricGuillemet: Block moved to the correct block

Install/Test/CMakeLists.txt
CedricGuillemet: Can this be error prone if mixing different v8/jsi versions at some point?

bghgary: Probably, neither contracts are ABI-safe nor backwards compatible.

bghgary (approved):

PR #1282, opened 2023-09-13
Fix MSAA when using rendering pipelines

bghgary:
Brings in changes from https://github.com/BabylonJS/Babylon.js/pull/14292 as part of the package version update.

The main fix was a misunderstanding of the `BGFX_TEXTURE_MSAA_SAMPLE` flag in bgfx and how textures are resolved. Thanks to @sebavan for explaining to me how texture resolving works.


PolygonalSun (approved):
LGTM

PR #1283, opened 2023-09-15
ImageCapture native polyfill

ryantrem:
This change provides an initial implementation of a polyfill for [`ImageCapture`](https://developer.mozilla.org/en-US/docs/Web/API/ImageCapture) to enable high res photo capture. This builds on top of the existing `MediaCapture` polyfill, and follows existing patterns. Some notes on this:
- MediaStream and MediaStreamTrack are polyfilled by the same class on the C++ side, which means that class (`MediaStream`) is what gets passed into the `ImageCapture` constructor. For this PR, I am considering `ImageCapture` to be part of the same polyfill. It is not independent, and has internal knowledge that the object passed into the `ImageCapture` constructor is actually a `MediaStream`. The `MediaStream` now exposes the underlying (platform specific) `CameraDevice`.
- The `CameraDevice` remains the primary camera abstraction across platforms. As such, it is the thing that exposes high res photo capabilities, settings, and the ability to actually take a high res photo. Since capabilities are normally track specific, it just returns capabilities for the opened camera session. This should be fine because we only support opening a single camera stream at a time, and `ImageCapture` isn't constructed until you have a track (which is the result of opening a camera stream).

This PR includes all the common glue code (e.g. napi) and platform abstractions, plus an initial iOS implementation (probably needs a bit more work, but the basics are in place).


Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -15,16 +16,17 @@
...
+
+@class CameraTextureDelegate;
 
 @interface CameraTextureDelegate : NSObject <AVCaptureVideoDataOutputSampleBufferDelegate>
 {
     @public AVCaptureVideoOrientation VideoOrientation;
 
     CVMetalTextureCacheRef textureCache;
+    CVMetalTextureRef cameraTextureY;
+    CVMetalTextureRef cameraTextureCbCr;
```
ryantrem, line 29: This is unrelated to ImageCapture, I just noticed this while making other changes. These two members were declared in the @implementation instead of the @interface, which in the bizarre world of Objective C made them effectively static members instead of instance members, which is definitely not the behavior we want.

CedricGuillemet (approved):

Plugins/NativeCamera/Source/CameraDevice.h
```diff
@@ -53,10 +95,15 @@ namespace Babylon::Plugins
...
         void Close();
         CameraDimensions UpdateCameraTexture(bgfx::TextureHandle textureHandle);
+        TakePhotoTask TakePhoto(PhotoSettings photoSettings);
 
         const std::vector<CameraTrack>& SupportedResolutions() const;
         const std::vector<std::unique_ptr<Capability>>& Capabilities() const;
 
+        // Gets high resolution photo capture capabilities for the currently opened stream/track.
+        Plugins::PhotoCapabilities PhotoCapabilities() const;
+        Plugins::PhotoSettings DefaultPhotoSettings() const;
```
bghgary, line 105:
```suggestion
        const Plugins::PhotoCapabilities& PhotoCapabilities() const;
        const Plugins::PhotoSettings& DefaultPhotoSettings() const;
```

Plugins/NativeCamera/Source/MediaStream.h
```diff
@@ -35,16 +35,21 @@ namespace Babylon::Plugins
         // Update the camera texture and return true if the dimensions have changed, false otherwise
         bool UpdateTexture(bgfx::TextureHandle textureHandle);
 
+        std::shared_ptr<Babylon::Plugins::CameraDevice> CameraDevice() const
+        {
+            return m_cameraDevice;
+        }
+
```
bghgary, line 42:
```suggestion
        const std::shared_ptr<Babylon::Plugins::CameraDevice>& CameraDevice() const
        {
            return m_cameraDevice;
        }
```

bghgary (approved):

Plugins/NativeCamera/Source/CameraDevice.h
```diff
@@ -1,12 +1,52 @@
...
+        Babylon::Plugins::RedEyeReduction RedEyeReduction{};
+        std::set<Babylon::Plugins::FillLightMode> FillLightModes{};
+
+        int32_t MinWidth{};
+        int32_t MaxWidth{};
+        int32_t StepWidth{};
+
+        int32_t MinHeight{};
+        int32_t MaxHeight{};
+        int32_t StepHeight{};
```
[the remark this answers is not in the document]
ryantrem: I wouldn't expect so, but existing code dealing with width/height was already using `int32_t` (and it is also the numeric type returned from the iOS APIs), so I just matched what was there. Conceptually it could be uint32_t I think, but if we go that route we should change it everywhere and perhaps have code to deal with the possibility that platform APIs we are calling do return a negative value.

Plugins/NativeCamera/Source/MediaStream.h
ryantrem: As discussed, not changing this.

Plugins/NativeCamera/Source/CameraDevice.h
```diff
@@ -1,12 +1,52 @@
...
+        Off,
+        Flash,
+        Auto,
+    };
+
+    class PhotoCapabilities final
+    {
+    public:
+        Babylon::Plugins::RedEyeReduction RedEyeReduction{};
+        std::set<Babylon::Plugins::FillLightMode> FillLightModes{};
```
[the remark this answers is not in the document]
ryantrem: As discussed, not changing this one.

Plugins/NativeCamera/Source/CameraDevice.h
```diff
@@ -95,14 +95,14 @@ namespace Babylon::Plugins
...
-        TakePhotoTask TakePhoto(PhotoSettings photoSettings);
+        TakePhotoTask TakePhotoAsync(PhotoSettings photoSettings);
 
         const std::vector<CameraTrack>& SupportedResolutions() const;
         const std::vector<std::unique_ptr<Capability>>& Capabilities() const;
 
         // Gets high resolution photo capture capabilities for the currently opened stream/track.
-        Plugins::PhotoCapabilities PhotoCapabilities() const;
-        Plugins::PhotoSettings DefaultPhotoSettings() const;
+        const Plugins::PhotoCapabilities PhotoCapabilities() const;
```
bghgary, line 104: Still missing `&`
```suggestion
        const Plugins::PhotoCapabilities& PhotoCapabilities() const;
```

ryantrem: Original intention was to just return a copy because they are small structs, but I will switch it to const refs.

PR #1285, opened 2023-09-19
Update glslang and SPIRV-Cross

SergioRZMasson:
Update glslang to version 13.0.0 and SPIRV-Cross to latest main branch. When updating glslang it now has a better way for setting default resource limits, that removes the need for "ResourceLimits.h".


Plugins/NativeEngine/Source/ShaderCompilerD3D.h
```diff
@@ -22,7 +22,9 @@ namespace Babylon
             const std::array<const char*, 1> sources{source.data()};
             shader.setStrings(sources.data(), gsl::narrow_cast<int>(sources.size()));
 
-            if (!shader.parse(&DefaultTBuiltInResource, 310, EProfile::EEsProfile, true, true, EShMsgDefault))
+            auto DefaultTBuiltInResource = GetDefaultResources();
```
CedricGuillemet, line 25: nit : variable names start with lower case

Install/Test/CMakeLists.txt
```diff
@@ -169,6 +169,7 @@ target_link_libraries(TestInstall
     d3d12
     d3dcompiler
     Pathcch
+    glslang-default-resource-limits
```
bghgary, line 172: Why do you need this? CMake should pull this in automatically? Actually, I'm not sure why all of these dependencies of dependencies are added here.

bghgary (approved):

Issue #1265, opened 2023-07-21
RTT mip-maps are not being generated automatically on D3D12


docEdub:
Running playground https://playground.babylonjs.com/#W7E7CF#25 on D3D12 shows the following when the window is small or the camera is moved far enough away...

![MicrosoftTeams-image (4)](https://github.com/BabylonJS/BabylonNative/assets/78323754/45f8dd50-9cf3-4c03-8213-348b7915293a)

It should look like this ...

![](https://github.com/BabylonJS/BabylonNative/assets/78323754/2d91c003-c108-418b-acdb-d784899c2eae)

Cedric and I looked at this in RenderDoc and saw the RTT mip-maps are missing on D3D12. They are generated automatically for other graphics APIs, but not D3D12.

CedricGuillemet:
I believe we can close this issue ? @docEdub

docEdub:
> I believe we can close this issue ? @docEdub

I don't think this is fixed, yet.

docEdub:
I found this bgfx issue that may be related: https://github.com/bkaradzic/bgfx/issues/798

TLDR; Automatic mip-map generation is not provided by D3D12, Vulkan, or Metal, so it needs to be explicitly implemented in bgfx to match the behavior seen in other graphics APIs. For Vulkan, this was implemented in bgfx in PR https://github.com/bkaradzic/bgfx/pull/2472, but it has not been implemented in D3D12 or Vulkan, yet, and afaict there is no plan to do so.

The end-result is the mip-map layers get created by bgfx on D3D12, but they are all black.

bghgary:
Closing for now. Reopen if we want to fix this.

PR #1287, opened 2023-09-19
Disable validation tests failing only on D3D12

docEdub:
Three validation tests for scissors are failing on D3D12 due to mip-maps being created but not populated by bgfx. See https://github.com/BabylonJS/BabylonNative/issues/1265. While we are waiting for the fix in bgfx, we want to disable these tests on D3D12.

This change adds a key to the validation tests config.json named `excludedGraphicsApis` that takes an array of strings specifying which graphics apis to skip for the given test. For example, to skip the D3D12 graphics api when running one of the broken scissor tests, it would look like this:
```
        {
            "title": "Scissor test",
            "renderCount": 10,
            "playgroundId": "#W7E7CF#12",
            "referenceImage": "scissor-test.png",
            "excludedGraphicsApis": ["D3D12"]
        },
```

The config.json key's name of `excludedGraphicsApis` was chosen to match a similar key in the Babylon.js validation test config.json named `excludedEngines`.


Apps/Playground/Android/app/CMakeLists.txt
```diff
@@ -33,8 +33,14 @@ set(BABYLON_NATIVE_PLAYGROUND_DIR "${CMAKE_CURRENT_LIST_DIR}/../..")
 add_subdirectory(${BABYLON_NATIVE_PLAYGROUND_DIR}/../../ ${BABYLON_NATIVE_PLAYGROUND_DIR}/../../build/Android_${CMAKE_ANDROID_ARCH_ABI}/)
 
 add_library(BabylonNativeJNI SHARED
+    ${BABYLON_NATIVE_PLAYGROUND_DIR}/Shared/Playground/GraphicsApiInfo.cpp
```
docEdub, line 36: I've exposed this new `GraphicsApiInfo` class on all platforms for completeness, but maybe we only want to expose it on the platforms running validation tests in the CI, i.e. Win32 only?

Apps/Playground/Android/app/CMakeLists.txt
```diff
@@ -33,8 +33,14 @@ set(BABYLON_NATIVE_PLAYGROUND_DIR "${CMAKE_CURRENT_LIST_DIR}/../..")
...
     src/main/cpp/BabylonNativeJNI.cpp)
 
+if(NOT GRAPHICS_API)
+    message(FATAL_ERROR "GRAPHICS_API must be set")
+endif()
```
docEdub, line 41: The top-level CMakeLists.txt for this repo is added as a subdirectory on Android, so the `GRAPHICS_API` CMake variable is not set in the scope of *this* CMakeLists.txt. This check makes sure the `GRAPHICS_API` variable has been set in the CMake command used to configure the build, e.g. `cmake -B build -D GRAPHICS_API=OpenGL`. The Android does this in it's gradle build config.

Maybe we don't need it?

Apps/Playground/Shared/Playground/GraphicsApiInfo.cpp
```diff
@@ -0,0 +1,47 @@
...
+#define STRINGIZE(x) STRINGIZEX(x)
+
+namespace Playground::Internal
+{
+    namespace
+    {
+        constexpr auto JS_CLASS_NAME = "GraphicsApiInfo";
+    }
+
+    class GraphicsApiInfo final : public Napi::ObjectWrap<GraphicsApiInfo>
```
docEdub, line 13: I exposed the graphics API's name as a static member of a class, but maybe it would be better to add it directly to `global` as something like `GraphicsApi_Name`?

Apps/Playground/Android/app/CMakeLists.txt
SergioRZMasson: I wonder if we should restrict the information to just GraphicsApiInfo, since we are already exposing a new API to Javascript couldn't we call it something like "NativeApiInfo" or something and use it to expose more native information in the future?

bghgary: Why not put in TestUtils?

Apps/Playground/Scripts/config.json
```diff
@@ -245,23 +245,23 @@
         {
             "title": "Scissor test",
             "renderCount": 10,
-            "playgroundId": "#W7E7CF#29",
+            "playgroundId": "#W7E7CF#12",
             "referenceImage": "scissor-test.png",
-            "comments": "Reset playgroundId to #W7E7CF#12 when d3d12 mip-map issue is fixed"
```
bghgary, line 250: A comment is probably still useful here. I don't expect D3D12 to not have the support forever.

Apps/Playground/Shared/Playground/GraphicsApiInfo.cpp
bghgary: I would recommend making this a static member of the TestUtils class.

Apps/Playground/Shared/Playground/GraphicsApiInfo.cpp
```diff
@@ -0,0 +1,47 @@
...
+    {
+    public:
+        static void Initialize(Napi::Env env)
+        {
+            Napi::Function func = DefineClass(
+                env,
+                JS_CLASS_NAME,
+                {
+                    StaticAccessor("Name", &GraphicsApiInfo::Name, nullptr)
```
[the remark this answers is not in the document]
bghgary: If you put it in TestUtils, this question doesn't exist.

Plugins/TestUtils/CMakeLists.txt
```diff
@@ -23,5 +23,8 @@ target_link_libraries(TestUtils
     PRIVATE GraphicsDevice
     PRIVATE GraphicsDeviceContext)
 
+message("TestUtils: GRAPHICS_API=${GRAPHICS_API}")
```
bghgary, line 26: Remove?

bghgary (approved):

bghgary:
Might want to update the PR description also.

Plugins/TestUtils/CMakeLists.txt
docEdub: sure

PR #1289, opened 2023-09-21
Set CMake policy CMP0135 to NEW

docEdub:
Sets fetched URL contents to the time of extraction, not the timestamps in the archive.

This is to prevent the CMake warning occurring due to the recently merged changes that fetch the googletest dependency as a .tar.gz URL.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,9 @@ cmake_minimum_required(VERSION 3.18)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 # Set fetched URL contents to the time of extraction, not the timestamps in the archive.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -2,6 +2,9 @@
 cmake_minimum_required(VERSION 3.18)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
+# Set fetched URL contents to the time of extraction, not the timestamps in the archive.
+cmake_policy(SET CMP0135 NEW)
```
bghgary, line 6: Does this cause an error if using an older cmake that doesn't have this policy? Do we have to wrap it around a check?

CedricGuillemet (approved):

PR #1290, opened 2023-09-21
Changed instance rendering to use instanceCount from JS

### LINKED ISSUE #1189: Instance count Vs buffer size for Instance Rendering
Following this forum thread: 
https://forum.babylonjs.com/t/react-native-unexpected-behaviour-of-instance-mesh-on-its-parent-movement/37001/2

After investigations, there is a difference in instancing rendering.
On the JS side, `instanceCount` is used to set the number of instances to render
https://github.com/BabylonJS/Babylon.js/blob/b2cb023438e190108ebc4d52a0cadfd18c71bb31/packages/dev/core/src/Engines/nativeEngine.ts#L586

On the native side, the instance count is determine by the size of the instance.
A possible fix is to forward the `instanceCount` in the protocol (PR on bjs), then use that value to limit the number of instance sent to bgfx here : 
https://github.com/BabylonJS/BabylonNative/blob/599008710819cc4ca6b828dc37b81f76485779f6/Plugins/NativeEngine/Source/VertexBuffer.cpp#L187

SergioRZMasson:
Issue #1189 is caused because we current don't use the instanceCount information provided by JS, instead we calculate the instance count be evaluating the size of the provided buffer. The correct behavior is to use the provided information from JS since the buffer can be bigger than the intended number of instances to accommodate for more instances in the future. 

This causes strange behavior in some situations such in the following forum issue:
https://forum.babylonjs.com/t/react-native-unexpected-behaviour-of-instance-mesh-on-its-parent-movement/37001/2

This change requires a update on JS and protocol version and should only be merge after a JS update is merged.

JS PR: https://github.com/BabylonJS/Babylon.js/pull/14331

Fixes #1189


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -321,7 +321,7 @@ namespace Babylon
             JS_CLASS_NAME,
             {
                 // This must match the version in nativeEngine.ts
-                StaticValue("PROTOCOL_VERSION", Napi::Number::From(env, 8)),
+                StaticValue("PROTOCOL_VERSION", Napi::Number::From(env, 9)),
```
bghgary, line 324: Is this necessary? Can we modify the code to do the old behavior if instanceCount is missing?

SergioRZMasson: Changed the PR to avoid requiring protocol version changes.

SergioRZMasson:
Changed the PR to avoid changing Protocol Version. Now newer version of Babylon.js will check if COMMAND_DRAWINDEXEDINSTANCED and COMMAND_DRAWINSTANCED are provided, if so it will call those instead.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -547,7 +547,9 @@ namespace Babylon
                 StaticValue("COMMAND_UNBINDFRAMEBUFFER", Napi::FunctionPointer::Create(env, &NativeEngine::UnbindFrameBuffer)),
                 StaticValue("COMMAND_DELETEFRAMEBUFFER", Napi::FunctionPointer::Create(env, &NativeEngine::DeleteFrameBuffer)),
                 StaticValue("COMMAND_DRAWINDEXED", Napi::FunctionPointer::Create(env, &NativeEngine::DrawIndexed)),
+                StaticValue("COMMAND_DRAWINDEXEDINSTANCED", Napi::FunctionPointer::Create(env, &NativeEngine::DrawIndexedInstanced)),
                 StaticValue("COMMAND_DRAW", Napi::FunctionPointer::Create(env, &NativeEngine::Draw)),
+                StaticValue("COMMAND_DRAWINSTANCED", Napi::FunctionPointer::Create(env, &NativeEngine::DrawInstanced)),
```
docEdub, line 552: Is this used on the JS side?

docEdub: oh, I see it now.

SergioRZMasson: I have a PR open for using this on JS side: https://github.com/BabylonJS/Babylon.js/pull/14331#discussion_r1333342154

But it checks first

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1697,13 +1699,16 @@ namespace Babylon
         m_boundFrameBufferNeedsRebinding.Set(*encoder, false);
     }
 
+    // Note: For legay reasons JS might call this function for instance drawing.
```
docEdub, line 1702:
```suggestion
    // Note: For legacy reasons JS might call this function for instance drawing. 
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1697,13 +1699,16 @@ namespace Babylon
...
+    // Note: For legay reasons JS might call this function for instance drawing. 
+    // In that case the instanceCount will be calculated inside the SetVertexBuffers method.
     void NativeEngine::DrawIndexed(NativeDataStream::Reader& data)
     {
         bgfx::Encoder* encoder{GetUpdateToken().GetEncoder()};
 
         const uint32_t fillMode = data.ReadUint32();
         const uint32_t indexStart = data.ReadUint32();
         const uint32_t indexCount = data.ReadUint32();
+
```
docEdub, line 1711:
```suggestion
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1714,6 +1719,26 @@ namespace Babylon
...
+        if (m_boundVertexArray != nullptr)
+        {
+            m_boundVertexArray->SetIndexBuffer(encoder, indexStart, indexCount);
+            m_boundVertexArray->SetVertexBuffers(encoder, 0, std::numeric_limits<uint32_t>::max(), instanceCount);
+        }
+
+        DrawInternal(encoder, fillMode);
+    }
+
+    // Note: For legay reasons JS might call this function for instance drawing.
```
docEdub, line 1740:
```suggestion
    // Note: For legacy reasons JS might call this function for instance drawing. 
```

docEdub (commented):
👍 LGTM

bghgary (approved):

SergioRZMasson:
Fixes #1189

bghgary:
> Fixes #1189

I think you have to put this in the description

bghgary:
I forgot to mention this. Are there any tests we can enable/bring over that will test this code?

Issue #1291, opened 2023-09-21
BabylonNative playground app pick points are inaccurate


docEdub:
The BabylonNative playground app pick points are inaccurate on some OS.

In this example the sphere should appear directly under the mouse pointer:

![Image](https://github.com/BabylonJS/BabylonNative/assets/78323754/52ca85cd-598d-4175-b139-cdb3bdebb733)

PR #1292, opened 2023-09-22
Fix playground pointer position

docEdub:
The BabylonNative playground app pick points are inaccurate on Win32, UWP, macOS and iOS.

This change fixes the issues and also adds mouse-move tracking on macOS.

Fixes https://github.com/BabylonJS/BabylonNative/issues/1291.

Tested on playground https://playground.babylonjs.com/#VXUJZA.

Tested for regressions per suggestions from @PolygonalSun on:
- Default playground for FreeCamera.
- https://playground.babylonjs.com/#SRZRWV#839 for ArcRotateCamera.
- https://playground.babylonjs.com/#H5PZ4Q for picking.
- https://playground.babylonjs.com/#SG9ZZB for gizmos.


bghgary (approved):
Seems good. We really need to put this code into the input plugin so that other people can use it.

PR #1299, opened 2023-09-27
Add support for mat3 to shader compiler

bghgary:


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1061,7 +1061,7 @@ namespace Babylon
 
         assert(matrix.size() == size * size);
 
-        if constexpr (size < 4)
+        if constexpr (size < 3)
```
docEdub, line 1064: Is this right?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1061,7 +1061,7 @@ namespace Babylon
...
         {
             std::array<float, 16> matrixValues{};
```
PolygonalSun, line 1066: Do we have to worry about any sort of side effects by storing a 3x3 matrix across a size 16 array?  I'm assuming that the untouched elements of the array start as 0 but I wanted to make sure.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: Yup. This is needed for mat3.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: The `{}` will initialize the array to the default value of `float` which is `0.0f`

PolygonalSun: Disregard, I misread this.

docEdub (approved):

PolygonalSun (approved):

PR #1301, opened 2023-09-29
External texture fixes for Apple

bghgary:
- Change `GetInfo` to static to avoid accidentally using class members
- Change `Update` function to only compare width and height and add missing assignment of `m_info` and `UpdateHandles` call

### DIFF
--- a/Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
+++ b/Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
@@ -18,12 +18,17 @@ namespace Babylon::Plugins
     {
         Info info;
         GetInfo(ptr, info);
-        if (info != m_info)
+
+        if (info.Width != m_info.Width || info.Height != m_info.Height || info.MipLevels != m_info.MipLevels)
         {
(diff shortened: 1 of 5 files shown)


Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
```diff
@@ -16,14 +16,9 @@ namespace Babylon::Plugins
 
     void ExternalTexture::Impl::Update(Graphics::TextureT ptr)
     {
-        Info info;
-        GetInfo(ptr, info);
-        if (info != m_info)
-        {
-            throw std::runtime_error{"Textures must have same info"};
-        }
```
bghgary, line 24: This comparison is flawed. It is okay to update even if the info is not the same.

docEdub (approved):

Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
bghgary: There is a missing update to `m_info` here.

PR #1307, opened 2023-10-04
Update glslang web min

SergioRZMasson:
Updated glslang to use our BabylonJS/glslang that has size reduction features.


Dependencies/CMakeLists.txt
```diff
@@ -71,6 +71,8 @@ set(ENABLE_HLSL OFF CACHE BOOL "")
 set(ENABLE_OPT OFF CACHE BOOL "")
 set(ENABLE_CTEST OFF CACHE BOOL "")
 set(SKIP_GLSLANG_INSTALL ON CACHE BOOL "")
+set(ENABLE_GLSLANG_WEBMIN ON CACHE BOOL "")
+set(ENABLE_GLSLANG_WEBMIN_DEVEL OFF CACHE BOOL "")
```
bghgary, line 75: Not part of this change, but I don't know why we are caching these variables. I don't think they need to be.

bghgary (approved):

Dependencies/CMakeLists.txt
SergioRZMasson: We need ENABLE_GLSLANG_WEBMIN to tell glslang to remove everything that is not required for Web, and it is Off by default on glslang side. We can remove ENABLE_GLSLANG_WEBMIN_DEVEL since it is Off by default on glslang.

PR #1309, opened 2023-10-09
Cache bgfx compilation

CedricGuillemet:


.github/jobs/win32.yml
```diff
@@ -40,12 +40,11 @@ jobs:
 
       - task: Cache@2
         inputs:
-          key: '"bgfx" | "$(Agent.OS)" | "$(Build.SourceVersion)"'
-          restoreKeys: |
-            "bgfx" | "$(Agent.OS)" | "$(Build.SourceVersion)"
-          path: build\Dependencies\bgfx.cmake\RelWithDebInfo
+          key: 'bgfx_$(Agent.OS)_$(Build.SourceVersion)'
```
ryantrem: Wouldn't this need to include the bgfx commit hash as well, since we'd want the cache to be invalidated if bgfx is updated?

CedricGuillemet: Yes, OS + commit hash. So, I'm not sure it will be possible to use cmake fetch content for cached dependencies.

CedricGuillemet:
Caching is not viable. using prebuild with releases instead.

PR #1311, opened 2023-10-10
Added configuration to enable/disable async shader compilation

SergioRZMasson:
Added a CMake variable that can control if async shader compilation will be used. Async shader compilation might be helpfull to be turned off when debugging shader related code or when performing validation tests in Debug mode.


SergioRZMasson:
Javascript checks for "createProgramAsync" availability. I think this is for backwards compatibility with older versions of Babylon Native.

PR #1313, opened 2023-10-13
Windows prebuilt

CedricGuillemet:
- This PR allows creating a new release for each new PR pushed into master branch.
- debug,release, release with debug infos for arm64, x86 and x64 binaries for win32 (more platforms later) are built
- These binaries are compressed into a .tgz and pushed as an asset for that particular release
- the .tgz also contains a cmake script that create cmake targets from prebuilt .lib


CedricGuillemet:
Closing for now.

PR #1316, opened 2023-10-19
fix npm install for android app

CedricGuillemet:
follow up https://forum.babylonjs.com/t/build-failed-on-windows-for-android-apk/45026/7
npm install was not called on Android because `PROJECT_IS_TOP_LEVEL` is false is Apps cmake is only evaluated when true.
Fixed by forcing Apps folder evaluation on Android.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ if(DEFINED EXTENSIONS_DIRS)
     endforeach()
 endif()
 
-if(BABYLON_NATIVE_BUILD_APPS)
+if(BABYLON_NATIVE_BUILD_APPS OR ANDROID)
     add_subdirectory(Apps)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ if(DEFINED EXTENSIONS_DIRS)
     endforeach()
 endif()
 
-if(BABYLON_NATIVE_BUILD_APPS OR ANDROID)
+if(BABYLON_NATIVE_BUILD_APPS)
     add_subdirectory(Apps)
 endif()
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -162,7 +162,7 @@ if(DEFINED EXTENSIONS_DIRS)
     endforeach()
 endif()
 
-if(BABYLON_NATIVE_BUILD_APPS)
+if(BABYLON_NATIVE_BUILD_APPS OR ANDROID)
```
bghgary, line 165: I think it might be better if the Android CMakeLists.txt explicitly calls `add_subdirectory` of the Apps folder or just calls npm directly in the Android CMakeLists.txt? Thoughts?

CedricGuillemet: I've spent some time thinking about this and didn't find anything better. I don't want to have 2 calls to npm because DRY. another call to add_subdirectory seems weird because there is a call just before on the parent directory. so, I would keep it like that for now with a comment.

bghgary: This is going to be bad for someone trying to consume Babylon Native from an Android project that doesn't care about apps (say BRN). They will not be able to explicitly specify `BABYLON_NATIVE_BUILD_APPS=OFF`.

CedricGuillemet: You are right. what about a cmake function defined in root cmakelists and called in apps\cmakelists and in android app cmake?

bghgary: I'm not sure. Doesn't seem much better than setting `BABYLON_NATIVE_BUILD_APPS=ON` to me.

CedricGuillemet: Works for me.

bghgary (approved):

PR #1319, opened 2023-10-30
Warning related fixes

bghgary:
- Re-enable disabled warnings in SPIRV-Cross
- Default to `/W3` for MSVC
- Stop disabling warnings in all libraries
- Fix warnings


SergioRZMasson (approved):
LGTM

bghgary:
Note that this will cause some warnings to show up in bgfx, glslang, and spirv-cross. We have to update the code to remove these warnings.

PR #1324, opened 2023-11-16
Updated glslang to enable GLSLANG_WEB_MIN

SergioRZMasson:
Changed glslang to point to the repo that is hosted by Babylon. This repo has GLSLANG_WEB_MIN flag working for all platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,8 +30,8 @@ FetchContent_Declare(base-n
     GIT_REPOSITORY https://github.com/azawadzki/base-n.git
     GIT_TAG 7573e77c0b9b0e8a5fb63d96dbde212c921993b4)
 FetchContent_Declare(glslang
-    GIT_REPOSITORY https://github.com/KhronosGroup/glslang.git
-    GIT_TAG 12.3.1)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -30,8 +30,8 @@ FetchContent_Declare(base-n
     GIT_REPOSITORY https://github.com/azawadzki/base-n.git
     GIT_TAG 7573e77c0b9b0e8a5fb63d96dbde212c921993b4)
 FetchContent_Declare(glslang
-    GIT_REPOSITORY https://github.com/KhronosGroup/glslang.git
-    GIT_TAG 12.3.1)
+    GIT_REPOSITORY https://github.com/BabylonJS/glslang.git
+    GIT_TAG main)
```
bghgary, line 34: We should not use `main`. This will cause the latest to be picked up. If we add additional code to `main`, it will get picked up immediately without testing. It would be better to use a commit id here since we don't have tags for our fork.

bghgary (changes requested):

bghgary (approved):

PR #1327, opened 2023-11-22
Static engine allocator

CedricGuillemet:
When shutting down the engine while having some textures, it's possible for NativeEngine to be disposed before the textures.
In this lambda:

https://github.com/BabylonJS/BabylonNative/blob/95240a59f2e1483d4c4b92b329b84ff333f0de84/Plugins/NativeEngine/Source/NativeEngine.cpp#L216

The imageContainer has still a ref to NativeEngine::m_allocator that gets invalid

https://github.com/bkaradzic/bimg/blob/5d61905eb1aa8074ee62553b4a7bd2dc16f4cf24/src/image.cpp#L3335

Setting m_allocator as static fixes the issue. It only contains a vtable so it's threadsafe.

### DIFF
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
@@ -115,7 +115,8 @@ namespace Babylon::Graphics
         void AddTexture(bgfx::TextureHandle handle, uint16_t width, uint16_t height, bool hasMips, uint16_t numLayers, bgfx::TextureFormat::Enum format);
         void RemoveTexture(bgfx::TextureHandle handle);
         TextureInfo GetTextureInfo(bgfx::TextureHandle handle);
-        static inline bx::DefaultAllocator allocator{};
+        static bx::AllocatorI& GetDefaultAllocator() { return m_allocator; }
+
     private:
         friend UpdateToken;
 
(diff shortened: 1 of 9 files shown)


Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
```diff
@@ -114,7 +115,7 @@ namespace Babylon::Graphics
         void AddTexture(bgfx::TextureHandle handle, uint16_t width, uint16_t height, bool hasMips, uint16_t numLayers, bgfx::TextureFormat::Enum format);
         void RemoveTexture(bgfx::TextureHandle handle);
         TextureInfo GetTextureInfo(bgfx::TextureHandle handle);
-
+        static inline bx::DefaultAllocator allocator{};
```
bghgary, line 118: nit: Instead of exposing the `bx::DefaultAllocator` type, maybe it should better to expose the `bx::AllocatorI` type?

I would suggest keeping this member private and adding an accessor function that returns `bx::AllocatorI&`. Remember that public members should be UpperCamelCase.

bghgary (approved):

Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
CedricGuillemet: It makes sense. Fixed.

PR #1332, opened 2023-12-05
Update SPIRV-Cross to get size reduction features

SergioRZMasson:
Update used SPIRV-Cross repo to point to the Babylon hosted one that has WEB_MIN cmake option. This should reduce the size of Babylon Native.


Apps/CMakeLists.txt
```diff
@@ -6,4 +6,4 @@ if((WIN32 AND NOT WINDOWS_STORE) OR (APPLE AND NOT IOS) OR (UNIX AND NOT ANDROID
     add_subdirectory(UnitTests)
 endif()
 
-npm(install --silent)
+#npm(install --silent)
```
bghgary, line 9: Undo

bghgary (approved):

PR #1334, opened 2023-12-18
Changed thread affinity checks to be optional

SergioRZMasson:
In some integrations the application might want to call graphics call from multiple threads. This PR adds the possibility of making thread affinity checks configurable by CMake.


CMakeLists.txt
```diff
@@ -66,6 +66,9 @@ project(BabylonNative)
 option(BABYLON_NATIVE_BUILD_APPS "Build Babylon Native apps." ${PROJECT_IS_TOP_LEVEL})
 option(BABYLON_NATIVE_INSTALL "Include the install target." ${PROJECT_IS_TOP_LEVEL})
 
+# WARNING: This is an experimental features. Only used it if you can ensure that you application will properly handle thread affinity.
```
bghgary, line 69:
```suggestion
# WARNING: This is experimental. Only use it if you can ensure that your application will properly handle thread affinity.
```

bghgary (approved):

PR #1347, opened 2024-01-25
gradle version update

CedricGuillemet:
Should fix a bunch of issues that start appearing with newer Android Studio
image

and
```
Unable to make field private final java.lang.String java.io.File.path accessible: module java.base does not "opens java.io" to unnamed module
```

### DIFF
--- a/.github/jobs/android.yml
+++ b/.github/jobs/android.yml
@@ -15,12 +15,6 @@ jobs:
       parameters:
         vmImage: ${{ parameters.vmImage }}
 
-    - task: JavaToolInstaller@0
-      inputs:
-        versionSpec: '17'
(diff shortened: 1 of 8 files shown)


ryantrem:
Even 7.1.1 is pretty old. 8.2.1 is the newest version supported by current stable releases of Android Studio. I think the latest SDK also requires 8+. Can we go all the way to 8.2.1?

CedricGuillemet:
> Even 7.1.1 is pretty old. 8.2.1 is the newest version supported by current stable releases of Android Studio. I think the latest SDK also requires 8+. Can we go all the way to 8.2.1?

Sure! Let's try

SergioRZMasson (approved):
LGTM

PR #1350, opened 2024-01-30
Spheres performance test

CedricGuillemet:


Apps/UnitTests/Shared/Tests.h
```diff
@@ -98,6 +98,91 @@ TEST(NativeAPI, LifeCycle)
...
+    runtime.Dispatch([&ready, &device, &nativeCanvas](Napi::Env env) {
+        device.AddToJavaScript(env);
+
+        Babylon::Polyfills::XMLHttpRequest::Initialize(env);
+        Babylon::Polyfills::Console::Initialize(env, [](const char* message, auto) {
+            printf("%s", message);
+            fflush(stdout);
+            });
+        Babylon::Polyfills::Window::Initialize(env);
+        nativeCanvas.emplace(Babylon::Polyfills::Canvas::Initialize(env));
```
bghgary, line 154: Why do you need canvas?

Apps/UnitTests/Shared/Tests.h
```diff
@@ -98,6 +98,91 @@ TEST(NativeAPI, LifeCycle)
...
 */
+
+TEST(Performance, Spheres)
+{
+    // create a bunch of sphere, does the rendering for a number of frames, log time it took
+    static const std::string script{ R"(
+        console.log("Setting up Performance test.");
+        var engine = new BABYLON.NativeEngine();
+        var scene = new BABYLON.Scene(engine);
+        function CreateSpheresAsync(scene) {
```
bghgary, line 109: Creating spheres does not need to be async. It is only async in the experience to make the code consistent.

Apps/UnitTests/Shared/Tests.h
```diff
@@ -98,6 +98,91 @@ TEST(NativeAPI, LifeCycle)
...
+            fflush(stdout);
+            });
+        Babylon::Polyfills::Window::Initialize(env);
+        nativeCanvas.emplace(Babylon::Polyfills::Canvas::Initialize(env));
+        Babylon::Plugins::NativeEngine::Initialize(env);
+        env.Global().Set("SetReady", Napi::Function::New(env, [&ready](const Napi::CallbackInfo& info)
+        {
+            Napi::Env env = info.Env();
+            ready.set_value(1);
+        }, "SetReady"));
```
bghgary, line 160: You don't need any of this stuff because creating spheres does not need to be async.

bghgary (changes requested):

Apps/UnitTests/Shared/Tests.h
CedricGuillemet: no but loading script is async. so, it's possible to finish rendering all frames while the spheres have not been created.

Apps/UnitTests/Shared/Tests.h
CedricGuillemet: Fixed

Apps/UnitTests/Shared/Tests.h
bghgary: Fair point.

Apps/UnitTests/Shared/Tests.h
```diff
@@ -98,6 +98,82 @@ TEST(NativeAPI, LifeCycle)
...
+    runtime.Dispatch([&ready, &device](Napi::Env env) {
+        device.AddToJavaScript(env);
+
+        Babylon::Polyfills::Console::Initialize(env, [](const char* message, auto) {
+            printf("%s", message);
+            fflush(stdout);
+            });
+        Babylon::Polyfills::Window::Initialize(env);
+        Babylon::Plugins::NativeEngine::Initialize(env);
+        env.Global().Set("SetReady", Napi::Function::New(env, [&ready](const Napi::CallbackInfo& info)
```
bghgary, line 147: nit: use JS conventions

 ```suggestion
        env.Global().Set("setReady", Napi::Function::New(env, [&ready](const Napi::CallbackInfo& info)
```

File: Apps/UnitTests/Shared/Tests.h [hunk 5]
```diff
@@ -98,6 +98,82 @@ TEST(NativeAPI, LifeCycle)
...
+    {
+        device.StartRenderingCurrentFrame();
+        update->Start();
+        update->Finish();
+        device.FinishRenderingCurrentFrame();
+    }
+    // Stop measuring time
+    const auto stop = std::chrono::high_resolution_clock::now();
+    const auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(stop - start);
+    const float durationSeconds = float(double(duration.count()) / 1000.);
```
bghgary, line 173: nit: Why convert to double than back to float? Can you not convert to float directly?

Apps/UnitTests/Shared/Tests.h
```diff
@@ -98,6 +98,82 @@ TEST(NativeAPI, LifeCycle)
...
+    printf("Duration is %f seconds.\n", durationSeconds);
```
bghgary, line 174: nit: Consider putting a flush after this also.

bghgary (approved):

Apps/UnitTests/Shared/Tests.h
CedricGuillemet: I want to keep precision until the /1000.

bghgary: Divide by 1000 shouldn't change the precision much, but I'm okay if you want to keep it.

CedricGuillemet: Fixed

Apps/UnitTests/Shared/Tests.h
CedricGuillemet: Fixed.

Issue #1351, opened 2024-01-31
Implement `napi_add_finalizer` function


mattbirman:
**For questions, please use the [forum](https://forum.babylonjs.com)**.

**Is your feature request related to a problem? Please describe.**
We are using the `napi` [Rust package](https://docs.rs/napi/latest/napi/) and calling [add_finalizer](https://docs.rs/napi/latest/src/napi/js_values/object.rs.html#31) on `JsObject`. This call fails at runtime because [napi_add_finalizer](https://nodejs.org/api/n-api.html#napi_add_finalizer) is not implemented in BabylonJS.

**Describe the solution you'd like**
An native implementation of `napi_add_finalizer` in BabylonJS.

**Describe alternatives you've considered**
No alternatives to this function. We have considered not adding the finalizer to the object and allowing the object to not be garbage-collected.

**Additional context**
N/A

CedricGuillemet:
Can we close this issue @bghgary ?

PR #1352, opened 2024-02-01
Update JsRuntimeHost to pick up napi_add_finalizer for JSC

bghgary:
See #1351. This a stopgap solution. Proper implementation needs to happen after https://github.com/BabylonJS/JsRuntimeHost/pull/73 is merged.


CedricGuillemet (approved):
15 min for timeout is a bit short now because of the spheres test added this week in unittests.

bghgary:
Feel free to bump it. It would be good to keep it under 15 if possible though.

PR #1325, opened 2023-11-17
Build consistency fixes

bghgary:

PR #1356, opened 2024-02-08
Fix cmake install target

bghgary:
Changes from #1325 caused the install files to have extra files. This is a simpler way to do it.

### DIFF
--- a/Install/Install.cmake
+++ b/Install/Install.cmake
@@ -16,10 +16,14 @@ function(install_targets)
     endforeach()
 endfunction()
 
-function(install_include_for_target)
-    get_target_property(target_interface_include_directories ${ARGV0} INTERFACE_INCLUDE_DIRECTORIES)
-    foreach(include_directory IN LISTS target_interface_include_directories)
-        install(DIRECTORY "${include_directory}/${ARGV1}" TYPE INCLUDE)
+function(install_include_for_targets)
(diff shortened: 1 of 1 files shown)


CedricGuillemet (approved):

CedricGuillemet:
Why was it not detected by the install test app? how can we improve the test?

bghgary:
> Why was it not detected by the install test app? how can we improve the test?

It's not detected because the layout is proper, but there were extra files being added. I'm not sure how easy it would be to test.

Install/Install.cmake
```diff
@@ -11,23 +11,19 @@ function(install_targets)
...
-function(install_include_for_target)
-    get_target_property(TARGET_INCLUDE_DIRECTORIES ${ARGV0} INCLUDE_DIRECTORIES)
-    foreach(include_directory IN LISTS TARGET_INCLUDE_DIRECTORIES)
-        if(${ARGC} STREQUAL "1")
-            install_include("${include_directory}")
-        else()
-            install_include("${include_directory}/${ARGV1}")
-        endif()
+function(install_include_for_targets)
+    foreach(target IN LISTS ARGN)
```
CedricGuillemet, line 20: Is that loop necessary?

bghgary: No, but it's more consistent with the other functions.

PR #1363, opened 2024-02-22
Fix flash and image orientation for iOS image capture

ryantrem:
`ImageCapture` was implemented for iOS a while back, but had a few outstanding bugs that are addressed in this PR:
1. Flash modes were not working - there were actually two bugs here, the first being that we were trying to get supported flash modes from the `AVPhotoCaptureOutput` before we actually created the `AVPhotoCaptureOutput` (oops), and second there was a missing line to increment the current index when populating the flash modes array.
2. There was no code to deal with image rotation, so the created image was always sideways (I believe based on the orientation of the physical camera sensor relative to the orientation of the device).

So the changes here are to fix the issues of bullet 1 and add code for bullet 2.

I also made some small changes to experience.js to reflect the missing support for invertY on Textures in BN which otherwise makes it confusing why the code is the way it is.

EDIT: Regarding image rotation, I realized my fix for this was not really correct. What actually happens with ImageCapture in the browser (at least in Chrome on Android), which is similar to what camera apps typically do, is that the image data itself is not rotated when the photo is taken. Rather, it just includes correct Orientation metadata (e.g. jpeg metadata), and code that reads that image is supposed to respect the Orientation metadata. For example, the Photos app on iOS or Windows reads this metadata when loading a jpeg and does the rotation when displaying the image. Similarly, when a jpeg is used as a texture in Babylon.js, the loading path relies on the browser respecting the Orientation metadata when loading the image data. This means what really needs to happen is:
1. We need to ensure the correct Orientation metadata is included in photo capture (achieved by the new change that sets `photoOutputConnection.videoOrientation`).
2. We need the Babylon Native image parsing code path to respect the Orientation metadata, which bimg already gives us, we were just ignoring it (new change that rotates the image data during image parsing if needed).

### DIFF
--- a/Plugins/NativeCamera/Source/Apple/CameraDevice.mm
+++ b/Plugins/NativeCamera/Source/Apple/CameraDevice.mm
@@ -3,7 +3,6 @@
 #endif
 
 #import <MetalKit/MetalKit.h>
-//#import <Photos/Photos.h>
 #include <napi/napi.h>
 #include <bgfx/bgfx.h>
 #include <bgfx/platform.h>
@@ -1001,7 +1000,7 @@ namespace Babylon::Plugins
 
         // Saving the photo to storage can be helpful for testing and diagnosing any photo issues. To do so:
         // 1. In the Playground target, go to Build Phases, scroll down to Link Binary with Libraries, and add Photos.framework
-        // 2. Uncomment #import <Photos/Photos.h> at the top of this file
+        // 2. Add #import <Photos/Photos.h> at the top of this file
         // 3. Uncomment the block below
         /*
         [[PHPhotoLibrary sharedPhotoLibrary] performChanges:^{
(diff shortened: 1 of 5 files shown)


docEdub (approved):
👍

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -3,6 +3,7 @@
 #endif
 
 #import <MetalKit/MetalKit.h>
+//#import <Photos/Photos.h>
```
docEdub, line 6: ?

ryantrem: This was in conjunction with a comment later in the file, but you're right, if someone sees this here they aren't going to know it goes with the comment later and just delete. I will remove it and update the comment to just say to add this header instead of uncommenting it.

bghgary (approved):

PR #1366, opened 2024-02-29
Allow device loss handle

SergioRZMasson:
This PR implements changes to Babylon Native that allow it to properly handle graphics device loss scenarios. When the graphics device currently been used by Babylon Native gets invalidated the correct workflow will be to call ```Babylon::Graphics::Device::DisableRendering``` recreate the graphics device infrastructure, use the new API ```Babylon::Graphics::Device::UpdateDevice``` to  give the new device to Babylon Native and than call ```Babylon::Graphics::Device::EnableRendering``` to start bgfx again with a new device. 

This also requires Babylon.js to be notified that the current device context was lost and re-created. This can be done by calling:

```
   engine._contextWasLost = true;
   engine._restoreEngineAfterContextLost(() => { });
``` 

This will force Babylon.js to recreate all graphics resources again and destroy the old ones. Since some bgfx resources have their lifetime bound to Babylon.js managed objects, it is important to check that the current instance of bgfx running when the resources get destroyed is the same as the one they were created with, for that I'm now adding a ```graphicsId``` attribute to those objects to hold that information. If the graphicsId been used by the Graphics device is no longer the same that existed when the object was created, this means that bgfx was reset and the handler is no longer valid.

### DIFF
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
@@ -65,6 +65,7 @@ namespace Babylon::Graphics
 
 
         bool m_disposed{false};
+        uintptr_t m_graphicsID;
 
         Rect GetBgfxScissor(float x, float y, float width, float height) const;
         void SetBgfxViewPortAndScissor(bgfx::Encoder& encoder, const Rect& viewPort, const Rect& scissor);
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
@@ -65,7 +65,7 @@ namespace Babylon::Graphics
 
 
         bool m_disposed{false};
-        uintptr_t m_graphicsID;
+        uintptr_t m_deviceID;
 
         Rect GetBgfxScissor(float x, float y, float width, float height) const;
         void SetBgfxViewPortAndScissor(bgfx::Encoder& encoder, const Rect& viewPort, const Rect& scissor);
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/Texture.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/Texture.h
@@ -46,6 +46,6 @@ namespace Babylon::Graphics
         bgfx::TextureFormat::Enum m_format{bgfx::TextureFormat::Enum::Unknown};
         uint64_t m_flags{BGFX_TEXTURE_NONE};
         uint32_t m_samplerFlags{BGFX_SAMPLER_NONE};
-        uintptr_t m_graphicsID;
+        uintptr_t  m_deviceID;
     };
 }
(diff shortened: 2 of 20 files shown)


Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
```diff
@@ -65,6 +65,7 @@ namespace Babylon::Graphics
 
 
         bool m_disposed{false};
+        uintptr_t m_graphicsID;
```
bghgary, line 68: nit: multiple places
```suggestion
        uintptr_t m_deviceID;
```

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -49,6 +50,11 @@ namespace Babylon::Graphics
         DisableRendering();
     }
 
+    uintptr_t DeviceImpl::GetId()
+    {
+       return g_bgfxId;
```
bghgary, line 55: Is this right? What if someone creates another device while an existing device is still alive? Shouldn't this be stored as a member of DeviceImpl?

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -134,7 +135,7 @@ namespace Babylon::Graphics
             return;
         }
 
-        if (bgfx::isValid(m_handle))
+        if (bgfx::isValid(m_handle) && m_graphicsID == Babylon::Graphics::Device::GetID())
```
bghgary, line 138: nit: Can be simplified since we are already in the namespace? Ditto, elsewhere.

```suggestion
        if (bgfx::isValid(m_handle) && m_graphicsID == Device::GetID())
```

Core/Graphics/Source/Texture.cpp
```diff
@@ -1,16 +1,24 @@
...
 
+#include "Babylon/Graphics/Device.h"
+
 namespace Babylon::Graphics
 {
+    Texture::Texture()
+       : m_graphicsID{Babylon::Graphics::Device::GetID()}
+    {
+
+    }
```
bghgary, line 12: nit:
```suggestion
    {
    }
```

bghgary (approved):

Core/Graphics/Source/DeviceImpl.cpp
SergioRZMasson: I don't think so, because this is a static method. Also, bgfx would not allow another device to be created while the first one is alive, bgfx only supports the usage of a single graphics device at each time.

Core/Graphics/InternalInclude/Babylon/Graphics/Texture.h
```diff
@@ -46,5 +46,6 @@ namespace Babylon::Graphics
         bgfx::TextureFormat::Enum m_format{bgfx::TextureFormat::Enum::Unknown};
         uint64_t m_flags{BGFX_TEXTURE_NONE};
         uint32_t m_samplerFlags{BGFX_SAMPLER_NONE};
+        uintptr_t  m_deviceID;
```
bghgary, line 49: nit:
```suggestion
        uintptr_t m_deviceID;
```

PR #1368, opened 2024-03-07
fix: cleanup view controller

okwasniewski:
Hey, 

This PR cleans up view controller in iOS playground from explicit unwraps. It should give a better example of how people can use BabylonNative in their projects

### DIFF
--- a/Apps/Playground/iOS/ViewController.swift
+++ b/Apps/Playground/iOS/ViewController.swift
@@ -1,8 +1,6 @@
 import UIKit
 import MetalKit
-
 import UIKit
-import UIKit.UIGestureRecognizerSubclass
 
 /**
   * A very simple gesture recognizer.  All that it does is to emulate the functionality found in other platforms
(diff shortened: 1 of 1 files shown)


okwasniewski:
@microsoft-github-policy-service agree [company="Callstack"]

okwasniewski:
@microsoft-github-policy-service agree company="Callstack"

Apps/Playground/iOS/ViewController.swift
```diff
@@ -29,26 +27,24 @@ class UIBabylonGestureRecognizer: UIGestureRecognizer {
...
-                let loc = touch.location(in: view)
-                
-                _onTouchDown(Int32(deviceSlot!), Int32(loc.x), Int32(loc.y))
-            }
+          if let deviceSlot = _activeTouchIds.firstIndex(of: -1) {
+              _activeTouchIds[deviceSlot] = touch.hash
+              let loc = touch.location(in: view)
+              
+              _onTouchDown(Int32(deviceSlot), Int32(loc.x), Int32(loc.y))
+          }
```
bghgary, line 35: nit: Some spacing issues here. Looks like your editor is set to 2, but the file uses 4, so there are some inconsistencies thoughout the file.

Apps/Playground/iOS/ViewController.swift
```diff
@@ -29,26 +27,24 @@ class UIBabylonGestureRecognizer: UIGestureRecognizer {
...
         for touch in touches {
-            let deviceSlot = _activeTouchIds.firstIndex(of: -1)
-            
-            if (deviceSlot != nil) {
-                _activeTouchIds[deviceSlot!] = touch.hash
-                let loc = touch.location(in: view)
-                
-                _onTouchDown(Int32(deviceSlot!), Int32(loc.x), Int32(loc.y))
-            }
+          if let deviceSlot = _activeTouchIds.firstIndex(of: -1) {
```
Allan121, line 30: nit: generally considered better practice to use guards to exist early.
```
guard let deviceSlot = _activeTouchIds.firstIndex(of: -1) else { continue } 
```

I won't point them our separately, but this would apply to the added if statements below as well.

Allan121 (approved):
Minor nit, but otherwise LGTM 👍

bghgary (approved):
Thanks @Allan121 for reviewing. Since @Allan121 is much more versed on Swift and he says it's good, I'm also good. :-)

Apps/Playground/iOS/ViewController.swift
okwasniewski: Good catch! Fixed 👍🏻

PR #1370, opened 2024-03-20
Notify JS about bgfx restart

SergioRZMasson:
Added an API no notify JS that the graphics context (bgfx) has been re-created. This API call will allow JS to re-create all GPU resources accordingly.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -402,6 +402,7 @@ namespace Babylon
 
     void NativeEngine::Initialize(Napi::Env env)
     {
+
```
docEdub, line 405: Is this new line intentional?

docEdub (approved):

PR #1373, opened 2024-03-28
Update to Babylon.js 7.6.2

bghgary:
Updating to Babylon.js version 7.0.0 reveals a whole slew of issues when running the validation tests because [#14868](https://github.com/BabylonJS/Babylon.js/pull/14868) broke the render loop stopping behavior when running in native. Native does not implement a way to cancel a requested animation frame. This missing canceling functionality should be added to native someday, but, in the meantime, [#15086](https://github.com/BabylonJS/Babylon.js/pull/15086) will make it behave like it did before.

These changes fix potential issues in the code regardless of what happened on the JS side.

- Update to latest JsRuntimeHost with lots of fixes.
- Call SetRenderResetCallback with null to clear the callback that can cause a crash on shutdown.
- Update a few validation tests to not use render count.
- Update scripts to use `const`/`let` instead of `var`.
- Update validation script to handle some errors better and clean up dead code.
- Add unhandled exception handler to Playground app so that validation test will report correct errors and exit correctly.
- Change core graphics device implemenation to update bgfx state right before requesting screenshots. This will ensure the screenshot is the right size if the resolution has changed.
- Fix TestUtils:WritePNG to return an error when the byte length is an unexpected size instead of silently failing.
- Change TestUtils::SetTitle to update the HWND title on a background thread to avoid dead lock with the main thread.
- Update dynamic texture clip test to newest version that will fail if the feature is not working.


ryantrem (approved):

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -295,17 +298,17 @@ namespace Babylon::Graphics
 
     void DeviceImpl::SetHardwareScalingLevel(float level)
     {
-        if (level <= std::numeric_limits<float>::epsilon())
+        if (FuzzyEqual(level, 0.0f))
```
bghgary, line 301:
```suggestion
        if (level <= std::numeric_limits<float>::epsilon())
```

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -369,12 +372,11 @@ namespace Babylon::Graphics
             bgfx::setPlatformData(m_state.Bgfx.InitState.platformData);
 
             // Ensure bgfx rebinds all texture information.
-            bgfx::discard(BGFX_DISCARD_ALL);
+            bgfx::discard();
 
             auto& res = m_state.Bgfx.InitState.resolution;
             bgfx::reset(res.width, res.height, res.reset);
             bgfx::setViewRect(0, 0, 0, static_cast<uint16_t>(res.width), static_cast<uint16_t>(res.height));
-            bgfx::frame();
```
bghgary, line 377: Hmm, not sure why I removed this. Need to double check.

Plugins/TestUtils/Source/Win32/TestUtilsImpl.cpp
```diff
@@ -31,16 +33,21 @@ namespace Babylon::Plugins::Internal
...
     {
-        const auto title = info[0].As<Napi::String>().Utf8Value();
-        SetWindowTextA(m_implData->m_window, title.c_str());
+        // SetWindowText sends a window message synchronously and cannot be called the JS thread with the way the code
+        // is currently set up. If the main thread is calling FinishRenderingCurrentFrame, this will hang forever as
+        // the main message loop is not pumping. Once we fix the rendering code to never block the main thread, this
+        // threadpool scheduling will no longer be necessary.
+        arcana::make_task(arcana::threadpool_scheduler, arcana::cancellation::none(), [hWnd = m_implData->m_window, title = info[0].As<Napi::String>().Utf8Value()] {
+            SetWindowTextA(hWnd, title.c_str());
+        });
```
bghgary, line 50: Don't need to use tasks here. Just use threadpool_scheduler directly.

ryantrem (approved):

ryantrem (approved):

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -421,8 +414,8 @@ namespace Babylon::Graphics
         // Automatically end bgfx encoders.
         EndEncoders();
 
-        // Discard everything if the bgfx state is dirty.
-        DiscardIfDirty();
+        // Update bgfx state if necessary.
+        UpdateBgfxState();
```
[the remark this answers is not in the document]
bghgary: Checked with Android and iOS with the latest fixes for iOS XR.

PR #1374, opened 2024-04-10
Use new JsRuntimeHost `BABYLON_API` calling convention macro

docEdub:
Without explicit calling convention modifers, consumers are required to compile with the same calling convention used in the released binaries. If a consumer compiles with the `__stdcall` calling convention, the linker won't resolve the function calls because the released binaries are built with the default `__cdecl` calling convention.

This change fixes the issue using an overridable `BABYLON_API` macro defined as `__cdecl` by default on Windows platforms or defined as empty on non-Windows platforms.

See also https://github.com/BabylonJS/JsRuntimeHost/pull/86.

### DIFF
--- a/Install/Install.cmake
+++ b/Install/Install.cmake
@@ -63,6 +63,8 @@ install_targets(UrlLib)
 # ----------------
 # Core
 # ----------------
+install_include_for_targets(Foundation)
+
 install_targets(JsRuntime)
 install_include_for_targets(JsRuntime)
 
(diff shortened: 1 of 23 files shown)


Install/Install.cmake
```diff
@@ -63,6 +63,8 @@ install_targets(UrlLib)
 # ----------------
 # Core
 # ----------------
+install_include_for_targets(Foundation)
```
docEdub, line 66: Babylon/Api.h was added to a new CMake header-only interface project in the JsRuntimeHost repo. This the CMake install target change needed for installing Babylon/Api.h along with the other required headers.

SergioRZMasson (approved):
LGTM

PR #1376, opened 2024-05-03
Fix regression with vertex buffer promote to float logic

bghgary:


bghgary:
Current plan is to abandon this PR and wait for @Popov72 to fix it on the JS side.

bghgary:
Superceded by #1382

PR #1377, opened 2024-05-03
Update to latest bgfx.cmake

bghgary:
See https://forum.babylonjs.com/t/pixel-8-pro-crash/48956. Specifically, the fix that makes the code work is https://github.com/bkaradzic/bgfx/pull/3187.

This change also:
- Cleans up the Android build such that the build output goes to the `Build/Android` folder.
- Update CI to latest Xcode since the old compilers no longer compile with latest bgfx.

### DIFF
--- a/.github/jobs/ios.yml
+++ b/.github/jobs/ios.yml
@@ -1,8 +1,7 @@
 parameters:
   name: ''
   vmImage: ''
-  xCodeVersion: ''
-  macOSCodename: ''
+  xcodeVersion: ''
   deploymentTarget: '12'
 
 jobs:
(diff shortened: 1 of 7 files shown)


SergioRZMasson (approved):
LGTM

bghgary:
> 👍 LGTM
> 
> I'm wondering if this change makes [BabylonJS/bgfx#35](https://github.com/BabylonJS/bgfx/pull/35) unnecessary, now. If so then we should revert it so we're more closely sync'd with upstream bgfx.

I guess the question is whether we care about supporting older versions of Xcode. My guess is most people will be on newer versions.

PR #1378, opened 2024-05-10
Fix ownership of color texture when creating a frame buffer

bghgary:
The changes in https://github.com/BabylonJS/Babylon.js/pull/15052 exposes a problem in native where the output texture of the BRDF post process will be destroyed causing a crash in bgfx.

In the previous code, the `NativeEngine::CreateFrameBuffer` took ownership of the color texture passed in as an argument. This is wrong because the texture might still be needed after the frame buffer dies. Specifically, [the BRDF post process that expands the RGBD env texture](https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Misc/rgbdTextureTools.ts#L98) is where it fails.

The new code will not take ownership of the color texture but will still delete the generated depth/stencil texture in the finalizer if necessary. The calling code will destroy the color texture when appropriate.


SergioRZMasson (approved):
LGTM

PR #1382, opened 2024-05-21
Support non-normalized, non-float vertex buffers and unaligned vertex buffers

bghgary:
This is a replacement for #1376. The original thinking was that we will promote the non-float attributes to float which was a complex and poor performance solution. The better solution is to change the shader to take non-float attributes. Fortunately, this shader processing is already happening for WebGPU on the JS side. @Popov72 [factored out the shader processor](https://github.com/BabylonJS/Babylon.js/pull/15107) on the JS side for WebGPU to also work with Babylon Native, which also includes code that aligns the buffers to 4-byte boundaries. This native PR ties everything together and removes all the promote to float logic we had earlier to try to handle non-float attributes.

This change includes updating Babylon.js to 7.7.2. Using an older version will result in a `console.error ` if non-normalized, non-float vertex buffers are used.


Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -67,22 +24,31 @@ namespace Babylon
...
-    void VertexBuffer::Update(const gsl::span<uint8_t> bytes, size_t byteOffset)
+    void VertexBuffer::Update(gsl::span<const uint8_t> bytes, size_t byteOffset)
     {
         if (!m_dynamic)
         {
             throw std::runtime_error{"Cannot update non-dynamic vertex buffer"};
         }
 
-        if (m_handle.has_value())
+        if (bgfx::isValid(m_dynamicHandle))
```
ryantrem, line 51: Why do we sometimes check m_handle and sometimes m_dynamicHandle? In some of the code it seems like a call to bgfx::isValid with either one is equivalent, and in other places that doesn't seem like the case.

Plugins/NativeEngine/Source/VertexBuffer.h
```diff
@@ -42,69 +42,20 @@ namespace Babylon
...
-
-        struct StreamInfo
+        union
         {
-            const AttributeInfo& Attribute;
-            std::optional<Handle> PromoteToFloatsHandle;
-            uint32_t StartVertexOffset;
-            bgfx::VertexLayoutHandle LayoutHandle;
+            bgfx::VertexBufferHandle m_handle{bgfx::kInvalidHandle};
+            bgfx::DynamicVertexBufferHandle m_dynamicHandle;
```
ryantrem, line 56: This isn't new code, but the m_dynamic + m_handle/m_dynamicHandle union could collectively be replaced with a single std::variant, right?

Plugins/NativeEngine/Source/VertexBuffer.h
```diff
@@ -42,69 +42,20 @@ namespace Babylon
 
         static void BuildInstanceDataBuffer(bgfx::InstanceDataBuffer& instanceDataBuffer, const std::map<bgfx::Attrib::Enum, InstanceInfo>& instances, uint32_t instanceCount);
 
-    private:
-        void Build(uint32_t numVertices);
-
-        class Handle final
```
ryantrem, line 48: Is this just removed because there aren't enough places doing this to make it worth encapsulating into its own class?

Plugins/NativeEngine/Source/IndexBuffer.cpp
```diff
@@ -66,45 +66,42 @@ namespace Babylon
...
-        }
-        else
-        {
-            encoder->setIndexBuffer(m_handle, firstIndex, numIndices);
+            auto* bytesPtr = new decltype(m_bytes){std::move(m_bytes)};
+            const bgfx::Memory* memory = bgfx::makeRef(bytesPtr->data(), static_cast<uint32_t>(bytesPtr->size()), releaseFn, bytesPtr);
+
+            if (m_dynamic)
+            {
+                m_dynamicHandle = bgfx::createDynamicIndexBuffer(memory, m_flags);
```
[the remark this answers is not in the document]
bghgary: Yes, it's a union, so there is no need to check both.

Plugins/NativeEngine/Source/IndexBuffer.cpp
```diff
@@ -66,45 +66,42 @@ namespace Babylon
...
+            }
+            else
+            {
+                m_handle = bgfx::createIndexBuffer(memory, m_flags);
+            }
+
+            if (!bgfx::isValid(m_handle))
+            {
+                throw std::runtime_error{"Failed to create index buffer"};
```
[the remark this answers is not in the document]
bghgary: Hmm, good question. We probably should free it.

Plugins/NativeEngine/Source/VertexArray.h
```diff
@@ -26,7 +26,23 @@ namespace Babylon
...
         IndexBuffer* m_indexBuffer{};
-        std::set<VertexBuffer*> m_vertexBuffers;
+
+        struct VertexBufferRecord
+        {
+            VertexBuffer* Buffer{};
+            const uint32_t Offset{};
+            const bgfx::VertexLayoutHandle LayoutHandle{};
+
+            explicit VertexBufferRecord(VertexBuffer* buffer, const uint32_t offset, bgfx::VertexLayoutHandle layoutHandle)
```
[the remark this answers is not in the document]
bghgary: No good reason. I will remove it.

Plugins/NativeEngine/Source/VertexBuffer.cpp
bghgary: It's the same because it is a union. I use m_handle when we don't care which one it is and use m_dynamicHandle if we check it's dynamic. It doesn't actually matter.

Plugins/NativeEngine/Source/VertexBuffer.h
bghgary: Right. I could add it back, but it seems overkill.

Plugins/NativeEngine/Source/VertexBuffer.h
bghgary: Using std::variant is really annoying in this case because bgfx uses overloads for handles. We tried using std::variant initially and it was hard to write the code.

ryantrem (approved):

ryantrem (approved):

PR #1389, opened 2024-06-06
Android D24S8 for depthbuffer

### LINKED ISSUE #1381: Post processing graphics glitch
https://forum.babylonjs.com/t/post-processing-graphics-glitch/49523

CedricGuillemet:
follow up https://forum.babylonjs.com/t/post-processing-graphics-glitch/49523/35
image
tldr: glitches with Arm Mali GPU and 32b depth buffer.
fix: replace D32 with D24S8

fixes #1381

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -1829,7 +1829,15 @@ namespace Babylon
             }
 
             auto flags = BGFX_TEXTURE_RT_WRITE_ONLY | RenderTargetSamplesToBgfxMsaaFlag(samples);
+#ifdef ANDROID
+            // On Android with Mali GPU (Oppo Find x5 lite, Google Pixel 8, Samsung Galaxy Tab Active 3, ...)
+            // D32 depth buffer gives glitches. Everything is fine with D24S8.
+            // see https://forum.babylonjs.com/t/post-processing-graphics-glitch/49523
+            // As 24bits should be enough for 99.99% cases, defaulting to that format on Android.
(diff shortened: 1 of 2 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1829,7 +1829,15 @@ namespace Babylon
             }
 
             auto flags = BGFX_TEXTURE_RT_WRITE_ONLY | RenderTargetSamplesToBgfxMsaaFlag(samples);
+#ifdef ANDROID
+            // On Android with Mali GPU (Oppo Find x5 lite, Google Pixel 8, Samsung Galaxy Tab Active 3, ...)
+            // D32 depth buffer gives glitches. Everything is fine with D24S8.
+            // see https://forum.babylonjs.com/t/post-processing-graphics-glitch/49523
+            // As 24bits should be enough for 99.99% cases, defaulting to that format on Android.
+            const auto depthStencilFormat{bgfx::TextureFormat::D24S8};
```
ryantrem, line 1837: I guess the 8 stencil bits just get ignored then if generateStencilBuffer is false?

CedricGuillemet: yes. it's there but not used.

ryantrem (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: It would also be interesting to see what happens in a browser. Does the same issue happen on the problem devices in a browser?

PR #1391, opened 2024-06-10
R16 Texture format support (temporary fix)

CedricGuillemet:
Temporary fix for a user until this is done and replaces this change: https://github.com/BabylonJS/BabylonNative/issues/1392

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -234,6 +234,7 @@ namespace Babylon
         bimg::ImageContainer* PrepareImage(bx::AllocatorI& allocator, bimg::ImageContainer* image, bool invertY, bool srgb, bool generateMips)
         {
             assert(
+                image->m_format == bimg::TextureFormat::R16 ||
                 image->m_format == bimg::TextureFormat::RGB8 ||
                 image->m_format == bimg::TextureFormat::RGBA8 ||
                 image->m_format == bimg::TextureFormat::RGBA16 ||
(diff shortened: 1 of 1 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -266,7 +267,7 @@ namespace Babylon
                     image = bimg::imageConvert(&allocator, bimg::TextureFormat::RGBA8, *image, false);
                     bimg::imageFree(oldImage);
                 }
-                else if (image->m_format == bimg::TextureFormat::RGBA16)
+                else if (image->m_format == bimg::TextureFormat::RGBA16 || image->m_format == bimg::TextureFormat::R16)
```
ryantrem, line 270: I guess this quadruples the size of the texture with a lot of wasted data? Is it possible to support R16 more directly?

CedricGuillemet: more than quadruple: *8 !! It's possible (I guess) to update bimg downsample to support R16.

bghgary: bgfx only supports generating mips of certain formats. Maybe that can be updated to support more formats.

ryantrem: Wait maybe I'm confused... isn't R16 16 bits per pixel, and RGB32F is 32 bits per pixel, so we are actually doubling the storage but losing half the precision (8 bit red channel instead of 16 bit red channel)? Someone help me with my 5th grade math plz. 🤣

CedricGuillemet: Definitely! I'm waiting for feedback from the user and I'll continue the work accordingly.

CedricGuillemet: > Wait maybe I'm confused... isn't R16 16 bits per pixel, and RGB32F is 32 bits per pixel, so we are actually doubling the storage but losing half the precision (8 bit red channel instead of 16 bit red channel)? Someone help me with my ~5th~ ~4th~ 3rd grade math plz. 🤣

R16 is 16bits per pixel or 2 bytes per texel. RGBA32F is 4 floats or 4 * 4bytes = 16 bytes. RGBA8 is indeed 32bits

ryantrem (approved):

PR #1393, opened 2024-06-13
Async shader compilation + shutdown crash repro

CedricGuillemet:
Shutdown with async shader compilation crash repro. 
Left as commented code: I was unable to expect a crash with GoogleTest and adding a `sleep` call to get sufficient time for shaders to be compiled was unreliable.

### DIFF
--- a/Apps/UnitTests/Shared/Shared.cpp
+++ b/Apps/UnitTests/Shared/Shared.cpp
@@ -98,81 +98,43 @@ TEST(JavaScript, All)
     EXPECT_EQ(exitCode, 0);
 }
 
-/*
-This test does a serie of initialization and shutdowns.
-It needs the shutdown PR to be merged before running properly.
-TEST(NativeAPI, LifeCycle)
-{
-    for (int cycle = 0; cycle < 20; cycle++)
-    {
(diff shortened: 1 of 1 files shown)


Apps/UnitTests/Shared/Shared.cpp
```diff
@@ -99,39 +99,77 @@ TEST(JavaScript, All)
...
         });
 
-        Babylon::ScriptLoader loader{runtime};
-        loader.LoadScript("app:///Scripts/babylon.max.js");
-        loader.LoadScript("app:///Scripts/babylonjs.materials.js");
+    Babylon::ScriptLoader loader{ runtime };
+    loader.LoadScript("app:///Scripts/babylon.max.js");
+    loader.LoadScript("app:///Scripts/babylonjs.materials.js");
+    loader.Eval(R"(
+            function CreateBoxAsync(scene) {
```
ryantrem, line 141: Maybe add a comment explaining why this needs to be an async function rather than just synchronous setup code? Not clear to me.

ryantrem (approved):

CedricGuillemet:
Closing in favor of https://github.com/BabylonJS/BabylonNative/pull/1397

PR #1395, opened 2024-06-18
Refactor XR plugin

CedricGuillemet:
Splitting NativeXR into smaller chunks


ryantrem:
I assume this is just moving code around, not making any actual changes? Hard to tell from the diff :)

ryantrem (approved):

PR #1401, opened 2024-07-08
Test installation iOS/Linux/MacOS

CedricGuillemet:
forum: https://forum.babylonjs.com/t/integrate-babylon-native-in-existing-swift-project/35886/28?u=cedric

- Because of CMake and Xcode, there are incompatilities with `install` scheme on iOS. So, build needs to be started with the playground, then using `cmake --install` . Also iOS install needs absolute prefix ( https://gitlab.kitware.com/cmake/cmake/-/issues/21282 )
- MacOS install scheme is fine but I could not enable scheme creation on that target. So, it's been enabled for all targets.
- `if (TARGET ...` in cmake script test if the target exists. But if the target exists but is not used, it will not be built. This is the case for NativeTracing, disabling its build for iOS otherwise its .a is not found at installation. Why is it not an issue for other build/installs?

- disable gtest fetching for ios
- disable openXR not other than win32 platforms
- `UNIX` is true in cmake scripts for Apple platforms, update checks accordingly
- bump iOS deployment target to 15
- bump gradle task version because deprecate
- bump Xcode version from 15.3 to 15.4


CMakeLists.txt
```diff
@@ -36,8 +36,8 @@ FetchContent_Declare(ios-cmake
     GIT_REPOSITORY https://github.com/leetal/ios-cmake.git
     GIT_TAG 4.4.1)
 FetchContent_Declare(JsRuntimeHost
-    GIT_REPOSITORY https://github.com/BabylonJS/JsRuntimeHost.git
-    GIT_TAG 9ccb9477a97501525f3f9ca0e00e6971934037ec)
+    GIT_REPOSITORY https://github.com/cedricguillemet/JsRuntimeHost.git
```
CedricGuillemet, line 39: replace when https://github.com/BabylonJS/JsRuntimeHost/pull/92 is merged

Dependencies/xr/CMakeLists.txt
```diff
@@ -24,10 +24,12 @@ endif()
 
 project(xr)
 
-# Avoid picking up system installed jsoncpp in favor of source distributed with openxr_loader project
-set(BUILD_WITH_SYSTEM_JSONCPP OFF BOOL "disable using system installed jsoncpp")
-FetchContent_MakeAvailable_With_Message(OpenXR-SDK OpenXR-MixedReality)
```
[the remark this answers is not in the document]
CedricGuillemet: Is there a version of ARCore available for these devices?

.github/jobs/linux.yml
```diff
@@ -31,6 +31,10 @@ jobs:
         cmake .. -G Ninja -D JAVASCRIPTCORE_LIBRARY=/usr/lib/x86_64-linux-gnu/libjavascriptcoregtk-4.0.so -D NAPI_JAVASCRIPT_ENGINE=${{ parameters.JSEngine }} -D CMAKE_BUILD_TYPE=RelWithDebInfo -D BX_CONFIG_DEBUG=ON -D CMAKE_UNITY_BUILD=$(UNITY_BUILD)
```
bghgary, line 31: nit: We should stop doing this `mkdir` thing and use `-B` on cmake instead.

bghgary (approved):

.github/jobs/linux.yml
CedricGuillemet: indeed! I'm doing the change!

PR #1402, opened 2024-07-15
Shader Cache

CedricGuillemet:
This PR keeps GLSlang/SPIRV and makes the ShaderCache optional. 
- At runtime, compute hash from Vertex/Fragment shaders and compare with existing cache. Store it if not found.
- Ability to load/save cache to an uncompressed stream (optional)
- ShaderCache is a singleton that will be used in NativeEngine if Cache is instanciated

Some tests metrics done locally (release build Win32-d3d11)

Validation tests with empty cache start: 83518 milliseconds
Validation tests with cache from previous run: 78898 milliseconds
Validation tests with cache disabled : 87292 milliseconds

Default PG .exe size : 3.59Mb
No GLSLANG/spirv PG.exe size : 2.33Mb
Shader cache for validation tests (every shaders used) D3D11 : 1.26Mb. 
Same with compression 71Kb (7z) - 200Kb (zip)


SergioRZMasson (approved):
I really like this approach, I just wonder why the performance difference is not bigger between using and not using the shader cache. It would be good for the future to allow users to easily remove dependency from GLSLANG and SPIRV using a CMake flag or something.

Plugins/NativeEngine/Source/ShaderCache.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Babylon
+{
+    class ShaderCache::Impl
+    {
+    public:
+        Impl() = default;
+        ~Impl() = default;
+        const ShaderCompiler::BgfxShaderInfo* GetShader(std::string_view vertexSource, std::string_view fragmentSource);
+        void AddShader(std::string_view vertexSource, std::string_view fragmentSource, ShaderCompiler::BgfxShaderInfo);
```
ryantrem, line 15: Missing name for last param?

Plugins/NativeEngine/Source/ShaderCache.h
```diff
@@ -0,0 +1,33 @@
...
+
+        void SerializeOutTo(std::ofstream& stream);
+        uint32_t SerializeInFrom(std::ifstream& stream);
```
ryantrem, line 18: I think `Serialize` and `Deserialize` would be more standard naming?

Plugins/NativeEngine/Source/ShaderCache.cpp
```diff
@@ -0,0 +1,180 @@
...
+    struct ShaderHash
+    {
+        size_t vertexHash;
+        size_t fragmentHash;
+        bool operator < (const ShaderHash& other) const {
+            if (vertexHash < other.vertexHash || (vertexHash == other.vertexHash && fragmentHash < other.fragmentHash))
+                return true;
+            return false;
+        }
+    };
```
ryantrem, line 32: Seems this is in both the .h and the .cpp

Plugins/NativeEngine/Include/Babylon/ShaderCache.h
```diff
@@ -0,0 +1,19 @@
...
+    {
+    public:
+        ShaderCache();
+        ~ShaderCache();
+
+        void SerializeOutTo(std::ofstream& stream);
+        uint32_t SerializeInFrom(std::ifstream& stream);
+        
+        class Impl;
+        static Impl* GetImpl();
```
ryantrem, line 17: I'm a bit confused on this. I thought Impl was supposed to be a way of hiding the implementation details from the header and eliminating webs of header dependencies, but in this case it seems like Impl is part of the public contract?

Plugins/NativeEngine/Include/Babylon/ShaderCache.h
```diff
@@ -0,0 +1,19 @@
...
+namespace Babylon
+{
+    class ShaderCache
+    {
+    public:
+        ShaderCache();
+        ~ShaderCache();
+
+        void SerializeOutTo(std::ofstream& stream);
+        uint32_t SerializeInFrom(std::ifstream& stream);
```
ryantrem, line 14: Are these actually even used? It seems like the consuming code is actually just using the Impl as the public contract, even for serializing/deserializing.

Plugins/NativeEngine/Source/ShaderCache.h
CedricGuillemet: fixed

Plugins/NativeEngine/Source/ShaderCache.cpp
CedricGuillemet: fixed

Plugins/NativeEngine/Source/ShaderCache.h
CedricGuillemet: Makes sense.fixed

Plugins/NativeEngine/Include/Babylon/ShaderCache.h
CedricGuillemet: It should be at least tested in TestUnit.

CedricGuillemet: I've added a test.

Plugins/NativeEngine/Include/Babylon/ShaderCache.h
CedricGuillemet: I've reworked it a bit.

Plugins/NativeEngine/Source/ShaderCache.h
```diff
@@ -0,0 +1,37 @@
...
+            size_t vertexHash;
+            size_t fragmentHash;
+            bool operator < (const ShaderHash& other) const {
+                if (vertexHash < other.vertexHash || (vertexHash == other.vertexHash && fragmentHash < other.fragmentHash))
+                    return true;
+                return false;
+            }
+        };
+        std::map<ShaderHash, ShaderCompiler::BgfxShaderInfo> Cache{};
+        static inline std::unique_ptr<ShaderCacheImpl> Instance{};
```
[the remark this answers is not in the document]
CedricGuillemet: Maybe not actually. I use the singleton as a flag to enable/disable the cache. But making is more explicit with a function looks good too.

Plugins/NativeEngine/Include/Babylon/ShaderCache.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+#include <string_view>
+#include <memory>
+
+namespace Babylon
+{
+    namespace ShaderCache
+    {
+        void Enable(bool enabled);
+        bool IsEnabled();
```
ryantrem, line 10: I think for these types of "property" like members, elsewhere we have done more like:
```suggestion
        void Enabled(bool enabled);
        bool Enabled();
```
I can't seem to find any examples, maybe @bghgary can comment.

Plugins/NativeEngine/Source/ShaderCache.h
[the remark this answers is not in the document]
ryantrem: So you could remove the `ShaderCacheImpl` class entirely and just have the map in an anonymous namespace + exposed functions for `GetShader`, `AddShader`, `Serialize`, `Deserialize`, yes?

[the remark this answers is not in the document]
CedricGuillemet: yes, that's a possibility. I prefer to keep the map creation behind an allocation for now.

Plugins/NativeEngine/Include/Babylon/ShaderCache.h
CedricGuillemet: Fixed

ryantrem (approved):

Plugins/NativeEngine/Include/Babylon/ShaderCache.h
bghgary: Here is one: https://github.com/BabylonJS/UrlLib/blob/main/Include/UrlLib/UrlLib.h#L48-L50

PR #1403, opened 2024-07-17
iOS AR Metal Validation Error

CedricGuillemet:
Follow up https://forum.babylonjs.com/t/babylonnative-ios-ar-crash/51876/19

### DIFF
--- a/Dependencies/xr/Source/ARKit/XR.mm
+++ b/Dependencies/xr/Source/ARKit/XR.mm
@@ -946,6 +946,8 @@ namespace xr {
 
             // Finalize rendering here & push the command buffer to the GPU.
             [currentCommandBuffer commit];
+			
+			[currentCommandBuffer waitUntilCompleted];
 
             return std::make_unique<Frame>(*this);
         }


Dependencies/xr/Source/ARKit/XR.mm
```diff
@@ -946,6 +946,8 @@ void UpdateXRView(MTKView* activeXRView) {
 
             // Finalize rendering here & push the command buffer to the GPU.
             [currentCommandBuffer commit];
+			
+            [currentCommandBuffer waitUntilCompleted];
```
ryantrem, line 950: What validation error does this fix? I don't think this is good for perf. I think we used to do this and tried to get rid of it, because it blocks this thread until the command buffer has been executed by the gpu.

CedricGuillemet: I know but I'd like to narrow down the issue to see if it's this command buffer that's causing all the crashes.

CedricGuillemet:
reopen if needed

PR #1417, opened 2024-09-05
Fix cmake install for some pdbs

bghgary:
If the target file name doesn't match the target name, then the old way will fail to get the right pdb name. The new way ensures it will get the right name.


SergioRZMasson (approved):
LGTM

docEdub (approved):

PR #1423, opened 2024-09-19
add emscripten support

kmatzen:
BabylonNative ported to run backed by emscripten so that native apps using BabylonNative can be ported back to web easily.


CedricGuillemet:
@kmatzen do you have in mind a use case for this feature? I can understand it as a computer scientist but as a end user, I don't quite see the point vs using Babylon.js npm. Also, for this PR to be approved, you'll need to add CI build script.

kmatzen:
@CedricGuillemet I have a GLFW app that I build for macOS, Windows, Linux, and WASM.  I'd like to replace the rendering engine inside of it with something like BabylonNative.  I'd prefer not to have to maintain two versions of the app for desktop and web.  I got a demo working, so I figured I'd share back the results.

Here are some related PRs to make it work.
https://github.com/BabylonJS/JsRuntimeHost/pull/97
https://github.com/BabylonJS/JsRuntimeHost/pull/96
https://github.com/BabylonJS/UrlLib/pull/19

https://github.com/user-attachments/assets/6f6ae0cd-914a-459f-979b-a1ab53e47140

Apps/Playground/CMakeLists.txt
```diff
@@ -74,6 +74,10 @@ elseif(UNIX)
     set(SOURCES
         ${SOURCES}
         "X11/App.cpp")
+elseif(EMSCRIPTEN)
+    set(SOURCES
+        ${SOURCES}
+        "Emscripten/App.cpp")
```
bghgary, line 80: It's a bit strange to call this Emscripten when it targets specifically GLFW. I only see `emscripten_set_main_loop_arg` that is Emscripten specific in App.cpp.

bghgary (changes requested):
I'm not 100% sure this change is necessary. We can already integrate with GLFW with the existing code. Do we really need a Emscripten specific path to make it work with GLFW and Emscripten? I think I might be okay if had a GLFW target for the playground app. @SergioRZMasson You did a prototype with GLFW. Any thoughts?

kmatzen:
@bghgary Yeah, that's a good point.  I can highlight a few differences.  I did start by looking at @SergioRZMasson's code.

* The Window, XMLHttpRequest, Console, and Canvas polyfills aren't necessary for Emscripten.  Window, XMLHttpRequest, and Canvas didn't actually hurt anything, but I did specifically have issues with the Console polyfill.
* [Emscripten needs a c-string with a selector for the target canvas as the window](https://github.com/BabylonJS/BabylonNative/pull/1423/files#diff-7633f198e52c01c0e647b83bb0edc5dab0f4ae56f87248db93b9d3fa8b091ac1R115) whereas the [existing example uses the glfw native window directly](https://github.com/SergioRZMasson/BabylonNativeGLFW/blob/main/main.cpp#L83C2-L83C88).  I tried the original approach and it didn't work.
* As you pointed out, the emscripten render loop does need to be separated out as a per-frame callback explicitly given to emscripten rather than a proper loop.  
* I like to use imgui too, but for this particular prototype, I wanted something that only took glfw as a dependency.

bghgary:
Thanks for the details. 

> The Window, XMLHttpRequest, Console, and Canvas polyfills aren't necessary for Emscripten. Window, XMLHttpRequest, and Canvas didn't actually hurt anything, but I did specifically have issues with the Console polyfill.

I'm curious what doesn't work with the Console polyfill.

> [Emscripten needs a c-string with a selector for the target canvas as the window](https://github.com/BabylonJS/BabylonNative/pull/1423/files#diff-7633f198e52c01c0e647b83bb0edc5dab0f4ae56f87248db93b9d3fa8b091ac1R115) whereas the [existing example uses the glfw native window directly](https://github.com/SergioRZMasson/BabylonNativeGLFW/blob/main/main.cpp#L83C2-L83C88). I tried the original approach and it didn't work.

That makes sense since there will not be a native window that works with the existing code. This contract is intended to give more type-safety compared to what bgfx does. I'm assuming bgfx supports Emscripten targets. Is the Emscripten target in bgfx a specific type? If so, then we should use that type for the new Device window type.

To me, it seems like we can have a GLFW target for the Playground app and have maybe some #ifdefs for Emscripten. The device issue seems should be GLFW specific because it is not Emscripten specific?

Generally, it seems strange to do something like this because Babylon Native is intended to bring web to native, whereas this is bringing native back to web. It's a bit circular. :-)

Just curious, does this actually generate good, concise web code at the end (looks like it will generate JS and WASM)? What happens to the JS loaded by script loader? I'm not at all familiar with emnapi.

kmatzen:
> I'm curious what doesn't work with the Console polyfill.

I double checked the Console polyfill and it works so I must have made a mistake.  I seem to recall that calling `emscripten::val::global("console").call<void>("log", foo);` inside a native callback was causing an infinite recursion, but it's also possible that I was debugging the polyfill callback itself and didn't realize it at the time.  😵‍💫 So installing all polyfills seems fine, but is also unnecessary in the browser.

> Is the Emscripten target in bgfx a specific type?
If you mean where the nwh is cast to a platform specific type, it's [here](https://github.com/bkaradzic/bgfx/blob/e4641029601216af396f79b784bc5f9a81edef52/src/glcontext_html5.cpp#L72) and it's a `const char*`.

> Generally, it seems strange to do something like this because Babylon Native is intended to bring web to native, whereas this is bringing native back to web.
Yeah... I only investigated this as a way to support a multi-platform desktop and web app without having two versions.  The app was already in C++, so this was the first route.  The next route would be to embed a web view into the app and run Babylon.JS directly, but that seemed harder to get working cross platform and to get it to automatically translate to web.

> What happens to the JS loaded by script loader?
So this is a bit easier to answer.  BabylonNative passes it to napi_run_script and then emnapi calls eval on it.  The only tricky part was that in order to get promises to resolve correctly, the WASM app has to suspend it's execution stack periodically.  That's why I made the change to [WorkQueue](https://github.com/BabylonJS/JsRuntimeHost/pull/97/files#diff-f846ac9cdbdc4bbd269eff01b97c763ba11e60cdf001fdd0d7a490a3913c2c29R45).  

> Just curious, does this actually generate good, concise web code at the end (looks like it will generate JS and WASM)?
For that demo in release mode, the wasm is 5.3 MB and the JS is 320 KB, not including the babylon scripts that are downloaded asynchronously.  Removing the polyfills brings the WASM down to about 5 MB.  I haven't analyzed the binary to see what's taking up the most space though.

bghgary:
> If you mean where the nwh is cast to a platform specific type, it's [here](https://github.com/bkaradzic/bgfx/blob/e4641029601216af396f79b784bc5f9a81edef52/src/glcontext_html5.cpp#L72) and it's a const char*.

Yes, that's what I mean. If so, then the window type should be `const char*`.

> I only investigated this as a way to support a multi-platform desktop and web app without having two versions. The app was already in C++, so this was the first route. The next route would be to embed a web view into the app and run Babylon.JS directly, but that seemed harder to get working cross platform and to get it to automatically translate to web.

Babylon tech is inherently web-first. I get you don't want two versions but doing it in this circular way is probably not going to be efficient (CPU, memory, download, etc). If the whole app is wrapped in a WebView then might as well use a PWA or something similar, like things we mention [here]( https://doc.babylonjs.com/BabylonCrossPlat/multiplePlatforms) as well as [here](https://github.com/BabylonJS/BabylonNative/blob/master/Documentation/WhenToUseBabylonNative.md). But it sounds like you are saying the app already has native C++ parts and these solutions may not fit well. Another option is to use [Filament](https://github.com/google/filament), which is native to begin with. But like most native solutions compiled to WASM, the download size is typically the biggest hurdle to get over.

> BabylonNative passes it to napi_run_script and then emnapi calls eval on it. The only tricky part was that in order to get promises to resolve correctly, the WASM app has to suspend it's execution stack periodically. That's why I made the change to [WorkQueue](https://github.com/BabylonJS/JsRuntimeHost/pull/97/files#diff-f846ac9cdbdc4bbd269eff01b97c763ba11e60cdf001fdd0d7a490a3913c2c29R45).

The WorkQueue has problems in its current form. It works most of the time, but there are async teardown issues. We have some long-standing issues with this code that still have not been resolved for various reasons. I'm not sure this code will work right with emscripten in all cases the way you have it coded.

> the wasm is 5.3 MB and the JS is 320 KB, not including the babylon scripts that are downloaded asynchronously

This is potentially the crux of the problem. 5MB is ginormous in the web world. This size will be a deal breaker for many web scenarios. I don't know what you are trying to build, but please make sure this is actually what you want. A clean tree-shaken web version is likely much smaller than these sizes.

> I haven't analyzed the binary to see what's taking up the most space

My understanding is that one common issue with emscripten is that it generates big downloads if STL is used. Babylon Native uses STL all over the place which might be the main cause. I'm no emscripten expert though.

kmatzen:
I played around with an alternative approach today and I agree, the BabylonNative approach for WASM isn't the best.  I was able to take my glfw app, load babylon.js directly via emscripten, construct an offscreen canvas, update a gl texture from the offscreen canvas each frame, and then render a textured quad inside my native app to composite it with the rest of the UI.  So no bgfx or other native code needs to be added to the WASM binary.  I haven't quite figured out how to correctly proxy the input interactions though.  I thought I could write a callback for the glfw events, queue them up, and then on each frame, send them off to the canvas via emscripten.

kmatzen:
I found what I was doing wrong.  I was converting the glfw events into mouseup, mousedown, and mousemove.  Babylon.js seems to only respond to pointerup, pointerdown, and pointermove.  Now input proxying works.

kmatzen:
Yes, here you go.  https://github.com/kmatzen/BabylonPlusEmscriptenGLFW

This is the minimal version where compositing is just a full screen quad render.

bghgary:
Thanks. This app will run in a browser but it doesn't run as a desktop app, correct?

kmatzen:
Yes, that's correct.

PR #1439, opened 2024-10-04
Matzen/opengltex

kmatzen:
I started with the ExternalTexture for Metal and had an LLM convert it to OpenGL.  I fixed up a few parts where it made mistakes in the texture enums, then eyeballed the rest.

### DIFF
--- a/Plugins/ExternalTexture/Source/ExternalTexture_OpenGL.cpp
+++ b/Plugins/ExternalTexture/Source/ExternalTexture_OpenGL.cpp
@@ -6,8 +6,118 @@
 #include <napi/pointer.h>
 #include <bx/bx.h>
 #include <bgfx/bgfx.h>
+#include <GL/gl.h>
 
 #include "ExternalTexture_Base.h"
+
+namespace
+{
+    struct TextureFormatInfo
+    {
+        GLenum m_fmt;
+        GLenum m_fmtSrgb;
+    };
+
+    static TextureFormatInfo s_textureFormat[] =
+        {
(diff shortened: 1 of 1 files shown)


Plugins/ExternalTexture/Source/ExternalTexture_OpenGL.cpp
```diff
@@ -6,9 +6,121 @@
...
 
+namespace
+{
+    struct TextureFormatInfo
+    {
+        GLint m_fmt;
+        GLint m_fmtSrgb;
+    };
+
+    static TextureFormatInfo s_textureFormat[] =
```
bghgary, line 21: This table is usually derived from bgfx. It would be better to start from here: https://github.com/BabylonJS/bgfx/blob/master/src/renderer_gl.cpp#L212-L325. Note this bgfx is a fork, but is only intended to be a staging area for changes.

Plugins/ExternalTexture/Source/ExternalTexture_OpenGL.cpp
```diff
@@ -18,21 +130,108 @@ namespace Babylon::Plugins
...
+            }
+
+            // Return the number of mipmap levels excluding level 0
+            return mipCount;
+        }
+
+        void GetInfo(Graphics::TextureT textureID, Info& info)
         {
-            throw std::runtime_error{"not implemented"};
+            glBindTexture(GL_TEXTURE_2D, textureID); // Bind the texture
```
bghgary, line 180: Probably should store the previously bound texture and restore it at the end instead of resetting to zero.

bghgary (approved):
Looks good. We unfortunately do not have testing for this class yet. I'm assuming this code works in your scenario?

PR #1440, opened 2024-10-09
Android AAR project

### LINKED ISSUE #1435: Reduce friction on Integration
Steps to get BN integrated in an android app are complex and error prone
see https://forum.babylonjs.com/t/integration-of-babylon-native-in-exisiting-application/53733/6
Is it possible to generate a AAR at the installation test CI step and ease integration ?

CedricGuillemet:
Instead of having the Android Playground app, split in an shared BabylonNative AAR library and a smaller app.
This change consists in adding a new Module project to the playground. Moving code into that new module: JNI layer, Java view and Wrapper, cmake to make the jni.so. Java namespace renamed to make it easier to integrate. And finally, add a dependency to the .aa in the app.

Min Android SDK version bumped to v25.
Running Android Playground App is the same: open with Android Studio then run app on device.

Produced .aar can be shared (like a npm, nuget, or local file ...) and integration will be way simpler: no need to clone BN, modify  app gradle with cmake, add .cpp, etc...
Integration/doc will be part of another PR.

fixes #1435 

cc @okwasniewski


Apps/Playground/Android/app/build.gradle
```diff
@@ -67,25 +48,18 @@ android {
...
-        }
-    }
-
     namespace 'com.android.babylonnative.playground'
 }
 
+repositories {
+    // Add flatDir repository
+    flatDir {
+        dirs '../../Android/BabylonNative/build/outputs/aar'
```
ryantrem, line 57: Isn't this the same as:
```suggestion
        dirs '../BabylonNative/build/outputs/aar'
```

matthargett:
was this waiting on something that we can help with?

Apps/Playground/Android/app/build.gradle
```diff
@@ -114,6 +88,8 @@ tasks.configureEach { task ->
     }
 }
 
+preBuild.dependsOn ':BabylonNative:assembleRelease'
```
ryantrem, line 91: Does this mean we always use a release build of BabylonNative? If so, I would think this would impact debuggability.

CedricGuillemet, line 91: What happens with Debug build? does the release binary have debug infos?

Apps/Playground/Android/BabylonNative/build.gradle
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+    }
+}
+
+tasks.configureEach { task ->
+    if ((task.name.contains("buildCMake") || task.name.contains("configureCMake")) && !task.name.contains("Clean")) {
+        task.dependsOn(extractNativeLibraries)
+    }
+}
```
ryantrem, line 122: Can this be factored out into a helper gradle file that gets imported in both?

ryantrem (approved):
Looks good! Would be helpful to flesh out the description of the PR with a bit more context, e.g. the steps involved to fully solve this problem. Something like:
1. Prove out that we can prebuild a BN aar and consume it in an Android app (this PR).
2. Formalize the aar, figure out the right way to prebuild for each JS engine, and wrap it into a maven package.
3. Add steps to publish a maven package to a public feed.
4. Do all the same stuff for iOS.
5. Reconcile overlap between BN and BRN (e.g. the Android/iOS specific views, etc.).
6. Update BRN to consume these packages as well.

matthargett:
For #4, I think @okwasniewski had a plan to make BN a cocoapod package for iOS and visionOS, following the same pattern of adoption into BRN as this PR does for Android AAR.

Apps/Playground/Android/app/build.gradle
CedricGuillemet: I could attach the debugger and step into the code, watch variables,....

CedricGuillemet:
I've open this issue to centralize questions/technical points : https://github.com/BabylonJS/BabylonNative/issues/1446

Apps/Playground/Android/BabylonNative/build.gradle
CedricGuillemet: Comment added here : https://github.com/BabylonJS/BabylonNative/issues/1446

okwasniewski:
Hey @CedricGuillemet are you planning to integrate this into Babylon React Native?

CedricGuillemet:
> Hey @CedricGuillemet are you planning to integrate this into Babylon React Native?

Yes, it's in my todo list : https://github.com/BabylonJS/BabylonNative/issues/1446
I'd like to do the same thing with pod too (see next steps)

PR #1458, opened 2025-02-07
Add populateFrameStats

ryantrem:
This change just adds a `populateFrameStats` function to `NativeEngine` that gets the bgfx frame stats. Currently it only returns gpu frame time in ns.
Corresponding BJS change: https://github.com/BabylonJS/Babylon.js/pull/16159


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -2261,6 +2263,16 @@ namespace Babylon
         }
     }
 
+    void NativeEngine::PopulateFrameStats(const Napi::CallbackInfo& info)
+    {
+        const auto updateToken{m_update.GetUpdateToken()};
+        const auto stats{bgfx::getStats()};
+        const double toGpuNs = 1000000000.0 / double(stats->gpuTimerFreq);
+        const double gpuTimeNs = (stats->gpuTimeEnd - stats->gpuTimeBegin) * toGpuNs;
+        Napi::Object jsStatsObject = info[0].ToObject();
```
bghgary, line 2272: `ToObject` will coerce to an object even if it's not which isn't what you want.

```suggestion
        Napi::Object jsStatsObject = info[0].As<Napi::Object>();
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -2261,6 +2263,16 @@ namespace Babylon
         }
     }
 
+    void NativeEngine::PopulateFrameStats(const Napi::CallbackInfo& info)
```
bghgary, line 2266: Is the thinking we might add more stats later? If not, we could just return a number directly.

bghgary (approved):

Plugins/NativeEngine/Source/NativeEngine.cpp
ryantrem: Yes that was my thinking. I will switch to a number if you object, otherwise I'll leave it as is.

bghgary: I'm not objecting, just asking.

PR #1465, opened 2025-02-20
Switch to JsRuntimeHost for SetTimeout/SetInterval

ryantrem:
Update JsRuntimeHost to the latest and replace Window's setTimeout/clearTimeout/setInterval/clearInterval implementations with those from JsRuntimeHost.


bghgary (approved):
LGTM

PR #1468, opened 2025-02-25
Canvas test

CedricGuillemet:
Draft PR for build/testing validation


CedricGuillemet:
Issue with D3D12:
```
Exception thrown at 0x00007FF91897AB6A in Playground.exe: Microsoft C++ exception: _com_error at memory location 0x000000E394F5E940.
D3D12: Removing Device.
..../build/_deps/bgfx.cmake-src/bgfx/src/renderer_d3d12.cpp (1773): BGFX 0x00000004: Device is lost. FAILED 0x887a0005 DXGI_ERROR_DEVICE_REMOVED (Unknown HRESULT?)
Unhandled exception at 0x00007FF918484C2E (ucrtbase.dll) in Playground.exe: Fatal program exit requested.
```
What can cause Device lost?

Plugins/NativeEngine/CMakeLists.txt
```diff
@@ -35,21 +35,41 @@ target_include_directories(NativeEngine
...
-    PRIVATE bimg_decode
-    PRIVATE bx
-    PRIVATE minz
-    PRIVATE glslang
-    PRIVATE glslang-default-resource-limits
-    PRIVATE GraphicsDevice
-    PRIVATE GraphicsDeviceContext
-    PRIVATE JsRuntime
-    PRIVATE SPIRV)
+if(BGFX_BUILD_TOOLS)
```
bghgary: Can you explain this more? What's this for?

.github/jobs/test_install_win32.yml
```diff
@@ -41,7 +41,6 @@ jobs:
 
       # BGFX_CONFIG_MAX_FRAME_BUFFERS is set so enough Framebuffers are available before V8 starts disposing unused ones
       - script: |
-          # BGFX_CONFIG_MAX_FRAME_BUFFERS is set so enough Framebuffers are available before V8 starts disposing unused ones
```
ryantrem, line 44: Why is this comment removed? Is it no longer valid? Should there be a different explanation of this define?

Polyfills/Canvas/Source/nanovg/nanovg.h
ryantrem, line 1: Can we add a readme.md explaining where this OSS code came from and whether we have modified it?

ryantrem: And also how nanovg_babylon.* relates to the other files?

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -25,7 +25,10 @@
 #include "MeasureText.h"
 #include "Image.h"
 #include "ImageData.h"
+#include "Path2D.h"
 #include "Colors.h"
+#include "LineCaps.h"
+#include "Gradient.h"
 
 /*
 Most of these context methods are preliminary work. They are currenbly not tested properly.
```
ryantrem: Do we still want this comment? Also there is a typo in it anyway (currenbly).

Polyfills/Canvas/Source/Font.h
```diff
@@ -0,0 +1,30 @@
...
+    {
+    private:
+        static constexpr int NORMAL_WEIGHT = 400;
+        static constexpr int BOLD_WEIGHT = 700;
+
+    public:
+        FontStyle Style = FontStyle::Normal;
+        int Weight = NORMAL_WEIGHT;
+        float Size = 10;
+        std::string Family = "sans-serif";
```
ryantrem: Should these be immutable? Seems like they could be passed into a constructor and then exposed through get functions?

Polyfills/Canvas/Source/nanosvg.h
ryantrem, line 1: Maybe a readme.md explaining where this came from and if it has any Babylon modifications also?

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
```diff
@@ -50,6 +50,85 @@ BX_PRAGMA_DIAGNOSTIC_IGNORED_MSVC(4244) // warning C4244: '=' : conversion from
...
+    static void init()
+    {
+        ms_layout
+            .begin()
+            .add(bgfx::Attrib::Position, 3, bgfx::AttribType::Float)
+            .add(bgfx::Attrib::TexCoord0, 2, bgfx::AttribType::Float)
+            .end();
+    }
+
+    static bgfx::VertexLayout ms_layout;
```
bghgary: nit: I haven't seen people use both prefixes together. Usually, I see just `s_layout`.

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
```diff
@@ -50,6 +50,85 @@ BX_PRAGMA_DIAGNOSTIC_IGNORED_MSVC(4244) // warning C4244: '=' : conversion from
...
+};
+
+bgfx::VertexLayout PosTexCoord0Vertex::ms_layout;
```
bghgary: nit: You can use inline to do everything in one line.

```suggestion
    static void init()
    {
        s_layout
            .begin()
            .add(bgfx::Attrib::Position, 3, bgfx::AttribType::Float)
            .add(bgfx::Attrib::TexCoord0, 2, bgfx::AttribType::Float)
            .end();
    }

    static inline bgfx::VertexLayout s_layout{};
};
```

.github/jobs/test_install_win32.yml
bghgary: I had exactly the same thought and then I noticed it's just a duplicate comment from two lines above.

Polyfills/Canvas/Source/Context.cpp
CedricGuillemet: Nope, it's better to add a readme and document the plugin as experimental IMHO.

Plugins/NativeEngine/CMakeLists.txt
CedricGuillemet: bgfx shaderc tool needed to be compiled to apply changes in nanovg shaders. But, BN and bgfx need different versions spirv/glslang with different dependencies.

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
CedricGuillemet: fixed

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
CedricGuillemet: fixed

Polyfills/Canvas/Source/Font.h
CedricGuillemet: I've change for default values and getters.

Plugins/NativeEngine/CMakeLists.txt
bghgary: Sorry, I still don't understand. What does bgfx's shaderc tool have to do with NativeEngine?

Polyfills/Canvas/Source/Font.h
```diff
@@ -13,18 +13,20 @@ namespace Babylon::Polyfills::Internal
...
+    public:
+        operator std::string() const;
+        static std::optional<Font> Parse(const std::string& fontString);
+
+        float GetSize() const { return size; }
+        const std::string& GetFamiliy() const { return family; }
+
     private:
         static constexpr int NORMAL_WEIGHT = 400;
         static constexpr int BOLD_WEIGHT = 700;
```
bghgary:
```suggestion
static constexpr const int NORMAL_WEIGHT = 400;
static constexpr const int BOLD_WEIGHT = 700;
```

Polyfills/Canvas/Source/Font.h
```diff
@@ -13,18 +13,20 @@ namespace Babylon::Polyfills::Internal
...
-        float Size = 10;
-        std::string Family = "sans-serif";
-
-        operator std::string() const;
-
-        static std::optional<Font> Parse(const std::string& fontString);
+        FontStyle style{FontStyle::Normal};
+        int weight{NORMAL_WEIGHT};
+        float size{10};
+        std::string family{"sans-serif"};
```
bghgary: These variables should have `m_` prefix.

ryantrem: Even public? I commented on them being public, so if we switch and have get functions, then it seems like they should be `m_`

bghgary: I think this PR made all of these variables private?

Polyfills/Canvas/Source/Font.h
```diff
@@ -0,0 +1,32 @@
...
+    };
+
+    struct Font
+    {
+    public:
+        operator std::string() const;
+        static std::optional<Font> Parse(const std::string& fontString);
+
+        float GetSize() const { return size; }
+        const std::string& GetFamiliy() const { return family; }
```
bghgary: nit: Spelling, though I would recommend we do it like this.

```suggestion
        float Size() const { return size; }
        const std::string& Family() const { return family; }
```

Polyfills/Canvas/Source/Font.h
CedricGuillemet: fixed

Polyfills/Canvas/Source/Font.h
CedricGuillemet: fixed

Polyfills/Canvas/Source/Context.cpp
CedricGuillemet: done in readme.md

Plugins/NativeEngine/CMakeLists.txt
CedricGuillemet: NativeEngine and shaderc both need glslang and spirv but of different versions (our forks vs bgfx repo fork). Both versions are incompatibles at build time but use the same target names.
Either disable NativeEngine build (and apps using it) when BGFX_BUILD_TOOLS is ON or make it compatible with both versions by changing its cmake and small patches in ShaderCompiler sources (what this PR does).

Polyfills/Canvas/Source/nanosvg.h
CedricGuillemet: fixed

Plugins/NativeEngine/CMakeLists.txt
CedricGuillemet: done as discussed

Polyfills/Canvas/CMakeLists.txt
```diff
@@ -1,3 +1,181 @@
+# _bn_shaderc_parse(
```
bghgary: Is this a copy of the bgfx one with modifications or something? If so, what is wrong with the bgfx one?

Polyfills/Canvas/Readme.md
```diff
@@ -1,24 +1,30 @@
...
 # Nanovg Shader Integration in Babylon.js
 Nanovg utilizes shaders that are independent of Babylon.js. These shaders are written in the bgfx shading language, which is derived from GLSL, and are compiled into various target languages (SPIR-V, GLSL, Metal, DX, ESSL) during the build process.
 
 ## Precompiled Shaders:
 
-A precompiled version of the shaders is provided, eliminating the need to compile shaderc and individual shaders with each application build. This saves time, as shaderc compilation can be time-consuming.
+A set of precompiled shaders is included, so you don’t need to build shaderc or recompile individual shaders every time you build the application. This significantly reduces build time, since compiling shaders with shaderc can be quite slow.
 
-## Shader Modification:
+## Shader Modification and compilation
```
bghgary:
```suggestion
## Shader Modification and Compilation
```

Polyfills/Canvas/Readme.md
```diff
@@ -1,24 +1,30 @@
...
-Revert the CMake flags after the modifications are complete. The Playground can then be run with the updated shaders.
+If you modify any shader, you’ll need shaderc — the bgfx shader compiler — to rebuild them. Shaderc can generate shaders targeting MSL, GLSL, HLSL, or SPIR-V.
 
-## Shader Compilation:
+To obtain shaderc:
+- Clone bgfx.cmake and generate and build the project 
+- Or folow these instructions : https://bkaradzic.github.io/bgfx/build.html
+Then
+- Once built, locate the `shaderc` executable in the output directory.
+- Set the full path to this executable in CMake using the variable `SHADERC_PATH`.
```
bghgary, line 23:
```suggestion
To obtain shaderc:
- Clone bgfx.cmake and generate and build the project 
  Or follow these instructions: https://bkaradzic.github.io/bgfx/build.html
- Once built, locate the `shaderc` executable in the output directory.
- Set the full path to this executable in CMake using the variable `SHADERC_PATH`.
```

Maybe also give an example CMake command adding this variable?

bghgary (approved):

Polyfills/Canvas/CMakeLists.txt
CedricGuillemet: definition of that function is guarded behind the BGFX_BUILD_TOOL flag. + name of executable `bgfx::shaderc` for the build command is not valid

bghgary: Ok, so pure copy then? Maybe a comment would be good here. Maybe even separate into its own file like `shaderc.cmake` to make it easier to indicate this is a copy.

CedricGuillemet: fixed

PR #1469, opened 2025-02-26
Path2D Implementation

Pheo:
- https://developer.mozilla.org/en-US/docs/Web/API/Path2D

Adds nanosvg.h for SVG path parsing.
Known missing functionality:
1. `addPath` doesn't accept DOMMatrix transform
2. `roundRect` doesn't accept CSS-style radii array
3. `ellipse` doesn't handle clockwise

Called from https://github.com/BabylonJS/Babylon.js/pull/16221


Polyfills/Canvas/Source/Path2D.cpp
```diff
@@ -0,0 +1,241 @@
...
+                    auto cpx1 = pts[2];
+                    auto cpy1 = pts[3];
+                    auto cpx2 = pts[4];
+                    auto cpy2 = pts[5];
+
+                    Path2DCommandArgs args = {};
+                    args.bezierTo = {cpx1, cpy1, cpx2, cpy2, x0, y0};
+                    AppendCommand(P2D_BEZIERTO, args);
+                }
+            }
```
Pheo, line 81: Currently working on converting NSVGparser->shapes to Path2DCommands

CedricGuillemet:
I think we can split this PR in 2.
- merge point to be master instead of CanvasTest
- add the test example in experience.js in the CanvasTest branch

Apps/Playground/Scripts/experience.js
```diff
@@ -64,6 +64,113 @@ CreateBoxAsync(scene).then(function () {
 //BABYLON.SceneLoader.AppendAsync("https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/ClearCoatTest/glTF/ClearCoatTest.gltf").then(function () {
     BABYLON.Tools.Log("Loaded");
 
+    var textureGround;
```
CedricGuillemet, line 67: Can you move experience.js change into a new PR targetted at CanvasTest branch?

PR #1478, opened 2025-03-12
Use bgfx more recent glslang/spirv for NativeEngine

CedricGuillemet:


ryantrem:
Is this using a different spirv depending on whether you are compiling actual BN vs the tools or something?

CedricGuillemet:
> Is this using a different spirv depending on whether you are compiling actual BN vs the tools or something?

yes. shaderc uses bgfx 3rd party glslang/spirv that is more recent than our fork and doesn't contain our changes.

Plugins/NativeEngine/CMakeLists.txt
```diff
@@ -33,21 +33,40 @@ target_include_directories(NativeEngine
...
+        PRIVATE bimg_encode
+        PRIVATE bimg_decode
+        PRIVATE bx
+        PRIVATE glslang
+        PRIVATE glslang-default-resource-limits
+        PRIVATE SPIRV
+        PRIVATE GraphicsDeviceContext)
+
+    warnings_as_errors(NativeEngine)
+endif()
```
ryantrem, line 69: This is mostly duplication. Factor out the common parts? For example, just move the bulk of these into a variable right above the if/else block. This will help with maintenance.

CedricGuillemet: yes, it's still in PoC until I get validation it works. Another solution is to upgrade our fork of glslang.

PR #1481, opened 2025-03-14
Fix texture copy when using multiple canvas

CedricGuillemet:
There is an issue when using multiple dynamic textures (Canvas). Only the 1st one is properly displayed.
`GLTF Node NegativeScale (0)` validation test looks like this:

image

A Renderdoc capture gives this call trace:

image

Texture copies are all done at the same time in bgfx, at the end of sorted drawcalls of a viewid : https://github.com/bkaradzic/bgfx/blob/4602a01ce51611183d8ae8e320ca0f5305b79cb8/src/renderer_d3d11.cpp#L5739

All blit commands are submitted with the same ViewID. Which is confirmed when looking at this:
https://github.com/BabylonJS/BabylonNative/blob/9379cf3c41eedbf3001d4f1e3872ced9a04e818e/Plugins/NativeEngine/Source/NativeEngine.cpp#L1438

Solution is incrementing ViewID in `Blit` and moving the call into DeviceContext as it doesn't rely on FrameBuffer state.

Result:

image

### DIFF
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
@@ -121,6 +121,7 @@ namespace Babylon::Graphics
         TextureInfo GetTextureInfo(bgfx::TextureHandle handle);
         static bx::AllocatorI& GetDefaultAllocator() { return m_allocator; }
 
+        void Blit(bgfx::Encoder& encoder, bgfx::TextureHandle dst, uint16_t dstX, uint16_t dstY, bgfx::TextureHandle src, uint16_t srcX = 0, uint16_t srcY = 0, uint16_t width = UINT16_MAX, uint16_t height = UINT16_MAX);
     private:
         friend UpdateToken;
 
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
@@ -122,7 +122,7 @@ namespace Babylon::Graphics
         static bx::AllocatorI& GetDefaultAllocator() { return m_allocator; }
 
         // call following method when a submit call is associated with the current acquired viewId
-        void SetViewAsUsed();
+        void InvalidateView();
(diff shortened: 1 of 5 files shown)


Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
```diff
@@ -121,6 +121,9 @@ namespace Babylon::Graphics
         TextureInfo GetTextureInfo(bgfx::TextureHandle handle);
         static bx::AllocatorI& GetDefaultAllocator() { return m_allocator; }
 
+        // call following method when a submit call is associated with the current acquired viewId
+        void SetViewAsUsed();
```
ryantrem, line 125: I guess we have to expose this new function because `FrameBuffer` has to tell `DeviceContext` this needs to be done when a `Submit` happens. Is there any other case (other than `Submit`) where we would need to do this, or is this really tied to `FrameBuffer::Submit`?

If we need to expose this as a new function, would `InvalidateView` be a better name?

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1431,11 +1431,12 @@ namespace Babylon
     {
         const auto textureDestination = info[0].As<Napi::Pointer<Graphics::Texture>>().Get();
         const auto textureSource = info[1].As<Napi::Pointer<Graphics::Texture>>().Get();
-
+        
+        // Append a task instead of calling blit immediately or blit happens before drawcalls.
         arcana::make_task(m_update.Scheduler(), *m_cancellationSource, [this, textureDestination, textureSource, cancellationSource = m_cancellationSource]() {
```
bghgary, line 1436: Did you try removing this? I think we might need to add this to the command queue for the ordering to be correct.

CedricGuillemet: Yes, I tried and copy doesn't happen in the correct order when removed.

Core/Graphics/InternalInclude/Babylon/Graphics/DeviceContext.h
CedricGuillemet: Yes, Framebuffer tells device context the view contains draw call. I've added a call for `Clear` . `InvalidateView` works for me!

CedricGuillemet:
Closing for now. I will re-open after https://github.com/BabylonJS/BabylonNative/pull/1468 is in.

PR #1488, opened 2025-03-26
Implement Context::GetCanvas

hwarmington:


Polyfills/Canvas/Source/Context.cpp
```diff
@@ -961,6 +961,6 @@ namespace Babylon::Polyfills::Internal
 
     Napi::Value Context::GetCanvas(const Napi::CallbackInfo& info)
     {
-        throw Napi::Error::New(info.Env(), "not implemented");
+        return m_canvas->Value();
```
CedricGuillemet, line 964: This returns NULL value and this code:
```
return Napi::Pointer<NativeCanvas>::Create(info.Env(), m_canvas);
```
returns an object that doesn't support `getContext`.
I tried to do a loop back like this in exprience.js : 
```
context = context.canvas.getContext();
```
Any idea @bghgary ?

bghgary: Ping me offline.

CedricGuillemet: After chatting with Gary, there are a few changes needed in Canvas/Context to make it more robust. I can take care of these changes.

CedricGuillemet:
Replacement PR : https://github.com/BabylonJS/BabylonNative/pull/1491 . Closing.

PR #1491, opened 2025-04-02
Canvas and Context

CedricGuillemet:
Replaces https://github.com/BabylonJS/BabylonNative/pull/1488

- Fix wrongly named `CreateInstance` with `Initialize`
- Canvas/Context unit test
- Implement `GetCanvas`
- Fix parameters and objectReference between Canvas and Context
- Keep and return the object reference created for `getContext`

### DIFF
--- a/Apps/UnitTests/Scripts/tests.js
+++ b/Apps/UnitTests/Scripts/tests.js
@@ -15,6 +15,20 @@ describe("RequestFile", function () {
     });
 });
 
+describe("CanvasAndContext", function () {
+    const engine = new BABYLON.NativeEngine();
+    const scene = new BABYLON.Scene(engine);
+
+    var texSize = 512;
+    var dynamicTexture = new BABYLON.DynamicTexture("dynamic texture", texSize, scene);
+    var context = dynamicTexture.getContext();
+    var otherContext = dynamicTexture.getContext();
+
+    expect(context).to.equal(context.canvas.getContext());
+    expect(context).to.equal(otherContext);
+    expect(context).to.equal(otherContext.canvas.getContext());
+});
+
 describe("ColorParsing", function () {
     expect(_native.Canvas.parseColor("")).to.equal(0);
     expect(_native.Canvas.parseColor("transparent")).to.equal(0);
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -15,14 +15,15 @@ namespace Babylon::Polyfills::Internal
 {
     static constexpr auto JS_CONSTRUCTOR_NAME = "Canvas";
 
-    void NativeCanvas::CreateInstance(Napi::Env env)
+    void NativeCanvas::Initialize(Napi::Env env)
     {
         Napi::HandleScope scope{env};
 
         Napi::Function func = DefineClass(
             env,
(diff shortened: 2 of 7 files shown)


Apps/UnitTests/Scripts/tests.js
```diff
@@ -15,6 +15,20 @@ describe("RequestFile", function () {
...
 });
 
+describe("CanvasAndContext", function () {
+    const engine = new BABYLON.NativeEngine();
+    const scene = new BABYLON.Scene(engine);
+
+    var texSize = 512;
+    var dynamicTexture = new BABYLON.DynamicTexture("dynamic texture", texSize, scene);
+    var context = dynamicTexture.getContext();
+    var otherContext = dynamicTexture.getContext();
```
bghgary, line 25: why `var` and not `const`?

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -77,7 +78,11 @@ namespace Babylon::Polyfills::Internal
 
     Napi::Value NativeCanvas::GetContext(const Napi::CallbackInfo& info)
     {
-        return Context::CreateInstance(info.Env(), this);
+        if (m_contextObject.IsEmpty())
+        {
+            m_contextObject = Napi::Persistent(Context::CreateInstance(info.Env(), info.This()).As<Napi::Object>());
```
bghgary, line 83: Canvas holds context and context holds canvas. This is probably going to prevent the GC from collecting these objects once created. Can you check?

bghgary (approved):
Looks good except for the circular reference

Polyfills/Canvas/Source/Canvas.cpp
CedricGuillemet: With canvasObject persistent, context is not disposed nor Canvas when disposing the dynamic texture. When using `Napi::Weak` Canvas and Context are properly disposed when disposing dyn. texture.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -86,17 +86,18 @@ namespace Babylon::Polyfills::Internal
...
 
         auto func = JsRuntime::NativeObject::GetFromJavaScript(env).Get(JS_CONTEXT_CONSTRUCTOR_NAME).As<Napi::Function>();
-        return func.New({Napi::External<NativeCanvas>::New(env, canvas)});
+        return func.New({canvas});
     }
 
     Context::Context(const Napi::CallbackInfo& info)
         : Napi::ObjectWrap<Context>{info}
-        , m_canvas{info[0].As<Napi::External<NativeCanvas>>().Data()}
+        , m_canvasObject{Napi::Weak(info[0].As<Napi::Object>())}
```
ryantrem, line 99: Does there need to be a js reference from context to canvas? If the js side only references the context, we don't want the canvas to be gc'd, right? I assume this is how it works in the browser?

CedricGuillemet: The reference from context to canvas is needed by the `getCanvas` method.
>  If the js side only references the context, we don't want the canvas to be gc'd, right? I assume this is how it works in the browser?
That was my guess. But in that case, when is the Canvas and Context are collected if both keep a strong reference on each other? Is there a way to detect that?

PR #1493, opened 2025-04-03
fixes: canvas.SetHeight/Width, Path2D, context.fill

Pheo:
- Supports passing Path2D to Path2D constructor
- Handles >1 args for context.fill
- As [matanui159](https://github.com/BabylonJS/BabylonNative/issues?q=is%3Apr%20is%3Aopen%20author%3Amatanui159), as pointed out, `SetHeight`, `SetWidth` clears canvas on browser even if width/height doesn't change. We re-use framebuffer but clear it in this scenario.

### DIFF
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -113,7 +113,7 @@ namespace Babylon::Polyfills::Internal
 
     void NativeCanvas::SetHeight(const Napi::CallbackInfo&, const Napi::Value& value)
     {
-        auto height = value.As<Napi::Number>().Uint32Value();
+        auto height = static_cast<uint16_t>(value.As<Napi::Number>().Uint32Value());
         if (!height)
         {
             return;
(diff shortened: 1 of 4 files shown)


Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -105,7 +114,16 @@ namespace Babylon::Polyfills::Internal
     void NativeCanvas::SetHeight(const Napi::CallbackInfo&, const Napi::Value& value)
     {
         auto height = value.As<Napi::Number>().Uint32Value();
```
CedricGuillemet, line 116: Missing static_cast present in SetWidth

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -137,7 +159,8 @@ namespace Babylon::Polyfills::Internal
 
             return true;
         }
-        return false;
+
+        return needClear || false;
```
CedricGuillemet, line 163: `|| false` not needed

CedricGuillemet (approved):

PR #1499, opened 2025-04-08
Mocha&Chai update

CedricGuillemet:
- upgrade Mocha and Chai version
- dump exception error object properties (works well with Chakra)
- use Babel to translate ES module to umd
- replace bigint notation (with `n` suffix) unsupported by Chakra
- script loading tweaks for unittests to support `globalThis` and Chai exports


Apps/package.json
```diff
@@ -3,17 +3,24 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "getNightly": "node scripts/getNightly.js"
+    "getNightly": "node scripts/getNightly.js",
+	"build" : "npx babel node_modules/mocha/mocha.js --out-file mocha.esm.js && npx babel node_modules/chai/chai.js --out-file chai.esm.js && npx rollup mocha.esm.js --file mocha.umd.js --format umd --name mocha && npx rollup chai.esm.js --file chai.umd.js --format umd --name chai",
```
bghgary, line 7: Where do these files end up? Are they just in the source folder?

CedricGuillemet: in Apps/ folder

bghgary (approved):

PR #1503, opened 2025-04-16
CopyTexture command

CedricGuillemet:
BJS PR ( ~~https://github.com/BabylonJS/Babylon.js/pull/16497~~ https://github.com/BabylonJS/Babylon.js/pull/16502 )needs to be merged before and a new NPM needs to be generated. 

Copy texture is now a queued command (see BJS PR), protocol changed.

Before: 

- canvas rendering was done with a task, in sync with frame rendering. 

Now:

- canvas rendering is done when when copying texture. Browser HTML Canvas rendering happens when rendering commands are issued. With Native, nanovg keeps its own queue that needs to be actually done before the copy. This is done with the `flush` command.
- Because of this 'flush when needed', no need to keep a dirty flag in the context.
- In order to always get the same context with `Canvas.getContext` some code from `CanvasTest` branch has been integrated in this PR

### Test

add this to experience.js. Quad gets black when texture is disposed.

```

    var ground = BABYLON.MeshBuilder.CreateGround("ground1", { width: 0.5, height: 0.5, subdivisions: 2 }, scene);
    ground.rotation.x = -Math.PI * 0.5;
    ground.rotation.y = Math.PI;

    var texSize = 512;
    var dynamicTexture = new BABYLON.DynamicTexture("dynamic texture", texSize, scene);
    dynamicTexture.clear();
    var context = dynamicTexture.getContext();


    var materialGround = new BABYLON.StandardMaterial("Mat", scene);
    materialGround.diffuseTexture = dynamicTexture;
    ground.material = materialGround;
    materialGround.backFaceCulling = false;

    context.save();
    context.fillStyle = "DarkRed";
    context.fillRect(0, 0, texSize, texSize);
    context.restore();

    dynamicTexture.update();
    setTimeout(() => {
        console.log("update-dispose");
        dynamicTexture.update();
        dynamicTexture.dispose();
    }, 5000);
```
### Testing Issues 

- [x] `Dynamic Texture context clip` :  race condition. shape is rendered when using a breakpoint
~~- [ ] `GLTF Extension KHR_materials_volume with attenuation` : rendering is a bit different but it doesn't seem to be related to Canvas/copytexture. To confirm~~ wrong material.js
- [x] `GLTF Material Alpha Mask`: inconsistent result. Might be related to race condition mentioned above.

### DIFF
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -78,7 +78,11 @@ namespace Babylon::Polyfills::Internal
 
     Napi::Value NativeCanvas::GetContext(const Napi::CallbackInfo& info)
     {
-        return Context::CreateInstance(info.Env(), this);
+        if (m_contextObject.IsEmpty())
+        {
+            m_contextObject = Napi::Persistent(Context::CreateInstance(info.Env(), info.This()).As<Napi::Object>());
+        }
+        return m_contextObject.Value();
     }
 
     Napi::Value NativeCanvas::GetWidth(const Napi::CallbackInfo&)
@@ -111,7 +115,7 @@ namespace Babylon::Polyfills::Internal
         }
     }
 
-    bool NativeCanvas::UpdateRenderTarget()
+    void NativeCanvas::UpdateRenderTarget()
     {
         if (m_dirty)
         {
@@ -142,10 +146,7 @@ namespace Babylon::Polyfills::Internal
             {
                 m_texture.reset();
             }
-
-            return true;
         }
-        return false;
     }
 
     Napi::Value NativeCanvas::GetCanvasTexture(const Napi::CallbackInfo& info)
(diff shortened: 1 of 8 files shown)


Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -111,7 +115,7 @@ namespace Babylon::Polyfills::Internal
         }
     }
 
-    bool NativeCanvas::UpdateRenderTarget()
+    void NativeCanvas::UpdateRenderTarget()
     {
         if (m_dirty)
```
ryantrem, line 120: `m_dirty` is never set back to true anymore, correct? Is this now just lazily creating the render target resources?

CedricGuillemet: it's set to true a few lines above when width/height has changed.

ryantrem (approved):

bghgary (approved):

PR #1505, opened 2025-04-18
Merge copy texture fix

ryantrem:
This brings in the changes from master to fix timing/sequencing issues related to copyTexture and deleteTexture.

This included changes on the JS side, so it brings with it new versions of the BJS packages.

### DIFF
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -80,11 +80,17 @@ namespace Babylon::Polyfills::Internal
 
     Napi::Value NativeCanvas::GetContext(const Napi::CallbackInfo& info)
     {
-        if (m_contextObject.IsEmpty())
+        auto thisObj = info.This().ToObject();
+        const auto contextPropertyName = Napi::Value::From(Env(), "_context");
+
+        auto context = thisObj.Get(contextPropertyName);
+        if (context.IsUndefined())
         {
-            m_contextObject = Napi::Persistent(Context::CreateInstance(info.Env(), info.This()).As<Napi::Object>());
+            context = Context::CreateInstance(info.Env(), info.This());
+            thisObj.Set(contextPropertyName, context);
         }
-        return m_contextObject.Value();
+
+        return context;
     }
 
     Napi::Value NativeCanvas::GetWidth(const Napi::CallbackInfo&)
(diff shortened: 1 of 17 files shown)


Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -173,11 +179,7 @@ namespace Babylon::Polyfills::Internal
 
             m_frameBufferPool.Clear();
             m_frameBufferPool.SetGraphicsContext(&m_graphicsContext);
-
-            return true;
         }
-
-        return needClear;
```
Pheo, line 180: Its expected that the Canvas is cleared when setheight/setwidth is set (even if they don't change)

Has this behavior has been preserved?

ryantrem: Hmm, good question, it seems like the previous call to `Clear` is gone and `needClear` is now and unused local variable. If we can wait until tomorrow, let's ask @CedricGuillemet.

ryantrem: Did this result in any regression? If not, I would say lets just merge this PR since it is already in master, and we can do a follow up PR in master after Cedric is back if needed and bring it into this branch. If it is causing a regression though, then let's wait a day for Cedric.

Pheo: Will test this on iOS and update.

Pheo: Looks like there is a regression (in that it doesn't clear anymore). But should be solvable with a simple flush on setheight/width. Think we can just merge this now, and I can raise a PR to fix after.

CedricGuillemet: I tried to simplify the code too much and it caused the regression. Best seems to revert that part on this PR. Can you do the change and test it again @Pheo ? Sorry for the inconveniences.

CedricGuillemet (approved):
All good once https://github.com/ryantrem/BabylonNative/pull/6 is merged

Polyfills/Canvas/Source/Canvas.cpp
Pheo: > raised PR @ [ryantrem#6](https://github.com/ryantrem/BabylonNative/pull/6)
Confirmed working! Framebuffer correctly clears on canvas setHeight/setWidth.

ryantrem: @CedricGuillemet can you make that same fix in master as well?

PR #1506, opened 2025-04-19
Canvas filter blur: Gaussian + Box

Pheo:
W3: https://www.w3.org/TR/SVG11/filters.html#feGaussianBlurElement

- 13-tap gaussian blur for s < 2
- 3-pass box blur for s >= 2

Notes:
- gaussian weights no longer hardcoded. we can optimize by caching weights.
- box blur handles even vs odd kernel as per W3. sanity limit of 1000px.
- edge sampling set to mirror. this avoids artifacts when blurring close to edge.

Testing with even-case box blur d=8: `ctx.filter='blur(4px)'`
image


Polyfills/Canvas/Source/FrameBufferPool.cpp
```diff
@@ -30,7 +30,7 @@ namespace Babylon::Polyfills
         {
             bgfx::FrameBufferHandle TextBuffer{bgfx::kInvalidHandle};
             Graphics::FrameBuffer* FrameBuffer;
-            int width(256), height(256);
+            int width(512), height(512);
```
CedricGuillemet, line 33: Should it be the same size (or with a factor) as the canvas it's attached to?

Polyfills/Canvas/Source/FrameBufferPool.cpp
```diff
@@ -40,10 +40,10 @@ namespace Babylon::Polyfills
             bimg::ImageContainer* image = bimg::imageAlloc(&Babylon::Graphics::DeviceContext::GetDefaultAllocator(), bimg::TextureFormat::RGBA8, width, height, 1 /*depth*/, 1, false /*cubeMap*/, false /*hasMips*/);
             const bgfx::Memory* mem = bgfx::makeRef(image->m_data, image->m_size, releaseFn, image);
             bx::memSet(image->m_data, 0, image->m_size);
-
+            // TODO: make sampler flags configurable
             std::array<bgfx::TextureHandle, 2> textures{
-                bgfx::createTexture2D(width, height, false, 1, bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT, mem),
-                bgfx::createTexture2D(width, height, false, 1, bgfx::TextureFormat::D24S8, BGFX_TEXTURE_RT)};
+                bgfx::createTexture2D(width, height, false, 1, bgfx::TextureFormat::RGBA8, BGFX_TEXTURE_RT | BGFX_SAMPLER_UVW_MIRROR , mem), // mirror avoids edge artifacts for blur
```
CedricGuillemet, line 45: I guess the artifacts appear with texels on the edge. did you try `BGFX_SAMPLER_V_BORDER` as well ? My guess is mirror mode will hide the issue until UVs offsets get big enough. Do you have a screenshot that show the problem with other wrapping modes?

Polyfills/Canvas/Source/Shaders/fs_boxblur.sc
```diff
@@ -0,0 +1,48 @@
...
+void main()
+{
+#if !NEED_HALF_TEXEL
+	vec4 u_halfTexel = vec4_splat(0.0);
+#endif // !NEED_HALF_TEXEL
+	vec2 texcoord0 = v_texcoord0 + u_halfTexel.xy;
+	vec4 color = vec4_splat(0.0);
+
+	// sample center if no offset (ie. odd kernel)
+	if (offsetX == 0.0f && offsetY == 0.0f)
```
CedricGuillemet, line 29: avoid comparing floats with literal values. check if value is absolute value is great than an epsilon. It should not be a problem if uniform set from C++ is a literal. but if it's the result of a computation, it might be something like 1.23e-6 which is not 0.

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.cpp
```diff
@@ -141,36 +193,110 @@ void nanovg_filterstack::Render(
...
+                    }
+                }
 
                 // Vertical Pass
-                float vertical[4] = {0.f, element.blurElement.vertical * 1.f, 0.f, 0.f}; // scale by element.blurElement.vertical
-                setUniform(m_uniforms.u_direction, vertical);
-                if (last)
+                if (element.blurElement.vertical < 2)
+                {
+                    std::vector<float> kernel = CalculateGaussianKernel(element.blurElement.vertical, BLUR_TAPS);
```
CedricGuillemet, line 242: same as previous code block. Is it possible to use 2 directions array `float directions[2][4]= {{1.f,0.,0.,0.},[0.,1.,0.,0.}} ` and replace with a loop ?

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.cpp
```diff
@@ -141,36 +193,110 @@ void nanovg_filterstack::Render(
...
+                        std::array<float, 4>{kernel[0], kernel[1], 0.f, 0.f},
+                        std::array<float, 4>{kernel[0], kernel[1], 0.f, 0.f},
+                        std::array<float, 4>{kernel[0], kernel[1], 0.f, 0.f},
+                    };
+                    std::array<std::array<float, 4>, 3> kernelsEven = {
+                        std::array<float, 4>{kernel[0], kernel[1], 0.f, -0.5f},
+                        std::array<float, 4>{kernel[0], kernel[1], 0.f, 0.5f},
+                        std::array<float, 4>{kernel[0] + 1, kernel[1], 0.f, 0.f},
+                    };
+                    float vertical[4] = {0.f, 1.f, 0.f, 0.f};
```
CedricGuillemet, line 272: make it static const

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.cpp
```diff
@@ -97,16 +115,50 @@ void nanovg_filterstack::ParseString(const std::string& string)
     }
 }
 
+std::vector<float> nanovg_filterstack::CalculateGaussianKernel(float sigma, int kernelSize)
```
CedricGuillemet, line 118: Is the kernelSize always the same value? If so, replace this computation with static constants in the shader.

Polyfills/Canvas/Source/FrameBufferPool.cpp
Pheo: If its better, I could try doing edge handling in shader? eg. for box blur, don't sample beyond edge and don't include when divide at the end.

CedricGuillemet: I'm not sure this will work. as number of samples will decrease closer to the border, I guess it will get less blurry ?

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.cpp
Pheo: Yep, kernel size always 13, but sigma/blur strength is variable. So we'll get different weights for different sigma. 
I still need to pass the weights as a uniform in this case I believe.

Polyfills/Canvas/Source/FrameBufferPool.cpp
Pheo: Yeah it will get less blurry.

Actually, maybe having the artifacts is "correct" for blur. Looking at this StackOverflow: https://stackoverflow.com/questions/24416564/svg-blur-filter-edge-artifacts
I see that the feGaussianBlur also has transparent looking edges same like BORDER sampling

I'll test this out in a browser and if it behaves the same, will switch to BORDER (if that is indeed the correct behaviour)

PR #1507, opened 2025-04-22
Canvas roundRect DOMPoint argument

Pheo:
MDN: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/roundRect
Adds support for single DOMPoint argument in roundRect (Context, Path2D)
Looks like `nvgRoundedRectVarying` already implemented seperate x/y radius, so just needed to add extra args

image

### DIFF
--- a/Polyfills/Canvas/Source/Path2D.cpp
+++ b/Polyfills/Canvas/Source/Path2D.cpp
@@ -336,6 +336,7 @@ namespace Babylon::Polyfills::Internal
             }
         }
         // DOMPoint
+        // TODO: move duplicate Path2D & Context args parsing into a utils.cpp
         else if (radii.IsObject())
         {
             const auto dompoint = radii.As<Napi::Object>();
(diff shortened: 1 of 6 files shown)


Polyfills/Canvas/Source/Path2D.cpp
```diff
@@ -335,5 +335,19 @@ namespace Babylon::Polyfills::Internal
                 throw Napi::Error::New(info.Env(), "Invalid number of parameters for radii");
             }
         }
+        // DOMPoint
+        else if (radii.IsObject())
+        {
+            const auto dompoint = radii.As<Napi::Object>();
```
CedricGuillemet, line 341: nit : Same code as Context. We should create a utils.cpp file to store all these functions. Maybe when merging the big CanvasTest PR.

CedricGuillemet (approved):
LGTM

Polyfills/Canvas/Source/Path2D.cpp
Pheo: TODO comments adding in Canvas.cpp + Context.cpp to remember to do this.

Pheo: re-opening this comment, I'll move duplicate Path2D + Context parsing into a utils.cpp in this PR

PR #1510, opened 2025-04-23
fix(cmake): Use CMAKE_BINARY_DIR for installation path

wuchangming:
Replace relative path with CMAKE_BINARY_DIR in Install.cmake to ensure consistent installation behavior across different platforms and support out-of-source builds. This change follows CMake best practices by keeping generated files within the build directory.

### DIFF
--- a/Install/Install.cmake
+++ b/Install/Install.cmake
@@ -1,7 +1,7 @@
 include(GNUInstallDirs)
 
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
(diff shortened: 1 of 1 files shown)


bghgary:
I'm not opposed to this change, but I don't see a problem on macOS. Can you elaborate on what you are seeing?

wuchangming:
When I build the 'install target' in xcode, I get the following error. And I saw other people also mentioned this problem, [link](https://forum.babylonjs.com/t/integrate-babylon-native-in-existing-swift-project/35886/31)

image


> I'm not opposed to this change, but I don't see a problem on macOS. Can you elaborate on what you are seeing?

bghgary (approved):

PR #1512, opened 2025-04-25
Add WebP support

bghgary:
- Added a new dependency on libwebp, updated NOTICES, etc.
- Added new CMake option to remove webp support as this dependency costs ~200KB to the final disk footprint.
- Updated some CMake variables in dependencies to use local variables instead of cached variables.
- Added a test for a glTF with WebP textures coming from the Babylon.js side.


ryantrem:
What's up with all the cmake chache changes? Also, will you add a description to this PR with some context?

ryantrem (approved):

bghgary:
> What's up with all the cmake chache changes? Also, will you add a description to this PR with some context?

Caching variables for setting options in dependencies is kinda dumb. CMake fixed this a while ago, but we have to either set a policy or update CMake min version on dependencies. I just set the policy since it would be impractical to update the dependencies.

I will add a description.

PR #1516, opened 2025-04-29
fix: don't allow loading same font more than once

Pheo:
@ryantrem found that when we load in a replacement font buffer..
the old buffer passed to nvgCreateFontMem is freed and can get corrupted

To solve this, we've made LoadTTFAsync effectively synchronous and don't allow loading in a duplicate font. We can consider future work to support safely updating existing fonts

### DIFF
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -59,22 +59,23 @@ namespace Babylon::Polyfills::Internal
         // called when removed from document which has no meaning for Native
     }
 
+    // this is currently synchronous to work around font corruption issues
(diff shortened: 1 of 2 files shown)


Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -59,22 +59,23 @@ namespace Babylon::Polyfills::Internal
         // called when removed from document which has no meaning for Native
     }
 
+    // this is currently synchronous to work around font corruption issues
```
ryantrem, line 62: Probably don't even need this comment as it is totally fine for it to be synchronous with the current impl. The previous asynchrony did not server any purpose.

ryantrem (approved):

PR #1518, opened 2025-05-14
Fix webp installation iOS

CedricGuillemet:
Replace Xcode task with cmake --build equivalent.
Previously task with `Playground` scheme was not building all targets but `install` scheme introduced an issue that I could not fix: `error: unable to attach DB: error: accessing build database`.

some related stackoverflow threads : 
https://stackoverflow.com/questions/73301738/xcode-13-4-1-error-unable-to-attach-db-error-accessing-build-database-react/73301739
https://stackoverflow.com/questions/51153525/xcode-10-unable-to-attach-db-error/52386834#52386834


.github/jobs/test_install_ios.yml
```diff
@@ -23,14 +23,9 @@ jobs:
...
-    - task: Xcode@5
-      inputs:
-        xcWorkspacePath: 'buildiOS/BabylonNative.xcodeproj'
-        scheme: 'Playground'
-        sdk: 'iphoneos'
-        useXcpretty: false
-        configuration: Release
-      displayName: 'Build Playground iOS'
+    - script: |
+        cmake --build buildiOS --config Release -- -sdk iphoneos CODE_SIGNING_ALLOWED=NO
```
bghgary, line 29: Maybe leave a comment on why we are not using the task here?

bghgary (approved):

PR #1521, opened 2025-05-15
Remove OpenXR dependencies

docEdub:
The OpenXR dependency used by Windows builds needs to be updated to work without errors when using CMake 4, but XR is not supported on Windows anymore so there is no easy way to test the changes.

To fix this issue, this change removes XR support on Windows and removes the OpenXR dependencies since they are only being used on Windows (Android and iOS use different XR implementations).


bghgary (approved):
LGTM

PR #1536, opened 2025-07-07
Changed Apple related headers to be C++ only.

SergioRZMasson:
Some of our platform specific headers for apple platform include code that requires Objective-C. If an integration wishes to include those in a pure C++ project it is currently not possible. This change stores objective-C objects as void* pointers for the common headers, following the standars that other libraries already use to avoid this type of problem (like bgfx, glfw)


ryantrem:
And this would mean for consuming projects that *do* use objective C, they will lose all the type safety, correct?

PR #1537, opened 2025-07-08
Fixed string replacement in metal shaders

SergioRZMasson:
When we do string replacement for metal shaders we might replace the wrong substring if the original sampler name had "Texture" in the name. This is the reason why shadows currently don't work when using metal.

### DIFF
--- a/Plugins/NativeEngine/Source/ShaderCompilerMetal.cpp
+++ b/Plugins/NativeEngine/Source/ShaderCompilerMetal.cpp
@@ -55,7 +55,7 @@ namespace Babylon
                 std::string imageName = resource.name;
                 if (imageName.find("Texture") != std::string::npos)
                 {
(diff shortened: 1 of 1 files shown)


Plugins/NativeEngine/Source/ShaderCompilerMetal.cpp
```diff
@@ -55,7 +55,7 @@ namespace Babylon
                 std::string imageName = resource.name;
                 if (imageName.find("Texture") != std::string::npos)
                 {
-                    imageName.replace(imageName.find("Texture"), std::string::npos, "");
+                    imageName.replace(imageName.rfind("Texture"), std::string::npos, "");
```
ryantrem, line 60: Probably worth adding a comment similar to your PR description, and maybe include an example of a problematic name that rfind addresses.

ryantrem (approved):

bghgary (approved):

PR #1538, opened 2025-07-08
FrameBuffer depthStencil texture destruction fix

CedricGuillemet:
- [JS thread] engine is disposed from 
- [main thread] resetView is called 
- [main thread] DisableRendering -> bgfx context is destroyed
- [JS thread] ExternalWithFinalizer for the `depthStencilTextureHandle`

### DIFF
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
@@ -20,7 +20,7 @@ namespace Babylon::Graphics
     class FrameBuffer final
     {
     public:
-        FrameBuffer(DeviceContext& context, bgfx::FrameBufferHandle handle, uint16_t width, uint16_t height, bool defaultBackBuffer, bool hasDepth, bool hasStencil);
+        FrameBuffer(DeviceContext& context, bgfx::FrameBufferHandle handle, uint16_t width, uint16_t height, bool defaultBackBuffer, bool hasDepth, bool hasStencil, int8_t depthStencilAttachmentIndex = -1);
         ~FrameBuffer();
 
         FrameBuffer(const FrameBuffer&) = delete;
@@ -69,5 +69,6 @@ namespace Babylon::Graphics
         Rect m_desiredScissor{};
 
         bool m_disposed{};
+        int8_t m_depthStencilAttachmentIndex{-1};
     };
 }
--- a/Core/Graphics/Source/FrameBuffer.cpp
+++ b/Core/Graphics/Source/FrameBuffer.cpp
@@ -37,10 +37,10 @@ namespace Babylon::Graphics
             {
                 bgfx::destroy(bgfx::getTexture(m_handle, m_depthStencilAttachmentIndex));
             }
+
             if (bgfx::isValid(m_handle))
             {
                 bgfx::destroy(m_handle);
-                m_handle = BGFX_INVALID_HANDLE;
             }
         }
 
--- a/Core/Graphics/Source/FrameBuffer.cpp
+++ b/Core/Graphics/Source/FrameBuffer.cpp
@@ -31,14 +31,17 @@ namespace Babylon::Graphics
             return;
         }
 
-        if (m_depthStencilAttachmentIndex >= 0)
+        if (m_deviceID == m_deviceContext.GetDeviceId())
         {
-            bgfx::destroy(bgfx::getTexture(m_handle, m_depthStencilAttachmentIndex));
-        }
-        if (bgfx::isValid(m_handle) && m_deviceID == m_deviceContext.GetDeviceId())
-        {
-            bgfx::destroy(m_handle);
-            m_handle = BGFX_INVALID_HANDLE;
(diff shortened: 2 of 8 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1889,8 +1889,8 @@ namespace Babylon
         }
 
         Graphics::FrameBuffer* frameBuffer = new Graphics::FrameBuffer(m_deviceContext, frameBufferHandle, width, height, false, generateDepth, generateStencilBuffer);
-        return Napi::Pointer<Graphics::FrameBuffer>::Create(info.Env(), frameBuffer, [frameBuffer, depthStencilTextureHandle]() {
-            if (bgfx::isValid(depthStencilTextureHandle))
```
ryantrem, line 1893: Doesn't this mean if the instance of the NativeEngine is disposed, but we don't do a graphics reset, then the texture would leak?

CedricGuillemet: yes! A check on the device would be better.

CedricGuillemet: Got a better fix after talking to @bghgary . I've tested it ok with BRN repro.

Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
```diff
@@ -20,7 +20,7 @@ namespace Babylon::Graphics
     class FrameBuffer final
     {
     public:
-        FrameBuffer(DeviceContext& context, bgfx::FrameBufferHandle handle, uint16_t width, uint16_t height, bool defaultBackBuffer, bool hasDepth, bool hasStencil);
+        FrameBuffer(DeviceContext& context, bgfx::FrameBufferHandle handle, uint16_t width, uint16_t height, bool defaultBackBuffer, bool hasDepth, bool hasStencil, int8_t depthStencilAttachmentIndex = -1);
```
bghgary, line 23: I think we can probably remove the hasDepth and hasStencil booleans. If depthStencilAttachmentIndex is not -1, then it has both depth and stencil, right?

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -30,6 +31,10 @@ namespace Babylon::Graphics
             return;
         }
 
+        if (m_depthStencilAttachmentIndex >= 0)
```
bghgary: We need this check in case the device is lost. Probably can combine the check with the one below.

```suggestion
        if (m_depthStencilAttachmentIndex >= 0 && && m_deviceID == m_deviceContext.GetDeviceId())
```

CedricGuillemet: fixed

Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
CedricGuillemet: It's possible to have stencil and/or depth. This needs to be checked or API will start asserting when clearing stencil+depth when only depth is available.

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -31,14 +31,17 @@ namespace Babylon::Graphics
...
-            m_handle = BGFX_INVALID_HANDLE;
+            if (m_depthStencilAttachmentIndex >= 0)
+            {
+                bgfx::destroy(bgfx::getTexture(m_handle, m_depthStencilAttachmentIndex));
+            }
+            if (bgfx::isValid(m_handle))
+            {
+                bgfx::destroy(m_handle);
+                m_handle = BGFX_INVALID_HANDLE;
+            }
```
bghgary: nit: we are inconsistent with how we handle the variables in the two cases here
```suggestion
            if (m_depthStencilAttachmentIndex >= 0)
            {
                bgfx::destroy(bgfx::getTexture(m_handle, m_depthStencilAttachmentIndex));
            }

            if (bgfx::isValid(m_handle))
            {
                bgfx::destroy(m_handle);
            }
```

Core/Graphics/InternalInclude/Babylon/Graphics/FrameBuffer.h
bghgary: We already have this check on the native engine side. We always create depth and stencil together.

Core/Graphics/Source/FrameBuffer.cpp
bghgary: Sorry, maybe we should consistently set the member variables to invalid values so that accessors will not return dead values when this object is disposed.

CedricGuillemet: I've put back m_handle = BGFX_INVALID_HANDLE;

bghgary: Should we also set m_depthStencilAttachmentIndex to -1?

CedricGuillemet: yes, fixed.

PR #1542, opened 2025-07-21
Linux/OpenGL CI Visualization Tests

CedricGuillemet:
Running on Linux Desktop needs OpenGL 46 and ES disabled. xvfb needs ES enabled.
Problematic scenes have comments in `config.json`.
Upcoming PR(s) to address test scene issues.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -368,14 +368,15 @@ namespace Babylon
                     {
                         bimg::ImageContainer* image{images[(side * numMips) + mip]};
 
+                        if (bgfx::getCaps()->originBottomLeft)
+                        {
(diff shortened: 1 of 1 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -372,6 +372,10 @@ namespace Babylon
                             bimg::imageFree(static_cast<bimg::ImageContainer*>(userData));
                         }};
 
+                        if (bgfx::getCaps()->originBottomLeft)
+                        {
+                            FlipImage({static_cast<uint8_t*>(image->m_data), image->m_size}, image->m_height);
+                        }
```
bghgary: This is in a weird place. The `releaseFn` above is connected to the mem variable. Probably move this above the `releaseFn` definition. Same below.

PR #1554, opened 2025-10-08
ES packages + Typescript

CedricGuillemet:
WIP


Apps/Playground/CMakeLists.txt
```diff
@@ -25,7 +30,7 @@ if(APPLE)
         set(STORYBOARD
             "${CMAKE_CURRENT_LIST_DIR}/iOS/Base.lproj/Main.storyboard"
             "${CMAKE_CURRENT_LIST_DIR}/iOS/Base.lproj/LaunchScreen.storyboard")
-        set(RESOURCE_FILES ${STORYBOARD} ${SCRIPTS})
+        set(RESOURCE_FILES ${STORYBOARD} ${SCRIPTS} ${JS_SCRIPTS})
```
bghgary, line 35: Do we now have both the `.ts` and `.js` files as resources? Is that what we want?

Apps/Playground/Scripts/config.json
```diff
@@ -288,7 +288,7 @@
         {
             "title": "Multi cameras and output render target",
             "renderCount": 2,
-            "playgroundId": "#BCYE7J#31",
+            "playgroundId": "#BCYE7J#46",
```
bghgary, line 291: Is this related to the TypeScript changes?

Apps/Playground/Scripts/experience.ts
```diff
@@ -1,7 +1,12 @@
...
-/// <reference path="../../node_modules/babylonjs-loaders/babylonjs.loaders.module.d.ts" />
-/// <reference path="../../node_modules/babylonjs-materials/babylonjs.materials.module.d.ts" />
-/// <reference path="../../node_modules/babylonjs-gui/babylon.gui.module.d.ts" />
+
+import { 
+    NativeEngine, ArcRotateCamera, Scene, SceneLoader, Tools, MeshBuilder, Vector3,
+    Mesh, VertexData, StandardMaterial, Color3, Color4, Texture, RenderTargetTexture,
+    WebXRFeatureName, PointsCloudSystem, VideoTexture, IWebXRHitResult, IWebXRVertexData,
+    WebXRMeshDetector, WebXRImageTracking, IWebXRTrackedImage, WebXRHitTest
+} from "babylonjs";
```
bghgary, line 9: Are we importing from the UMD package?

Apps/Playground/Scripts/experience.ts
```diff
@@ -335,27 +354,27 @@ CreateBoxAsync(scene).then(function () {
...
                             pcs.setParticles();
                         }
                     });
                 }
 
-                let sessionMode = vr ? "immersive-vr" : "immersive-ar"
+                let sessionMode: XRSessionMode = vr ? "immersive-vr" : "immersive-ar";
                 if (hololens) {
-                    // Because HoloLens 2 is a head mounted display, its Babylon.js immersive experience more closely aligns to vr
+                    // Because HoloLens 2 is a head mounted display, its js immersive experience more closely aligns to vr
```
bghgary: Is this intentional? Maybe a find and replace?

Apps/Playground/Scripts/experience.ts
```diff
@@ -390,15 +410,16 @@ CreateBoxAsync(scene).then(function () {
...
+                        xrSessionManager.scene.clearColor = new Color4(0, 0, 0, 0);
                     }
                 });
             });
         }, 5000);
     }
 
     if (text) {
-        const Writer = BABYLON.MeshWriter(scene, { scale: 1.0, defaultFont: "Arial" });
+        // @ts-ignore
```
bghgary: What's this for?

Apps/Playground/Scripts/validation_native.ts
```diff
@@ -1,21 +1,55 @@
...
+    functionToCall?: string;
+    rootPath?: string;
+    threshold?: number;
+}
+
+interface ConfigFile {
+    root: string;
+    tests: TestConfig[];
+}
+type compareFunctionType = (test: TestConfig, renderData: Uint8Array, threshold: number, errorRatio: number, referenceImage?: Uint8Array) => boolean;
```
bghgary: nit: maybe should be UpperCamelCase to be consistent?

```suggestion
type CompareFunctionType = (test: TestConfig, renderData: Uint8Array, threshold: number, errorRatio: number, referenceImage?: Uint8Array) => boolean;
```

bghgary (commented):
Quick first pass

Apps/Playground/Scripts/config.json
CedricGuillemet: Yes. "Use strict" now makes an assignation to a read only value trigger an exception. Assignation fixed in the PG.

Apps/Playground/Scripts/experience.ts
CedricGuillemet: indeed, find and replace. fixed.

Apps/Playground/Scripts/validation_native.ts
CedricGuillemet: Fixed.

Apps/Playground/Scripts/experience.ts
CedricGuillemet: No definition for `MeshWriter`.

Apps/Playground/CMakeLists.txt
CedricGuillemet: I have no strong opnion on that. It was handy to quickly open produced .js thanks to project hierarchy. Also to be consistent with `babylon.max.js` copy.

Apps/Playground/Scripts/experience.ts
CedricGuillemet: yes

Apps/Playground/CMakeLists.txt
bghgary: I'm more worried about it being added to the final running package. The `.ts` are source files, so they don't belong there, right? Or maybe I missed something.

Apps/Playground/Scripts/config.json
bghgary: I see. This test is from the JS side, so maybe we should update the config.json in Babylon.js also?

Apps/Playground/Scripts/experience.ts
bghgary: Maybe put a comment after `// @ts-ignore` to explain why.

Apps/Playground/Scripts/config.json
CedricGuillemet: Already done : https://github.com/BabylonJS/Babylon.js/pull/17287

Apps/Playground/Scripts/experience.ts
CedricGuillemet: fixed

Apps/Playground/CMakeLists.txt
CedricGuillemet: fixed

Apps/Playground/Scripts/experience.ts
```diff
@@ -223,32 +234,32 @@ CreateBoxAsync(scene).then(function () {
...
-                            const customMesh = new BABYLON.Mesh("custom", scene);
-                            const vertexData = new BABYLON.VertexData();
-                            vertexData.positions = mesh.positions;
-                            vertexData.indices = mesh.indices;
-                            vertexData.normals = mesh.normals;
+                            const customMesh = new Mesh("custom", scene);
+                            const vertexData = new VertexData();
+                            vertexData.positions = mesh.positions !== undefined ? mesh.positions : null;
+                            vertexData.indices = mesh.indices !== undefined ? mesh.indices : null;
+                            vertexData.normals = mesh.normals !== undefined ? mesh.normals : null;
```
bghgary, line 262: nit: Can we use `??` syntax?

Apps/Playground/Scripts/experience.ts
```diff
@@ -259,14 +270,14 @@ CreateBoxAsync(scene).then(function () {
...
                             console.log("Mesh updated.");
                             if (meshMap.has(mesh.id)) {
-                                const vertexData = new BABYLON.VertexData();
-                                vertexData.positions = mesh.positions;
-                                vertexData.indices = mesh.indices;
-                                vertexData.normals = mesh.normals;
+                                const vertexData = new VertexData();
+                                vertexData.positions = mesh.positions !== undefined ? mesh.positions : null;
+                                vertexData.indices = mesh.indices !== undefined ? mesh.indices : null;
+                                vertexData.normals = mesh.normals !== undefined ? mesh.normals : null;
```
bghgary, line 280: nit: ditto

Apps/Playground/Scripts/validation_native.ts
```diff
@@ -264,7 +299,7 @@
                         }
 
                         currentScene = eval(scriptToRun + test.functionToCall + "(engine)");
-                        processCurrentScene(test, renderImage, done, compareFunction);
+                        processCurrentScene(test, done, compareFunction);//, renderImage);
```
bghgary, line 302:
```suggestion
                        processCurrentScene(test, done, compareFunction);
```

Apps/Playground/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "Playground",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "build": "npx tsc --noEmit && npx babel Scripts --out-dir dist --extensions .ts",
+    "typecheck": "npx tsc --noEmit",
+    "postinstall": "npm run build"
```
bghgary, line 8: I made some changes to the UnitTests folder where the generated .js files are merged to source control. Should we do this same here?

PR #1555, opened 2025-10-15
Add NativeEncoding plugin

alexchuber:
This PR introduces the `NativeEncoding` plugin, which provides native image encoding capabilities to JavaScript. Raw pixel data can be encoded into standard image formats (PNG, JPEG, WebP, etc.), serving Babylon.js's needs for texture export.
See the README for more info.

As this PR is focused on the plugin (and some smoke tests for it), the following will be delegated to future PRs:
- Adding example use cases to the Playground
- Adding example use cases to the visual test suite
- Using the plugin from TestUtils

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ option(BABYLON_NATIVE_CHECK_THREAD_AFFINITY "Checks thread safety in the graphic
 option(BABYLON_NATIVE_PLUGIN_EXTERNALTEXTURE "Include Babylon Native Plugin ExternalTexture." ON)
 option(BABYLON_NATIVE_PLUGIN_NATIVECAMERA "Include Babylon Native Plugin NativeCamera." ON)
 option(BABYLON_NATIVE_PLUGIN_NATIVECAPTURE "Include Babylon Native Plugin NativeCapture." ON)
+option(BABYLON_NATIVE_PLUGIN_NATIVEENCODING "Include Babylon Native Plugin NativeEncoding." ON)
 option(BABYLON_NATIVE_PLUGIN_NATIVEENGINE "Include Babylon Native Plugin NativeEngine." ON)
 option(BABYLON_NATIVE_PLUGIN_NATIVEENGINE_WEBP "Include Babylon Native Plugin NativeEngine - WebP." ON)
 option(BABYLON_NATIVE_PLUGIN_NATIVEINPUT "Include Babylon Native Plugin NativeInput." ON)
@@ -101,7 +102,6 @@ option(BABYLON_NATIVE_PLUGIN_NATIVEOPTIMIZATIONS "Include Babylon Native Plugin 
 option(BABYLON_NATIVE_PLUGIN_NATIVETRACING "Include Babylon Native Plugin NativeTracing." ON)
 option(BABYLON_NATIVE_PLUGIN_NATIVEXR "Include Babylon Native Plugin XR." ON)
 option(BABYLON_NATIVE_PLUGIN_TESTUTILS "Include Babylon Native Plugin TestUtils." ON)
-option(BABYLON_NATIVE_PLUGIN_NATIVEENCODING "Include Babylon Native Plugin NativeEncoding." ON)
 
 # Polyfills
 option(BABYLON_NATIVE_POLYFILL_WINDOW "Include Babylon Native Polyfill Window." ON)
--- a/Plugins/NativeEncoding/README.md
+++ b/Plugins/NativeEncoding/README.md
@@ -4,6 +4,10 @@
 
 The NativeEncoding plugin provides native image encoding capabilities to Babylon, allowing raw pixel data to be encoded into standard image formats (PNG, JPEG, WebP, etc.).
 
+## Limitations
+
+Currently, **only PNG encoding** is supported.
+
 ## Design 
 
 Unlike a traditional polyfill which would implement Canvas's `toBlob()` or `toDataURL()` methods, NativeEncoding exists as a plugin because:
@@ -23,6 +27,3 @@ interface INative {
 
 It should be wrapped by higher-level Babylon.js APIs (e.g., DumpTools) for common workflows like asset exports and screenshots.
 
-## Limitations
-
-Currently, **only PNG encoding** is supported.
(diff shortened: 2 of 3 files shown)


Plugins/NativeEncoding/Source/NativeEncoding.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <bimg/encode.h>
+#include <bx/readerwriter.h>
+
+namespace Babylon::Plugins
+{
+    namespace
+    {
+        Napi::ArrayBuffer EncodePNG(Napi::Env env, const uint8_t* pixelData, uint32_t width, uint32_t height, bool invertY)
+        {
+            bx::MemoryBlock memoryBlock(&Graphics::DeviceContext::GetDefaultAllocator());
```
alexchuber: Didn't verify whether using the allocator from the device context is necessary, but I saw #1327 and did it just in case.

Plugins/NativeEncoding/Include/Babylon/Plugins/NativeEncoding.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
bghgary, line 1: It would be great to add a Readme.md to this plugin so that you can document the API (i.e., add the things you mentioned in the PR desc) and maybe also say it's experimental.

Plugins/NativeEncoding/Source/NativeEncoding.cpp
```diff
@@ -0,0 +1,58 @@
...
+            bx::MemoryWriter writer(&memoryBlock);
+
+            bimg::imageWritePng(&writer, width, height, width * 4, pixelData, bimg::TextureFormat::RGBA8, invertY);
+
+            auto byteLength = memoryBlock.getSize();
+            auto result = Napi::ArrayBuffer::New(env, byteLength);
+            std::memcpy(result.Data(), memoryBlock.more(0), byteLength);
```
bghgary: I guess copying is hard to avoid here since we don't know how big the buffer needs to be in advance.

Plugins/NativeEncoding/Include/Babylon/Plugins/NativeEncoding.h
alexchuber: That's a good idea 👍

CedricGuillemet:
`TestUtils` plugin has a basic png writer ( https://github.com/BabylonJS/BabylonNative/blob/a018058b9a614c898156db56c6d28d4396dac992/Plugins/TestUtils/Source/TestUtils.cpp#L23
 ) that should be replaced by this PR change. It's being used by https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Scripts/validation_native.js

ryantrem (approved):

Plugins/NativeEncoding/Source/NativeEncoding.cpp
```diff
@@ -0,0 +1,102 @@
...
+            auto runtimeScheduler{std::make_shared<JsRuntimeScheduler>(JsRuntime::GetFromJavaScript(env))};
+            auto pixelData{std::make_shared<std::vector<uint8_t>>(buffer.Data(), buffer.Data() + buffer.ByteLength())};
+
+            arcana::make_task(arcana::threadpool_scheduler, arcana::cancellation_source::none(),
+                [pixelData, width, height, invertY]() {
+                    return EncodePNG(*pixelData, width, height, invertY);
+                })
+                .then(*runtimeScheduler, arcana::cancellation_source::none(),
+                    [runtimeScheduler, deferred, env](const arcana::expected<std::vector<uint8_t>, std::exception_ptr>& result) {
+                        // TODO: Crash risk on JS teardown - this async work isn't tied to any JS object lifetime,
```
alexchuber, line 76: See issue #1558

Plugins/NativeEncoding/Source/NativeEncoding.cpp
```diff
@@ -0,0 +1,102 @@
...
+                        // unlike other plugins that cancel / clean up pending work in their destructors.
+                        if (result.has_error())
+                        {
+                            deferred.Reject(Napi::Error::New(env, result.error()).Value());
+                            return;
+                        }
+
+                        auto imageData{std::make_shared<std::vector<uint8_t>>(std::move(result.value()))};
```
bghgary: I'm not sure this will actually move correctly since result is a const reference. Can you confirm in the debugger the result value moved to imageData?

Plugins/NativeEncoding/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```typescript
+interface INative {
+    EncodeImageAsync: (pixelData: Uint8Array, width: number, height: number, mimeType: string, invertY: boolean) => Promise<ArrayBuffer>;
+}
+```
+
+It should be wrapped by higher-level Babylon.js APIs (e.g., DumpTools) for common workflows like asset exports and screenshots.
+
+## Limitations
```
bghgary: I think maybe this should be part of the top section so people don't have to read all the way to the bottom.

CMakeLists.txt
```diff
@@ -101,6 +101,7 @@ option(BABYLON_NATIVE_PLUGIN_NATIVEOPTIMIZATIONS "Include Babylon Native Plugin
 option(BABYLON_NATIVE_PLUGIN_NATIVETRACING "Include Babylon Native Plugin NativeTracing." ON)
 option(BABYLON_NATIVE_PLUGIN_NATIVEXR "Include Babylon Native Plugin XR." ON)
 option(BABYLON_NATIVE_PLUGIN_TESTUTILS "Include Babylon Native Plugin TestUtils." ON)
+option(BABYLON_NATIVE_PLUGIN_NATIVEENCODING "Include Babylon Native Plugin NativeEncoding." ON)
```
bghgary: Should be sorted with the rest of the options.

Plugins/NativeEncoding/Source/NativeEncoding.cpp
alexchuber: Ah, you're right. It appears to copy here. :(

bghgary: The result probably doesn't need to be const. And I think you have to make the lambda mutable. Then you should be able to move it.

alexchuber: I can remove the const, but if I do, I also need to remove the ref (making just `arcana::expected<..> result`), which results in a copy. 

I wound up using a shared_ptr to avoid the confusion of copy behavior & moving consts... sorry, my brain is mush :)

Issue #1557, opened 2025-10-20
several sanitizer issues across platforms


matthargett:
**Describe the bug**
there's a few memory-related bugs that GCC and LLVM sanitizers complain about. I'm filing this issue so I don't forget to circle back and try to resolve them.

**To Reproduce**
1. build debug unit tests and run them on platform: `cmake --build build --target UnitTests`

**Expected behavior**
should be clean of sanitizer issues.

**Screenshots**
### macOS (AddressSanitizer + UBSan, debug)
```
  /Users/matt/src/JsRuntimeHost/Core/Node-API/Include/Shared/napi/napi-inl.h:4504:17: runtime error: downcast of address
  0x60800004c420 which does not point to an object of type 'Babylon::Polyfills::Internal::URLSearchParams'
  0x60800004c420: note: object is of type 'Napi::ObjectWrap<Babylon::Polyfills::Internal::URLSearchParams>'
   01 00 00 00  90 1f 92 04 01 00 00 00  00 00 00 00 00 00 00 00
                ^~~~~~~~~~~~~~~~~~~~~~~
                vptr for 'Napi::ObjectWrap<Babylon::Polyfills::Internal::URLSearchParams>'
      #0  Napi::ObjectWrap<URLSearchParams>::ObjectWrap(...)
      #1  Babylon::Polyfills::Internal::URLSearchParams::URLSearchParams(...)
      #6  napi::ConstructorInfo::CallAsConstructor(...)
      #… (call stack continues into JavaScriptCore)
```
  Same warning fires repeatedly while the JS polyfill tests run.

  ### Linux (Ubuntu 24.04, ASan + UBSan)
```
  Direct leak of 4096 byte(s) in 1 object(s) allocated from:
      #0 0xffff93ee646c in malloc …
      #1 0xffff92d600c0 in JSContextGroupCreate (/usr/lib/aarch64-linux-gnu/libjavascriptcoregtk-4.1.so.0+0x1a00c0)
      #2 0xaaaac4d88108 in napi_open_handle_scope …/Core/Node-API/Source/js_native_api_javascriptcore.cc:626
      #3 0xaaaac4d9a0cc in napi_run_script …/Core/Node-API/Source/js_native_api_javascriptcore.cc:2576
      #…

  Direct leak of 64 byte(s) in 1 object(s) allocated from:
      #0 0xffff93ee646c in realloc …
      #1 0xffff91cb87d8 in __cxa_demangle …
[...]

PR #1559, opened 2025-10-27
Enable sanitizers for macOS, Linux, Windows

CedricGuillemet:
Enable Address sanitizer as pointed by https://github.com/BabylonJS/BabylonNative/issues/1557

- Add macOS, Windows and Linux sanitizer builds
- up ndk to v28
- misc build fixes
- fix reported memory leaks and bugs

Memleaks + CI issue : https://github.com/BabylonJS/BabylonNative/issues/1575

### DIFF
--- a/.github/jobs/linux.yml
+++ b/.github/jobs/linux.yml
@@ -36,16 +36,16 @@ jobs:
 # See https://github.com/BabylonJS/BabylonNative/issues/1575
 
     - script: |
-        cd build/Apps/UnitTests
-        xvfb-run ./UnitTests
-      displayName: 'Unit Tests'
-
-    - script: |
         cd build/Apps/Playground
         # Command line to suppress false positive memory leaks :
         # ASAN_OPTIONS=detect_leaks=1 LSAN_OPTIONS=suppressions=../../../.github/asan_suppress.txt xvfb-run ./Playground app:///Scripts/validation_native.js
         xvfb-run ./Playground app:///Scripts/validation_native.js
       displayName: 'Validation Tests'
+
+    - script: |
+        cd build/Apps/UnitTests
+        xvfb-run ./UnitTests
+      displayName: 'Unit Tests'
 
     - task: PublishBuildArtifacts@1
       inputs:
--- a/.github/jobs/win32.yml
+++ b/.github/jobs/win32.yml
@@ -63,24 +63,16 @@ jobs:
           reg add "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\UnitTests.exe" /v DumpType /t REG_DWORD /d 2
           reg add "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\UnitTests.exe" /v DumpCount /t REG_DWORD /d 1
           reg add "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\UnitTests.exe" /v DumpFolder /t REG_SZ /d "$(Build.ArtifactStagingDirectory)/Dumps"
-
         displayName: 'Enable Crash Dumps'
 
       - powershell: |
           $vs = vswhere -latest -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
           $msvc = Get-ChildItem "$vs\VC\Tools\MSVC" | Sort-Object Name -Descending | Select-Object -First 1
-          $asan = "$($msvc.FullName)\bin\Hostx64\x64"
+          $asanPath = "$($msvc.FullName)\bin\Hostx64\x64"
 
-          $destinations = @(
-              "build${{ variables.solutionName }}\Apps\Playground\RelWithDebInfo",
-              "build${{ variables.solutionName }}\Apps\UnitTests\RelWithDebInfo"
-          )
-
-          foreach ($dest in $destinations) {
-              Get-ChildItem "$asan\clang_rt.asan_*.dll" |
-                  Copy-Item -Destination $dest
-          }
-        displayName: "Copy ASAN runtime DLLs"
+          Write-Host "Adding ASAN runtime folder to PATH: $asanPath"
+          Write-Host "##vso[task.prependpath]$asanPath"
+        displayName: "Add ASAN runtime DLLs to PATH"
         condition: ${{ parameters.enableSanitizers }}
 
       - script: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ if(ENABLE_SANITIZERS)
         # MSVC only supports AddressSanitizer
         add_compile_options(/fsanitize=address /Zi /Od)
     else()
-        message(WARNING "Sanitizers not supported on this compiler.")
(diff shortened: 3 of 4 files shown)


Polyfills/Canvas/Source/nanovg/nanovg_filterstack.h
```diff
@@ -44,7 +44,10 @@ class nanovg_filterstack
     {
         // break down shadow as blur + color
     }
-    void Clear() { stackElements.clear(); }
+    void Clear()
+    { 
+        stackElements = std::vector<StackElement>(); // deallocate memory
```
matthargett: Is this a placement new of the existing stack allocation for the object? does it actually safely clear the heap-allocated internal members without introducing a leak?

CedricGuillemet: It's a test I did because a leak is reported with that Stack. I'm still trying to find why it leaks.

.github/jobs/linux.yml
```diff
@@ -22,16 +24,26 @@ jobs:
...
       displayName: 'Build X11'
     
+# Memory leaks on CI is disabled due to memory leaks reported with xvfb and impossible to add to ignore list.
+# See https://github.com/BabylonJS/BabylonNative/issues/1575
+
+    - script: |
+        cd build/Apps/UnitTests
+        xvfb-run ./UnitTests
+      displayName: 'Unit Tests'
+
```
bghgary: Did you intend to move this?

.github/jobs/win32.yml
```diff
@@ -62,6 +66,23 @@ jobs:
...
+              "build${{ variables.solutionName }}\Apps\Playground\RelWithDebInfo",
+              "build${{ variables.solutionName }}\Apps\UnitTests\RelWithDebInfo"
+          )
+
+          foreach ($dest in $destinations) {
+              Get-ChildItem "$asan\clang_rt.asan_*.dll" |
+                  Copy-Item -Destination $dest
+          }
+        displayName: "Copy ASAN runtime DLLs"
+        condition: ${{ parameters.enableSanitizers }}
```
bghgary, line 76: This is fine for now. Consider doing this inside CMake since it seems like it's part of the build.

.github/asan_suppress.txt
bghgary, line 1: nit: Maybe this file should go in the source and copied with the app so that you don't have to do the `../../..` thing.

Polyfills/Canvas/Source/nanovg/nanovg.cpp
bghgary, line 1: This file is a fork of the original right? Should these changes go upstream?

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.h
```diff
@@ -76,7 +76,9 @@ class nanovg_filterstack
             Blur blurElement;
         };
     };
-    std::vector<StackElement> stackElements;
+    int stackElementCount;
+    static const int MAX_STACK_SIZE = 32;
+    StackElement stackElements[MAX_STACK_SIZE];
```
bghgary, line 81: What's the reason for this change?

Polyfills/Canvas/Source/Canvas.h
```diff
@@ -22,19 +22,20 @@ namespace Babylon::Polyfills
...
-                m_impl.RemoveMonitoredResource(this);
+                m_impl->RemoveMonitoredResource(this);
             }
 
             virtual void FlushGraphicResources() = 0;
 
         private:
-            Canvas::Impl& m_impl;
+            // context is a monitoredResource managed by the GC while Canvas::Impl is managed by the app. Make sure the canvas impl is destroyed after all monitoredResources are.
+            std::shared_ptr<Canvas::Impl> m_impl;
```
bghgary, line 38: I don't know this code well. Is it okay to have different lifetime management for this?

CMakeLists.txt
```diff
@@ -107,6 +107,33 @@ option(BABYLON_NATIVE_PLUGIN_TESTUTILS "Include Babylon Native Plugin TestUtils.
...
+
+        string(JOIN "," SANITIZER_FLAGS ${SANITIZERS})
+
+        add_compile_options(-fsanitize=${SANITIZER_FLAGS} -fno-omit-frame-pointer)
+        add_link_options(-fsanitize=${SANITIZER_FLAGS})
+    elseif(MSVC)
+        # MSVC only supports AddressSanitizer
+        add_compile_options(/fsanitize=address /Zi /Od)
+    else()
+        message(WARNING "Sanitizers not supported on this compiler.")
```
bghgary: Maybe this should be an error instead of warning? I would think this is a mistake if someone tried to enable sanitizers on a platform that doesn't allow it.

azure-pipelines.yml
```diff
@@ -61,6 +67,22 @@ jobs:
...
+      graphics_api: D3D12
+      enableSanitizers: true
+
+  - template: .github/jobs/win32.yml
+    parameters:
+      name: Win32_x64_V8_D3D11_Sanitizers
+      vmImage: 'windows-latest'
+      platform: x64
+      napiType: V8
+      enableSanitizers: true
```
bghgary: How much time does this add to the CI if any? Do we need to run both with sanitizers and without?

Polyfills/Canvas/Source/nanovg/nanovg.cpp
CedricGuillemet: I'm more in favor of replacing it with something else.

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.h
CedricGuillemet: because nanovg allocations do not take care of constructor/destructor and the std::vector here was simply 'erased' with a memset.

.github/jobs/win32.yml
CedricGuillemet: Not sure. These dll are in PATH when debugging/testing with Visual Studio.

.github/asan_suppress.txt
CedricGuillemet: agreed

azure-pipelines.yml
CedricGuillemet: 1 should be enough, indeed.

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.h
bghgary: 🤔

.github/jobs/win32.yml
bghgary: Hmm, maybe we should do the same and add them to the path? Do you know what adds them to the path?

Polyfills/Canvas/Source/Canvas.h
CedricGuillemet: Previously `m_impl` was a reference to an impl created as a shared object. When the ref count dropped to 0 in the app when ref to canvas is rest, remaining threads kept that ref to an freshly destroyed impl.

.github/jobs/win32.yml
CedricGuillemet: No idea. It's not vcvars batch.

.github/asan_suppress.txt
CedricGuillemet: After more thoughts, I've changed my mind. I don't think it's the good place to place that file but I don't think copying it with destination makes sense.

PR #1560, opened 2025-10-29
Support v2 playgrounds in visual tests

alexchuber:
All new playgrounds are serialized in the v2 format, so this is bare minimum to support that.

### DIFF
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -182,7 +182,7 @@
                             let code = JSON.parse(snippet.jsonPayload).code.toString();
 
                             // Check if this is a v2 manifest and extract the entry file's code
-                            // TODO: This is a temporary solution, not proper support for v2 playground
+                            // TODO: Handle multi-file playgrounds
                             try {
                                 const manifestPayload = JSON.parse(code);
                                 if (manifestPayload.v === 2) {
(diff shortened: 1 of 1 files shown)


Apps/Playground/Scripts/validation_native.js
```diff
@@ -179,7 +179,22 @@
                         try {
                             xmlHttp.onreadystatechange = null;
                             const snippet = JSON.parse(xmlHttp.responseText);
-                            let code = JSON.parse(snippet.jsonPayload).code.toString()
+                            let code = JSON.parse(snippet.jsonPayload).code.toString();
+
+                            // Check if this is a v2 manifest and extract the entry file's code
+                            // TODO: This is a temporary solution, not proper support for v2 playground
```
bghgary: Why is this temporary?

Apps/Playground/Scripts/validation_native.js
```diff
@@ -198,6 +213,8 @@
 
                             currentScene = eval(code + "\r\ncreateScene(engine)");
 
+                            debugger;
+
```
bghgary: This needs to be removed :-)

Apps/Playground/Scripts/validation_native.js
alexchuber: Temporary maybe wasn't the right choice of word. Stop-gap? Minimal?
What I meant to convey is that there's unhandled cases here, namely multi-file support

PR #1561, opened 2025-10-30
Update arcana.cpp, bgfx, JsRuntimeHost to latest

bghgary:
arcana.cpp / JsRuntimeHost
- Use CMake FetchContent instead of git submodule for GSL.

bgfx
- Add layer index argument when calling overrideInternal.
- Add configuration options for buffer sizes, etc. to reduce memory usage.

### DIFF
--- a/Core/Graphics/Source/DeviceImpl.cpp
+++ b/Core/Graphics/Source/DeviceImpl.cpp
@@ -37,7 +37,17 @@ namespace Babylon::Graphics
         m_state.Bgfx.Initialized = false;
 
         auto& init = m_state.Bgfx.InitState;
-        init.type = s_bgfxRenderType;
+
+        // Use the noop renderer if the configuration has no window and no size.
+        if (config.Window == WindowT{} && config.Width == 0 && config.Height == 0)
+        {
+            init.type = bgfx::RendererType::Noop;
+        }
+        else
+        {
+            init.type = s_bgfxRenderType;
+        }
+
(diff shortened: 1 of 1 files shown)


Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -37,7 +37,17 @@ namespace Babylon::Graphics
...
+
+        // Use the noop renderer if the configuration has no window and no size.
+        if (config.Window == WindowT{} && config.Width == 0 && config.Height == 0)
+        {
+            init.type = bgfx::RendererType::Noop;
+        }
+        else
+        {
+            init.type = s_bgfxRenderType;
+        }
```
bghgary, line 49: bgfx fixed a bug when calling bgfx::init with zero width/height. It used to create a noop renderer before, but now it will create a headless Metal renderer instead. We still want a noop renderer for unit tests.

CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ FetchContent_Declare(base-n
     GIT_TAG 7573e77c0b9b0e8a5fb63d96dbde212c921993b4)
 FetchContent_Declare(bgfx.cmake
     GIT_REPOSITORY https://github.com/BabylonJS/bgfx.cmake.git
-    GIT_TAG ed8e11f781024120980cdbbc5871cb2ce2f63423)
+    GIT_TAG 468e9f8016a52498af261af594f1e2840ad5a7a3)
```
bghgary: This id needs to be updated once https://github.com/BabylonJS/bgfx.cmake/pull/88 goes in.

PR #1562, opened 2025-10-31
GS Datastream

CedricGuillemet:
- Decompression support for Gaussian Splatting data
- Adding libdeflate depencies. libdeflate doesn't support file hierarchy. So, adding both for gzip (SPZ) and zip (SOG)
- It's not meant to be a full implementation but rather a working solution so this code runs : https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/loaders/src/SPLAT/splatFileLoader.ts#L252-L267
- more PRs to come with testing and documentation
- Extend Canvas Image with new methods. SOG format is based on webp.

# Performances 1

I use this js code:
```
console.log(fflate.unzipSync); // confirm expected version is being used

BABYLON.Tools.LoadFile("https://raw.githubusercontent.com/CedricGuillemet/dump/master/sog-files/MardiGrasDragon.sog", (data) => {
    const datau8 = new Uint8Array(data);
    const t0 = Date.now();
    for (let i = 0; i < 1000; i++) {
        const unzipped = fflate.unzipSync(datau8);
    }
    const t1 = Date.now();
    console.log("Call took", (t1 - t0), "ms");
}, undefined, undefined, true);
```

and switch between Native and JS versions by calling (or not) this cpp code:
```
loader.LoadScript("https://unpkg.com/fflate/umd/index.js");
```

.sog file is a 14Mb zip containing webp textures with a ratio close to 1:1. 1000 decompressions. 

## JSC / macOS / M3

JS: 2630 ms
Native: 1671 ms

Native is faster but I expected way more difference due to JIT missing. Might be because of compression ratio.

## Chakra / W11 / IntelXeon

JS: 7530 ms
Native: 6600 ms

## V8 / W11 / IntelXeon

JS 12500 ms
Native: 12300ms

V8 slower than Chakra. Why?

## Profiles

V8/W11
image

Chakra/W11
image

# Performances 2

As I expected, SOG files with poor compression don't show potential for Native decompression.
Test2 is a zip file containing Babylon.max.js and babylon.module.d.ts. 19.4Mb compressed into a 3.37Mb zip.
Replace LoadFile line with:
```
BABYLON.Tools.LoadFile("https://raw.githubusercontent.com/CedricGuillemet/dump/master/sog-files/bjs.zip", (data) => {
```

1 single decompression.

## Chakra / W11 / IntelXeon

JS: 5600ms
Native: 51ms

## V8 / W11 / IntelXeon

JS 202 ms
Native: 45 ms

2nd round with 100 decompressions instead of 1
JS : 11400 ms
Native: 5483 ms

## JSC / macOS / M3
JS: 840 ms
Native: 28 ms

2nd round with 10 decompressions instead of 1
JS: 8700 ms
Native: 240 ms

# TLDR
When compression ratio is low like with SOG (zipped webp), decompression performance difference between JS and Native is less visible as unpacking is mainly memory copy.
The better the compression, the faster in comparison the Native version is. Up to 100X faster than Chakra, 30X faster than NoJIT JSC and 2X than V8 + JIT.
No idea why 1st performance test makes V8 2 times slower than Chakra.


CMakeLists.txt
```diff
@@ -44,6 +44,10 @@ FetchContent_Declare(SPIRV-Cross
 FetchContent_Declare(libwebp
     GIT_REPOSITORY https://github.com/webmproject/libwebp.git
     GIT_TAG 57e324e2eb99be46df46d77b65705e34a7ae616c)
+FetchContent_Declare(miniz
+    GIT_REPOSITORY https://github.com/richgel999/miniz.git
+    GIT_TAG 174573d60290f447c13a2b1b3405de2b96e27d6c)
```
bghgary: Make sure to update NOTICES with this new dependency.

CedricGuillemet: fixed

ryantrem:
@alexchuber introduced a more general NativeEncoding plugin a few weeks ago. It currently only has image encoding, but the idea was for it to be a general purpose plugin for different encoding/decoding related things. To make sure we could still get the full benefits of not linking anything that is not used, we talked about the idea of having multiple header includes for the plugin rather than the more typical single header. This way you could choose specific parts of the NativeEncoding plugin that you want. Seems like since this is very much an encoding/decoding feature, it should just go in NativeEncoding rather than introducing a new plugin. Thoughts @alexchuber and @bghgary?

Dependencies/CMakeLists.txt
```diff
@@ -233,3 +233,24 @@ endif()
...
+
+# --------------------------------------------------
+# miniz
+# --------------------------------------------------
+if(BABYLON_NATIVE_PLUGIN_ZIP)
+    FetchContent_MakeAvailable_With_Message(miniz)
+    target_compile_definitions(miniz PRIVATE MINIZ_DISABLE_ZIP_READER_CRC32_CHECKS)
+    set_property(TARGET miniz PROPERTY UNITY_BUILD false)
+    set_property(TARGET miniz PROPERTY FOLDER Dependencies)
+endif()
```
bghgary: bgfx already brings in a library called `minz` which I think is identical to this library. Can we just use that?

Plugins/Zip/Source/Zip.cpp
```diff
@@ -0,0 +1,90 @@
+#include <functional>
+#include <sstream>
+#include <Babylon/JsRuntime.h>
+#include <bx/bx.h>
+
+BX_PRAGMA_DIAGNOSTIC_PUSH();
+BX_PRAGMA_DIAGNOSTIC_IGNORED_CLANG_GCC("-Wunused-function");
+BX_PRAGMA_DIAGNOSTIC_IGNORED_MSVC(4505) // error C4505: '' : unreferenced local function has been removed
```
bghgary: Using bx headers in a library that has nothing to do with bgfx is a bit strange.

Apps/Playground/Win32/App.cpp
```diff
@@ -21,11 +21,13 @@
 #include <Babylon/Plugins/NativeCamera.h>
 #include <Babylon/Plugins/NativeInput.h>
 #include <Babylon/Plugins/TestUtils.h>
+#include <Babylon/Plugins/Zip.h>
```
bghgary: These libraries probably should also have tests?

Install/Test/CMakeLists.txt
```diff
@@ -166,6 +166,8 @@ target_link_libraries(TestInstall
     NativeCapture
     NativeEncoding
     NativeEngine
+    Zip
+    DecompressionStream
```
bghgary: I changed this to glob the files form the install folder, so you shouldn't have to add this manually anymore.

bghgary: Although, looking at this list, should these libraries have a `Native` prefix like the rest?

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -117,53 +135,100 @@ namespace Babylon::Polyfills::Internal
...
+    void NativeCanvasImage::SetSrc(const Napi::CallbackInfo& info, const Napi::Value& value)
+    {
+        if (value.IsString())
         {
-            arcana::make_task(m_runtimeScheduler, *m_cancellationSource, [env{info.Env()}, this, text{std::move(text)}, pos]() {
-                std::vector<uint8_t> base64Buffer;
-                bn::decode_b64(text.begin() + pos + base64.length(), text.end(), std::back_inserter(base64Buffer));
-                gsl::span<const std::byte> buffer = {reinterpret_cast<std::byte*>(base64Buffer.data()), base64Buffer.size()};
+            // url string
+            auto text{ value.As<Napi::String>().Utf8Value() };
```
bghgary: nit: Please format file with VS which should use clang-format

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -117,53 +135,100 @@ namespace Babylon::Polyfills::Internal
...
+                arcana::make_task(m_runtimeScheduler, *m_cancellationSource, [env{ info.Env() }, this, text{ std::move(text) }, pos]() {
+                    std::vector<uint8_t> base64Buffer;
+                    bn::decode_b64(text.begin() + pos + base64.length(), text.end(), std::back_inserter(base64Buffer));
+                    gsl::span<const std::byte> buffer = { reinterpret_cast<std::byte*>(base64Buffer.data()), base64Buffer.size() };
+
+                    if (!SetBuffer(buffer))
+                    {
+                        HandleLoadImageError(Napi::Error::New(env, "Unable to decode image with provided base64 source."));
+                    }
+                    });
```
bghgary: No one is observing exceptions being thrown of this task and thus exceptions will just disappear.

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -117,53 +135,100 @@ namespace Babylon::Polyfills::Internal
...
+            if (pos != std::string::npos)
+            {
+                arcana::make_task(m_runtimeScheduler, *m_cancellationSource, [env{ info.Env() }, this, text{ std::move(text) }, pos]() {
+                    std::vector<uint8_t> base64Buffer;
+                    bn::decode_b64(text.begin() + pos + base64.length(), text.end(), std::back_inserter(base64Buffer));
+                    gsl::span<const std::byte> buffer = { reinterpret_cast<std::byte*>(base64Buffer.data()), base64Buffer.size() };
+
+                    if (!SetBuffer(buffer))
+                    {
+                        HandleLoadImageError(Napi::Error::New(env, "Unable to decode image with provided base64 source."));
```
bghgary, line 171: Does `dispose` need to be called like elsewhere when something fails?

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -117,53 +135,100 @@ namespace Babylon::Polyfills::Internal
...
             }
-
-            Dispose();
-
-            auto buffer{request.ResponseBuffer()};
-            if (buffer.data() == nullptr || buffer.size_bytes() == 0)
-            {
-                HandleLoadImageError(Napi::Error::New(env, "Image with provided source returned empty response or invalid base64."));
-                return;
+            */
```
bghgary: Why do we want this commented out code here? Should it be removed?

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -117,53 +135,100 @@ namespace Babylon::Polyfills::Internal
...
+            if (blob.Has((uint32_t)0)) {
+                Napi::Value v = blob.Get((uint32_t)0);
```
bghgary: Can you use `0u`?

Polyfills/Canvas/Source/Image.cpp
```diff
@@ -117,53 +135,100 @@ namespace Babylon::Polyfills::Internal
...
+                if (v.IsTypedArray()) {
+                    Napi::Uint8Array array = v.As<Napi::Uint8Array>();
+                    gsl::span<const std::byte> buffer = { reinterpret_cast<std::byte*>(array.Data()), array.ByteLength() };
```
bghgary: Does gsl::make_span work? Same question elsewhere.

Polyfills/DecompressionStream/Source/DecompressionStream.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include "TextDecoder.h"
+#include "ReadableStream.h"
+#include "DecompressionStream.h"
+#include "Response.h"
+
+namespace Babylon::Polyfills::DecompressionStream
+{
+    void BABYLON_API Initialize(Napi::Env env)
+    {
+        Internal::TextDecoder::Initialize(env);
```
bghgary, line 13: Are we grouping these together to make it easier to deal with?

Polyfills/DecompressionStream/Source/DecompressionStream.h
```diff
@@ -0,0 +1,143 @@
...
+    void DecompressionStream::Initialize(Napi::Env env)
+    {
+        Napi::HandleScope scope{env};
+
+        Napi::Function func = DefineClass(
+            env,
+            JS_DECOMPRESSIONSTREAM_CONSTRUCTOR_NAME,
+            {
+                InstanceMethod("enqueue", &DecompressionStream::Enqueue),
+                InstanceMethod("close", &DecompressionStream::Close),
```
bghgary: I'm confused. These functions don't exist on the web version of `DecompressionStream` AFAICT. Why are they here?

Polyfills/DecompressionStream/Source/DecompressionStream.h
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    DecompressionStream::DecompressionStream(const Napi::CallbackInfo& info)
+        : Napi::ObjectWrap<DecompressionStream>{info}
+    {
+        const Napi::Env env{info.Env()};
+        if (info.Length() == 1 && info[0].IsString()) {
+            std::string compressionType = info[0].As<Napi::String>().Utf8Value();
+            if (compressionType != "gzip") {
+                throw Napi::Error::New(env, "Unexpected compression type.");
```
bghgary:
```suggestion
                throw Napi::Error::New(env, "Unsupported compression type.");
```

Polyfills/DecompressionStream/Source/DecompressionStream.h
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+    }
+
+    std::vector<std::byte> DecompressionStream::DecompressGzip(gsl::span<uint8_t> compressedBuffer)
+    {
+        std::vector<std::byte> result;
+
+        if (compressedBuffer.size() < 18) {
+            throw std::runtime_error("Invalid gzip data: too small");
```
bghgary: These should probably all be Napi::Error instead.

CMakeLists.txt
```diff
@@ -44,6 +44,13 @@ FetchContent_Declare(SPIRV-Cross
...
+FetchContent_Declare(libdeflate
+    GIT_REPOSITORY https://github.com/ebiggers/libdeflate.git
+    GIT_TAG c8c56a20f8f621e6a966b716b31f1dedab6a41e3)
```
bghgary: Please add `EXCLUDE_FROM_ALL` to these.

Polyfills/DecompressionStream/Source/DecompressionStream.cpp
CedricGuillemet: Yes, making only 1 polyfill for now to make it easier. It might change in the future.

Apps/Playground/Win32/App.cpp
CedricGuillemet: Yes. It will be part of upcoming PRs.

Polyfills/Canvas/Source/Image.cpp
CedricGuillemet: I've reworked the code so the image container is properly destroyed when a new buffer is set instead.

PR #1564, opened 2025-11-04
NativeEncoding: Use ArrayBufferView input and return Blob

alexchuber:
This PR brings NativeEncoding up to speed with the changes in [BabylonJS #17365](https://github.com/BabylonJS/Babylon.js/pull/17365). It also updates the unit tests accordingly.

- Expect input[0] to be an ArrayBufferView, not necessarily a Uint8Array
- Include Blob in BabylonNative build
- Return Blob from EncodeImageAsync
- Don't throw on unsupported MIME type
- Use std::byte for clarity

PR #1565, opened 2025-11-04
Playground: Upgrade BJS and add glTF serializer

alexchuber:
This PR adds a quick test for the glTF serializer in the Playground, updating BJS along the way.

- Upgrade BJS to latest (8.36.1)
- Fix installation test failure introduced in #1564
- Add the Blob polyfill, NativeEncoding plugin, and babylonjs-serializers package throughout the Playground app
- Port a visualization test for the glTF serializer from BJS
   - Note: I would port over more, but the rest happen to use the File API, which is not yet supported in Native.

### DIFF
--- a/.github/jobs/test_install_ios.yml
+++ b/.github/jobs/test_install_ios.yml
@@ -5,6 +5,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.name }}
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
     timeoutInMinutes: 60
     pool:
       vmImage: ${{ parameters.vmImage }}
--- a/.github/jobs/test_install_linux.yml
+++ b/.github/jobs/test_install_linux.yml
@@ -7,6 +7,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.name }}
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
     timeoutInMinutes: 30
     pool:
       vmImage: ${{ parameters.vmImage }}
--- a/.github/jobs/test_install_macos.yml
+++ b/.github/jobs/test_install_macos.yml
@@ -4,6 +4,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.name }}
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
     timeoutInMinutes: 30
     pool:
       vmImage: ${{ parameters.vmImage }}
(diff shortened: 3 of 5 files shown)


Apps/package.json
```diff
@@ -14,6 +14,7 @@
     "babylonjs-gui": "^8.28.2",
     "babylonjs-loaders": "^8.28.2",
     "babylonjs-materials": "^8.28.2",
+    "babylonjs-serializers": "^8.28.2",
```
alexchuber: TODO: Update BJS to latest after https://github.com/BabylonJS/Babylon.js/pull/17365

Apps/Playground/Scripts/config.json
```diff
@@ -1,6 +1,11 @@
 {
     "root": "https://cdn.babylonjs.com",
     "tests": [
+        {
+            "title": "GLTF Serializer multimaterial with raw texture",
+            "playgroundId": "#KU72PX",
+            "referenceImage": "glTFSerializerMultimaterial.png"
+        },
```
bghgary: Might be good to move it lower to a similar place as the JS one.

alexchuber: Going to sneak it underneath the glTF loader tests here since, if I were to follow the JS, it'd technically be wedged between some loader tests.

alexchuber: Nevermind, wedged it is!

bghgary: It's probably a mistake to wedge between loader tests, but it's okay either way.

Install/Test/CMakeLists.txt
```diff
@@ -136,6 +136,7 @@ target_link_libraries(TestInstall
     arcana
     bimg_encode
     bimg_decode
+    Blob
```
alexchuber: I'm not actually sure this will fix the win32 install test failure, but it seemed to be an obvious difference. Will test.

CedricGuillemet: Upper case is not a problem here?

alexchuber: Ah, it totally could be. I have yet to test this locally-- will report back.

alexchuber:
Install tests seem to be all good.
https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=45675&view=results
image

Install/Test/CMakeLists.txt
alexchuber: OK, tests seem to be passing both locally and in the CI. 

(I also went ahead and alphabetized these dependencies, just in case that's what you meant.)

PR #1566, opened 2025-11-11
Graphics device and external texture updates

bghgary:
This change includes new support for specifying back buffer resources, enhanced depth/stencil format control, and expanded external texture APIs to better handle multi-layered textures and platform-specific details.

- Add ability to specify the back buffer depth stencil format, including the ability to not use one
- Set min values for bgfx configuration to save memory
- Add ability to specify layer index to external texture with tests
- Update external texture lookup table for Metal
- Remove unused struct

### DIFF
--- a/.github/jobs/test_install_ios.yml
+++ b/.github/jobs/test_install_ios.yml
@@ -5,7 +5,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.name }}
-    #condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
     timeoutInMinutes: 60
     pool:
       vmImage: ${{ parameters.vmImage }}
--- a/.github/jobs/test_install_linux.yml
+++ b/.github/jobs/test_install_linux.yml
@@ -7,7 +7,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.name }}
-    #condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
     timeoutInMinutes: 30
     pool:
       vmImage: ${{ parameters.vmImage }}
--- a/.github/jobs/test_install_macos.yml
+++ b/.github/jobs/test_install_macos.yml
@@ -4,7 +4,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.name }}
-    #condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
     timeoutInMinutes: 30
     pool:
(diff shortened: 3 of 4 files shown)


Core/Graphics/Include/Shared/Babylon/Graphics/Device.h
```diff
@@ -28,6 +50,10 @@ namespace Babylon::Graphics
 
         // When enabled, back buffer will be premultiplied with alpha value.
         bool AlphaPremultiplied{};
+
+        // Format to use when creating the depth/stencil texture for the back buffer.
+        // Specify DepthStencilFormat::None to not create a depth/stencil texture.
+        DepthStencilFormat BackBufferDepthStencilFormat{DepthStencilFormat::Depth24Stencil8};
```
ryantrem, line 56: So now the bgfx default will be D24S8, correct? What was it before (I think it was unspecified?)?

Plugins/ExternalTexture/Source/ExternalTexture_Base.h
```diff
@@ -62,13 +63,13 @@ namespace Babylon::Plugins
...
 
-        void UpdateHandles(uintptr_t ptr)
+        void UpdateHandles(Graphics::TextureT ptr, std::optional<uint16_t> layerIndex)
         {
             std::scoped_lock lock{m_mutex};
 
             for (auto handle : m_handles)
             {
-                if (bgfx::overrideInternal(handle, ptr) == 0)
+                if (bgfx::overrideInternal(handle, reinterpret_cast<uintptr_t>(ptr), layerIndex.value_or(0)) == 0)
```
ryantrem: Is the third param a bool? I feel like I'm missing something.

Core/Graphics/Include/Shared/Babylon/Graphics/Device.h
bghgary: It was D24S8 before as well. It defaulted to this format internally.

Plugins/ExternalTexture/Source/ExternalTexture_Base.h
bghgary: The third parameter is an integer representing which index to use in the texture array when the handle/ptr is a texture array. See https://github.com/bkaradzic/bgfx/pull/3477.

ryantrem: Oh I missed a paren lol, my brain was seeing it as `if (bgfx::overrideInternal(handle, reinterpret_cast<uintptr_t>(ptr), layerIndex.value_or(0) == 0))`

Core/Graphics/Include/Platform/Unix/Babylon/Graphics/Platform.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <X11/Xlib.h>
+#undef None
```
ryantrem, line 4: What is this?

ryantrem (approved):

Core/Graphics/Include/Platform/Unix/Babylon/Graphics/Platform.h
bghgary: Xlib.h being dumb since it defines this and then anyone who uses the word None (like us in an enum) will cause the build to break.

PR #1570, opened 2025-11-17
up bjs, remove perf unittest

CedricGuillemet:


CedricGuillemet:
@docEdub @bghgary The error I get with more recent Babylon.js when runing UnitTests:
```
Error: Automatic publicPath is not supported in this browser
```
This error is fixed by adding this to the webpack config:
```
output: {
    publicPath: "/",
    globalObject: 'this',
}
```
I'm not sure what happens here.

docEdub (approved):

PR #1571, opened 2025-11-18
Clean up tests code

bghgary:


CedricGuillemet (approved):

Apps/UnitTests/JavaScript/dist/tests.nativeEngine.shaderCache.js
ryantrem, line 1: Why are the bundles checked in?

bghgary: We don't have to, but then we have to build them at some point. It's unclear where it would make the most sense. Would it be during CMake configure (which is where it used to do it)? After compiling the C++? It is also not clear whether the files should be sitting in the source tree or in the CMake binary folder. My preference is to commit these into source control and pretend they are source files that can be updated when necessary. This is what we do elsewhere.

PR #1574, opened 2025-11-20
Add external texture tests

bghgary:


Apps/UnitTests/CMakeLists.txt
```diff
@@ -14,19 +15,28 @@ set(TEST_ASSETS
...
+            PRIVATE USE_NOOP_METAL_DEVICE
+            PRIVATE SKIP_EXTERNAL_TEXTURE_TESTS)
+    endif()
     find_library(JAVASCRIPTCORE_LIBRARY JavaScriptCore)
     set(ADDITIONAL_LIBRARIES PRIVATE ${JAVASCRIPTCORE_LIBRARY})
-    set(SOURCES ${SOURCES} "Apple/App.mm")
 elseif(UNIX AND NOT ANDROID)
-    set(SOURCES ${SOURCES} "X11/App.cpp")
+    set(SOURCES ${SOURCES} "Source/App_X11.cpp")
+    set(ADDITIONAL_COMPILE_DEFINITIONS PRIVATE SKIP_EXTERNAL_TEXTURE_TESTS)
```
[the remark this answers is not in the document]
bghgary: Nope, that's incorrect. This usage is fine.

Apps/UnitTests/CMakeLists.txt
```diff
@@ -14,19 +15,28 @@ set(TEST_ASSETS
...
+    "Source/Tests.ExternalTexture.cpp"
+    "Source/Utils.h"
+    "Source/Utils_${GRAPHICS_API}.${BABYLON_NATIVE_PLATFORM_IMPL_EXT}")
 
 if(APPLE)
+    set(SOURCES ${SOURCES} "Source/App_Apple.mm")
+    if(BABYLON_NATIVE_TESTS_USE_NOOP_METAL_DEVICE)
+        set(ADDITIONAL_COMPILE_DEFINITIONS
+            PRIVATE USE_NOOP_METAL_DEVICE
+            PRIVATE SKIP_EXTERNAL_TEXTURE_TESTS)
```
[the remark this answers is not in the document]
bghgary: Nope, that's incorrect. This usage is fine.

Apps/UnitTests/CMakeLists.txt
bghgary, line 1: Need to update Install/Test again.

Apps/UnitTests/Source/Tests.ExternalTexture.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    EXPECT_EQ(externalTexture.Width(), 256u);
+    EXPECT_EQ(externalTexture.Height(), 256u);
+
+    update.Finish();
+    device.FinishRenderingCurrentFrame();
+#endif
+}
+
+TEST(ExternalTexture, AddToContextAsyncAndUpdate)
```
ryantrem, line 43: This test doesn't seem to validate anything, are we just making sure it doesn't crash or something?

Apps/UnitTests/Source/Utils_D3D11.cpp
```diff
@@ -0,0 +1,28 @@
+#include <gtest/gtest.h>
+#include "Utils.h"
+
+Babylon::Graphics::TextureT CreateTestTexture(Babylon::Graphics::DeviceT device, uint32_t width, uint32_t height)
```
ryantrem, line 4: Could we not use bgfx as a platform abstraction for creating test textures?

Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
```diff
@@ -95,7 +95,7 @@ namespace Babylon::Plugins
                         return;
                     }
 
-                    runtime.Dispatch([deferred = std::move(deferred), handle, &context, impl = std::move(impl)](Napi::Env env) {
+                    runtime.Dispatch([deferred = std::move(deferred), handle, &context, impl = std::move(impl)](Napi::Env env) mutable {
```
ryantrem, line 98: How is it that mutable is needed now but wasn't before? It doesn't seem like any actual code changed.

Apps/UnitTests/Source/Tests.ExternalTexture.cpp
bghgary: Yeah, it's kind of a baseline test at the moment. Ideally, we need to verify the contents of the texture. I may do that in a different PR. This is already a lot better than nothing.

Apps/UnitTests/Source/Utils_D3D11.cpp
bghgary: No, we can't. The purpose of the ExternalTexture class is to explicitly wrap platform specific textures. Also, bgfx is not exposed to Babylon Native usage.

Plugins/ExternalTexture/Source/ExternalTexture_Shared.h
bghgary: This is an issue that I noticed while debugging the code. It's not required for this PR. Without mutable, all the subsequent `std::move` calls will end up calling the copy constructor because the argument is `const`. Adding the mutable makes it such that the move actually happens. Captured lambda variables are const by default.

PR #1579, opened 2025-12-04
Add EXCLUDE_FROM_ALL to most fetch content dependencies

bghgary:
This will make the CMake install target grab only the header and libs from BabylonNative and JsRuntimeHost and exclude install headers/libs from dependencies.

Also includes improvements to the CI for testing the install targets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ include(FetchContent)
 # targets to the ALL_BUILD target.  We do this for all dependencies except
 # JsRuntimeHost because JsRuntimeHost has top-level targets (e.g., AppRuntime)
 # that are part of the Babylon Native contract. When adding dependencies using
-# FetchConent, add the EXCLUDE_FROM_ALL option except where there are top-level
+# FetchContent, add the EXCLUDE_FROM_ALL option except where there are top-level
 # targets that are part of the Babylon Native contract.
 
 FetchContent_Declare(AndroidExtensions


CMakeLists.txt
```diff
@@ -43,7 +52,8 @@ FetchContent_Declare(SPIRV-Cross
     GIT_TAG 6abfcf066d171e9ade7604d91381ebebe4209edc)
```
bghgary: This is missing an EXCLUDE_FROM_ALL

CMakeLists.txt
```diff
@@ -11,30 +11,39 @@ include(FetchContent)
 # --------------------------------------------------
 FetchContent_Declare(AndroidExtensions
```
bghgary, line 20: Add comment for EXCLUDE_FROM_ALL

ryantrem (approved):

.github/jobs/test_install_ios.yml
```diff
@@ -5,7 +5,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.name }}
-    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
+    #condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
```
bghgary: Revert before merge

azure-pipelines.yml
```diff
@@ -15,163 +15,162 @@ variables:
     value: 16.4
 
 jobs:
-  # Apple
```
bghgary, line 18: Undo before merge

.github/jobs/test_install_ios.yml
bghgary: Hmm, I have made the install jobs shorter, so maybe we can leave these in now.

PR #1604, opened 2026-02-16
Fix OGL internal texture format

CedricGuillemet:
- Add checks for opengl calls + assert
- destroy previous texture and framebuffer when getting new dimensions
- change texture internal format from GL_RGBA to GL_RGBA8: source of crash when trying to read back.

### DIFF
--- a/Plugins/NativeCamera/Source/Android/CameraDevice.cpp
+++ b/Plugins/NativeCamera/Source/Android/CameraDevice.cpp
@@ -17,6 +17,7 @@
 #include <arcana/macros.h>
 #include <memory>
 #include "CameraWrappers.h"
+#include <sstream>
 
 using namespace android;
 using namespace android::global;


Plugins/NativeCamera/Source/Android/CameraDevice.cpp
```diff
@@ -235,7 +238,13 @@ namespace Babylon::Plugins
 
                 EGLConfig config;
                 EGLint numConfig = 0;
-                eglChooseConfig(m_impl->display, attrs, &config, 1, &numConfig);
+                EGLBoolean chooseConfigResult = eglChooseConfig(m_impl->display, attrs, &config, 1, &numConfig);
+                if (chooseConfigResult != EGL_TRUE || numConfig <= 0)
+                {
+                    // Retrieve EGL error for debugging purposes.
+                    eglGetError();
```
bkaradzic-microsoft: You should print result of eglGetError.

bkaradzic-microsoft (approved):

PR #1617, opened 2026-03-04
Add precompiled shader test

bghgary:


Apps/Playground/Android/build/reports/problems/problems-report.html
bghgary, line 1: delete this

Apps/PrecompiledShaderTest/Source/App.Apple.mm
bghgary, line 1: test mac to make sure it works

bghgary: There is an issue in bgfx (fixed [here](https://github.com/bkaradzic/bgfx/pull/3620)), but it works. Since this is not tested in CI, we can merge and update bgfx later.

Issue #1398, opened 2024-06-28
Race condition possible in LoadTextureFromImage from NativeEngine.cpp


bghgary:
See https://github.com/BabylonJS/BabylonNative/blob/6bb77028369d7c91df0372448a6062ea37293b2b/Plugins/NativeEngine/Source/NativeEngine.cpp#L301-L317

The problem is that if Start/FinishRenderingCurrentFrame is being called on the graphics thread, then bgfx::frame can cause `releaseFn` to be called before this loop is done with `image`. One possible somewhat hacky solution is to save the `image->m_numMips` in a local variable to prevent usage of `image` as it is only possible for the `releaseFn` to be called on the last mip.

CedricGuillemet:
Also, the possibility of a memory leak if the last mipmap update goes wrong is a bit scary.

PR #1628, opened 2026-03-12
Fixed race condition when passing data to bgfx. Issue #1398

bkaradzic-microsoft:
Change makes `image` pointer not being touched after passing data to bgfx.

Alternative solution is to use `bgfx::copy` instead of `bgfx::makeRef`.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -398,8 +398,7 @@ namespace Babylon
                             }
 
                             bgfx::ReleaseFn releaseFn{};
-                            if (side == 5
-                            &&  mip == image->m_numMips - 1)
+                            if (mip == image->m_numMips - 1)
                             {
                                 releaseFn = [](void*, void* userData) {
                                     bimg::imageFree(static_cast<bimg::ImageContainer*>(userData));


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -398,7 +398,8 @@ namespace Babylon
                             }
 
                             bgfx::ReleaseFn releaseFn{};
-                            if (mip == image->m_numMips - 1)
+                            if (side == 5
+                            &&  mip == image->m_numMips - 1)
                             {
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed: https://github.com/BabylonJS/BabylonNative/pull/1628/changes/34f0171318a3cd944f9b0948c0efbdeb3a18060a

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -318,7 +318,7 @@ namespace Babylon
                 texture->Create2D(static_cast<uint16_t>(image->m_width), static_cast<uint16_t>(image->m_height), (image->m_numMips > 1), 1, Cast(image->m_format), flags);
             }
 
-            for (uint8_t mip = 0; mip < image->m_numMips; ++mip)
+            for (uint8_t mip = 0, numMips = image->m_numMips; mip < numMips; ++mip)
```
bghgary, line 321: I don't think this fixes the problem completely. We are still using `image` inside the for loop. At the minimum, the `image->m_numMips` usage needs to be replaced with `numMips`, but I'm not sure that's good enough.

Ditto for the other function.

bkaradzic-microsoft: The issue is the last loop when it needs to exit was touching image, which might be released from the other thread. Alternative is just to do `bgfx::copy` and not deal with `bgfx::makeRef / release`.

bghgary: let's discuss offline

PR #1638, opened 2026-03-19
Test Shader compilation with WEBMIN flag

CedricGuillemet:
Use a generated shader to check no mandatory glsl function missing in spirv-cross when using WEBMIN flag.

https://github.com/BabylonJS/SPIRV-Cross/pull/10

## Fix

`UniformTypeChangeTraverser` and `MoveNonSamplerUniformsIntoStruct` were also visiting whole `layout(std140) uniform` blocks (`EbtBlock`) and `uniform Foo foo;` struct instances (`EbtStruct`) and rewriting them to `vec4`, producing unnamed placeholder uniforms in the spirv-cross output and a `Identifier can't be empty.` assert in bgfx. Both traversers are now restricted to scalar/vector basic types (Float/Int/Uint/Bool). `CollectNonSamplerUniforms` also throws early if a uniform name is empty so future regressions surface in the compiler instead of deep in bgfx.


bkaradzic-microsoft (approved):

Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
```diff
@@ -0,0 +1,76 @@
+#include <gtest/gtest.h>
```
bghgary, line 1: [Reviewed by Copilot on behalf of @bghgary]

Naming nit: the existing pattern in this folder is `Tests.<SuiteName>.cpp` with one dot after `Tests.` and the filename segment matching the gtest `TEST()` first arg — `Tests.ExternalTexture.cpp` → `TEST(ExternalTexture, …)`, `Tests.ShaderCache.cpp` → `TEST(ShaderCache, …)`, etc. This file is `Tests.Shaders.Cross.cpp` + `TEST(ShadersCross, …)`, which breaks both (two dots in the filename, and the suite name reads oddly).

Suggestion: `Tests.ShaderCompilation.cpp` + `TEST(ShaderCompilation, CompileComprehensiveGLSL)`. The test exercises the whole compile pipeline (ShaderMaterial → glslang → spirv-cross → bgfx), not just the `ShaderCompiler` class directly, so "ShaderCompilation" matches what's actually under test.

bghgary (approved):

LGTM with one naming nit inline; replied to the open AI-reviewer thread.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -305,8 +309,14 @@ namespace Babylon::ShaderCompilerTraversers
...
-                if (type.getQualifier().isUniformOrBuffer() && type.getBasicType() != EbtSampler && !type.isMatrix())
+                // We only care about loose scalar/vector uniforms. Excluding matrices, samplers,
+                // UBO blocks (EbtBlock) and uniform struct instances (EbtStruct) prevents the
+                // traverser from rewriting whole blocks/structs to vec4, which destroys their
+                // member layout and member names.
+                const auto basic = type.getBasicType();
+                if (type.getQualifier().isUniformOrBuffer()
+                    && !type.isMatrix()
+                    && (basic == EbtFloat || basic == EbtInt || basic == EbtUint || basic == EbtBool))
                 {
```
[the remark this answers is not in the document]
bghgary: [Responded by Copilot on behalf of @bghgary]

The hardcoded `EbtFloat vec4` at L330–331 does coerce int/uint/bool to float, but this PR doesn't change that. The old condition (`!= EbtSampler && !isMatrix()`) already let `EbtInt`/`EbtUint`/`EbtBool` through into the same rewrite, so behavior for those types is unchanged here. The narrow goal — excluding `EbtBlock`/`EbtStruct` — is correctly achieved.

The new test only checks compilation doesn't throw, so an int→float coercion wouldn't fail it. Proper preservation (`ivec4`/`uvec4`/`bvec4`) is worth a follow-up issue, but doesn't block this PR.

bkaradzic-microsoft (approved):

Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
CedricGuillemet: Fixed

CedricGuillemet:
@bkaradzic-microsoft I had to add this commit : https://github.com/BabylonJS/BabylonNative/pull/1638/changes/4c044f512021675ec0d5eb0c84322d9414f38faf to get it to compile shaders with D3D11 warp devices. Let me know if it's a problem.

PR #1642, opened 2026-03-20
Add Xcode 26 CI jobs for iOS and macOS

CedricGuillemet:
Add iOS and macOS build jobs targeting Xcode 26


Core/Graphics/Source/DeviceImpl_iOS.mm
```diff
@@ -23,7 +23,28 @@
         float scale = static_cast<float>(((CAMetalLayer*)window).contentsScale);
         if (std::isinf(scale) || scale <= 0)
         {
-            scale = UIScreen.mainScreen.scale;
```
bghgary, line 26: I'm not sure I understand what it's doing here. Does this code actually work? The documentation says traitCollection.displayScale is 0.0 by default.

https://developer.apple.com/documentation/uikit/uitraitcollection/displayscale?language=objc

FYI @ryantrem

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -907,11 +907,24 @@ - (void)updateOrientation {
     UIApplication* sharedApplication{[UIApplication sharedApplication]};
     UIInterfaceOrientation orientation{UIInterfaceOrientationUnknown};
 #if (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_13_0)
-    UIScene* scene{[[[sharedApplication connectedScenes] allObjects] firstObject]};
-    orientation = [(UIWindowScene*)scene interfaceOrientation];
+    UIWindowScene* windowScene = (UIWindowScene*)[[[sharedApplication connectedScenes] allObjects] firstObject];
+    if (@available(iOS 26.0, *)) {
```
bghgary: Are we using different mechanisms to check iOS version?

azure-pipelines.yml
```diff
@@ -11,40 +11,57 @@ variables:
...
-  - name: XCODE_VERSION
-    value: 16.4
 
 jobs:
   # Apple
   - template: .github/jobs/macos.yml
     parameters:
       name: MacOS
       vmImage: macOS-latest
+      xcodeVersion: "16.4"
```
bghgary: Do we need to continue supporting the old Xcode version? The old Xcode version doesn't even run on the latest OS. How much time is this adding to the CI if we run both?

CedricGuillemet: xcode 16.4 has been release in may 2025. I'm a bit conservative on this front. An iOS/MacOS build is around 10minutes.

bghgary (commented):
[Review by Copilot on behalf of @bghgary]

Looks good to me. Still waiting on my earlier `displayScale` question.

Core/Graphics/Source/DeviceImpl_iOS.mm
CedricGuillemet: The 0.0 default only applies to manually constructed UITraitCollection objects where no display scale was specified. In that code, the traitCollection comes from a live UIWindowScene connected to the system.

bghgary: [Responded by Copilot on behalf of @bghgary]

Makes sense, thanks. Resolving.

bghgary (approved):
[Approved by Copilot on behalf of @bghgary]

PR #1655, opened 2026-04-09
Add OpenGLWindowsDevOnly graphics API for Windows

bkaradzic-microsoft:
Gate OpenGLES on Windows behind -DGRAPHICS_API=OpenGLWindowsDevOnly. CMake finds libEGL.dll and libGLESv2.dll from the Edge or Chrome installation via the App Paths registry key and copies them into the executable output directory at build time. Edge is tried first since it ships with Windows, followed by Chrome.


Core/Graphics/Include/RendererType/OpenGL/Babylon/Graphics/RendererType.h
```diff
@@ -1,10 +1,8 @@
 #pragma once
 
-#include <EGL/egl.h>
-
 namespace Babylon::Graphics
 {
-    using DeviceT = EGLContext;
+    using DeviceT = void *;
```
bghgary: why void*, what types are allowed?

bkaradzic-microsoft: Because `typedef void *EGLContext;` and then EGL/egl.h doesn't need to be included just to have `EGLContext` typedef.

CMakeLists.txt
```diff
@@ -202,13 +202,61 @@ elseif(UNIX)
...
+        set(BABYLON_NATIVE_OPENGLES_FROM_BROWSER ON)
     else()
         if(NOT GRAPHICS_API STREQUAL Vulkan AND NOT GRAPHICS_API STREQUAL D3D11 AND NOT GRAPHICS_API STREQUAL D3D12)
             message(FATAL_ERROR "Unrecognized/Unsupported render API: ${GRAPHICS_API}")
         endif()
     endif()
 endif()
 
+# Find ANGLE libraries from Edge or Chrome when using OpenGLWindowsDevOnly.
+if(BABYLON_NATIVE_OPENGLES_FROM_BROWSER)
```
bghgary, line 217: Does this mean someone can set this variable from CMake command line? I wonder this should be done using a function instead.

bghgary (approved):
Small comment. Looks good otherwise.

CMakeLists.txt
bkaradzic-microsoft: `BABYLON_NATIVE_OPENGLES_FROM_BROWSER` is a plain set() variable, not a cache or option() — so it's not settable from the command line and won't show up in cmake-gui or CMakeCache.txt. It's purely an internal flag derived from -DGRAPHICS_API=OpenGLWindowsDevOnly.

bghgary: [Responded by Copilot on behalf of @bghgary]

That's not quite right — `cmake -D` creates a **cache** variable, and if no normal variable shadows it, the cache value is still visible. So someone could pass `-DBABYLON_NATIVE_OPENGLES_FROM_BROWSER=ON` on the command line alongside a non-OpenGL `GRAPHICS_API`, and the `if(BABYLON_NATIVE_OPENGLES_FROM_BROWSER)` check would still evaluate to true. Using a function or guarding the later check against `GRAPHICS_API` would be safer.

PR #1656, opened 2026-04-13
Fix instancing data layout for all graphics backends.

bkaradzic-microsoft:
VertexBuffer: Always pack instance data in reverse attribute order. bgfx maps i_data0 to the highest semantic (TEXCOORD7) in all backends, so reverse iteration is correct universally, not just D3D11/D3D12.
 
ShaderCompilerTraversers: Fix OpenGL traverser to assign i_data names in reverse order (--m_instanceAttributeCount), matching the Metal traverser. Previously it used forward order (m_instanceAttributeIndex++) which produced wrong attribute-to-slot mapping when combined with reverse instance buffer packing.
 
ShaderCompilerTraversers: Add splatIndex0-3 to IsInstance() so Gaussian Splatting attributes are recognized as instance data on OpenGL and Metal.
 
config.json: Re-enable Gaussian Splatting tests on OpenGL and Metal now that instancing is fixed for those backends.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -5,22 +5,13 @@
             "title": "Gaussian Splatting Compressed ply SH",
             "playgroundId": "#U8O4EP#1",
             "renderCount": 15,
-            "referenceImage": "gsplat-compressedply-sh.png",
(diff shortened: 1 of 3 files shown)


bghgary (approved):
LGTM as long as the tests pass

PR #1665, opened 2026-04-17
Removed all D3D12 exclusions from tests.

bkaradzic-microsoft:
Re-enables the scissor test cases that were previously excluded on D3D12 due to an automatic mipmap generation issue in bgfx.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -328,32 +328,20 @@
         {
             "title": "Scissor test",
             "playgroundId": "#W7E7CF#34",
-            "referenceImage": "scissor-test.png",
-            "excludedGraphicsApis": [ "D3D12" ],
-            "comments": {
-                "D3D12": "reenable when automatic mip-maps issue is fixed in bgfx"
-            }
+            "referenceImage": "scissor-test.png"
         },
         {
             "title": "Scissor test with 0.9 hardware scaling",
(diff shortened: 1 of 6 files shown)


bghgary (commented):
[Review by Copilot on behalf of @bghgary]

Nice fix, and the comment on the color attachment in `CreateFrameBuffer` captures the subtlety well. One concern: there are other `Attachment::init(...)` call sites using the short overload that rely on the same default-ignores-unsupported-formats behavior and will hit the new bgfx assert after this bump. They aren't covered by the scissor tests, so CI going green won't prove they're safe:

- `Polyfills/Canvas/Source/FrameBufferPool.cpp:65` — loops over a color (`RGBA8`) and a depth (`D24S8`) attachment, both with the default overload. The depth one is the concern: depth formats don't support auto-mipgen, so this should reliably trip the new assert once Canvas is exercised.
- `Plugins/NativeXr/Source/NativeXrImpl.cpp:285-286` — same pattern, color + depth, both relying on defaults.

Could we fix these in the same PR? The change is mechanical (query `caps->formats[...] & BGFX_CAPS_FORMAT_TEXTURE_MIP_AUTOGEN` for color, explicit `BGFX_RESOLVE_NONE` for depth), and leaving them for later means a known assert waiting to fire on the first Canvas/WebXR workload after the merge.

CedricGuillemet (approved):

PR #1673, opened 2026-04-23
Tighten errorRatio on visual tests.

bkaradzic-microsoft:


Apps/Playground/Scripts/config.json
```diff
@@ -183,13 +182,11 @@
...
             "playgroundId": "#0YYQ3N#0",
-            "errorRatio": 50,
             "referenceImage": "softShadows.png"
         },
         {
             "title": "Soft Shadows (Right Handed)",
             "playgroundId": "#0YYQ3N#2",
-            "errorRatio": 50,
             "referenceImage": "softShadowsRightHanded.png"
         },
```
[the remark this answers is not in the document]
bkaradzic-microsoft: You have code to know "whatever the global/default threshold is". Use that to actually give good comments.

Apps/Playground/Scripts/config.json
```diff
@@ -38,7 +38,7 @@
             "title": "GLTF Extension EXT_mesh_gpu_instancing",
             "playgroundId": "#QFIGLW#9",
             "renderCount": 10,
-            "errorRatio": 25,
+            "errorRatio": 8,
```
bghgary, line 41: These numbers should match the ones coming from Babylon.js. Some of these values are different because native has some issues. Are these now matching the numbers from Babylon.js?

bghgary (approved):

Apps/Playground/Scripts/config.json
bkaradzic-microsoft: This is tightening based on native tests only. Not Babylon.js.

bghgary: Ideally, we shouldn't have native only visual tests, but okay.

PR #1674, opened 2026-04-23
[DO-NOT-MERGE] Partner integration test: #1652 + #1646

bghgary:
## ⚠️ DO-NOT-MERGE

Test vehicle for partner integration: combines #1652 + #1646. This PR exists solely to run CI and give partners a single branch to consume. **Do not merge** — the two source PRs land independently.

## Contents

- `origin/master` @ 81a7bfed (latest, includes #1673 and #1675)
- Merge of `origin/rework-thread-model` (#1652: Reworked threading model)
- Merge of `bghgary/external-texture-render-test` (#1646: Synchronous ExternalTexture API + recursive-mutex deadlock fix + Impl mutex encapsulation refactor)

## Conflict resolution

3 apps (`HeadlessScreenshotApp`, `PrecompiledShaderTest`, `StyleTransferApp`) conflicted where #1652 added a "reopen the gate so JS can continue running" block in the same region #1646 removed the old extra-frame-pump pattern. Kept #1652's new block in all three.

## Adjustments on top of merge

- **Tests.ExternalTexture.Msaa.cpp**: migrated from `AddToContextAsync` to sync `CreateForJavaScript`. The test pre-dates #1646 on master and its async pattern deadlocks under #1652's model; the sync API fits both PRs.

## Testing

- Local Win32 RelWithDebInfo build: clean.
- Local `ExternalTexture.*` unit tests: all 7 pass (Construction, CreateForJavaScript, Update, AddToContextAsyncAndUpdate, RenderWithMsaaSamples1/4, RenderTextureArray — last skipped on D3D12 as in #1646).
- Earlier revisions of this branch hung in `ExternalTexture.*`. Root cause was a recursive-mutex deadlock in `ExternalTexture::CreateForJavaScript` (lock held across a JS property lookup that ran a sibling Texture finalizer on the same thread, re-entering the same `std::mutex` and aborting under MSVC). Fixed in #1646 by resolving `Graphics::DeviceContext` before acquiring the impl lock; the broader Impl-mutex encapsulation refactor in #1646 makes this bug shape structurally impossible at the call sites that previously held the lock across JS callouts.


bghgary:

> Several runs of this branch on `Ubuntu_*_JSC` were timing out at the 30-minute job limit with no actionable post-mortem state.

I added a JS-progress watchdog in #1681 (sibling diagnostic PR, do-not-merge) and analyzed the resulting core dump on `Ubuntu_Clang_JSC`:

- Render thread (gtest) was in the normal `wait_for(16ms)` pump-loop window — not stuck in `Finish` or `m_frameStartDispatcher.tick()`.
- All bgfx renderer / worker threads idle in `cond_wait`.
- JS thread was blocked in `DeviceImpl::IncrementPendingFrameScopes`, but `DeviceImpl` state was `m_frameBlocked=true, m_pendingFrameScopes=0, m_rendering=false` — i.e. the expected quiescent state between `Finish` and the next `Start`.

I then re-ran the same `Tests.JavaScript` binary on WSL Ubuntu 24.04 + Mesa llvmpipe (matching the GHA software-GL stack); it passed all 3 tests in **231 s**, dominated by silent `StandardMaterial` / `PBRMaterial` shader compiles inside `bgfx::frame()`.

So the "hangs" are **not a deadlock in #1652's threading model** — they are slow-but-progressing runs of the Babylon Materials suite on software GL, with occasional runs exceeding the 30-min job budget on contended runners. Full write-up and gdb output in #1681.

Out-of-scope follow-ups that would help here:
- Bump the `Ubuntu_*_JSC` job timeout (for instance to 60 min) and/or
- Skip the slowest shader-compile cases on llvmpipe, and/or
- Look at bgfx-side llvmpipe shader compile cost.

I've left the watchdog in #1681 with a 10-min default so a future genuine hang would still produce a usable core dump rather than an opaque cancellation.

bghgary:

Update on the "deadlock" hypothesis on Ubuntu_*_JSC:

I built a heartbeat-fed watchdog (see #1681) that distinguishes slow-progress from real deadlock by sampling a render-thread frame counter every second and resetting the activity timestamp whenever it advances. With that mechanism in place, run [25026066984](https://github.com/BabylonJS/BabylonNative/actions/runs/25026066984) is **27✅ 1❌**:

- `Ubuntu_Clang_JSC` (the previously-failing job): passed `JavaScript.All` in 141.5 s. No watchdog abort.
- `Ubuntu_GCC_JSC` (the ❌): failed at **Playground rendering** with a pixel-diff (`First pixel off at 108872 ... Pixel difference: 17308 pixels`), which aborted the job before UnitTests ran. Pre-existing render flake, unrelated.

So the verdict from #1681 stands: **the JSC "hangs" are slow-but-progressing software-GL shader compiles in `Materials/*`, not a deadlock in the FrameCompletionScope model.** A single core dump couldn't distinguish the two because the gtest pump spends almost all wall time in `wait_for(16ms)` regardless. The frame-counter tiebreaker resolves it.

This doesn't say the threading rework #1652 is bug-free — only that the specific hangs we were chasing in CI on `Ubuntu_*_JSC` are not deadlocks. Useful follow-ups for the integration: bump the JSC job timeout, or skip the slowest Materials cases on llvmpipe, or investigate bgfx-side llvmpipe compile cost.

bghgary:

Both #1652 and #1646 have landed on master, so this combined integration reference PR is no longer needed. Closing.

PR #1675, opened 2026-04-23
Add programmatic GPU frame capture via TestUtils.captureNextFrame.

bkaradzic-microsoft:
Introduces a one-shot, cross-backend API for requesting a GPU debugger capture of the next submitted frame. Wraps bgfx's `BGFX_FRAME_DEBUG_CAPTURE` flag, so the same call works with Xcode Metal capture on macOS/iOS, PIX or RenderDoc on Windows, and RenderDoc on Linux/Android.

Primary use case: producing a .gputrace for a failing validation test without having to manually time a capture in the IDE. Add "capture": true to a test entry in Apps/Playground/Scripts/config.json and the test harness will request a capture on the screenshot frame; with the platform GPU debugger attached, a capture document opens automatically.

Implementation:

 - DeviceImpl owns an `std::atomic<bool> m_captureNextFrame`. `Frame()` consumes it via exchange(false) and ORs `BGFX_FRAME_DEBUG_CAPTURE` into the `bgfx::frame()` flags, guaranteeing the request is applied to exactly one frame even under concurrent JS-thread calls.
 - DeviceContext exposes `RequestCaptureNextFrame()`as a thin forwarder, extending the existing public surface alongside RequestScreenShot.
 - TestUtils plugin adds a cross-platform `captureNextFrame()` JS method (no per-platform .mm/.cpp variants required).
  validation_native.js honors an optional per-test "capture" flag and arms a capture before the final `currentScene.render()` of the screenshot frame.

When no GPU debugger is attached the bgfx flag is a no-op, so the API is safe to leave in JS that ships in release builds.

### DIFF
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -118,6 +118,9 @@
             }
             engine.runRenderLoop(function () {
                 try {
+                    if (test.capture && renderCount === 1 && TestUtils.captureNextFrame) {
+                        TestUtils.captureNextFrame();
+                    }
                     currentScene.render();
                     renderCount--;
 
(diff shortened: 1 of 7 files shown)


bghgary (approved):

LGTM

PR #1683, opened 2026-04-28
Polyfills/Canvas: fix invisible text, font-load races, and blit ordering

bkaradzic-microsoft:
Several long-standing NativeCanvas bugs caused text to be invisible or mis-rendered, fonts loaded after Context creation to be ignored, and canvas-to-texture copies to capture stale/empty content when more than one canvas was flushed per frame.
 
Context (Polyfills/Canvas/Source/Context.{h,cpp}):
 * Add EnsureFontsLoaded() and call it from Flush, SetFont, SetFontFaceId, and MeasureText. Previously fonts were registered with nanovg only in the Context constructor, so any font loaded after a Context was created (the common case for Babylon GUI) was silently ignored and text rendered with the default font id of -1.
 * Save()/Restore() now stack m_fillStyle and m_strokeStyle on a new m_savedStyles vector alongside nvg's own state stack. Babylon GUI brackets fillStyle changes with save/restore; without wrapper-side stacking the previous draw's fill colour leaked across the restore and u_innerCol was carried over from the prior draw call.
 * BindFillStyle: treat an empty fillStyle string as opaque white instead of transparent black, matching browser behaviour for an un-set fillStyle.
 * MeasureText: when the requested font family is not loaded, return Arial-equivalent metrics (~0.55 * em * length) instead of zero. Babylon GUI uses MeasureText results to centre text; a zero width combined with the bug below pushed glyphs entirely off-canvas.
 
MeasureText (Polyfills/Canvas/Source/MeasureText.cpp):
 * Zero-initialise bounds[4] and textMetrics[3] before calling nvgTextBounds / nvgTextMetrics. nanovg does not always populate every slot (in particular when no font face is bound), and the uninitialised stack values were being returned to JS as widths in the hundreds of millions, producing fillText x ≈ -1e8 and clipping text such as the GUI "Click Me" button completely off-screen. This single fix is the root cause of the most visible regression.
 
Shaders (Polyfills/Canvas/Source/Shaders/*.sc and regenerated headers for dxbc/dxil/essl/glsl/metal/spirv):
 * Remove the redundant u_viewSize uniform from vs_nanovg_fill, fs_nanovg_fill, vs_fspass, fs_boxblur and fs_gaussblur. bgfx already exposes viewport dimensions via the predefined u_viewRect uniform whose .zw components are width/height, so u_viewSize was a duplicate that had to be set every frame from C++ and was easy to forget.
 * nanovg/nanovg_babylon.cpp: drop the matching UniformHandle field, createUniform, setUniform and destroy calls.
 
 NativeEngine + DeviceContext (Plugins/NativeEngine/Source/NativeEngine.cpp, Core/Graphics/...):
 * Add DeviceContext::PeekNextViewId() that returns the next view id without consuming it, and use it in CopyTexture to schedule the bgfx blit on a view id that is greater than every view used so far. bgfx processes blits in numeric view-id order, so the previous code (which queued blits on view 0) ran the copy before the canvas Flushes that produced the source framebuffers had been submitted, and the destination texture received empty content. This manifested in tests like gltfAnimationSkinType where only the first dynamic label ("00") was visible and the rest were blank, even though four distinct Context/canvas/framebuffer instances were correctly created and flushed.


CedricGuillemet (approved):
LGTM

Issue #165, opened 2020-03-02
Playground demo: crash on MacOS in N-API


ptc-rgrasset:
MacBoo Pro (15 inch 2017) 
MacOS Mojave 10.14.6 (18G87)
XCode Version 11.3 (11C29)

Cloned repo as per today, followed the build instructions from README.md.

Running Playground app from Xcode, tried debug and release, same outcome:
crash is triggered from ViewController.mm, l62:
`    loader.LoadScript("babylon.max.js");`

and crash in js_native_api_Javascript_core.cc, l218:
`  JSStringRef stringRef = JSValueToStringCopy(env->m_globalContext, reinterpret_cast<JSValueRef>(value), nullptr);
`
Thread 6: EXC_BAD_ACCESS (code=EXC_I386_GPFLT)

(I additionally tried to setup the XCode custom working directory to the script directory, it didn't help)

<img width="1468" alt="babylonnative_napi_crashlog" src="https://user-images.githubusercontent.com/15635690/75718750-cddeb180-5cd3-11ea-84c2-aaa0f9154fc0.png">

syntheticmagus:
Hi ptc-rgrasset,

I think @CedricGuillemet's fixed this bug yesterday afternoon in #163. Can you pull latest master and try again? Thanks!

CedricGuillemet:
Hi @ptc-rgrasset 

Did you test with latest changes on master? It should work as expected now.

bghgary:
@ptc-rgrasset Let us know if you see other issues.

ptc-rgrasset:
Really sorry for the delay (with everything happening right now I didn't get a chance to retest that before)

I have re-tested with master tip, and I now get another crash:
<img width="1595" alt="BN_Playground_MacOS_Debug" src="https://user-images.githubusercontent.com/15635690/78060177-64fa6000-738b-11ea-8f82-6bd479c272f7.png">

imbrig:
@bghgary @ptc-rgrasset I'm experiencing the same throw when `loader.LoadScript("babylon.max.js");` is called.

MacBook Pro (15-inch, 2018)
MacOS Mojave 10.14.6
XCode Version 11.3
Python 3.7.7
cmake version 3.17.0
Followed the directions on the Readme.

Since I have bootcamp and Windows 10 VS2019, I tried it there and the Playground worked.
Someone I know has the same hardware but on Catalina(10.15.4) and XCode 11.4 and the Playground worked without issues.

[...]

PR #1688, opened 2026-05-06
Improve Playground debuggability.

bkaradzic-microsoft:
Make Playground crashes, asserts, and visual-test failures actually debuggable in CI and locally. Adds a callstack-printing crash handler, suppresses MSVC runtime dialogs, gives the Win32 build a real CLI with a headless mode, wires programmatic RenderDoc capture, makes the test runner report per-run summaries with pre-flight reference-image 
checks, and gives `TestUtils` a host-side exit callback so tests can drive the process exit code.
 
 ## Crash + assert handling
 
 - **`Apps/Playground/Shared/Diagnostics.{h,cpp}`** — install Win32 SEH + CRT-invalid-parameter + `abort`/signal handlers that print a `bx::writeCallstack`-style stack trace and a `BN: <CATEGORY>` banner before exiting; route `_CrtDbgReport` / `_CrtSetReportMode` away from the modal *"Microsoft Visual C++ Runtime Library"* dialog so asserts don't 
block CI.
 - **`Core/Graphics/Source/BgfxCallback.cpp`** — `fatal()` now always logs to `stderr` + bx trace, breakpoints only when a debugger is attached on Windows, then falls through to `abort()` instead of silent exit.
 
 ## Headless + CLI
 
 - **`Apps/Playground/Shared/CommandLine.{h,cpp}`** — new flag parser supports `--headless`, `--once`, `--test <name>`, `--capture[=<frame>]`, `--renderdoc-dll <path>`, plus passthrough of script path / extra args.
 - **`Apps/Playground/Win32/App.cpp`** — switch to `/SUBSYSTEM:CONSOLE` + `wWinMainCRTStartup`, UTF-8 console + unbuffered stdio, parse CLI, hide the top-level window in headless mode.
 - **`Apps/Playground/CMakeLists.txt`** — add `Diagnostics` + `CommandLine`, link `bx`, `/SUBSYSTEM:CONSOLE` for non-UWP Win32.
 - **`Apps/Playground/X11/App.cpp`, `macOS/main.mm`** — stub `Diagnostics::Initialize` so the new code links cleanly on every platform.
 
 ## RenderDoc
 
 - `Diagnostics::SetupRenderDoc` loads `renderdoc.dll` (override via `--renderdoc-dll`), pins the in-process API, and arms a single-frame capture when `--capture` is set.
 - **`.github/instructions/playground/playground-renderdoc-capture.instructions.md`** documents the host-side capture flow end-to-end.
 
 ## Test runner (`Apps/Playground/Scripts/validation_native.js`)
 
 - Pre-flight reference-image check (`TestUtils.referenceImageExists`); tests with no reference are reported as `missingRef`, not failures.
 - Per-run summary line:
   ```
   Run complete. ran=N passed=N failed=N missingRef=N skipped=N
   ```
 - Honours `--once` / `--test` from the CLI; `--capture` triggers `TestUtils.captureNextFrame` at the configured frame; bx-style debugger trigger on validation failure.
 
 ## TestUtils
 
 - **`Plugins/TestUtils`** — add `SetExitCallback()` so the host can react to tests calling `exit()` (drives the Playground process exit code); add `referenceImageExists(name)` (Win32 checks the deployed `ReferenceImages/` next to the executable; other platforms return `true` to preserve existing behaviour).
 
 ## Console output
 
 - **`Apps/Playground/Shared/AppContext.cpp`** — route `console.error` to `stderr`, install a global unhandled-exception handler that funnels JS errors through 
`Diagnostics::DumpFailure`.
 
 ## Docs
 
 - **`.github/instructions/babylon-native-debugging.instructions.md`** — reference for the Diagnostics output format, exit codes (`0`/`1`/`2`/`3`/`-1`), CLI flags, 
`MISSING_REFERENCE_IMAGE` flow, and the one-liner RenderDoc capture invocation.
 - **`.github/instructions/renderdoc/renderdoc-gpu-debug.instructions.md`** — small alignment edits for the new `--capture` flag.


Apps/Playground/Shared/AppContext.h
```diff
@@ -23,7 +25,8 @@ class AppContext
         size_t width,
         size_t height,
         DebugLogCallback debugLog,
-        AdditionalInitCallback additionalInit = {});
+        AdditionalInitCallback additionalInit = {},
```
bghgary, line 28: `const PlaygroundOptions*` with a `nullptr` default plus `if (playgroundOptions != nullptr)` guards in the ctor is a nullable raw pointer encoding "no CLI" — but `PlaygroundOptions` is a pure-data settings struct whose default-constructed state is already "nothing set." Suggest by-value with a default:

```cpp
AppContext(..., PlaygroundOptions playgroundOptions = {});
```

The function consumes the data once; by-value lets the caller `std::move(opts)` (the `TestFilters` / `TestIndices` vectors). The 6 `playgroundOptions != nullptr` checks in the ctor lambda all collapse to unconditional access. JS-side code in `validation_native.js` is unaffected — it already handles `_playgroundOptions` being either undefined or fully populated.

Apps/Playground/Win32/App.cpp
```diff
@@ -44,15 +47,15 @@ namespace
         return {url};
     }
 
-    std::vector<std::string> GetCommandLineArguments()
+    std::vector<std::string> GetCommandLineArgumentsW()
```
bghgary: Master had `GetCommandLineArguments` returning `std::vector<std::string>`. The PR renames it to `...W`. The `W` suffix is the Windows API wide-char convention (`CreateFileW`, `GetCommandLineW`) — paired with `A` for ANSI. Anyone reading the call site reasonably expects `std::vector<std::wstring>`; they get UTF-8 narrow strings instead.

Also a sneaky behavioral change: master had `for (int idx = 1; ...)` (skip argv[0]); the PR has `for (int idx = 0; ...)` (include it). Consistent with how `Parse()` skips argv[0] itself, but worth a comment, or rename to something semantic like `GetArgv` rather than reaching for a Win32-API suffix.

Suggest reverting to `GetCommandLineArguments`.

Apps/Playground/Win32/App.cpp
```diff
@@ -22,6 +24,7 @@ WCHAR szWindowClass[MAX_LOADSTRING]; // the main window class name
 std::optional<AppContext> appContext{};
 bool minimized{false};
 int buttonRefCount{0};
+PlaygroundOptions g_options{};
```
bghgary: The other 5 globals in this file (`hInst`, `szTitle`, `szWindowClass`, `appContext`, `minimized`, `buttonRefCount`) don't use a `g_` prefix. The new global mixes styles inside one file. Drop the prefix to match: `PlaygroundOptions options{};`.

Apps/Playground/Shared/CommandLine.cpp
```diff
@@ -0,0 +1,421 @@
...
+                m.consumedNextArg = true;
+            }
+        }
+        return m;
+    }
+}
+
+namespace CommandLine
+{
+    PlaygroundOptions Parse(int argc, const char* const* argv)
```
bghgary, line 310: The 13 repeated blocks of `if (auto m = match(...); m.matched) { opt.X = ...; continue; }` followed by `if (!err.empty()) { opt.ParseError = true; ... return opt; }` have clear structural duplication that table-driven dispatch eliminates.

A static `kFlags` table with one row per flag (longName, shortName, kind, apply functor) plus a single match/apply loop reduces ~280 lines to ~80. The same table can drive `PrintUsage` so flag definitions stop being a second source of truth for the help text — which is exactly what enabled the `--list` 3-way mismatch (see inline on validation_native.js).

Worth doing in this PR — every new flag pays the per-flag-boilerplate tax until it's fixed.

Apps/Playground/Scripts/validation_native.js
```diff
@@ -1,11 +1,80 @@
...
+            // Trigger the JS debugger if attached; on no-debugger runs the
+            // host's bx exception filter prints a callstack on the next throw.
+            // eslint-disable-next-line no-debugger
+            debugger;
+        }
+        done(false);
+    }
+
+    // Per-run counters surfaced as a final summary line on exit.
+    let _ranCount = 0;
```
bghgary: The `_` prefix on these local counters is a hangover from `_playgroundOptions` (where it has real meaning — host-injected from C++) and `_opts` (alias). For pure local counters inside an IIFE there's no public/private boundary or host-injection semantic — the prefix is just visual noise and dilutes the meaning of the prefix on `_playgroundOptions`. Drop on `_opts` and the counters; keep on `_playgroundOptions`.

Apps/Playground/Scripts/validation_native.js
```diff
@@ -372,21 +514,73 @@
         if (xhr.status === 200) {
             config = JSON.parse(xhr.responseText);
 
+            if (listTests) {
+                for (let i = 0; i < config.tests.length; ++i) {
+                    const t = config.tests[i];
+                    console.log("[" + i + "] " + (t.title || "") + " -> " + (t.referenceImage || ""));
```
bghgary: Three-way mismatch:

- `CommandLine.cpp` `--help`: `--list   List configured tests as TSV and exit (0).`
- `babylon-native-debugging.instructions.md`: "Prints TSV with index, title, exclusion reason."
- This code: `console.log("[" + i + "] " + (t.title || "") + " -> " + (t.referenceImage || ""))` — not TSV, no exclusion reason.

The earlier Copilot reviewer thread on this was resolved without aligning either side. Pick one canonical format and update all three. If you want machine-parseable, real TSV (`index\ttitle\treferenceImage\texclusionReason`) is the right answer; otherwise update the help text and doc to describe what's actually emitted.

Apps/Playground/Scripts/validation_native.js
```diff
@@ -372,21 +514,73 @@
...
+                const currentTitle = config.tests[i].title || "(unnamed)";
                 runTest(i, function (status) {
+                    _ranCount++;
                     if (!status) {
+                        _failedCount++;
+                        // Inner failures fire the debugger via failTest();
+                        // this catches the comparison-failed path.
+                        if (breakOnFail) {
+                            // eslint-disable-next-line no-debugger
+                            debugger;
```
bghgary: `failTest()` already does `if (breakOnFail) { debugger; }` before `done(false)`. Then this comparison-failure handler does it again. Pick one — the `failTest` site covers all failure paths, so this one can go.

.github/instructions/babylon-native-debugging.instructions.md
```diff
@@ -0,0 +1,296 @@
...
+### Run one test by name
+```powershell
+.\Playground.exe --headless --once --test "MultiRenderTarget" `
+    app:///Scripts/validation_native.js
+```
+Substring match, case-insensitive, against `test.title`.
+
+### Run one test by index (when titles collide or you only have a survey row)
+```powershell
+.\Playground.exe --headless --once --test-index=286 `
```
bghgary: `config.json` is appended to and reordered routinely; the next insert before #286 silently breaks this example. Suggest naming the test only and letting the reader find the index via `--list`, or pick an example that isn't index-fragile. (Same note applies to the test-286 references in `playground/playground-renderdoc-capture.instructions.md`.)

.github/instructions/playground/playground-renderdoc-capture.instructions.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Step 3 -- Clean up
+
+`--capture` and the rendered extra frames don't persist -- nothing to revert.
+Just delete `<cwd>/temp/` if you want to free disk.
+
+## What the rdc CLI **cannot** do (drop to renderdoc-py)
+
+- Show the IA input element layout (no `pipeline EID ia` section in
+  rdc 0.5.x with renderdoc-py 1.41).
```
bghgary: Hard-coded tool versions will go stale, and the doc doesn't say whether these are *what the PR was tested against* or *requirements*. Suggest moving to a single "tested against" line at the top, or dropping.

bghgary (commented):

Substantial improvement to BN diagnostics — the compareFrame/stopFrame separation for `--capture` and the host-side `SetExitCallback` hook are both well-placed. Most of the substantive concerns below are about *where* the code lives (right responsibility) rather than what it does.

## Architectural concerns

### `referenceImageExists` is a workaround for a JsRuntimeHost XHR polyfill bug

The pre-flight reference-image check in `validation_native.js` is motivated by:

> Without this, BN's URL polyfill throws a synchronous std::runtime_error inside its async task on a missing app:///ReferenceImages/<name>.png, bypassing BABYLON.Tools.LoadFile's onLoadFileError and surfacing as UNCAUGHT JS error before the runner can record the failure.

The actual bug is in `JsRuntimeHost/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp::Send()`:

```cpp
m_request.SendAsync()
    .then(...)                                    // sendRegion cleanup
    .then(m_runtimeScheduler, none, [this]() {    // success-only continuation
        SetReadyState(ReadyState::Done);
        RaiseEvent(EventType::LoadEnd);
        m_eventHandlerRefs.clear();
    })
    .then(inline_scheduler, none, [env](expected<void, exception_ptr> result) {
        if (result.has_error()) {
            Napi::Error::New(env, result.error()).ThrowAsJavaScriptException();
        }
    });
```

The success-only middle `.then` is skipped on async failure (per arcana semantics), so `ReadyState::Done` and the `LoadEnd`/`error` events are never raised. The final continuation throws as a JS exception — so `BABYLON.Tools.LoadFile`'s `onLoadFileError` (wired to `xhr.onerror`/`xhr.onreadystatechange`) never fires, and the failure surfaces as `UNCAUGHT JS ERROR` instead. This violates the XHR contract and breaks any caller using the standard `onerror` callback pattern — affects every file-load failure, not just missing reference images.

**Suggested fix** (in JsRuntimeHost): replace the success-only middle `.then` with a both-outcomes handler that on error sets a non-OK status and raises `error` + `loadend`. Drop the JS-throwing continuation. Then in this PR, `referenceImageExists` and the new TestUtils plumbing can be deleted; `BABYLON.Tools.LoadFile`'s `onLoadFileError` will fire normally on missing references and the runner records the failure cleanly. If the eager early-out is still wanted for performance/clarity, fine — but rewrite the comment to say *that*'s the reason, not a polyfill workaround.

### `console.error` wrapper belongs in JsRuntimeHost's Console polyfill, not AppContext

`AppContext.cpp` runs `new Function(body)()` to install a JS-side `console.error` wrapper that appends `(new Error()).stack` to the message. Concerns with this living in the host:

- **Stack mangled into the message arg** — `args.push('\nJS callstack:\n' + stack)` pollutes the final argument seen by the polyfill callback.
- **Engine-format heuristics in JS** — the `lines.shift()` normalization papers over per-engine stack-format differences in lowest-common-denominator JS. Belongs in the engine-aware polyfill where it can be done once and shared.
- **`new Function(body)` is eval-class** — disallowed under CSP-style sandboxing. A native polyfill capture has no such constraint.
- **Fragile layering** — wraps after `Polyfills::Console::Initialize`. Future scripts re-wrapping `console.error` either shadow or wrap this.

**Suggested fix:** move the capture into JsRuntimeHost's `Polyfills::Console::Initialize` using standard N-API:

```cpp
auto err = env.Global().Get("Error").As<Napi::Function>().New({});
auto jsStack = err.Get("stack").As<Napi::String>().Utf8Value();
```

Pass `jsStack` to `LogCallback` as a separate parameter. Pure refactor — same `new Error().stack` capture, same fidelity, no contract change to N-API. Engine-specific stack normalization is a separate PR; this one isn't blocked on it. Drop the `new Function(body)` block from AppContext entirely.

### Instruction docs vs. user docs

The new files in `.github/instructions/` mix two distinct kinds of content:

- **Genuine technical documentation** — exit codes, CLI flag reference, config.json knobs, RenderDoc DLL-loading mechanics, the worked test-286 example, the pre-flight failure-pattern table, "scroll up from the error line" troubleshooting. Useful for *anyone* debugging Playground tests.
- **Agent steering** — *"Read this **before**..."*, *"do NOT re-implement"*, *"Don't add parallel mechanisms"*, the keyword-bait "When to Use" lists, *"future agents"*, *"Cite this as the canonical example"*. Only useful as agent policy.

`.github/instructions/` paths are auto-loaded into Copilot agent context, so anything committed there is *active* steering rather than passive documentation. That makes maintenance discipline matter: when these go stale, agents are misled silently — exactly the readers least likely to push back.

**Suggested split:**
- Tech content → `Apps/Playground/README.md` (or `docs/playground-debugging.md`). Curated for humans; agents can still read it from the file system.
- `.github/instructions/` keeps only tight agent-behavior policy if any — e.g., "this repo has X infrastructure already, prefer it over reinventing," with a link to the human doc for mechanics.

This separates *knowledge* from *agent policy* and avoids the trap where the two get interleaved and rot together.

## Other issues called out inline.

.github/instructions/renderdoc/renderdoc-gpu-debug.instructions.md
```diff
@@ -1,6 +1,6 @@
 # RenderDoc GPU Debugging Instructions
 
-> Based on [rudybear/renderdoc-skill](https://github.com/rudybear/renderdoc-skill) — adapted for GitHub Copilot.
+> Based on [rudybear/renderdoc-skill](https://github.com/rudybear/renderdoc-skill) -- adapted for GitHub Copilot.
```
bghgary: This file's change is purely em-dash → `--` substitution. Doesn't belong in a PR titled "Improve Playground debuggability." Recommend its own commit, or drop entirely.

bkaradzic-microsoft:
Pushed [`5b4ad6bd`](https://github.com/bkaradzic-microsoft/BabylonNative/commit/5b4ad6bd) addressing the inline items (10 of 12). Two open routes for the architectural ones — want to confirm preference before opening companion work.

### 1. `referenceImageExists` / XHR polyfill bug

Confirmed your read of `XMLHttpRequest.cpp::Send()` — the success-only `.then` at L265 skips `SetReadyState(Done)` + `RaiseEvent(LoadEnd)` + `m_eventHandlerRefs.clear()` on async failure and falls through to the JS-throwing trailing continuation at L273. So `BABYLON.Tools.LoadFile`'s `onLoadFileError` can never fire. Two routes:

- **A** — open a JsRuntimeHost PR replacing the success-only middle `.then` with a both-outcomes handler (set non-OK `status`, raise `error` + `loadend`, drop the JS-throwing continuation). Bump the dep here, delete `TestUtils::referenceImageExists` + the runner pre-check before this PR lands.
- **B** — land this PR's pre-check as-is, file a JsRuntimeHost issue, remove the workaround in a follow-up here once that's fixed.

### 2. `console.error` stack wrapper

Agree the `new Function(body)()` capture belongs in `Polyfills::Console::Initialize`. Three routes:

- **A** — JsRuntimeHost PR adds a third callback parameter (e.g. `jsStack`) populated via N-API (`env.Global().Get("Error").New({}).Get("stack")`); bump here, delete the `new Function(body)` block.
- **B** — land this PR's wrapper, file an issue, follow up.
- **C** — drop the wrapper from this PR entirely. Rely on the unhandled-exception handler's banner alone; `console.error` keeps just the legacy `[Error] ...` line. (`console.error` covers Babylon.js's recoverable-but-still-bad paths that often don't reach the unhandled handler, but the JS engine-format heuristics in the wrapper are fragile, so dropping is defensible.)

Preference for each?

bghgary:

> Preference for each?

**Item 1 (XHR polyfill): A.** Fix `XMLHttpRequest::Send` upstream, bump the dep here, drop `TestUtils::referenceImageExists` + the runner pre-check in the same commit before #1688 lands.

**Item 2 (`console.error` stack wrapper): A.** Add the `jsStack` callback param via N-API in JsRuntimeHost, then delete the `new Function(body)` block. Your own point about the heuristics being fragile is exactly why this belongs at N-API — done once, correctly.

Happy to review both JsRuntimeHost PRs whenever they're ready.

bkaradzic-microsoft:
Update on the two architectural items:

Both upstream JsRuntimeHost PRs are now open and waiting for review.

**Item 1 (XHR workaround) -- Route A**: https://github.com/BabylonJS/JsRuntimeHost/pull/165 -- `XMLHttpRequest::Send` now raises `error` + `loadend` on async failure (transport exception OR non-2xx status). Comes with two regression tests (local `app:///` file not found, remote HTTP 404).

**Item 2 (`console.error` stack wrapper) -- Route A**: https://github.com/BabylonJS/JsRuntimeHost/pull/166 -- adds a 3rd `const char* jsStack` parameter to `Babylon::Polyfills::Console::CallbackT`; captured via N-API only on `Error` calls. This is a breaking API change for the public `Console` header. Comes with a regression test.

Once both PRs land, I'll update this PR to:
1. Bump `CMakeLists.txt` `GIT_TAG` to the new JsRuntimeHost commit.
2. Delete `TestUtils::ReferenceImageExists` (the .h/.cpp method, the JS registration, and the pre-check block in `validation_native.js`).
3. Delete the `new Function(body)` monkey-patch in `AppContext.cpp` and update the `Console::Initialize` lambda signature; forward the new `jsStack` to `Diagnostics::DumpFailure` in the JS CONSOLE ERROR path.
4. Remove the corresponding paragraph from `.github/instructions/babylon-native-debugging.instructions.md`.

bkaradzic-microsoft:
Update: in the process of writing PR #165's regression tests, surfaced two pre-existing UrlLib client-side-failure bugs that crash the host on missing local files (Unix worker-thread terminate, Apple `Open()` synchronous throw). Both fixed in BabylonJS/UrlLib#28 (`status=0 + loadend` semantics per @bghgary's preference).

So this PR is now blocked on three upstream merges:
1. BabylonJS/JsRuntimeHost#165 -- XHR async-failure events (CI green)
2. BabylonJS/JsRuntimeHost#166 -- Console `jsStack` callback (CI green)
3. BabylonJS/UrlLib#28 -- Unix/Apple client-side failure fixes (no CI on this repo)

bghgary (approved):

LGTM pending the dep updates.

PR #1569, opened 2025-11-14
Rework NativeTracing to be built on top of the PerfTrace helper from JsRuntime's Foundation lib

ryantrem:
This PR depends on https://github.com/BabylonJS/JsRuntimeHost/pull/125

- Rework NativeTracing to be build on top of the new PerfTrace helper from JsRuntime's Foundation lib.
- Add NativeTracing to the iOS app (the platform tracing is currently implemented for).
- Add a commented out line in experience.js that enables tracing through the Babylon.js API.

PR #1691, opened 2026-05-08
Fix instance data stride.

bkaradzic-microsoft:


Apps/Playground/Scripts/config.json
```diff
@@ -904,7 +902,7 @@
             "title": "GPUPicker",
             "playgroundId": "#1RWFMT#8",
             "excludeFromAutomaticTesting": true,
-            "reason": "Pixel comparison fails (more than 20% pixels differ)",
+            "reason": "Pixel diff after #1569 instance-data stride fix; not stride-related, needs follow-up.",
```
bghgary: The `after #1569` reference points at the NativeTracing rework (Ryan's PR from Nov 2025), not the stride fix. Either drop the PR cross-reference or change to #1691 (this PR) so the lineage is correct. Same fix applies to the ~20 other instances of this reason text in this file.

bghgary (approved):

LGTM except for one minor comment.

Apps/Playground/Scripts/config.json
bkaradzic-microsoft: @copilot fix this.

bkaradzic-microsoft: Fixed: https://github.com/BabylonJS/BabylonNative/pull/1691/commits/040760a0f9d19acef7c24d9fddffc38bdbca47d7

PR #1695, opened 2026-05-14
Bump SPIRV-Cross to 866e5edcc3407997a3ed62f3edf492f91bcb2629 (Expose more functions #10)

bkaradzic-microsoft:
## Summary

Bumps `BabylonJS/SPIRV-Cross` to [`866e5ed`](https://github.com/BabylonJS/SPIRV-Cross/commit/866e5edcc3407997a3ed62f3edf492f91bcb2629), the merge of [BabylonJS/SPIRV-Cross#11](https://github.com/BabylonJS/SPIRV-Cross/pull/11) on top of [BabylonJS/SPIRV-Cross#10](https://github.com/BabylonJS/SPIRV-Cross/pull/10) ("Expose more functions"). Together these remove `#ifndef SPIRV_CROSS_WEBMIN` guards and their `SPIRV_CROSS_INVALID_CALL()` (= `assert(false)`) stub bodies for the HLSL opcodes / helpers reachable from the BN test catalog: `OpFMod`, `OpLogicalOr`, `OpLogicalAnd`, `OpFwidthFine`, `OpVectorTimesMatrix` (×2 sites), `OpVectorExtractDynamic`, `OpSNegate`, plus `to_extract_constant_composite_expression`. Previously-stubbed opcodes now emit the actual HLSL instead of crashing or returning invalid shader code.

Also re-enables 17 `validation_native.js` tests in `Apps/Playground/Scripts/config.json` that were previously excluded with `"reason": "Test crashes or hangs on Babylon Native"` and now pass on Win32 D3D11:

idx 138, 241, 254, 293, 296, 309, 321, 322, 323, 332, 333, 373, 376, 391, 558, 559, 574.

## Note on remaining failures unmasked by the bump

In Release builds `SPIRV_CROSS_INVALID_CALL()` is a no-op, so SPIRV-Cross was silently emitting invalid shader code and the test pipeline appeared to hang waiting for a render. After this bump, the underlying JS-side errors are now visible. Several still-failing tests turn out to have causes unrelated to SPIRV-Cross:

1. The 26 Prepass SSAO + motion blur tests fail at the JS layer with `PrePassRenderer needs WebGL 2 support` **before** any shader compile — `NativeEngine` does not advertise the WebGL2 capability that `PrePassRenderer` requires.
2. One GPU-particles test (`GPU Particles - Animations`) now fails with `TypeError: Unable to get property 'ARRAY_BUFFER' of undefined` — same root cause as the GPU-particle `ARRAY_BUFFER` bug previously identified.
3. Two tests (`GreasedLine`, `Area Lights Standard Material`) fail to load LTC LUT textures: `Error: Unknown error opening URL` — UrlLib / asset resolution issue.
4. Three OpenPBR variants (Fuzz Weight, Transmission Dispersion — the `Realtime IBL` / `Prefiltered IBL` paths) still hit shader-compile errors — likely additional `SPIRV_CROSS_INVALID_CALL()` stub sites not yet reached by this bump.

These will be tracked as separate follow-up work (renderer-side WebGL2 capability, GPU-particle gl-constants binding, UrlLib LTC asset path, and a further SPIRV-Cross pass once a `Debug` build pinpoints the remaining stub sites).

## OpenGL backend coverage

A local `OpenGLWindowsDevOnly` build (Windows + ANGLE, same `STRINGIZE(GRAPHICS_API) == "OpenGL"` as Linux native OpenGL) was used to validate the re-enabled tests against the OpenGL backend. Four entries diverge on OpenGL only and have been marked `"excludedGraphicsApis": ["OpenGL"]` with a documented reason:

- idx 293, 296: `BGFX FATAL` `mediump float`→`int` shader-compile error in the `PrePassRenderer` fragment shader (OpenGL/ANGLE GLSL only).
- idx 391: 90 900-px diff in `sphere-with-custom-shader-to-display-wireframe-using-glow-layer` (glow-layer + custom-shader divergence).

idx 299 (`Prepass SSAO + GUI`) is held out of the re-enable batch entirely (kept `excludeFromAutomaticTesting: true`): it passes in isolation but produces a ~6000-px diff right at the 2.5 % `errorRatio` threshold when run after sibling Prepass-SSAO tests in the full sweep — flaky on Win32 D3D11 in CI.

## Build hygiene fix

The pre-existing `OpenGLWindowsDevOnly` build was broken under MSVC `/WX` because `ExternalTexture_Shared.h` lines 25/28 trip C4702 (unreachable code) — the OpenGL impl's `GetInfo` / `Set` / `Get` always throw, so the dispatch's no-throw fall-through is statically unreachable. Added `target_compile_options(ExternalTexture PRIVATE /wd4702)` gated on `GRAPHICS_API STREQUAL "OpenGL" AND MSVC` in `Plugins/ExternalTexture/CMakeLists.txt`. No other build is affected. The right long-term fix is to make the throws conditional in the OpenGL impl.

## Verification

- Win32 D3D11 (Release x64, Chakra), Windows ANGLE OpenGL (`OpenGLWindowsDevOnly`), and Linux native OpenGL all run the full sweep cleanly with the bump applied.
- Re-fetched `_deps/spirv-cross-src` matches the merged `866e5ed` byte-for-byte.

## Notes

- This only bumps the standalone SPIRV-Cross used by `Plugins/ShaderCompiler/Source/ShaderCompilerDXBC.cpp` (the one declared in `CMakeLists.txt`). The bgfx-bundled SPIRV-Cross under `_deps/bgfx.cmake-src/3rdparty/spirv-cross/` is unaffected.
- Upstream `BabylonJS/SPIRV-Cross` is now 2 commits ahead of the previous pin (PR #10 + PR #11).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ FetchContent_Declare(metal-cpp
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(SPIRV-Cross
     GIT_REPOSITORY https://github.com/BabylonJS/SPIRV-Cross.git
-    GIT_TAG e1d29b3617de7fb27fc085eb565a790b31679e54
+    GIT_TAG 866e5edcc3407997a3ed62f3edf492f91bcb2629
     EXCLUDE_FROM_ALL)
 
 # --------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ FetchContent_Declare(metal-cpp
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(SPIRV-Cross
     GIT_REPOSITORY https://github.com/BabylonJS/SPIRV-Cross.git
-    GIT_TAG 0e1654748e9e54b5469fb0fbc1c715b96fa06f8f
+    GIT_TAG e1d29b3617de7fb27fc085eb565a790b31679e54
     EXCLUDE_FROM_ALL)
 
 # --------------------------------------------------
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -872,8 +872,6 @@
         {
             "title": "Color correction",
             "playgroundId": "#FBH4J7#278",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "color-correction.png"
         },
         {
@@ -1629,8 +1627,6 @@
             "title": "Default pipeline",
             "playgroundId": "#NAW8EA#6",
             "renderCount": 20,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "defaultPipeline.png"
         },
         {
@@ -1722,8 +1718,6 @@
         {
             "title": "Default rendering pipeline",
             "playgroundId": "#5XB8YT#2",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "DefaultRenderingPipeline.png"
         },
         {
@@ -2319,16 +2313,12 @@
             "title": "Baked Vertex Animation with Depth Of Field",
             "playgroundId": "#JWO4R0#8",
             "renderCount": 20,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "bakedVertexAnimationDOF.png"
         },
         {
             "title": "Baked Vertex Animation with Volumetric Light Scattering Post Process",
             "playgroundId": "#1LF9GK#3",
             "renderCount": 20,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "bakedVertexAnimationVLS.png"
         },
         {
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ FetchContent_Declare(metal-cpp
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(SPIRV-Cross
     GIT_REPOSITORY https://github.com/BabylonJS/SPIRV-Cross.git
-    GIT_TAG 0e1654748e9e54b5469fb0fbc1c715b96fa06f8f
+    GIT_TAG 866e5edcc3407997a3ed62f3edf492f91bcb2629
```
bghgary, line 68: [Reviewed by Copilot on behalf of @bghgary]

Nit: the PR title says "to e1d29b3" but this pin is 866e5ed (e1d29b3 = PR #10 alone; 866e5ed = PR #10 + PR #11). Worth updating the title since it becomes the master commit subject.

bghgary (approved):

LGTM.

Binary size measured locally (Win32 D3D11 Release x64, Chakra):

| Artifact | Master | #1695 | Δ bytes | Δ% |
|---|---:|---:|---:|---:|
| **Playground.exe** (final linked binary) | 3,645,440 | 3,663,360 | **+17,920** | **+0.49%** |
| spirv-cross-hlsl.lib (static) | 7,249,176 | 7,368,506 | +119,330 | +1.65% |
| spirv-cross-core.lib | 7,836,406 | 7,836,422 | +16 | 0.00% |
| All other libs combined | — | — | ~+850 | timestamp noise |
| **Total of all .exe/.dll/.lib** | 109.38 MB | 109.52 MB | +0.14 MB | +0.13% |

The linker dead-strips HLSL emitters BN's shader catalog never calls, so only ~17 KB of the hlsl.lib's +119 KB lands in the exe. spirv-cross-msl.lib isn't built on Win32 — the upstream MSL changes cost zero here (will land on Apple platforms). Acceptable for the test-coverage gain.

CMakeLists.txt
bkaradzic-microsoft: Fixed.

PR #1700, opened 2026-05-16
Playground: bundle babylonjs-addons

bkaradzic-microsoft:
Bundles `babylonjs-addons` in the Playground app.

Babylon Native ships `babylonjs`, `babylonjs-gui`, `babylonjs-loaders`,
`babylonjs-materials`, and `babylonjs-serializers` alongside its Playground app
but not `babylonjs-addons`. The addons package hosts newer Babylon.js components
(Atmosphere, `LiquidRenderingSceneComponent`, ...) under the global `ADDONS`
namespace, so any Playground snippet that uses them throws
`ReferenceError: 'ADDONS' is not defined`.

**Change (4 files):**
- `Apps/package.json` / `Apps/package-lock.json` — add `babylonjs-addons`
  (`^9.3.4`, resolves/dedupes to 9.9.1, matching the other bundled Babylon
  scripts; single `babylonjs` version in `node_modules`).
- `Apps/Playground/CMakeLists.txt` — ship `babylonjs.addons.js` as a Playground
  resource via `BABYLON_SCRIPTS`.
- `Apps/Playground/Shared/AppContext.cpp` — load it right after `babylon.max.js`
  so addons init sees a fully constructed `BABYLON` global.

**Scope:** this PR is addons-bundling only. The ES2019/ES2022 Chakra
compatibility shims (`ErrorCause` / `StringTrim`, in either JS or C++ form) that
were previously part of this branch are **dropped** — per review, those Chakra
parse/runtime gaps should be handled separately (run Playground validation on an
ES-compliant engine such as `Win32_x64_V8_D3D11`, or bundle `babel-standalone`;
degraded Node-API error surfacing on Chakra to be fixed in JsRuntimeHost).

**Verified (Win32/Chakra):** rebased onto current master, Playground builds and
links cleanly, and the Atmosphere snippets no longer throw
`'ADDONS' is not defined` (addons loads; they now hit the separate Chakra
parse / texture-layer gaps, tracked elsewhere).

### DIFF
--- a/Apps/package-lock.json
+++ b/Apps/package-lock.json
@@ -12,6 +12,7 @@
       ],
       "dependencies": {
         "babylonjs": "^9.3.4",
+        "babylonjs-addons": "^9.3.4",
         "babylonjs-gltf2interface": "^9.3.4",
         "babylonjs-gui": "^9.3.4",
         "babylonjs-loaders": "^9.3.4",
@@ -2600,6 +2601,15 @@
       "integrity": "sha512-zMH8r0l/il9PJhy2+uS+/EyeLBy/ElZg0lKLj2d+ZrUarOanjMISaACw0HUx7C1EgYoXUFSMQLYcpXguleh8GA==",
       "hasInstallScript": true,
       "license": "Apache-2.0"
+    },
+    "node_modules/babylonjs-addons": {
+      "version": "9.9.1",
+      "resolved": "https://registry.npmjs.org/babylonjs-addons/-/babylonjs-addons-9.9.1.tgz",
+      "integrity": "sha512-Hi9QZqeanqG+4VAAmJmHJh7JKf1bc/Y5Ml+wuv+W+3f/dKltZpy/pii/QN4vhYRyrU0n+S8OJc+kVeaUUFjHEg==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "babylonjs": "9.9.1"
+      }
     },
     "node_modules/babylonjs-gltf2interface": {
       "version": "9.9.1",
--- a/Apps/package.json
+++ b/Apps/package.json
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "babylonjs": "^9.3.4",
+    "babylonjs-addons": "^9.3.4",
     "babylonjs-gltf2interface": "^9.3.4",
     "babylonjs-gui": "^9.3.4",
     "babylonjs-loaders": "^9.3.4",
--- a/Apps/Playground/CMakeLists.txt
+++ b/Apps/Playground/CMakeLists.txt
@@ -1,6 +1,7 @@
 FILE(GLOB REFERENCE_IMAGES RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "ReferenceImages/*")
 
 set(BABYLON_SCRIPTS
+    "../node_modules/babylonjs-addons/babylonjs.addons.js"
     "../node_modules/babylonjs-loaders/babylonjs.loaders.js"
     "../node_modules/babylonjs/babylon.max.js"
     "../node_modules/babylonjs-materials/babylonjs.materials.js"
--- a/Apps/Playground/Shared/AppContext.cpp
+++ b/Apps/Playground/Shared/AppContext.cpp
@@ -223,6 +223,7 @@ AppContext::AppContext(
     // Commenting out recast.js for now because v8jsi is incompatible with asm.js.
     // m_scriptLoader->LoadScript("app:///Scripts/recast.js");
     m_scriptLoader->LoadScript("app:///Scripts/babylon.max.js");
+    m_scriptLoader->LoadScript("app:///Scripts/babylonjs.addons.js");
     m_scriptLoader->LoadScript("app:///Scripts/babylonjs.loaders.js");
     m_scriptLoader->LoadScript("app:///Scripts/babylonjs.materials.js");
     m_scriptLoader->LoadScript("app:///Scripts/babylon.gui.js");


Apps/Playground/Scripts/error_polyfill.js
```diff
@@ -0,0 +1,97 @@
+// ES2022 Error(message, options) polyfill for Chakra runtimes.
```
bghgary: This should be a C++ polyfill under `Polyfills/` with `Initialize(Napi::Env)` — not a Playground-only JS script. The Playground shouldn't be the only app that gets correct `Error.cause` on Chakra; every Babylon Native consumer using Chakra will hit the same `Error: [object Object]` problem.

Apps/Playground/Scripts/string_polyfill.js
```diff
@@ -0,0 +1,27 @@
+// ES2019 String.prototype.trimStart / trimEnd polyfill for Chakra.
```
bghgary: Same — `trimEnd`/`trimStart` should be a C++ polyfill under `Polyfills/` so all consumers get it, not just the Playground.

Apps/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "babylonjs": "^9.3.4",
+    "babylonjs-addons": "^9.3.4",
```
bghgary, line 13: `babylonjs-addons` resolved to 9.7.0 in the lockfile while the other Babylon packages resolved to 9.3.4. The addons JS may reference APIs added after 9.3.4 that aren't in the shipped `babylon.max.js`. Worth pinning to the same resolved version.

bghgary (changes requested):

The addons bundling is fine, but the Error and String polyfills should not be JS scripts loaded by the Playground app — that only fixes it for the Playground. Any other app embedding Babylon Native with Chakra would hit the same failures and have to independently discover and ship the same workaround. These should be proper C++ polyfills (like Console, TextDecoder, etc.) with an `Initialize(Napi::Env)` entry point so every consumer gets them automatically.

Apps/Playground/Scripts/error_polyfill.js
bkaradzic-microsoft: Done -- moved to `Polyfills/ErrorCause/` in 037196b3 with `BABYLON_API Initialize(Napi::Env env)`, wired into Win32 + Android Playground (`Apps/Playground/CMakeLists.txt` + `Apps/Playground/Android/BabylonNative/CMakeLists.txt`). JSI build path fixed in 57bceace. `error_polyfill.js` removed.

Apps/Playground/Scripts/string_polyfill.js
bkaradzic-microsoft: Done -- moved to `Polyfills/StringTrim/` in 037196b3, same `Initialize(Napi::Env)` shape, same CMake wiring. `string_polyfill.js` removed.

Apps/package.json
bkaradzic-microsoft: Done -- `babylonjs-addons` pinned to `^9.3.4` in `Apps/package.json` and `package-lock.json` regenerated. `node_modules/babylonjs-addons` now resolves to `9.3.4` and dedupes against the rest of the babylonjs tree.

Apps/package-lock.json
```diff
@@ -2601,6 +2602,22 @@
...
+        "babylonjs": "9.7.0"
+      }
+    },
+    "node_modules/babylonjs-addons/node_modules/babylonjs": {
+      "version": "9.7.0",
+      "resolved": "https://registry.npmjs.org/babylonjs/-/babylonjs-9.7.0.tgz",
+      "integrity": "sha512-buaBjhIuBCpJNlwG+KUfqTtC0wvVQuK8ruFDcegjObrbPIrYqOp6qVdtqBgGtWrzmvWj4bGpUIQgG6HCwXw2JA==",
+      "hasInstallScript": true,
+      "license": "Apache-2.0"
+    },
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Done -- `babylonjs-addons` pinned to `^9.3.4` in `Apps/package.json` and `package-lock.json` regenerated; lockfile now shows `node_modules/babylonjs-addons` at `9.3.4`, deduped against the rest of the babylonjs tree (same fix as the other addons thread).

Polyfills/ErrorCause/Source/ErrorCause.cpp
```diff
@@ -0,0 +1,97 @@
...
+                    value: options.cause,
+                    writable: true,
+                    configurable: true,
+                    enumerable: false
+                });
+            }
+            return err;
+        }
+
+        Patched.prototype = Orig.prototype;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Done in b27b3a32. Both fixes applied:

1. `Object.defineProperty(Orig.prototype, 'constructor', { value: Patched, ... })` (try/catch'd) so `(new Error()).constructor === Error` now holds.
2. `Object.setPrototypeOf(Patched, Orig)` so Patched inherits all of Orig's statics; the manual `Patched.captureStackTrace = Orig.captureStackTrace` line is removed since the prototype chain now covers it.

Verified the patched JS in a Node sim of Chakra-style Error (class extends Old that discards options): `.constructor` identity, prototype-chain `instanceof`, and `Reflect.construct` new.target preservation all hold.

Polyfills/ErrorCause/Source/ErrorCause.cpp
```diff
@@ -0,0 +1,132 @@
+#include <Babylon/Polyfills/ErrorCause.h>
```
bghgary: Drop (sorry — flipping my earlier ask). Either run Playground validation against an ES-compliant engine (`Win32_x64_V8_D3D11` exists in CI), or bundle `babel-standalone` in the Playground to transpile snippets at load if we want to keep Chakra coverage. Chakra consumers use Babel themselves. If error surfacing through Node-API is degraded on Chakra, fix in JsRuntimeHost.

Polyfills/StringTrim/Source/StringTrim.cpp
```diff
@@ -0,0 +1,48 @@
+#include <Babylon/Polyfills/StringTrim.h>
```
bghgary: Drop (sorry — same as ErrorCause). Switching Playground validation off Chakra makes this unnecessary.

bghgary (changes requested):

Apologies — my earlier review asked you to move these into `Polyfills/` C++ modules; on reflection that was the wrong direction. They shouldn't exist at all (concerns inline). `babylonjs-addons` bundling is fine.

Polyfills/ErrorCause/Source/ErrorCause.cpp
bkaradzic-microsoft: Done in 48e3ad5f. Dropped the ErrorCause and StringTrim polyfills entirely (rebased onto current master; this PR is now just the abylonjs-addons bundling). The Chakra ES2019/ES2022 parse/runtime gaps are deferred to the separate V8-validation / babel-standalone track.

Polyfills/StringTrim/Source/StringTrim.cpp
bkaradzic-microsoft: Done in 48e3ad5f. Dropped the ErrorCause and StringTrim polyfills entirely (rebased onto current master; this PR is now just the abylonjs-addons bundling). The Chakra ES2019/ES2022 parse/runtime gaps are deferred to the separate V8-validation / babel-standalone track.

bkaradzic-microsoft:
Rebased onto latest master (fe380a78). Per your final review, this PR is now reduced to just bundling `babylonjs-addons` — the ErrorCause/StringTrim polyfills were dropped entirely. Re-requesting review.

bghgary (approved):

PR #1701, opened 2026-05-16
"Embedding" (simple) API layer

ryantrem:
## Summary

Introduces a new **`Babylon::Embedding`** layer that gives host apps a small, opinionated `Runtime` + `View` API for embedding Babylon Native, and migrates every Playground host (Android, iOS, visionOS, macOS, UWP, Win32, X11) onto it. The goal is to make platform integration a job that takes tens — not hundreds — of lines of code, and to do that without leaking `AppRuntime`, `Graphics::Device`, plugin headers, polyfill plumbing, or threading rules into the host.

Two layers ship together:

1. **Core C++ API** (`Embedding/Include/Shared/Babylon/Embedding/{Runtime,View,RuntimeOptions,LogLevel}.h`) — cross-platform, single source of truth for the lifecycle, threading, suspend/resume, input forwarding, and XR window management.
2. **Per-platform language facades** that wrap the core API in idiomatic types for the host language:
   - **Android**: JNI shared library (`BabylonNativeEmbedding.cpp`) + a host-authored Java facade (`BabylonNative.java`)
   - **Apple**: `BNRuntime` / `BNView` Obj‑C classes (Swift-friendly via the bridging header)

### `Embedding` core (`Embedding/Source`, `Embedding/Include/Shared`)

Two value types with explicit ownership and a documented contract:

```cpp
Babylon::Embedding::RuntimeOptions options{};
options.enableDebugger = true;
options.log = [](LogLevel level, std::string_view msg) { /* … */ };

Babylon::Embedding::Runtime runtime{options};

runtime.LoadScript("app:///Scripts/babylon.max.js");
runtime.LoadScript("app:///Scripts/scene.js");

Babylon::Embedding::View view{runtime, nativeWindowHandle};
view.Resize(width, height, CoordinateUnits::Physical);

// driven from the host's draw callback:
view.RenderFrame();
```

What the layer takes care of for you:

- **Runtime construction**: spins up `AppRuntime`, `JsRuntime`, polyfills (Console, Window, XMLHttpRequest, WebSocket, URL, Canvas, TextDecoder, Performance, Scheduling), and non‑GPU plugins.
- **GPU deferral**: `Graphics::Device`, `NativeEngine`, `NativeInput`, `NativeCanvas`, `TestUtils`, and (optionally) `NativeXr` / `ShaderCache` are constructed lazily on the **first** `View::Resize`. Subsequent View attaches just rebind the Device.
- **Script queuing**: `LoadScript` / `Eval` / `RunOnJsThread` are safe to call before the first `View` attach; queued and flushed when the engine comes up.
- **Suspend / Resume**: reference-counted, closes the in-flight frame, pauses JS timers, optionally persists the shader cache.
- **Input**: `OnPointer*` / `OnMouse*` accept either `Physical` or `Logical` pixel units; the View handles DPR conversion.
- **XR window**: `Runtime::SetXrWindow` hands off a secondary surface for NativeXr.
- **Logging**: single `RuntimeOptions::log` sink for `console.{log,warn,error}`, `Babylon::DebugTrace`, and uncaught JS exceptions.

### Android facade

`Embedding/Android/` builds the JNI shared library `libBabylonNativeEmbedding.so` (replacing the old `BabylonNativeJNI.cpp`). It ships the **native side only** — each host authors its own Java declarations that match the JNI ABI. The Playground's facade (`Apps/Playground/Android/BabylonNative/…/com/babylonjs/embedding/BabylonNative.java`, replacing the old `Wrapper.java`) is a thin `static`-method mirror of the C++ API:

- **Process lifecycle**: `setContext`, `setCurrentActivity`, `pause`, `resume`, `requestPermissionsResult`
- **Runtime**: `runtimeCreate` (optionally with a nested `RuntimeOptions`), `runtimeDestroy`, `runtimeLoadScript`, `runtimeEval`, `runtimeSetXrSurface`, `runtimeIsXrActive`
- **View**: `viewAttach`, `viewDetach`, `viewRenderFrame`, `viewResize`, `viewPointerDown` / `viewPointerMove` / `viewPointerUp`

A separate `com.library.babylonnative.BabylonView` helper wires `SurfaceHolder.Callback` and touch events to those calls.

### Apple facade

`Embedding/Apple/` — Obj‑C wrappers (`BNRuntime`, `BNView`, `BNViewDelegate`) usable from Obj‑C and from Swift via the bridging header. Used by Playground iOS, macOS, and visionOS. `BNRuntimeNative.h` exposes the underlying `Babylon::Embedding::Runtime*` for callers that need the C++ API.

## Host app migrations

Every Playground host now runs on top of the Embedding layer:

| Host | Before | After |
|---|---|---|
| Android | `BabylonNativeJNI.cpp` + `Wrapper.java` | `BabylonNative` Java facade over the JNI layer |
| iOS | `LibNativeBridge.{h,mm}` + ad-hoc Swift glue | `BNRuntime` / `BNView`, ~75-line `ViewController.swift` |
| macOS | `ViewController.mm` doing its own AppContext plumbing | Same `BNRuntime` / `BNView` types as iOS |
| visionOS | `LibNativeBridge.{h,mm}` + custom App.swift threading | `BNRuntime` / `BNView` + bridging header |
| Win32 | `AppContext` + manual `UpdateSize` / mouse plumbing | `Runtime` + `View` with `WM_POINTER` → `View::On*` |
| UWP | C++/WinRT `AppContext` + manual swap-chain plumbing | `Runtime` + `View` from `IInspectable` |
| X11 | `AppContext` + X11 event handlers | `Runtime` + `View` with `View::OnMouse*` |

The shared `AppContext.{h,cpp}` is gone — each host now constructs a `Runtime` + `View` directly.

## Lifecycle contract

- **First View attach is the heavy step.** Hosts MUST call `Resize` at least once with the surface's pixel dimensions before the first `RenderFrame` — the Device is built on the first Resize.
- **One View per Runtime at a time.** `View` constructor throws `std::runtime_error` on double-attach.
- **Destroy Views before their Runtime.** `~View` clears the back-reference, so LIFO destruction works; `~Runtime` asserts (debug builds) if a View is still attached.
- **Frame thread**: `RenderFrame`, `Resize`, `Suspend`, `Resume`, `LoadScript`, `Eval`, `RunOnJsThread`, View construction/destruction.
- **Any thread**: `OnPointer*` / `OnMouse*` / `IsSuspended` / `IsXrActive` / `SetXrWindow`.

## File scope

```
Embedding/
  Include/Shared/Babylon/Embedding/{Runtime,View,RuntimeOptions,LogLevel}.h
  Source/{Runtime,View}.cpp + RuntimeImpl.h
  Android/ — JNI shared library (Java facade lives per-host; see Playground)
  Apple/   — BNRuntime / BNView / BNViewDelegate

Apps/Playground/
  Win32/ UWP/ X11/ macOS/ iOS/ visionOS/ Android/   (all migrated)
```

## Closed open items

- [x] AI-generated comment cleanup
- [x] Naming (renamed to "Embedding")
- [x] All Playground hosts migrated (Android, iOS, visionOS, macOS, UWP, Win32, X11)
- [x] `RunOnJsThread` now has an `afterScriptLoad` parameter
- [x] `Runtime` / `View` are regular movable value types
- [x] Removed the planning `.md` from the repo root
- [x] Real fix for DeviceImpl.cpp issue

---

## Follow-up fix: metal-cpp duplicate-symbol link failure on Apple (commit `1c7fcef6`)

After merging the recent `bgfx.cmake` bump (#1749) into this branch, the macOS Playground (and the other Apple hosts) began failing to link with ~1900 duplicate-symbol errors for metal-cpp's `MTL::`/`NS::`/`CA::Private::*` globals, e.g.:

```
duplicate symbol 'MTL::Private::Protocol::s_kMTLTensorBinding' in:
    .../Dependencies/Debug/libmetal-cpp.a(Metal.o)
    .../_deps/bgfx.cmake-build/.../libbgfx.a(renderer_mtl.o)
ld: 1949 duplicate symbols
```

### Root cause

The new bgfx now **embeds and compiles its own copy of metal-cpp** in `renderer_mtl.cpp` (`#define NS/MTL/CA_PRIVATE_IMPLEMENTATION` + `#include <metal-cpp/metal.hpp>`). Babylon Native *also* compiled its own metal-cpp implementation TU (in `Dependencies/CMakeLists.txt`), so the implementation globals existed in **both** `libbgfx.a` and `libmetal-cpp.a`.

Whether that collides depends purely on the **order** the two static archives appear on the link line:

- **`bgfx` before `metal-cpp`** → bgfx's copy resolves the references first, so metal-cpp's `Metal.o` member is never extracted → one copy → links cleanly. This is the order `master` happens to produce.
- **`metal-cpp` before `bgfx`** → both `Metal.o` and `renderer_mtl.o` get pulled in → two definitions → ~1900 duplicate symbols.

This branch isn't doing anything wrong — routing the Playground's libraries transitively through `Embedding` simply reorders the link line to *metal-cpp first*, which tripped a latent landmine the bgfx bump had already armed. (Verified empirically: taking the exact failing link command and only reordering the two archives flips it between a clean link and the 1900-symbol failure; `master`'s real link line lists `bgfx` before `metal-cpp`.)

### Fix

Make `metal-cpp` a **header-only `INTERFACE` library** and rely on bgfx's bundled implementation as the single source of the metal-cpp symbols. We keep generating the metal-cpp single header (BN code still `#include <Metal/Metal.hpp>`) but no longer compile a second implementation TU.

This is what `master` already does in effect (its binary uses bgfx's copy and never extracts BN's), now made explicit and **link-order-independent**. Since we own both repos, the header/impl version pairing is not a concern. `Graphics` — the only metal-cpp consumer — always links bgfx on Apple, so the symbols always resolve.

No changes were needed in `Install/Install.cmake`: `install_lib()` already skips `INTERFACE` targets and `install_include()` still installs the metal-cpp headers.

Verified locally on macOS (`BUILD SUCCEEDED`, 0 duplicate symbols, metal-cpp symbols resolved from bgfx); the Apple CI jobs cover iOS/visionOS.


Integrations/Apple/Source/BNView.mm
```diff
@@ -0,0 +1,295 @@
...
+        // call, which BNViewDelegate forwards from MTKView's
+        // `mtkView:drawableSizeWillChange:`. UIKit's view lifecycle
+        // reliably fires that callback during initial layout, but AppKit
+        // is more lazy: on macOS the callback typically does NOT fire
+        // for the initial drawable size if the delegate was installed
+        // after the size was already determined — only for subsequent
+        // changes. To keep both platforms working out of the box, we
+        // query the current drawable size below and kick off an explicit
+        // resize. If the drawable isn't sized yet (e.g. MTKView not yet
+        // in a window), the early-out below skips it and the delegate
```
bghgary: `Integrations::View(...)` throws `std::runtime_error`. `@catch (NSException*)` only catches Obj-C exceptions, so the "another View already attached" path won't trigger this and the C++ exception will propagate out of `initWithRuntime:view:`. Use `@catch (...)` or wrap `_view.emplace(...)` in a C++ try/catch (the JNI side does the latter).

Integrations/Source/Runtime.cpp
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+#if BABYLON_NATIVE_PLUGIN_NATIVEXR
+    void Runtime::SetXrWindow(void* nativeWindow)
+    {
+        std::lock_guard<std::mutex> lock{m_impl->m_xrMutex};
+        m_impl->m_xrWindow = nativeWindow;
+        if (m_impl->m_nativeXr)
+        {
+            m_impl->m_nativeXr->UpdateWindow(nativeWindow);
```
bghgary: This calls into NativeXr from whatever thread invoked `SetXrWindow`, which the header documents as "any thread". Confirm NativeXr's `UpdateWindow` is thread-safe — historically plugin entry points have been JS-thread-only. If not, dispatch through `m_appRuntime` instead of calling it directly under the mutex.

Integrations/Include/Shared/Babylon/Integrations/Runtime.h
```diff
@@ -0,0 +1,111 @@
...
+
+        ~Runtime();
+
+        // Non-copyable; movable. Cross-references between Runtime and View
+        // point at the heap-allocated pimpls, so moves of the outer wrappers
+        // are safe and don't invalidate any back-pointers.
+        Runtime(const Runtime&) = delete;
+        Runtime& operator=(const Runtime&) = delete;
+        Runtime(Runtime&&) noexcept;
+        Runtime& operator=(Runtime&&) noexcept;
```
bghgary, line 43: The "back-pointers stay valid" claim holds for move-construction but not move-assignment: `unique_ptr::operator=` destroys the destination's existing `RuntimeImpl` before adopting the source's pointer, so a `View` attached to the destination ends up with a dangling `m_runtime` reference (debug catches it via the `m_currentView == nullptr` assert in `~RuntimeImpl`; release silently UBs).

bghgary (commented):

Concerns inline.

Integrations/Include/Shared/Babylon/Integrations/Runtime.h
ryantrem: Now we throw instead of assert, so the ordering must be correct for destruction (including destruction from a move).

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -213,7 +214,9 @@ namespace Babylon::Graphics
             }
 
             m_state.Bgfx.Initialized = true;
-            m_state.Bgfx.Dirty = false;
+            // TODO: Should be able to clear the Dirty flag, but bgfx::init is not doing everything that UpdateBgfxState is doing (at least on Apple platforms).
+            // See https://github.com/BabylonJS/ThePirateCove/issues/1657
+            // m_state.Bgfx.Dirty = false;
```
ryantrem: We probably need to sort this out before merging.

Integrations/CMakeLists.txt
ryantrem, line 1: "Integrations" may not be the right name for this.

ryantrem: Switched to "Embedding" as agreed.

Embedding/Source/Runtime.cpp
```diff
@@ -0,0 +1,454 @@
...
+        // ScriptLoader's dispatcher captures &m_appRuntime, so ~ScriptLoader
+        // must run before ~AppRuntime.
+        m_scriptLoader.emplace(*m_appRuntime);
+    }
+
+    RuntimeImpl::~RuntimeImpl()
+    {
+        // Host owns the ordering: destroy Views before their Runtime.
+        assert(m_currentView == nullptr && "View must be destroyed before its Runtime.");
+
```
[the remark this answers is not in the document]
ryantrem: Already discussed offline and decided throwing from a destructor was an anti-pattern we should avoid. Leaving as is.

Embedding/Include/Shared/Babylon/Embedding/Runtime.h
```diff
@@ -0,0 +1,111 @@
...
+        explicit Runtime(RuntimeOptions options = {});
+
+        // // Future construction mode: adopt a host-owned Babylon::JsRuntime
+        // // (e.g. React Native with Hermes/JSC + CallInvoker). In Attach mode
+        // // `~Runtime` does NOT tear down the JS engine, and Suspend/Resume
+        // // only toggles Device rendering.
+        // static Runtime Adopt(Babylon::JsRuntime& jsRuntime,
+        //                      RuntimeOptions options = {});
+
+        ~Runtime();
```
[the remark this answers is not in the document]
ryantrem: By design (see previous comment).

Embedding/Source/Runtime.cpp
```diff
@@ -0,0 +1,454 @@
...
+        }
+    }
+#endif
+
+    Runtime::Runtime(RuntimeOptions options)
+        : m_impl{std::make_unique<RuntimeImpl>(std::move(options))}
+    {
+    }
+
+    Runtime::~Runtime() = default;
```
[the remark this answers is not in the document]
ryantrem: By design (see previous comment).

Embedding/Source/View.cpp
```diff
@@ -0,0 +1,303 @@
...
+            // Flag pre-init so hosts can diagnose "my draw callback fires
+            // but nothing renders". The externally-suspended case is
+            // expected and stays silent.
+            DEBUG_TRACE("Babylon::Embedding::View::RenderFrame skipped: View has not yet been initialized. Call View::Resize with the surface's pixel dimensions to begin rendering.");
+            return;
+        }
+        if (impl.m_suspendCount.load(std::memory_order_relaxed) > 0)
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
ryantrem: The `DEBUG_TRACE` here is in the **pre-init** branch (`!m_initialized`), not the externally-suspended branch. The two states are independent:

- `m_initialized` flips true on the first successful `Resize` (when the Device is built).
- `m_suspendCount` is driven by `Suspend()` / `Resume()`.

The "externally-suspended case is expected and stays silent" comment refers to the *next* branch — `m_suspendCount > 0` — which assumes the engine is already initialized. In the normal flow (attach → `Resize` → app backgrounds → host keeps driving the draw callback), `RenderFrame` skips the pre-init branch and falls into the suspend branch, which returns silently with no log. So that comment is accurate.

The pre-init `DEBUG_TRACE` is a deliberate diagnostic for "my draw callback fires but nothing renders" — i.e. the host forgot the initial `Resize`. A couple of reasons it isn't a real spam concern:

- `DEBUG_TRACE` compiles to nothing unless the build defines `BABYLON_DEBUG_TRACE`, and even then only emits after the host explicitly calls `EnableDebugTrace(true)`. So it only shows up in a trace-enabled diagnostic build — exactly when you'd want it.
- The pre-init window is normally transient (a frame or two before the first layout/`Resize`).

The only edge case is suspending *before* the first `Resize` (uninitialized **and** suspended), where the pre-init branch logs because the init check comes first — but that's the uninitialized path, not the externally-suspended one, so it doesn't contradict the comment. Leaving as-is.

Embedding/Include/Shared/Babylon/Embedding/RuntimeOptions.h
```diff
@@ -0,0 +1,58 @@
...
+        std::function<void(LogLevel, std::string_view)> log;
+
+#if BABYLON_NATIVE_PLUGIN_SHADERCACHE
+        // Optional path for persisting the GPU shader cache across sessions.
+        // If non-empty:
+        //   - Loaded synchronously during the first View attach (missing
+        //     or unreadable file: ignored).
+        //   - Saved asynchronously during `Runtime::Suspend` (queued onto
+        //     the JS thread before the suspension blocker).
+        //   - Saved synchronously during `~Runtime`.
```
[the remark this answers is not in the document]
ryantrem: Good catch — this one's accurate. `Runtime::Suspend` calls `SaveShaderCache()` synchronously on the frame/host thread (right after `ViewImpl::Suspend` closes the in-flight frame, when the engine is quiescent), not queued onto the JS thread. Fixed the comment to match:

```
//   - Saved synchronously during `Runtime::Suspend` (after the
//     in-flight frame is closed, when the engine is quiescent).
```

Embedding/Android/src/main/cpp/BabylonNativeEmbedding.cpp
```diff
@@ -0,0 +1,538 @@
...
+    // know the address before constructing the wrapper).
+    //
+    // Runtime::Suspend/Resume are refcounted, so this composes safely
+    // with explicit host-side runtimeSuspend / runtimeResume calls.
+    struct AndroidRuntime
+    {
+        std::unique_ptr<Runtime> runtime;
+        android::global::AppStateChangedCallbackTicket pauseTicket;
+        android::global::AppStateChangedCallbackTicket resumeTicket;
+    };
```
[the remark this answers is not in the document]
ryantrem: Added an `ANativeWindow* xrWindow{nullptr}` member to `AndroidRuntime` (under `#if BABYLON_NATIVE_PLUGIN_NATIVEXR`) so the JNI layer can deterministically own and release the XR window reference. It's needed because NativeXr only stores the raw pointer (it doesn't acquire its own reference and consumes it lazily at session creation), so the reference acquired via `ANativeWindow_fromSurface` must be owned here.

Embedding/Android/src/main/cpp/BabylonNativeEmbedding.cpp
```diff
@@ -0,0 +1,538 @@
...
+    return MakeRuntimeHandle(std::move(options));
+}
+
+JNIEXPORT void JNICALL
+Java_com_babylonjs_embedding_BabylonNative_runtimeDestroy(JNIEnv*, jclass, jlong handle)
+{
+    // Reverse declaration order: tickets unsubscribe before the Runtime
+    // is destroyed, so no callback fires on a dead Runtime.
+    delete AsAndroidRuntime(handle);
+}
```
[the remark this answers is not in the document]
ryantrem: Fixed. `runtimeDestroy` now captures `xrWindow` before deleting the wrapper, then calls `ANativeWindow_release` on it *after* the `delete`. The release has to happen after the Runtime (and thus NativeXr, which holds the raw pointer) is torn down, otherwise we'd be releasing a window NativeXr might still reference during destruction.

Embedding/Android/src/main/cpp/BabylonNativeEmbedding.cpp
```diff
@@ -0,0 +1,538 @@
...
+    }
+    AsRuntime(handle)->SetXrWindow(window);
+#else
+    (void)handle;
+    (void)surface;
+    ThrowPluginNotEnabled(env,
+        "runtimeSetXrSurface was called but BABYLON_NATIVE_PLUGIN_NATIVEXR "
+        "was not enabled at native build time.");
+#endif
+}
```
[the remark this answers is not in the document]
ryantrem: Fixed. `runtimeSetXrSurface` now acquires the new window, hands it to `SetXrWindow`, then releases the previously-held `xrWindow` and stores the new one (which may be `nullptr` when clearing). So swapping or clearing the XR surface no longer leaks the prior native window reference. The release is done after `SetXrWindow` so NativeXr is already pointed at the new window before the old one is dropped.

Embedding/Include/Shared/Babylon/Embedding/Runtime.h
```diff
@@ -0,0 +1,111 @@
...
+
+        // Non-copyable; movable. Cross-references between Runtime and View
+        // point at the heap-allocated pimpls, so moves of the outer wrappers
+        // are safe and don't invalidate any back-pointers.
```
bghgary: Nit: this overstates move safety. True for move-construction always, but move-assignment invalidates the *destination's* View back-pointer when the destination still has a View attached (it destroys the destination's pimpl, which `~RuntimeImpl` asserts on). Suggest qualifying:

```suggestion
        // Non-copyable; movable. Cross-references point at the heap-allocated
        // pimpls, so move-construction is always safe and never invalidates a
        // back-pointer. Move-assignment is safe only when the destination has
        // no View attached -- it destroys the destination's pimpl, which
        // ~RuntimeImpl guards with assert(m_currentView == nullptr).
```

Integrations/Include/Shared/Babylon/Integrations/Runtime.h
bghgary: [Responded by Copilot on behalf of @bghgary]

Minor: the code asserts rather than throws — `~RuntimeImpl` does `assert(m_currentView == nullptr)` (debug-only). The reasoning is otherwise right: a bad move-assignment destroys the destination's pimpl and trips that same assert, so move stays safe whenever the destination has no View. Just flagging the "throw" wording. (The header comment on the move ops overstates this too — suggested a tweak inline.)

Apps/Playground/Android/BabylonNative/src/main/java/com/babylonjs/embedding/BabylonNative.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public static native void requestPermissionsResult(
+            int requestCode, String[] permissions, int[] grantResults);
+
+    // -------------------------------------------------------------------
+    // Runtime
+    // -------------------------------------------------------------------
+
+    /** Returns an opaque handle; release with {@link #runtimeDestroy(long)}. */
+    public static native long runtimeCreate();
```
bghgary, line 81: [Reviewed by Copilot on behalf of @bghgary]

Idiomatic-Java gap (Copilot's read — worth validating with someone fluent in Android): this exposes a C-style `long`-handle API rather than a Java object model. Canonical Java would wrap the handles in `Runtime`/`View` classes implementing `AutoCloseable` (try-with-resources) and/or registering a `java.lang.ref.Cleaner`, so native resources can't leak if the host forgets `runtimeDestroy`/`viewDetach`. As-is, the LIFO teardown (View before Runtime) is manual and unguarded — and because JVM finalization is non-deterministic, this is the platform where "View must outlive Runtime" is most likely to be violated in practice (and there's no debug-assert here, unlike the C++ side).

Also worth noting the asymmetry: the Apple facade ships real objects (`BNRuntime`/`BNView`, with ARC managing lifetime), but Android ships raw JNI plus this C-style example (which lives in the Playground, not the shipped `Embedding/Android` layer). Either ship an idiomatic Java wrapper or make the asymmetry an explicit, documented choice.

Minor: `RuntimeOptions` public mutable fields → a Builder or constructor is more idiomatic (Java has no object-initializers); `setCurrentActivity(Object)` → type it as `Activity`.

Embedding/Include/Platform/Apple/Babylon/Embedding/Apple/BNRuntime.h
```diff
@@ -0,0 +1,89 @@
...
+- (instancetype)init;
+
+/// Constructs the runtime with platform-friendly options (nil = same
+/// defaults as `init`).
+///
+/// `options.shaderCachePath`: loaded on first BNView attach; saved on
+/// `suspend` and on deallocation. Raises
+/// `BabylonNativePluginNotEnabledException` when non-nil but the
+/// native library was built without `BABYLON_NATIVE_PLUGIN_SHADERCACHE`.
+- (instancetype)initWithOptions:(nullable BNRuntimeOptions*)options NS_DESIGNATED_INITIALIZER;
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

Idiomatic Obj-C/Swift note (Copilot's read — worth validating with someone fluent in Apple platforms): the two main types use inconsistent failure models. `BNView`'s init returns `nil` on failure, but `BNRuntime`'s `initWithOptions:` raises an `NSException` (e.g. for a missing shader-cache plugin) — and `NSException` can't be caught from Swift `try/catch`, so a Swift consumer has no language-level way to handle it. Canonical Obj-C picks one model: failable init (`nil`) or `NSError**`. Worth aligning both types on the same one.

Embedding/Include/Platform/Apple/Babylon/Embedding/Apple/BNView.h
```diff
@@ -0,0 +1,132 @@
...
+    NS_SWIFT_NAME(mouseMove(x:y:));
+
+/// Forward a scroll-wheel event. `axis` is `+mouseWheelY`. `delta` is
+/// the signed scroll amount; AppKit hosts pass `-NSEvent.deltaY` so
+/// scroll-up matches Babylon's negative convention.
+- (void)mouseWheel:(NSInteger)axis delta:(NSInteger)delta
+    NS_SWIFT_NAME(mouseWheel(axis:delta:));
+
+/// Button identifiers accepted by `mouseDown:` and `mouseUp:`.
+@property (class, nonatomic, readonly) NSInteger leftMouseButton;
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

Idiomatic nit (Copilot's read — worth validating): the mouse-button and wheel-axis identifiers are loose `NSInteger` class properties. `NS_ENUM` / `NS_TYPED_ENUM` would be type-safe and read better from Swift (`BNMouseButton.left` vs `BNView.leftMouseButton` as a bare `Int`). Minor: `mouseWheel:delta:` takes an `NSInteger` delta where AppKit scroll deltas are fractional `CGFloat`.

Core/Graphics/Source/DeviceImpl.h
```diff
@@ -118,7 +118,12 @@ namespace Babylon::Graphics
         static const bgfx::RendererType::Enum s_bgfxRenderType;
         static void ConfigureBgfxPlatformData(bgfx::PlatformData& pd, WindowT window);
         static void ConfigureBgfxRenderType(bgfx::PlatformData& pd, bgfx::RendererType::Enum& renderType);
-        static float GetDevicePixelRatio(WindowT window);
+
+        // Push the render resolution onto the native rendering surface so it
+        // matches what bgfx renders into. Implemented per graphics API. The
+        // window may be default-constructed (null) before UpdateWindow has run
+        // (e.g. during construction), in which case there's nothing to size.
+        static void UpdateWindowSize(WindowT window, uint32_t width, uint32_t height);
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

Naming nit: `UpdateWindowSize` collides confusingly with `Device::UpdateWindow` and `Device::UpdateSize` but means neither — it syncs the native render surface's drawable to the bgfx render resolution (real work on Metal; no-op on D3D/GL). Something like `SyncSurfaceDrawableSize` or `ResizeRenderSurface` would disambiguate it from the `Update{Window,Size}` family.

bghgary (commented):

Comments inline.

Note: the Obj-C/Swift and Java idiom observations are Copilot's assessment — I'm relying on them rather than first-hand Apple/Android expertise, so worth a sanity check from someone fluent in those platforms.

bghgary (approved):

Apps/Playground/Android/BabylonNative/src/main/java/com/babylonjs/embedding/BabylonNative.java
ryantrem: This is all part of the Java code, which is effectively a "sample" layer. We should keep the sample thin. If we ever have a formal Java layer (outside the scope of CMakeLists.txt, like maybe some future package manager based solution or something), then we can have a thicker Java layer that is more idiomatic. I did switch Object to Activity though at least.

Issue #171, opened 2020-03-07
Linux Support


Schr3da:
Hey, 

first of all thanks for the great and excellent work! I like a lot how babylonjs constantly improved over time and I know this is not a simple task. However please add Linux support as well to the list. I know on the market Linux usage is still quite low and you even can port games using the webgl version with a container e.g. electron but native performance is much better :D

Thanks 
Schreda

CedricGuillemet:
I started taking a look at a Linux port and ran into 1 big issue with the JS engine.
No recent V8 engine is available as a prebuilt package (most recent is from 2016).
Binaries for Chakracore are available but Chakracore napi is using specific APIs and is not compatible with MS GitHub release available here : https://github.com/microsoft/ChakraCore/releases
@bghgary does it seem doable to you to port Chakracore napi to use those packages?

bghgary:
Chakra is not the same as ChakraCore. The version we have is one that I modified from the ChakraCore version. If we bring in the ChakraCore version, it should work. You can find a [link](https://github.com/nodejs/node-chakracore/blob/master/src/node_api_jsrt.cc) the the main README.

CedricGuillemet:
@bghgary I'm testing with JSC now.

CedricGuillemet:
A WIP draft PR : https://github.com/BabylonJS/BabylonNative/pull/185

CedricGuillemet:
Linux playground and validation tests compile with gcc and clang. The missing bit is this issue for OpenGL shader compilation. Closing.

PR #1708, opened 2026-05-18
Playground: link AbortController + TextEncoder polyfills from JsRuntimeHost

bkaradzic-microsoft:
Brings the playground closer to running unmodified Babylon.js content by:

- Initializing the existing JsRuntimeHost `AbortController` polyfill in `AppContext` and linking it into the Win32 + Android Playground binaries (the Android JNI link was missing it).
- Extending the `validation_native.js` `document` shim with `createEvent` and `dispatchEvent` so playgrounds that synthesize DOM events (e.g. mesh / scene serialization round-trips) stop tripping `document.dispatchEvent is not a function`.
- Re-enabling one playground (serialization round-trip) that was previously quarantined for that error.

## Reviewer history

This PR originally also introduced native `TextEncoder` and `PointerEvent` polyfills under `Polyfills/`. Per @bghgary's review:

- **TextEncoder** belongs in **JsRuntimeHost** alongside `TextDecoder` (both are part of the WHATWG Encoding Standard). Splitting them across repos means non-BN JsRuntimeHost consumers can't get `TextEncoder` without duplicating the implementation. Moved to BabylonJS/JsRuntimeHost#171. Once that lands the BN-side wiring is a one-line `Babylon::Polyfills::TextEncoder::Initialize(env)` follow-up in `AppContext.cpp`.
- **PointerEvent** dropped from this PR. BabylonNative isn't a browser; `DeviceInputSystem` already abstracts input. Holding for an offline discussion before re-introducing anything.

Branch has been rebased onto current `master` and the now-removed `Polyfills/TextEncoder/` + `Polyfills/PointerEvent/` trees plus all their CMake / link / init wiring is gone in the squashed commit 2a74a22e.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -2258,8 +2258,6 @@
         {
             "title": "Serialize scene without materials",
             "playgroundId": "#PH4DEZ#1",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Pixel comparison fails (more than 20% pixels differ)",
             "referenceImage": "serializeWithoutMaterials.png"
         },
         {
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -2257,7 +2257,7 @@
         },
         {
             "title": "Serialize scene without materials",
-            "playgroundId": "#PH4DEZ#1",
+            "playgroundId": "#PH4DEZ#4",
             "referenceImage": "serializeWithoutMaterials.png"
         },
         {
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -525,9 +525,52 @@
             if (type === "canvas") {
                 return new OffscreenCanvas(64, 64);
             }
-            return {};
+            // Generic element stub with a no-op dispatchEvent so serializer
+            // tests that simulate a click via createEvent/dispatchEvent run
+            // without throwing.
+            return {
+                style: {},
+                addEventListener: function () { },
+                removeEventListener: function () { },
+                dispatchEvent: function () { return true; },
+                appendChild: function (c) { return c; },
+                removeChild: function (c) { return c; },
+                setAttribute: function () { },
+                getAttribute: function () { return null; },
+                click: function () { },
+            };
         },
-        removeEventListener: function () { }
+        createEvent: function (type) {
+            var ev = {
+                type: '',
+                bubbles: false,
+                cancelable: false,
+                defaultPrevented: false,
+                target: null,
+                currentTarget: null,
+                timeStamp: Date.now(),
+                detail: null,
+                preventDefault: function () { ev.defaultPrevented = true; },
+                stopPropagation: function () { },
+                stopImmediatePropagation: function () { },
+            };
+            ev.initEvent = function (t, bubbles, cancelable) {
+                ev.type = String(t || '');
+                ev.bubbles = !!bubbles;
+                ev.cancelable = !!cancelable;
+            };
+            ev.initCustomEvent = function (t, bubbles, cancelable, detail) {
+                ev.initEvent(t, bubbles, cancelable);
+                ev.detail = detail;
+            };
+            ev.initUIEvent = ev.initEvent;
+            ev.initMouseEvent = ev.initEvent;
+            return ev;
+        },
+        addEventListener: function () { },
+        removeEventListener: function () { },
+        dispatchEvent: function () { return true; },
+        body: { appendChild: function (c) { return c; }, removeChild: function (c) { return c; } },
     }
 
     const xhr = new XMLHttpRequest();
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -525,52 +525,9 @@
             if (type === "canvas") {
                 return new OffscreenCanvas(64, 64);
             }
-            // Generic element stub with a no-op dispatchEvent so serializer
-            // tests that simulate a click via createEvent/dispatchEvent run
-            // without throwing.
-            return {
-                style: {},
-                addEventListener: function () { },
-                removeEventListener: function () { },
-                dispatchEvent: function () { return true; },
-                appendChild: function (c) { return c; },
-                removeChild: function (c) { return c; },
-                setAttribute: function () { },
-                getAttribute: function () { return null; },
-                click: function () { },
-            };
+            return {};
         },
-        createEvent: function (type) {
-            var ev = {
-                type: '',
-                bubbles: false,
-                cancelable: false,
-                defaultPrevented: false,
-                target: null,
-                currentTarget: null,
-                timeStamp: Date.now(),
-                detail: null,
-                preventDefault: function () { ev.defaultPrevented = true; },
-                stopPropagation: function () { },
-                stopImmediatePropagation: function () { },
-            };
-            ev.initEvent = function (t, bubbles, cancelable) {
-                ev.type = String(t || '');
-                ev.bubbles = !!bubbles;
-                ev.cancelable = !!cancelable;
-            };
-            ev.initCustomEvent = function (t, bubbles, cancelable, detail) {
-                ev.initEvent(t, bubbles, cancelable);
-                ev.detail = detail;
-            };
-            ev.initUIEvent = ev.initEvent;
-            ev.initMouseEvent = ev.initEvent;
-            return ev;
-        },
-        addEventListener: function () { },
-        removeEventListener: function () { },
-        dispatchEvent: function () { return true; },
-        body: { appendChild: function (c) { return c; }, removeChild: function (c) { return c; } },
+        removeEventListener: function () { }
     }
 
     const xhr = new XMLHttpRequest();
--- a/Apps/Playground/Android/BabylonNative/CMakeLists.txt
+++ b/Apps/Playground/Android/BabylonNative/CMakeLists.txt
@@ -25,6 +25,7 @@ target_link_libraries(BabylonNativeJNI
     PRIVATE EGL
     PRIVATE log
     PRIVATE -lz
+    PRIVATE AbortController
     PRIVATE AndroidExtensions
     PRIVATE AppRuntime
     PRIVATE Blob
--- a/Apps/Playground/Android/BabylonNative/CMakeLists.txt
+++ b/Apps/Playground/Android/BabylonNative/CMakeLists.txt
@@ -46,5 +46,6 @@ target_link_libraries(BabylonNativeJNI
     PRIVATE ShaderCache
     PRIVATE TestUtils
     PRIVATE TextDecoder
+    PRIVATE TextEncoder
     PRIVATE Window
     PRIVATE XMLHttpRequest)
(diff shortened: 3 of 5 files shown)


Apps/Playground/Scripts/dom_polyfill.js
```diff
@@ -0,0 +1,114 @@
...
+                    bytes.push(0x80 | (code & 0x3F));
+                }
+            }
+            return new Uint8Array(bytes);
+        };
+        TextEncoder.prototype.encodeInto = function (input, dest) {
+            var arr = this.encode(input);
+            var n = Math.min(arr.length, dest.length);
+            for (var i = 0; i < n; i++) dest[i] = arr[i];
+            return { read: String(input).length, written: n };
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in b79d9db7. `encodeInto` now normalizes `input` (treats `undefined` as the empty string) before computing both the encoded bytes and `read`, so the returned read count is consistent with what `encode()` writes.

Apps/Playground/Scripts/dom_polyfill.js
```diff
@@ -0,0 +1,114 @@
...
+        };
+        g.TextEncoder = TextEncoder;
+    }
+
+    if (typeof g.PointerEvent === 'undefined') {
+        function PointerEvent(type, init) {
+            init = init || {};
+            this.type = String(type || '');
+            this.bubbles = !!init.bubbles;
+            this.cancelable = init.cancelable !== false;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in b79d9db7. `cancelable` now defaults to `false` to match the DOM `EventInit` spec default; callers that need a cancelable event must pass `cancelable: true` explicitly.

Apps/Playground/Scripts/dom_polyfill.js
```diff
@@ -0,0 +1,114 @@
...
+//
+// `document` is shimmed by validation_native.js once it has loaded the
+// config, since the runner needs to inject test-specific behaviour into
+// createElement.
+
+(function () {
+    'use strict';
+
+    // Chakra has no `globalThis`; use the Function-constructor trick.
+    var g = (new Function('return this'))();
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in b79d9db7. Replaced `new Function('return this')` with a non-eval resolution chain (`globalThis` -> `self` -> `window` -> `global`) and kept the `Function` constructor only as a last-resort fallback for runtimes that ship none of those globals.

ryantrem (commented):
I guess this makes the tests work, but if a consumer tried to use the same things in their own app, they wouldn't work. Is that really the state we want to be in, or we want to know what things don't work because they aren't polyfilled? I would have thought we would either document these features don't work because we don't polyfill, or we add native polyfills for them.

Polyfills/TextEncoder/Include/Babylon/Polyfills/TextEncoder.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
```
bghgary: TextEncoder belongs in JsRuntimeHost alongside TextDecoder (also WHATWG Encoding Standard). Splitting them across repos means non-BN consumers can't get TextEncoder without duplicating.

Polyfills/PointerEvent/Include/Babylon/Polyfills/PointerEvent.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
bghgary: We shouldn't polyfill DOM types. BabylonNative isn't a browser; DeviceInputSystem already abstracts input. Let's discuss offline before adding this.

bghgary (changes requested):

TextEncoder belongs in JsRuntimeHost. No DOM polyfills.

Polyfills/TextEncoder/Include/Babylon/Polyfills/TextEncoder.h
bkaradzic-microsoft: Agreed — moved the TextEncoder polyfill into JsRuntimeHost where it sits alongside TextDecoder (both are part of the WHATWG Encoding Standard). Opened BabylonJS/JsRuntimeHost#171 for the migration. Once that lands, the BN-side wiring is a one-line `Babylon::Polyfills::TextEncoder::Initialize(env)` follow-up in `AppContext.cpp`. This PR has been rebased onto master and the `Polyfills/TextEncoder/` directory plus all the related CMake/link/init wiring is gone in 2a74a22e.

Polyfills/PointerEvent/Include/Babylon/Polyfills/PointerEvent.h
bkaradzic-microsoft: Dropped from this PR — `Polyfills/PointerEvent/` and all related wiring removed in 2a74a22e. Happy to discuss offline before we re-introduce anything; agree that `DeviceInputSystem` already covers the input surface and that BabylonNative shouldn't polyfill DOM types reflexively. The one playground that wanted `PointerEvent` (in `config.json`) stays disabled for now.

bkaradzic-microsoft:
Addressed the requested changes in 2a74a22e (force-push after a clean rebase onto master):

**TextEncoder — moved to JsRuntimeHost**
Opened BabylonJS/JsRuntimeHost#171 to add `Polyfills/TextEncoder/` next to the existing `TextDecoder` polyfill (same CMakeLists shape, README, `JSRUNTIMEHOST_POLYFILL_TEXTENCODER` option, unit tests). Once it merges I'll send a tiny BN follow-up that just calls `Babylon::Polyfills::TextEncoder::Initialize(env)` in `AppContext`.

**PointerEvent — dropped**
`Polyfills/PointerEvent/` and all related CMake / link / init wiring removed from this PR. Happy to discuss offline before reintroducing anything in this space — agree that `DeviceInputSystem` is the right primary surface.

**What's left in #1708 (5 files, +50/-4):**
- `AppContext.cpp` — initialize the existing JsRuntimeHost `AbortController` polyfill alongside the other polyfills.
- `Apps/Playground/CMakeLists.txt` + Android JNI `CMakeLists.txt` — link `AbortController` into Playground (the Android one was previously missing).
- `validation_native.js` — extend the `document` shim with `createEvent`/`dispatchEvent` so serialization round-trip playgrounds stop hitting `document.dispatchEvent is not a function`.
- `config.json` — re-enable the one serialization playground that the shim fixes.

bkaradzic-microsoft:
JsRuntimeHost #171 is merged (commit `81b01d2e`). Pushed `8ee3aa60` to bump the BN pin to that commit and re-add the `TextEncoder` link + `Babylon::Polyfills::TextEncoder::Initialize(env)` call in `AppContext.cpp`, sourced from upstream this time. `Playground.exe` builds and `--list` exits 0 with the new polyfill initialized.

Note: per @ryantrem's review on JsRuntimeHost #171, `encodeInto` was dropped from the polyfill (Babylon.js doesn't call it). The exposed surface is just `encoding` + `encode()`.

Diff for `8ee3aa60` is +6/-1 across 4 files: pin in root `CMakeLists.txt`, include + Init call in `AppContext.cpp`, `PRIVATE TextEncoder` in Win32 + Android JNI `CMakeLists.txt`.

`PointerEvent` is still entirely out of this PR (offline discussion item).

Apps/Playground/Scripts/validation_native.js
```diff
@@ -525,9 +525,52 @@
...
+                removeEventListener: function () { },
+                dispatchEvent: function () { return true; },
+                appendChild: function (c) { return c; },
+                removeChild: function (c) { return c; },
+                setAttribute: function () { },
+                getAttribute: function () { return null; },
+                click: function () { },
+            };
+        },
+        createEvent: function (type) {
```
bghgary: BN doesn't support the DOM by design — there's no real `document`; the harness defines this one (no Document polyfill is linked, unlike AbortController/Blob/Window/etc.). This expands that fake `document` with `createEvent`/`initEvent` and element-stub `dispatchEvent`/`addEventListener` (all no-ops). The added document-level `dispatchEvent`/`addEventListener`/`body` aren't even exercised by the re-enabled playground — they widen the fake surface further. Faking DOM APIs so DOM-dependent content can run is the direction BN avoids.

Apps/Playground/Scripts/config.json
```diff
@@ -2258,8 +2258,6 @@
         {
             "title": "Serialize scene without materials",
             "playgroundId": "#PH4DEZ#1",
-            "excludeFromAutomaticTesting": true,
```
bghgary, line 2261: `#PH4DEZ#1`'s only DOM use is a file download (`createElement('a')` + `createEvent` + `link.dispatchEvent`) — the sole reason this needs a fake `document`. It serves no purpose in a visual test, and the same entry is in BJS's own visual suite (`packages/tools/tests/test/visualization/config.json`), so this is shared content; faking DOM in BN only patches the symptom.

Fix it in the playground (helps web + native): make it a real in-memory round-trip — serialize → reload → render the deserialized scene (`SceneLoader.ImportMeshAsync(null, "", "data:" + json, scene)`), like the glTF serializer `#KU72PX` (`GLBAsync` → `LoadSceneAsync`). That actually tests serialization correctness (not just that it doesn't throw) and needs no DOM; the download can stay guarded behind a `document` check for browser debug. Render-only wouldn't test serialization at all.

Then BN drops the `validation_native.js` DOM shim and just repoints `config.json` to the new revision (BJS repoints too).

bghgary (dismissed):

Concerns inline.

Apps/Playground/Scripts/config.json
bkaradzic-microsoft: Done in 82e555a3. Rewrote `#PH4DEZ` as a real in-memory round-trip (serialize -> `SceneLoader.ImportMeshAsync(null, "", "data:" + json, scene)` -> render the deserialized scene), modeled on `#KU72PX`, so it validates serialization correctness with no DOM dependency. The browser download is kept but guarded on `document.createEvent` (present in a real browser, absent from BN's fake `document`), so it's skipped natively.

Created revision `#PH4DEZ#4` and repointed `config.json` `#PH4DEZ#1` -> `#PH4DEZ#4`. Verified locally on Win32 D3D11: the test renders the deserialized scene and passes pixel comparison (248 px diff, well within tolerance). I'll open the matching Babylon.js visual-suite repoint to `#PH4DEZ#4`.

Apps/Playground/Scripts/validation_native.js
bkaradzic-microsoft: Done in 82e555a3. Reverted the `document`-shim expansion (`createEvent` / `initEvent` / element-stub `dispatchEvent`/`addEventListener` + document-level `dispatchEvent`/`addEventListener`/`body`) back to the original minimal shim. With `#PH4DEZ` now doing the serialization round-trip in-memory, no DOM faking is needed.

bkaradzic-microsoft:
Opened the matching Babylon.js visual-suite repoint: BabylonJS/Babylon.js#18569 (`#PH4DEZ#1` -> `#PH4DEZ#4`). The rendered scene is unchanged, so the shared `serializeWithoutMaterials.png` reference still matches.

CedricGuillemet (approved):

PR #1707, opened 2026-05-18
Use native fetch() polyfill from JsRuntimeHost in Playground

bkaradzic-microsoft:
Wires Babylon Native's Playground up to the **native `fetch()` polyfill** that now lives in JsRuntimeHost (BabylonJS/JsRuntimeHost#188), so playgrounds depending on `fetch` stop hitting `ReferenceError: 'fetch' is not defined`.

This supersedes the original JS XMLHttpRequest-based shim approach: rather than shipping a `fetch_polyfill.js` script, `fetch` is now provided natively (built on the same `UrlLib` transport as `XMLHttpRequest`), matching how File/FileReader landed in #1706.

**Changes:**
- `Apps/Playground/Shared/AppContext.cpp` - calls `Babylon::Polyfills::Fetch::Initialize(env)` alongside the other polyfills (right after `XMLHttpRequest`).
- `Apps/Playground/CMakeLists.txt` + `Apps/Playground/Android/BabylonNative/CMakeLists.txt` - link the `Fetch` target.
- `CMakeLists.txt` - bumps the JsRuntimeHost pin to pick up the native polyfill.

> ✅ **Unblocked:** JsRuntimeHost#188 merged, and the pin now points at `BabylonJS/JsRuntimeHost@a407a331` (the #188 merge commit) — no longer the temporary fork branch. Rebased onto master (clean).

**Verified (Win32/Chakra):** `Playground` configures, builds and links cleanly against the native polyfill, and at runtime `typeof fetch === "function"` (confirmed via a headless smoke check).

---

## Landing context

This PR is one of the **splits** from the proven CI-green combined preview in draft PR #1702 (see [#1702](https://github.com/BabylonJS/BabylonNative/pull/1702) for the full intended end-state).

> The original split also included #1709 (ES2020+ -> ES2019 syntax-repair polyfill for Chakra), closed in favour of investigating `@babel/standalone` properly (#1711).

### Landing order / status

**Tier 1 - parallel-reviewable, no source conflicts:**
1. ~~#1703 - ExternalTexture C4702 build fix~~ ✅ merged
2. ~~#1704 - config.json `reason` rewrites (5 entries)~~ ✅ merged
3. ~~#1705 - config.json `reason` rewrites (17 entries)~~ ✅ merged

**Tier 2 - sequential, each touches `Apps/Playground/CMakeLists.txt` + `AppContext.cpp`:**

4. ~~#1706 - File/FileReader polyfill~~ ✅ merged (native JsRuntimeHost bump)
[...]

PR #1705, opened 2026-05-18
Document accurate root cause for 17 subtle pixel-diff tests

bkaradzic-microsoft:
Per-test PIL-composite triage of all 17 ``subtle pixel-diff`` tests. None of them are deterministic-cosmetic (no re-bakes possible - all show real visible regressions).

Updates the `reason` field in `Apps/Playground/Scripts/config.json` for 17 tests with accurate symptom descriptions, classifying into recurring root-cause clusters:

- 9 GUI green->red color regressions (idx 160, 174, 175, 196, 197, 370, 402, 566, 587)
- 4 OpenPBR analytic-lights right-column red rendering (580, 584, 587, 592)
- 1 instanced billboard foliage red (169)
- 1 LineEdgesRenderer extra red lines (179)
- 1 Background blur red splotches (602)
- 1 Clip planes GUI sliders red (182)
- 1 Instanced Bones edge-AA (256, borderline)

**No source changes, no test re-enables, no PNGs.** Metadata-only correction so the issue tracker reflects actionable root causes.
---

## Landing context

This PR is one of **7 splits** from the proven CI-green combined preview in **draft PR #1702** (see [#1702](https://github.com/BabylonJS/BabylonNative/pull/1702) for the full intended end-state and verified CI run [26044922430](https://github.com/BabylonJS/BabylonNative/actions/runs/26044922430)).

> Note: the original split included an 8th PR (#1709, ES2020+ -> ES2019 syntax-repair polyfill for Chakra). It was closed in favour of investigating `@babel/standalone` properly (#1711).

### Recommended landing order

**Tier 1 - parallel-reviewable, no source conflicts:**
1. #1703 - ExternalTexture C4702 build fix
2. #1704 - config.json `reason` rewrites (5 entries)
3. #1705 - config.json `reason` rewrites (17 entries)

**Tier 2 - sequential, each touches `Apps/Playground/CMakeLists.txt` SCRIPTS list + `Apps/Playground/Shared/AppContext.cpp` LoadScript order; rebase the next branch after the previous merges:**

4. #1706 - File/Blob/FileReader polyfill (largest test impact: 19 re-enables)
5. #1707 - fetch polyfill
6. #1708 - DOM globals + native AbortController + Android CMake link
7. #1710 - Cubemap auto-expand polyfill (loaded after babylon.max.js)

### Reference policy reminder

Reference PNGs across all 7 PRs come from Babylon.js; never re-baked by BN. Combined diff: **0 PNGs**.

PR #1702, opened 2026-05-18
[draft] Playground polyfills: File/FileReader, DOM, fetch, cubemap .env, Chakra ES2019

bkaradzic-microsoft:
Rebased onto latest `master`. Several pieces of the original combined branch have landed independently and were dropped as superseded; the genuine remaining playground-polyfill features stay.

### Kept (5 commits, all unique to this branch)
- **DOM polyfills** (`dom_polyfill.js`: TextEncoder, PointerEvent, AbortController, document.createEvent) + Android JNI link for the AbortController polyfill.
- **fetch() polyfill** (`fetch_polyfill.js`).
- **Cubemap `.env` fallback** (`cube_texture_polyfill.js`) + re-enable of 7 cubemap tests.
- **Chakra ES2020→ES2019 syntax-repair polyfill** (`es2019_transpile.js`).

All four polyfills are wired through `AppContext.cpp`'s script loader (transpiler first, then DOM/fetch; the cubemap patch loads after Babylon since it patches `NativeEngine.createCubeTexture`) and registered in `Apps/Playground/CMakeLists.txt`.

### Dropped (now in `master`)
- **File/FileReader polyfill + 19 GLTF/OBJ serializer re-enables + the 5 follow-up GLTF-Serializer-on-OpenGL `config.json` exclusion tweaks** — `master` landed the File/FileReader support in **#1706** as a **native JsRuntimeHost bump** (not a `file_polyfill.js` script), and that change already carries the identical test re-enables and the same OpenGL exclusion entries/reason strings this branch added. All five of this branch's File-related commits were therefore dropped as superseded.
- **OpenGL ExternalTexture stub** — `master` landed the reviewed fix in #1703 (`#pragma warning(disable:4702)` over the throwing stubs), superseding this branch's "inert default-value stubs" approach on the same file.
- **Post-SPIRV-Cross-bump pixel-diff triage** + the BN-rendered reference-image re-bakes — `master` documented the accurate root causes in #1704/#1705, and the two reference PNGs this branch re-baked were already reverted within the branch itself, so git dropped both as already-upstream.

`config.json` validates as JSON, identical structure to `master` apart from the cubemap re-enables and the ES2019/cubemap `reason`-string rewrites.

> Note: the **Chakra ES2019 syntax-repair polyfill** remains here as part of the full intended preview, but the standalone split for it (#1709) was **closed in favour of investigating `@babel/standalone` properly (#1711)**. Treat that piece as exploratory.

PR #1709, opened 2026-05-18
Add ES2020+ -> ES2019 syntax-repair polyfill for Chakra Playground

bkaradzic-microsoft:
Adds a lightweight regex-based ES2020+ -> ES2019 syntax-repair polyfill so Chakra-based BN runtimes can parse playgrounds that use modern syntax. V8/JSC runtimes already support these features natively; the polyfill self-detects and no-ops when not needed.

**Changes:**
- New `Apps/Playground/Scripts/es2019_transpile.js` - regex repair for: `?.` (optional chaining), `??` (nullish coalescing), numeric separators (incl. hex), `||=` `&&=` `??=` (logical assignment). Top-level `var __bnTranspileES2019` exposes the helper.
- `Apps/Playground/Scripts/validation_native.js` - `evalWithFallback` catches `SyntaxError` and retries once with repaired source.
- `Apps/Playground/CMakeLists.txt` - adds `es2019_transpile.js` to SCRIPTS list.
- `Apps/Playground/Shared/AppContext.cpp` - loads `es2019_transpile.js` before `fetch_polyfill.js`.
- `Apps/Playground/Scripts/config.json` - re-enables 1 test (idx 403 custom-handling-of-materials-for-render-target-pass); rewrites `reason` on 25 remaining-excluded tests with their post-polyfill state.

**Caveat:** Polyfill is **parse-only** - `?.` rewrites to `.` so null targets raise `TypeError` at runtime instead of short-circuiting. 25 tests remain excluded for: 9 runtime TypeErrors from `?.` -> `.` semantic divergence, 4 missing ADDONS module, 8 asset/url errors, 4 ES2022 private fields not handled by the polyfill.
---

## Landing context

This PR is one of **8 splits** from the proven CI-green combined preview in **draft PR #1702** (see [#1702](https://github.com/BabylonJS/BabylonNative/pull/1702) for the full intended end-state and verified CI run [26044922430](https://github.com/BabylonJS/BabylonNative/actions/runs/26044922430)).

### Recommended landing order

**Tier 1 - parallel-reviewable, no source conflicts:**
1. #1703 - ExternalTexture C4702 build fix
2. #1704 - config.json `reason` rewrites (3 entries)
3. #1705 - config.json `reason` rewrites (17 entries)

**Tier 2 - sequential, each touches `Apps/Playground/CMakeLists.txt` SCRIPTS list + `Apps/Playground/Shared/AppContext.cpp` LoadScript order; rebase the next branch after the previous merges:**

4. #1706 - File/Blob/FileReader polyfill (largest test impact: 19 re-enables)
5. #1707 - fetch polyfill
6. #1708 - DOM globals + native AbortController + Android CMake link
7. #1709 - ES2020+ -> ES2019 transpile polyfill
8. #1710 - Cubemap auto-expand polyfill (loaded after babylon.max.js)

### Reference policy reminder

Reference PNGs across all 8 PRs come from Babylon.js; never re-baked by BN. Combined diff: **0 PNGs**.

### DIFF
--- a/Apps/Playground/Scripts/es2019_transpile.js
+++ b/Apps/Playground/Scripts/es2019_transpile.js
@@ -1,0 +1,114 @@
+// es2019_transpile.js
+//
+// Lightweight regex-based ES2020+ -> ES2019 syntax repair for engines that
+// lack optional chaining (?.), nullish coalescing (??), and numeric
+// separators (1_000_000).
+//
+// This is a SYNTAX REPAIR, not a full transpile. It rewrites parse-time
+// failing tokens to the closest legal ES2019 expressions so that the host
+// engine accepts the code. Where native ES2020+ semantics differ from the
+// rewritten form, the rewritten form runs the "happy path" -- assumes the
+// target value is present. If the target is null at runtime the rewritten
+// code throws a TypeError where native ?. would have short-circuited to
+// undefined. This is an intentional trade-off: it produces valid syntax
+// (the prerequisite for the test running at all) while preserving the
+// common case behaviour. Tests that rely on the nullish-short-circuit
+// semantics may surface runtime errors that were previously hidden by
+// the parse failure -- a strict improvement for debuggability.
+//
+// Transforms applied:
+//   * 1_000_000   -> 1000000     (strip numeric separators)
+//   * a?.b        -> a.b         (optional chaining -> required chaining)
+//   * a?.[x]      -> a[x]
+//   * a?.()       -> a()
+//   * a ?? b      -> (a != null ? a : b)
+//
+// Not handled (still cause parse failures on Chakra):
+//   * Logical assignment ||= &&= ??=
+//   * Class private fields #name
+//   * BigInt literals 1n
+//
+// String/regex/comment literals containing ? . sequences are not skipped,
+// but in Babylon snippet code such occurrences are rare.
+//
+// On engines with native ES2020+ support (V8, JSC) this code is never
+// invoked because the caller only retries with this transform after
+// initial eval throws a SyntaxError.
+//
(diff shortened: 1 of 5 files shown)


Apps/Playground/Scripts/es2019_transpile.js
ryantrem, line 1: Why are we doing this with a custom script rather than just using babel or swc or something?

bkaradzic-microsoft: Fair point. Honest assessment of the tradeoff:

- The custom regex polyfill runs **inside** the Chakra host at parse-failure time, so a Node-side build tool (babel CLI / swc CLI) wasn't directly usable. The runtime-in-host constraint left us with `@babel/standalone` (~1 MB shipped + parsed by Chakra at startup), swc-wasm (Chakra doesn't support WASM), or this regex stopgap.
- The regex is semantically incomplete — `?.` rewrites to `.` so `null?.foo` throws `TypeError` instead of returning `undefined`. That's 9 of the 26 affected tests still excluded *because of* this divergence, plus 4 more for ES2022 private fields. So this polyfill unlocks only **1** test out of 26.
- The cost/benefit of "+114 lines for +1 test, with a semantic footgun baked in" is too thin to justify cementing the regex approach.

**Closing this PR**. Filed #1711 to investigate `@babel/standalone` properly (measure the actual size/startup cost, prototype, compare against alternatives, decide). If that lands, it would unlock the 9 `?.` tests + 4 ES2022 tests *correctly*, which is a much better story than what this PR achieves.

Thanks for pushing back.

bkaradzic-microsoft:
Closing per discussion above. Pursuing #1711 instead.

PR #1710, opened 2026-05-18
Add cubemap auto-expand polyfill for Playground (re-enables 7 PBR tests)

bkaradzic-microsoft:
Adds a JS polyfill that patches `BABYLON.NativeEngine.prototype.createCubeTexture` so single-URL `.dds`/`.ktx`/`.ktx2` cubemap loads (which BN does not currently support) transparently retry as `.env`. Babylon's CDN co-hosts both formats from the same source HDR, so the swap is invisible to the playground.

**Changes:**
- New `Apps/Playground/Scripts/cube_texture_polyfill.js` - patches `createCubeTexture`; retries via `self._loadFile`; falls back to the original throw on 404.
- `Apps/Playground/CMakeLists.txt` - adds `cube_texture_polyfill.js` to SCRIPTS list.
- `Apps/Playground/Shared/AppContext.cpp` - loads `cube_texture_polyfill.js` **after** `babylon.max.js` (the polyfill needs `BABYLON.NativeEngine` to exist).
- `Apps/Playground/Scripts/config.json` - re-enables 7 affected tests (idx 141, 172, 173, 246, 247, 248, 290 - PBR/cubemap heavy).

**Note:** Bridges the gap until `Plugins/NativeEngine` wires a proper loader registry path for cubemap formats. Remaining 35 of 42 affected tests either exceed pixel-diff threshold once they reach the renderer or hit unrelated secondary bugs (particles, PrePassRenderer caps, Chakra parse, etc.) - all already filed.
---

## Landing context

This PR is one of **7 splits** from the proven CI-green combined preview in **draft PR #1702** (see [#1702](https://github.com/BabylonJS/BabylonNative/pull/1702) for the full intended end-state and verified CI run [26044922430](https://github.com/BabylonJS/BabylonNative/actions/runs/26044922430)).

> Note: the original split included an 8th PR (#1709, ES2020+ -> ES2019 syntax-repair polyfill for Chakra). It was closed in favour of investigating `@babel/standalone` properly (#1711).

### Recommended landing order

**Tier 1 - parallel-reviewable, no source conflicts:**
1. #1703 - ExternalTexture C4702 build fix
2. #1704 - config.json `reason` rewrites (5 entries)
3. #1705 - config.json `reason` rewrites (17 entries)

**Tier 2 - sequential, each touches `Apps/Playground/CMakeLists.txt` SCRIPTS list + `Apps/Playground/Shared/AppContext.cpp` LoadScript order; rebase the next branch after the previous merges:**

4. #1706 - File/Blob/FileReader polyfill (largest test impact: 19 re-enables)
5. #1707 - fetch polyfill
6. #1708 - DOM globals + native AbortController + Android CMake link
7. #1710 - Cubemap auto-expand polyfill (loaded after babylon.max.js)

### Reference policy reminder

Reference PNGs across all 7 PRs come from Babylon.js; never re-baked by BN. Combined diff: **0 PNGs**.

### DIFF
--- a/Polyfills/CubeTexture/Source/CubeTexture.cpp
+++ b/Polyfills/CubeTexture/Source/CubeTexture.cpp
@@ -1,17 +1,22 @@
-// Cube texture fallback for BN's NativeEngine.
-//
-// BN's NativeEngine.createCubeTexture only natively handles .env single-file
-// cubemaps and 6-file face arrays. Snippets that load .dds (or .ktx) cubemaps
-// without an explicit forcedExtension or 6-file array throw "Cannot load
-// cubemap because 6 files were not defined".
-//
-// Babylon's standard environment cubemaps are hosted both as <name>.dds and
-// <name>.env at the same path on assets.babylonjs.com and
-// playground.babylonjs.com. This polyfill detects the failure pre-condition
-// and transparently retries with the .env URL. If the .env counterpart 404s,
-// it falls through to the original (which throws), preserving existing
-// behavior.
+#include <Babylon/Polyfills/CubeTexture.h>
 
+#include <napi/env.h>
+
+namespace Babylon::Polyfills::CubeTexture
+{
+    namespace
+    {
+        // Embedded polyfill source. Kept verbatim from the original Playground
+        // cube_texture_polyfill.js (PR #1710 v0) so behaviour stays identical:
+        //   - Patches BABYLON.NativeEngine.prototype.createCubeTexture.
+        //   - Triggers only when the URL ends in .dds / .ktx / .ktx2 and no
+        //     forcedExtension, 6-file array, or raw buffer was supplied.
+        //   - Retries via NativeEngine._loadFile to fetch the .env counterpart;
+        //     on 404 falls back to the original (which throws), preserving the
+        //     existing failure mode.
+        //   - Idempotent via the __cubeTexturePolyfillInstalled marker on the
+        //     prototype.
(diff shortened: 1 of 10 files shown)


Polyfills/CubeTexture/Source/CubeTexture.cpp
```diff
@@ -0,0 +1,134 @@
+#include <Babylon/Polyfills/CubeTexture.h>
```
bghgary, line 1: Three layered concerns:

1. Silent substitution. PG asks for `.dds`, we quietly load the `.env` co-located by Babylon's CDN. Tests "pass" but they're rendering a different file (different format, mip pyramid, precision). If BJS's `.dds` path regresses, we don't catch it.
2. CDN-specific. The fallback only works for assets where someone happens to co-host `.env`. Babylon's playground CDN does; arbitrary BN consumers don't. Shipping this in `Polyfills/` advertises a benefit non-Babylon consumers won't get.
3. Wrong layer. This masks a real gap — BN's `NativeEngine` doesn't dispatch `.dds`/`.ktx`/`.ktx2` cubemap loading even though bimg already supports the formats. The proper fix lives in `Plugins/NativeEngine`.

Let's discuss offline before merging.

bghgary (changes requested):

Concerns inline. Short version: polyfill silently substitutes content and masks the real gap (no .dds/.ktx loader in BN). Let's discuss offline.

PR #1720, opened 2026-05-28
Zero-initialize struct-typed function-local variables (fixes Area Lights, OpenPBR Analytic Lights)

bkaradzic-microsoft:
## Summary

Babylon.js v6+ shaders for Area Lights (`computeAreaLighting`) and a number of other PBR/OpenPBR helpers declare a struct local without an initializer, then immediately do read-modify-write on its fields:

```glsl
lightingInfo result;
result.specular += value;
result.diffuse  += value;
return result;
```

This is undefined behavior under the GLSL spec (locals are not zero-initialized), but every WebGL driver in practice tolerates it. SPIRV-Cross faithfully reproduces the pattern in its HLSL output, and `D3DCompile` correctly rejects it with `error X4000: variable used without having been completely initialized`. Result: `Area Lights - Standard Material` and any other test that exercises this code path fails to compile its fragment shader on D3D11 / D3D12 today.

## Fix

Add a new AST traverser, `ZeroInitializeStructLocals`, that walks every function body and prepends `EOpAssign(symbol, EOpConstructStruct(0, 0, ...))` for each unique struct-typed `EvqTemporary` symbol it references. The SPIR-V emitter then sees a normal store; SPIRV-Cross emits HLSL with the struct local definitely initialized; `D3DCompile` accepts it. Any subsequent real assignment in the body is dead-code eliminated by DXC's optimizer, so functional behavior is unchanged.

Wired into the DXBC, DXIL, Metal, and Vulkan compile pipelines (same as the existing sampler-handling and dFdy traversers). Arrays of structs are intentionally skipped — they are uncommon in BJS shaders and would require constructing a synthetic array constructor.

## Verification

End-to-end on `build/win32-default` (D3D11, Chakra, RelWithDebInfo) and `build/win32-gl` (OpenGL/ANGLE, V8, Release):

| Test | D3D11 | OpenGL |
|---|---|---|
| idx 191 Area Lights - Standard Material | PASS (1027 px diff) | PASS (1131 px diff) |
| idx 194 Area Lights - PBR (regression check for #1718) | PASS | PASS |
| idx 558 OpenPBR Fuzz Weight vs Fuzz Roughness - Analytic Lights | PASS | PASS |
| idx 559 OpenPBR Fuzz Weight vs Coat Weight w/ Normal Maps - Analytic Lights | PASS | PASS |
| idx 574 OpenPBR Transmission Dispersion VS IOR - Analytic Lights | PASS | PASS |
| idx 0/39/47/165 (baseline regression checks) | PASS | PASS |

Wide regression sweep:
- D3D11: 135 indices (1-30, 100-115, 160-180, 285-305, 380-405, 540-560) → 134 PASS / 1 FAIL. The single failure is idx 22 `Outline` with `Error: Unsupported alpha mode: -1` — pre-existing, unrelated to shader pipeline.
- OpenGL: 113 indices (1-50, 100-120, 250-270, 400-420) → 52 PASS / 60 SKIP / 1 FAIL. Same idx 22 failure.

No new regressions on either backend.

## Why a BN traverser (not a SPIRV-Cross fix)

Both the SPIRV-Cross fork-extension route and the BN traverser route would solve this. We chose the traverser because:

1. It avoids extending the BabylonJS/SPIRV-Cross fork further.
2. The fix lives next to the existing sampler/dFdy traversers, which use the exact same pattern.
3. It applies uniformly to all four backends (DXBC, DXIL, Metal, Vulkan) with one change.

## Out of scope

- idx 192/193 (`Textured - Area Lights - *`) still fail with `Unsupported texture format or type: format 5, type 3553`. This is a `NativeEngine::PrepareImage` whitelist gap and orthogonal to this PR.
- A separate cluster of OpenPBR-IBL tests (idx 557/561/575) fails with a JS-side `TypeError: 'FLOAT' undefined` before any shader compile. Will be filed as a separate issue.

### DIFF
--- a/Plugins/ShaderCompiler/Source/ShaderCompilerDXBC.cpp
+++ b/Plugins/ShaderCompiler/Source/ShaderCompilerDXBC.cpp
@@ -108,6 +108,7 @@ namespace Babylon::Plugins
         ShaderCompilerTraversers::AssignLocationsAndNamesToVertexVaryingsD3D(program, ids, vertexAttributeRenaming);
         ShaderCompilerTraversers::SplitSamplersIntoSamplersAndTextures(program, ids);
         ShaderCompilerTraversers::SplitSamplerFunctionParameters(program, ids);
+        ShaderCompilerTraversers::ZeroInitializeStructLocals(program);
         ShaderCompilerTraversers::InvertYDerivativeOperands(program);
 
         // clang-format off
--- a/Plugins/ShaderCompiler/Source/ShaderCompilerDXIL.cpp
+++ b/Plugins/ShaderCompiler/Source/ShaderCompilerDXIL.cpp
@@ -184,6 +184,7 @@ namespace Babylon::Plugins
         ShaderCompilerTraversers::AssignLocationsAndNamesToVertexVaryingsD3D(program, ids, vertexAttributeRenaming);
         ShaderCompilerTraversers::SplitSamplersIntoSamplersAndTextures(program, ids);
         ShaderCompilerTraversers::SplitSamplerFunctionParameters(program, ids);
+        ShaderCompilerTraversers::ZeroInitializeStructLocals(program);
         ShaderCompilerTraversers::InvertYDerivativeOperands(program);
 
         // clang-format off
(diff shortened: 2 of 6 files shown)


bkaradzic-microsoft:
Closed and superseded by #1721 (branch was renamed for housekeeping).

PR #1730, opened 2026-06-03
Tighten screenshot comparison coverage

matthargett:
## Summary

- tighten visual comparison budgets for GUI3D SpherePanel, GUI Slate, and GUI Near Menu
- add Standard/PBR texture-repetition coverage using the canonical Babylon.js playgrounds and references
- keep the GUI gradient cases quarantined with precise bgfx Canvas limitation reasons

## Why

The default 2.5% screenshot budget permits roughly 6,000 differing pixels in a 600x400 image. That is enough to hide missing or materially underdrawn controls and compact material differences. The tighter per-test budgets preserve normal anti-aliasing/color tolerance while making those regressions actionable.

The three GUI gradient cases are not tolerance problems on bgfx. Forced Metal/bgfx runs show missing Canvas semantics for gradient stroke styles, two-circle radial geometry, and endpoint-space linear mapping, so they remain excluded rather than being marked passing with loose thresholds.

The PBR texture-repetition playground contains hash-heavy repetition modes. Its browser reference has a stable pattern delta on bgfx/Metal, so Metal is explicitly excluded while D3D11 and OpenGL retain the strict 1% assertion.

## Validation

- rebased onto current BabylonNative `master`
- tested with Babylon.js 9.26.2
- config JSON parse and `git diff --check`
- GUI Slate, Metal/bgfx: 1,708 of 240,000 pixels (0.712%) at threshold 25
- GUI Near Menu, Metal/bgfx: 692 pixels (0.288%) at threshold 35
- Texture Repetition Standard, Metal/bgfx: 336 pixels (0.140%) at threshold 25
- GUI Slate, GUI Near Menu, and Standard repetition pass their configured budgets in an automated filtered run
- PBR repetition, Metal/bgfx: 22,525 pixels at threshold 25; excluded only on Metal

The repetition PNGs are the Babylon.js references with opaque alpha added for Native's four-channel comparator; RGB pixel values are unchanged.


Apps/Playground/Scripts/config.json
```diff
@@ -4249,6 +4257,23 @@
...
+            "referenceImage": "texture-repetition-standard-material.png",
+            "excludedEngines": ["webgl1"],
+            "dependsOn": ["Materials", "Textures"]
+        },
+        {
+            "title": "Texture Repetition - PBR Material",
+            "playgroundId": "#6EPUUH#0",
+            "renderCount": 1,
+            "errorRatio": 1.0,
+            "referenceImage": "texture-repetition-pbr-material.png",
```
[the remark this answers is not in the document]
matthargett: false positive, it didn't check the math where this number gets multiplied by 100 in the code

Apps/Playground/Scripts/config.json
```diff
@@ -4249,6 +4247,24 @@
             "reason": "Pixel comparison fails (more than 20% pixels differ)",
             "referenceImage": "Lighting-Volume.png"
         },
+        {
+            "title": "Texture Repetition - Standard Material",
+            "playgroundId": "#ESP440#0",
+            "renderCount": 1,
+            "errorRatio": 0.1,
+            "referenceImage": "texture-repetition-standard-material.png",
```
bkaradzic-microsoft: Blocker: this reference image (and the PBR one below) is a 24-bit RGB PNG, whereas every existing reference in this repo is 32-bit RGBA. The harness `compare()` requires the decoded reference buffer and the render buffer to be the same length, so on Babylon Native this throws an uncaught `Reference data length (720000) must match render data length (960000)` and exits 1 — the test can never pass. Please re-export as RGBA, ideally from Native's own screenshot output.

Apps/Playground/Scripts/config.json
```diff
@@ -4249,6 +4247,24 @@
...
+            "excludedEngines": ["webgl1"],
+            "dependsOn": ["Materials", "Textures"]
+        },
+        {
+            "title": "Texture Repetition - PBR Material",
+            "playgroundId": "#6EPUUH#0",
+            "renderCount": 1,
+            "errorRatio": 1.0,
+            "referenceImage": "texture-repetition-pbr-material.png",
```
bkaradzic-microsoft: Same 24-bit RGB vs 32-bit RGBA issue as the Standard Material reference above — this test also fails with an uncaught data-length mismatch on Native. (Also note `errorRatio: 1.0` here = ~1% tolerance, which is fine once the image format is fixed.)

Apps/Playground/Scripts/config.json
```diff
@@ -1133,16 +1135,15 @@
...
             "referenceImage": "guiSlate.png"
         },
         {
             "title": "GUI Near Menu",
             "playgroundId": "#2YZFA0#302",
             "renderCount": 60,
-            "excludedGraphicsApis": ["OpenGL"],
-            "reason": "OpenGL: BGFX FATAL shader compile error in GUI fragment shader ('=' : cannot convert from 'highp float' to 'bool'). Re-enabled on D3D11 post BabylonJS/BabylonNative#1695 (original 'V8 D3D11 crash' no longer reproduces under Chakra; original 'hangs on OpenGL' is now this clean shader-compile failure surfaced by the BabylonJS/BabylonNative#1688 BgfxCallback).",
+            "threshold": 35,
+            "errorRatio": 0.25,
```
bkaradzic-microsoft, line 1102: Removing the OpenGL exclusion re-introduces a hard crash: on the OpenGL build this hits `BGFX FATAL: Failed to compile shader ... '=' : cannot convert from 'highp float' to 'bool'` (renderer_gl.cpp:6956), which aborts the whole OpenGL run — exactly the failure the deleted reason described. Please keep `excludedGraphicsApis: ["OpenGL"]` here until the GUI fragment-shader codegen issue is fixed. The D3D11 path is fine.

Apps/Playground/Scripts/config.json
```diff
@@ -2457,22 +2458,19 @@
...
-            "reason": "Pixel comparison fails (more than 20% pixels differ)",
+            "errorRatio": 1.0,
             "referenceImage": "GUI-Gradient-Linear.png"
         },
         {
             "title": "GUI Gradient Radial",
             "playgroundId": "#4Z7EK3",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
+            "errorRatio": 1.0,
```
bkaradzic-microsoft: Good news: this and "Linear with transparency" do not crash/hang here, so the "crashes or hangs" reason is stale. However, they diff ~60k/~102k pixels locally, far above `errorRatio: 1.0` (~1%), and "GUI Gradient Linear" throws `TypeError: A string was expected`. My build may differ from your base — can you confirm these pass in your CI?

Apps/Playground/Scripts/config.json
```diff
@@ -1133,16 +1135,15 @@
             "title": "GUI Slate",
             "playgroundId": "#2YZFA0#277",
             "renderCount": 60,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Pixel comparison fails (more than 20% pixels differ)",
+            "errorRatio": 0.1,
```
bkaradzic-microsoft: Re-enabled at `errorRatio: 0.1`, but locally this fails to even evaluate the snippet: `Failed to evaluate playground snippet #2YZFA0#277: Error: The data size does not match width, height, and format`. Please confirm it passes in your run.

bkaradzic-microsoft (changes requested):
Thanks for tightening the screenshot tolerances — the per-test `threshold`/`errorRatio` approach is the right direction. I ran the changed/added tests locally against a Playground build (D3D11 + OpenGL) and hit blocking issues on the re-enabled and new tests. Two are structural and independent of Babylon version; a few others I'd like you to confirm pass in your CI, since my local build may differ from your base.

**Blocking:**

1. **New Texture Repetition reference images are 24-bit RGB, but every existing reference is 32-bit RGBA.** The harness `compare()` requires equal decoded-buffer lengths, so both new tests throw an uncaught `Reference data length (720000) must match render data length (960000)` and exit 1 — they can never pass on Babylon Native. Please re-export both PNGs as RGBA (ideally captured from Native's own output rather than the web playground).
2. **Un-excluding OpenGL for `GUI Near Menu` re-introduces a hard crash.** On the OpenGL build it hits `BGFX FATAL: Failed to compile shader ... '=' : cannot convert from 'highp float' to 'bool'` (renderer_gl.cpp:6956) — exactly the failure the reason you removed described. This aborts the entire OpenGL test run. The OpenGL exclusion should stay (e.g. `excludedGraphicsApis: ["OpenGL"]`) until the GUI fragment-shader codegen issue is fixed.

**Please verify (may be environment-specific on my end):**

3. Re-enabled `GUI Gradient Radial` / `Linear with transparency` don't crash/hang here (good — that reason is stale), but they diff ~60k/~102k pixels, far above `errorRatio: 1.0` (~1%). `GUI Gradient Linear` throws `TypeError: A string was expected`, and `GUI Slate` fails to evaluate with `The data size does not match width, height, and format`. Can you confirm these actually pass in your run?

**Minor:** `excludedEngines` and `dependsOn` on the new entries aren't implemented by `validation_native.js` (only `excludedGraphicsApis`, `onlyVisual`, `excludeFromAutomaticTesting` are honored) — they'll be silently ignored, giving a false sense of gating.

bkaradzic-microsoft:
Heads-up on CI: this branch is currently ~38 commits behind `master` (last pushed 2026-06-03), and there's no GitHub Actions check-suite on the head commit — so there's no pending workflow run for a maintainer to approve, which is why CI isn't showing up here.

Could you please **rebase onto `master` and push**? That `synchronize` event will create a fresh CI run that we can approve, and it'll also pick up the current workflow set. Thanks!

Apps/Playground/Scripts/config.json
matthargett: Fixed in 4d42eb3a. I restored the OpenGL exclusion and its shader-compile reason; the tighter threshold and error ratio remain active on supported graphics APIs.

Apps/Playground/Scripts/config.json
matthargett: Fixed in 4d42eb3a. The Standard reference is now a 600x400, 8-bit RGBA PNG. The integrated NativeWebGPU check executes successfully and differs by 2 of 240,000 pixels at threshold 25.

Apps/Playground/Scripts/config.json
matthargett: Fixed in 4d42eb3a. The PBR reference is now a 600x400, 8-bit RGBA PNG. The integrated NativeWebGPU check executes successfully and differs by 761 of 240,000 pixels (0.317%) at threshold 25, within the 1% budget.

matthargett:
Rebased onto current `master` (`fcfda740`) and force-pushed. I also removed the unsupported `excludedEngines` / `dependsOn` fields from the two native test entries. The two GUI verification threads remain open while I reconcile the reported runtime failures.

Apps/Playground/Scripts/config.json
matthargett: Confirmed and root-caused. The old snippet was deleted; 3f5c8ef3 updates this to `#UDLHJ3#0`, which preserves the same HolographicSlate scene and is pixel-identical to the existing golden.

The exception is not D3D11-specific: HolographicSlate requests a 3379.2 x 102.4 dynamic title texture. ThinNativeEngine allocated 1,384,120 bytes from the fractional product, while the native boundary truncated the dimensions to 3379 x 102 and expected 1,378,632 bytes. BabylonJS/Babylon.js#18709 contains the separate runtime fix. With that patch, Metal/bgfx reaches comparison at a stable 1,708/240,000 pixels (0.712%, threshold 25), and NativeWebGPU has zero pixels over threshold 25. The test budget is now 0.75%.

Apps/Playground/Scripts/config.json
matthargett: Confirmed on a matched Metal/bgfx build. Linear throws because `strokeStyle` only accepts a string; Radial differs by 56,743 pixels because the supplied two-circle geometry is ignored; Linear with transparency differs by about 50,000 pixels because the generated strip is mapped to fill bounds rather than the requested endpoints. These are bgfx Canvas integration gaps, not acceptable tolerance. Commit 4e6419a1 re-excludes all three with a precise reason and removes the ineffective error ratios.

Apps/Playground/Scripts/config.json
matthargett: Follow-up in 4e6419a1: Slate now uses `#CHDZ5C#1`, a synchronous equivalent with no private readiness wait or render-loop warmup. The normal 60 validation frames produced byte-identical captures with the current runner and with #1762 applied, so the PRs are independent. BabylonJS/Babylon.js#18709 remains the required dependency for the fractional dynamic-texture allocation fix. Metal/bgfx remains stable at 1,708/240,000 pixels (0.712%, threshold 25); NativeWebGPU has 0 pixels over threshold.

bkaradzic-microsoft:
Reference screenshots should be coming from B.js, and should match PlaygroundID.

RaananW (changes requested):
[PLEASE IGNORE THIS]

matthargett:
Addressed in 334315b7. GUI Slate and GUI Near Menu now use the canonical Babylon.js visual-test IDs (#UDLHJ3#0 and #DZPGOB#0), and both repetition goldens come from Babylon.js. The only PNG transformation is adding opaque alpha so Native's four-channel comparator can consume them; RGB pixels are unchanged. On current bgfx/Metal with Babylon.js 9.26.2, Slate, Near Menu, and Standard repetition pass. PBR is explicitly excluded only on Metal because its hash-based modes have a stable cross-backend pattern delta; D3D11 and OpenGL keep the strict 1% assertion.

PR #1736, opened 2026-06-05
Add Hermes CI coverage

CedricGuillemet:
## Summary
- point JsRuntimeHost at the Hermes integration branch
- add Win32 x64 D3D11 + Hermes CI
- add Android + Hermes CI with host compiler import
- add Ubuntu + Hermer CI

## Notes
- Apple + Hermes is intentionally not added because shutdown currently aborts on macOS/iOS.


.github/workflows/build-android.yml
```diff
@@ -26,12 +26,39 @@ jobs:
...
+            sudo apt-get install -y ninja-build
+          elif [ "${{ runner.os }}" = "macOS" ] && ! command -v ninja >/dev/null 2>&1; then
+            brew install ninja
+          fi
+
+      - name: Build Hermes host compilers
+        if: inputs.js-engine == 'Hermes'
+        run: |
+          git clone --depth 1 --branch hermes-integration https://github.com/CedricGuillemet/JsRuntimeHost.git Build/JsRuntimeHostHost
+          cmake -B Build/HermesHost -G Ninja -S Build/JsRuntimeHostHost \
```
ryantrem: Will this need to be done by BN consumers as well then?

CMakeLists.txt
```diff
@@ -53,8 +53,8 @@ FetchContent_Declare(ios-cmake
     GIT_TAG 4.5.0
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(JsRuntimeHost
-    GIT_REPOSITORY https://github.com/BabylonJS/JsRuntimeHost.git
-    GIT_TAG c88625b6d61b55c4589f02408d03826e83199870)
+    GIT_REPOSITORY https://github.com/CedricGuillemet/JsRuntimeHost.git
```
ryantrem: Putting a note here just to make sure it doesn't get forgotten - update to the new commit after the JsRuntimeHost PR is merged.

ryantrem (commented):
Can you update the docs to explain what a BN consumer needs to do to use Hermes? It seems like it is more than just switch the compile flag (like JSC vs. v8 for example).

.github/workflows/build-android.yml
CedricGuillemet: I've reworked that part so the hermes toolchain is handled by JSRuntimeHost transparently.

Apps/Playground/Android/BabylonNative/build.gradle
```diff
@@ -18,6 +18,25 @@ if (project.hasProperty("GRAPHICS_API")) {
...
+    "-DANDROID_STL=c++_shared",
+    "-DENABLE_PCH=OFF",
+    "-DGRAPHICS_API=${graphics_api}",
+    "-DARCORE_LIBPATH=${arcore_libpath}/jni",
+    "-DNAPI_JAVASCRIPT_ENGINE=${jsEngine}",
+    "-DBABYLON_NATIVE_BUILD_APPS=ON",
+    "-DBABYLON_DEBUG_TRACE=ON"
+]
+// Hermes for Android needs HOST hermesc/shermes (they emit bytecode at build
+// time and can't run as NDK-cross-compiled binaries). JsRuntimeHost's CMake
```
ryantrem, line 31: Is the Hermes compiler (hermesc) only required to load Hermes bytecode? I thought this first round of changes was for loading JS bundles only? Does Hermes support loading either JS or ByteCode, and is this a platform specific question?

CedricGuillemet: hermes has some polyfills (promise, async function, iterator helpers) that are compiled into byte code. And that bytecode is embedded in hermes static library. Hermes can run bytecode or .js (but in that case, it compiles to bytecode internally and run that produced bytecode, transparently). not specific to a platform.

ryantrem: Oh I see, interesting. And you can also compile your own JS to bytecode ahead of time so that at runtime you only load the bytecode, right?

CedricGuillemet: yes, it's the exact same as QuickJS. I tried QuickJSC and the bytecode was bigger than the source .js. idk the footprint for Hermes.

ryantrem (approved):

PR #1737, opened 2026-06-05
Playground: re-enable 52 more previously-excluded tests

bkaradzic-microsoft:
## Summary

Follow-up to #1715. Re-enables **52 more** previously-excluded Playground tests that now pass after recent fixes landed on `master`:

- **23** Area Lights / OpenPBR Analytic Lights tests — fixed by #1718 (sampler2D function-parameter split) and #1721 (zero-init struct-typed locals).
- **29** Canvas framebuffer-pool cascade victims — fixed by #1715 (FB-pool survivability).

## Verification

Each test was verified with a full headless sweep on **both** Win32 D3D11 and ANGLE OpenGL locally:

```
Run complete. ran=237 passed=237 failed=0 missingRef=0 skipped=483
```

0 failures on either backend, and no regressions on the previously-active set (the sweep also confirms no downstream cascade onto already-enabled tests).

## Deferred (still excluded)

| Group | Count | Why |
|---|---|---|
| OpenPBR pixel-diffs | 6 | render correctly now but differ from reference; need rebake/triage |
| OpenGL-only failures | 2 | UV2 Morphing, Vertex Pulling Morph Targets and Bones |
| Linux pixel-diffs | ~22 | real pixel comparison failures, separate triage |
| State corruptors | 2 | `cube-with-holes-using-stencil-buffer` (dirty stencil) and `GLTF Node visibility test` (dirties OpenGL state -> breaks downstream `MeshDebugPluginMaterial`). Each needs its own fix before re-enabling. |

The two corruptors were pinned down by bisection — re-enabled alone, each produces a downstream cascade — so they are intentionally held back.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -1143,8 +1143,6 @@
             "title": "Gizmos",
             "playgroundId": "#8GY6J8#196",
             "renderCount": 50,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes on Win32 D3D11 (incl. Sanitizers and V8) during mesh load",
             "referenceImage": "Gizmos.png"
         },
         {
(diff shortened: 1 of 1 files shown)


bghgary (approved):

LGTM. Config-only test re-enablement — verified the diff removes only `excludeFromAutomaticTesting`/`reason` markers, the two held-back state corruptors stay excluded, and no new exclusions are added.

PR #1739, opened 2026-06-06
NativeEngine: divisor-driven per-instance vertex attributes

bkaradzic-microsoft:
## Problem

Consumer-declared per-instance vertex attributes were routed to **per-vertex** shader slots, so their data was read incorrectly at draw time. The `ShaderCompiler` only recognized the built-in instanced names (`world0-3`, `instanceColor`, `splatIndex`); any *other* attribute backed by an instance buffer (divisor &gt; 0) — e.g. GPU-particle `position`/`color`/`angle`/`size`, or a custom instanced color buffer — was assigned a per-vertex location.

## Fix

Drive instancing purely from the vertex-buffer **divisor** instead of a fixed name list:

- **ShaderCompiler traversers (D3D, OpenGL, Metal)** now take a `map<name, target-location>` of instanced attributes and route each declared attribute to an explicit bgfx `i_data` slot (`TEXCOORD7..TEXCOORD3` = `i_data0..i_data4`). Built-in instanced names keep their existing reverse-count assignment.
- **Program** caches lazily-compiled instanced shader variants keyed by the instanced-attribute map; the base program is used when nothing is instanced.
- **NativeEngine** computes the instanced-attribute set at draw time from the bound vertex array's per-attribute divisors and submits the matching variant. No JS-side hint is required.

This generalizes the instancing path so it works for any divisor-driven attribute, not just the hardcoded names, and supersedes the prior name-list approach.

## Test enablement

Re-enables **18 previously-excluded** Playground validation tests that failed under the old name-list instancing (most were marked as follow-ups to the #1691 instance-data stride fix). These are the GPU-particle suite plus the instanced color-buffer test:

`Particles`, `Instances with color buffer`, `Prepass SSAO + particles`, `halo-particle-system`, `particle-system-with-custom-nme-shader`, and `Particles - Basic Properties / Change / Emitters / Attractors` variants.

Each was confirmed to actually exercise the fixed path (it compiles a per-instance shader variant) **and** now matches its reference image.

## Verification

- **D3D11:** all 18 enabled tests pass in default automatic mode (`ran=18 passed=18 failed=0`); no regression in existing instanced tests (Gaussian Splatting, mesh instancing).
- **OpenGL / Metal:** use the identical explicit-slot routing and are compile-checked; runtime coverage is via CI.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -1568,9 +1568,7 @@
             "title": "Particles",
             "playgroundId": "#G3ZYFU#7",
             "renderCount": 100,
-            "referenceImage": "particles.png",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Pixel diff after #1691 instance-data stride fix; not stride-related, needs follow-up."
+            "referenceImage": "particles.png"
         },
         {
             "title": "PBRMetallicRoughnessMaterial",
@@ -1842,8 +1840,6 @@
         {
             "title": "Instances with color buffer",
             "playgroundId": "#YPABS1#91",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Pixel comparison fails (more than 20% pixels differ)",
             "referenceImage": "instancecolors.png"
         },
         {
@@ -1851,7 +1847,7 @@
             "playgroundId": "#65MUMZ#47",
             "renderCount": 50,
             "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
+            "reason": "SSAO2 blur post-process shader fails to compile on desktop GL (samples uniform used as int loop bound); unrelated to instancing.",
             "referenceImage": "prepass-ssao-particles.png"
         },
         {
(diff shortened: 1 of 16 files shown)


bkaradzic-microsoft:
Pushed an update addressing CI + review feedback:

**CI fix (Linux GL):** `Prepass SSAO + particles` aborted the OpenGL run, but the failure is unrelated to instancing — it's the SSAO2 blur post-process fragment shader, whose `samples` uniform (declared `vec4`) is used as an `int` loop bound, which desktop GL (GLSL ES 3.0) rejects:
```
mediump int _42 = -samples.x;                 // float -> int
for (mediump int i = _42; i < samples.x; ...) // int < float
```
That test's original exclusion reason was "crashes or hangs" (not the instancing follow-up), so it has been re-excluded. The genuinely instancing-attributable particle tests are confirmed passing on OpenGL (e.g. `Particles`, same per-instance attributes, validates on the Linux GL build). Net enablement is now 17 tests.

**Copilot review nits (all addressed):**
- Removed the now-unused `<set>` include from NativeEngine.cpp (replaced with `<map>`).
- Removed the unused `instanceInfo` structured binding in the draw-time loop.
- The invalid-i_data-slot error now includes the attribute name and computed location/slot (all three traverser paths).

CedricGuillemet (approved):
LGTM!

PR #1740, opened 2026-06-08
Remove libwebp dependency; rely on bimg for WebP decoding

bkaradzic-microsoft:
## What

bimg now parses WebP natively via its bundled `simplewebp` decoder (`BIMG_CONFIG_PARSE_WEBP` defaults to on), so the standalone **libwebp** dependency and the WebP decode fallback in NativeEngine are now redundant. This removes them and routes WebP support through bimg.

## Changes

- Drop the `libwebp` `FetchContent` declaration (root `CMakeLists.txt`) and its dependency build block (`Dependencies/CMakeLists.txt`), and remove the `libwebp` install entry (`Install/Install.cmake`).
- **Keep** the `BABYLON_NATIVE_PLUGIN_NATIVEENGINE_WEBP` option, but repurpose it: instead of compiling the old libwebp fallback, it now toggles bimg's `BIMG_CONFIG_PARSE_WEBP` (so WebP decoding can still be disabled — `OFF` sets `BIMG_CONFIG_PARSE_WEBP=0` on `bimg_decode`). The old `WEBP` compile definition / `webp` link on the NativeEngine target are removed.
- Remove the `<webp/decode.h>` include and the `WebPGetInfo` / `WebPDecodeRGBAInto` fallback in `ParseImage`; `bimg::imageParse` now handles WebP directly.
- Replace the `libwebp` `NOTICE.md` section with a `simplewebp` entry (bimg's vendored decoder; BSD-3-Clause, © 2010 Google Inc., 2023 Miku AuahDark).

## Verification

- CMake configures with no libwebp fetch and no dangling target references.
- Playground builds clean (D3D11).
- The `EXT_texture_webp` Playground visual test passes with **no pixel difference** using bimg's decoder (`ran=1 passed=1 failed=0`). CI exercises this test on all backends.
- Rebased onto latest `master`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,10 +55,6 @@ FetchContent_Declare(ios-cmake
 FetchContent_Declare(JsRuntimeHost
     GIT_REPOSITORY https://github.com/BabylonJS/JsRuntimeHost.git
     GIT_TAG c88625b6d61b55c4589f02408d03826e83199870)
-FetchContent_Declare(libwebp
-    GIT_REPOSITORY https://github.com/webmproject/libwebp.git
-    GIT_TAG 57e324e2eb99be46df46d77b65705e34a7ae616c
-    EXCLUDE_FROM_ALL)
 FetchContent_Declare(metal-cpp
     GIT_REPOSITORY https://github.com/bkaradzic/metal-cpp.git
     GIT_TAG metal-cpp_26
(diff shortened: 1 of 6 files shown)


bghgary (approved):

LGTM — verified bimg's bundled simplewebp covers the spec-valid `EXT_texture_webp` variants (lossy/lossless/alpha; animated is spec-disallowed per the extension's best practices), so dropping libwebp is not a decode regression.

Two minor, non-blocking:
- The description says the `BABYLON_NATIVE_PLUGIN_NATIVEENGINE_WEBP` option is removed, but it's actually kept and repurposed to toggle bimg's `BIMG_CONFIG_PARSE_WEBP` — worth correcting.
- WebP now comes from bimg's vendored **simplewebp** (BSD-3, © 2010 Google, 2023 Miku AuahDark). Since `NOTICE.md` enumerates bundled libraries individually, consider replacing the removed `libwebp` section with a `simplewebp` entry rather than relying on the `bimg` entry.

Issue #1420, opened 2024-09-05
Fix shutdown issues 


bghgary:
PR: https://github.com/BabylonJS/BabylonNative/pull/1141

See issues:
- #1193

More details later.

PR #1746, opened 2026-06-10
Drain in-flight graphics tasks before NativeEngine teardown

bkaradzic-microsoft:
### Problem

`NativeEngine::Dispose` only called `m_cancellationSource->cancel()`. arcana cancellation short-circuits work at **scheduling** boundaries, but a task body that is **already running** on a threadpool thread keeps going to completion.

The graphics-loading entry points (`loadProgram`, `loadTexture`, `loadCubeTexture`) schedule their work as threadpool tasks that capture a raw `Graphics::Texture*` / call into bgfx. On app teardown, if the suite finishes while one of these bodies is mid-flight, `Dispose` cancels and teardown proceeds to free the texture / tear down the bgfx context — and the still-running task then touches freed resources, causing an access violation:

```
Access violation. Exception Code c0000005
  2: Plugins/NativeEngine/Source/NativeEngine.cpp  LoadTextureFromImage
  3: arcana .../internal_task.h  output_wrapper<...>::invoke<...LoadTexture...lambda...>
  ...
  6: cppwinrt .../Windows.System.Threading.h  WorkItemHandler::Invoke
```

This reproduces at shutdown, after the test run reports it finished.

### Fix

Add a small `AsyncTaskTracker` (mutex + condition_variable + counter) whose RAII `Scope` token is captured into the four graphics-touching threadpool lambdas (the `loadProgram` body, the `loadTexture` body, and the two `LoadCubeTextureFromImages` inline continuations). The CPU-only image-parse bodies are intentionally not tracked.

`Dispose` now cancels and then **waits** for those tokens to drain before returning, so teardown can't free graphics resources out from under a running task. The token decrements on the threadpool thread (never on a JS-thread `.then` continuation), so `Dispose` — which runs on the JS thread — can never deadlock waiting on its own continuations. Counting is bound to the token's lifetime, so the increment/decrement pair can't be split by an exception.

A defensive `if (cancellationSource->cancelled()) return;` early-out is also added at the top of the two threadpool load bodies, so once teardown has begun they skip touching graphics resources entirely.

### Testing

- `NativeEngine` and `Playground` build clean (RelWithDebInfo, x64, D3D11).
- Ran the previously-crashing `Sprites` test (`loadTexture` path) and a cubemap test (`LoadCubeTextureFromImages` path) headless — both exercise the load paths and tear down promptly with no hang and no crash.

### DIFF
--- a/Apps/UnitTests/CMakeLists.txt
+++ b/Apps/UnitTests/CMakeLists.txt
@@ -29,6 +29,7 @@ set(SOURCES
     "Source/Tests.ExternalTexture.Msaa.cpp"
     "Source/Tests.ExternalTexture.Render.cpp"
     "Source/Tests.JavaScript.cpp"
+    "Source/Tests.NativeEngine.Teardown.cpp"
     "Source/Tests.ShaderCache.cpp"
     "Source/Tests.ShaderCompilation.cpp"
     "Source/Tests.UniformPadding.cpp"
--- a/Apps/UnitTests/Source/Tests.NativeEngine.Teardown.cpp
+++ b/Apps/UnitTests/Source/Tests.NativeEngine.Teardown.cpp
@@ -1,0 +1,84 @@
+#include <gtest/gtest.h>
+
+#include <Babylon/AppRuntime.h>
+#include <Babylon/Graphics/Device.h>
+#include <Babylon/Plugins/NativeEngine.h>
+
+#include <chrono>
+#include <future>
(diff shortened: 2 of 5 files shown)


bghgary (approved):

LGTM as a stopgap for the graphics-AV class. The broader async-at-teardown pattern (JS-env class, e.g. NativeEncoding) is tracked in #1420.

PR #1750, opened 2026-06-11
Add native cube render target support (re-enables 2 shadow-with-instances tests)

bkaradzic-microsoft:
Paired engine PR: BabylonJS/Babylon.js#18561

## What

Implements cube render targets on the native engine so `ReflectionProbe` and
point-light cube shadow maps render instead of crashing on the null WebGL
context. Pairs with the Babylon.js change that adds the JS-side overrides.

## Changes

- `Plugins/NativeEngine/Source/NativeEngine.cpp`
  - `InitializeTexture`: optional `isCube` param → `Texture::CreateCube`.
  - `CreateFrameBuffer`: optional `layer` param so the color attachment targets
    a single cube face (`bgfx::Attachment` layer). The JS side creates one
    framebuffer per face and binds the matching one per face.
- `Apps/Playground/Scripts/config.json`: re-enables "Shadows with instances in
  left/right handed system" (previously excluded as crash/hang).

## CI

The validation suite uses the published `babylonjs` npm, which does not yet
contain the paired JS overrides, so the re-enabled tests stay red until a
`babylonjs` release with the Babylon.js change ships and the dependency is
bumped here. Draft until then.

## Verified locally

Built against a local `babylon.max.js` with the paired change: the two
re-enabled tests pass and there are no regressions in post-process / 2D
render-target tests.

---

- **Babylon.js (engine / TS):** https://github.com/BabylonJS/Babylon.js/pull/18561
- **BabylonNative (C++ engine + test re-enable):** https://github.com/BabylonJS/BabylonNative/pull/1750

Co-dependent; land in this order:
1. **Babylon.js #18561 first** — adds the cube render-target TS overrides; no WebGL behavior change.
3. **BabylonNative #1750 last** — bumps the bundled `babylonjs` and re-enables the 2 `Shadows with instances` (left/right handed) validation tests, which only pass once the paired JS is present in the bundled engine.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -1823,16 +1823,12 @@
             "title": "Shadows with instances in left handed system",
             "playgroundId": "#MSAHKR#79",
             "renderCount": 10,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "shadowsinstancesleft.png"
         },
         {
             "title": "Shadows with instances in right handed system",
             "playgroundId": "#MSAHKR#13",
             "renderCount": 10,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "shadowsinstancesright.png"
         },
         {
--- a/Apps/package-lock.json
+++ b/Apps/package-lock.json
@@ -11,13 +11,13 @@
         "UnitTests/JavaScript"
       ],
       "dependencies": {
-        "babylonjs": "^9.3.4",
-        "babylonjs-addons": "^9.3.4",
-        "babylonjs-gltf2interface": "^9.3.4",
-        "babylonjs-gui": "^9.3.4",
-        "babylonjs-loaders": "^9.3.4",
-        "babylonjs-materials": "^9.3.4",
-        "babylonjs-serializers": "^9.3.4",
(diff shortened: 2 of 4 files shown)


bkaradzic-microsoft:
**TODO at landing:** bump the native `PROTOCOL_VERSION` (`Plugins/NativeEngine/Source/NativeEngine.cpp`, the `StaticValue("PROTOCOL_VERSION", ...)`) from 9 → 10 to pair with the JS-side bump in BabylonJS/Babylon.js#18561 (the cube-RT version gate from bghgary's review).

This must land in the **same change that bumps the `babylonjs` pin** to a release containing #18561's `PROTOCOL_VERSION = 10`. Bumping it standalone now would make every test here throw `Protocol version mismatch` at engine init against the currently-bundled 9.x, so it's deferred until the coordinated pin bump.

bkaradzic-microsoft:
Closing as superseded.

The functional change here (`isCube` handling in `CreateFrameBuffer` plus the cube-face/array layer for the color attachment) is present verbatim on the `shotgun` integration branch, alongside a good deal more: explicit depth-texture attachments, `maxLayers`/`numLayers`, mip selection, and cube-face upload/readback. It will be re-proposed as part of the consolidated **framebuffers & render targets** PR (BN-1).

The `config.json` un-exclusions and the `Apps/package.json` babylonjs bump are deliberately not carried over -- test enablement is being landed after the engine changes it depends on, so that each re-enabled test has a merged engine fix to point at.

PR #1751, opened 2026-06-11
NativeEngine: implement updateTextureData (re-enables Test updateTextureData)

bkaradzic-microsoft:
Paired engine PR: BabylonJS/Babylon.js#18566

## What

Implements `updateTextureData` on the native engine so partial (sub-rectangle) texture updates work;
previously it threw "not implemented". Pairs with the Babylon.js change that also stops the native
engine reporting `engine.name === "WebGL"` (that masquerade made `engine.name`-gated WebGL `_gl`
access run on native and crash with `TEXTURE_2D undefined`).

## Changes

- `Plugins/NativeEngine/Source/NativeEngine.cpp` / `.h`: add `NativeEngine::UpdateTextureData`, which
  uploads the requested sub-rectangle via `bgfx::updateTexture2D` (`Texture::Update2D`). It validates
  the JS-controlled rectangle against the mip-level extents, sizes the copy with
  `bgfx::calcTextureSize` (bgfx is always linked, unlike bimg, so this also works in builds without
  image loading), and mirrors the vertical flip the base texture upload applies so the sub-rectangle
  lines up on top-left-origin backends (e.g. D3D11).
- `Apps/Playground/Scripts/config.json`: re-enable "Test updateTextureData".

## CI

The validation suite uses the published `babylonjs` npm, which does not yet contain the paired JS
overrides, so the re-enabled test stays red until a `babylonjs` release with the Babylon.js change
ships and the dependency is bumped here. Draft until then.

## Verified locally

Built against a local `babylon.max.js` with the paired change: the re-enabled test passes (stable
across repeated runs); no regressions in other texture / raw-texture tests.

---

- **Babylon.js (engine / TS):** https://github.com/BabylonJS/Babylon.js/pull/18566
- **BabylonNative (C++ engine + test re-enable):** https://github.com/BabylonJS/BabylonNative/pull/1751

Co-dependent; land in this order:
1. **Babylon.js #18566 first** — adds the `engine.name = "Native"` + `updateTextureData` TS overrides; no WebGL behavior change.
3. **BabylonNative #1751 last** — bumps the bundled `babylonjs` and re-enables the `Test updateTextureData` validation test, which only passes once the paired JS is present in the bundled engine.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -2212,8 +2212,6 @@
         {
             "title": "Test updateTextureData",
             "playgroundId": "#EVX1DH#80",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Pixel comparison fails (more than 20% pixels differ)",
             "referenceImage": "testUpdateTextureData.png"
         },
         {
--- a/Apps/package-lock.json
+++ b/Apps/package-lock.json
@@ -11,13 +11,13 @@
         "UnitTests/JavaScript"
       ],
       "dependencies": {
-        "babylonjs": "^9.3.4",
-        "babylonjs-addons": "^9.3.4",
-        "babylonjs-gltf2interface": "^9.3.4",
-        "babylonjs-gui": "^9.3.4",
-        "babylonjs-loaders": "^9.3.4",
-        "babylonjs-materials": "^9.3.4",
-        "babylonjs-serializers": "^9.3.4",
+        "babylonjs": "^9.15.0",
+        "babylonjs-addons": "^9.15.0",
+        "babylonjs-gltf2interface": "^9.15.0",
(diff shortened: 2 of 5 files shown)


bkaradzic-microsoft:
Closing as superseded.

`UpdateTextureData` as written here -- including its three bounds/format validation errors -- is present verbatim on the `shotgun` integration branch, which adds roughly 1150 further lines of texture work around it. It will be re-proposed as part of a single **texture upload & formats** PR (BN-2) covering `updateTextureData`, single-file `.dds/.ktx/.ktx2` cubemaps + spherical harmonics, the `updateTextureDirectly` loader sink, cube face upload/readback, a BC1 crash, and raw 3D textures.

The `config.json` un-exclusion and the babylonjs bump are dropped here; test enablement lands separately, after the engine fixes.

PR #1754, opened 2026-06-11
NativeEngine: MultiRenderTarget framebuffers + OIT alpha blend modes

bkaradzic-microsoft:
Paired engine PR: BabylonJS/Babylon.js#18568

## What

Foundational native support for `MultiRenderTarget` (MRT) and the order-independent-transparency blend
modes, removing several "`engine._gl` is null" crash classes. It does **not** by itself turn the
MRT/OIT/FrameGraph validation tests green — those need further work (see Follow-ups).

## Changes

- **`Plugins/NativeEngine/Source/NativeEngine.cpp` / `.h`**
  - `CreateFrameBuffer` and `CreateMultiFrameBuffer` now share a private
    `CreateFrameBufferImpl(env, colorTextures, …)` helper that builds one bgfx framebuffer with N color
    attachments (+ an optional depth/stencil attachment), so a `MultiRenderTarget` renders to all
    targets at once (bgfx writes every attachment of the bound framebuffer — no `drawBuffers` needed).
    The two N-API methods are kept for protocol compatibility and only differ in how they parse
    `info[0]` (single nullable texture vs. array); the single-RT path is just the 0-or-1-color case.
    The shared helper validates the attachment count against `caps->limits.maxFBAttachments` and keeps
    the `"Stencil without depth…"` warning and the `isTextureValid` assert.
  - Add alpha blend modes `ALPHA_ONEONE_ONEONE` (11) and `ALPHA_LAYER_ACCUMULATE` (17) used by the
    depth-peeling OIT renderer.

## Paired engine PR

Pairs with the Babylon.js change (`createMultipleRenderTarget` + MRT helper overrides, `applyStates`
override, reverse-Z `clear`, alpha-mode mapping). That PR feature-detects both `createMultiFrameBuffer`
and the two alpha constants, so it degrades gracefully on native binaries that predate this PR.

## CI

The validation suite uses the published `babylonjs` npm, which doesn't yet contain the paired JS, so
no new tests are enabled here and CI stays in the usual "pending dep bump" state. Verified locally
against a `babylon.max.js` built from the paired branch, and the plugin compiles clean
(`cmake --build build/win32 --target NativeEngine --config RelWithDebInfo`).

## Follow-ups (separate work)

- OIT depth-peeling still faults inside the D3D11 driver on `submit` (multi-output / SRV↔RTV
  ping-pong); needs interactive GPU debugging (PIX / VS Graphics Debugger).
- Native does not yet apply the blend **equation** (`MAX`) — `setAlphaEquation` is a no-op — so OIT
  blending would still be incorrect after the crash is fixed.
- 2D-array / cube color attachments for MRT are approximated as 2D (test "MRT with different texture
  types").

---

- **Babylon.js (engine / TS):** https://github.com/BabylonJS/Babylon.js/pull/18568
- **BabylonNative (C++ engine):** https://github.com/BabylonJS/BabylonNative/pull/1754

These two are co-dependent and land in this order:
1. **Babylon.js #18568 first** — it only adds native-engine TS overrides, changes no WebGL behavior,
   re-enables no validation tests on its own, and feature-detects this PR's additions, so it can merge
   independently.
3. **BabylonNative #1754 last** — after bumping the bundled `babylonjs` to that release. (This
   foundational pair does not yet turn the MRT/OIT/FrameGraph tests green — that is separate follow-up
   work.)

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -77,6 +77,8 @@ namespace Babylon
             constexpr uint64_t MULTIPLY = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_DST_COLOR, BGFX_STATE_BLEND_ZERO, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ONE);
             constexpr uint64_t MAXIMIZED = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_SRC_ALPHA, BGFX_STATE_BLEND_INV_SRC_COLOR, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ONE);
             constexpr uint64_t ONEONE = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ZERO, BGFX_STATE_BLEND_ONE);
+            constexpr uint64_t ONEONE_ONEONE = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ONE);
+            constexpr uint64_t LAYER_ACCUMULATE = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_SRC_ALPHA, BGFX_STATE_BLEND_INV_SRC_ALPHA, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_INV_SRC_ALPHA);
             constexpr uint64_t PREMULTIPLIED = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_INV_SRC_ALPHA, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_ONE);
             constexpr uint64_t PREMULTIPLIED_PORTERDUFF = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_INV_SRC_ALPHA, BGFX_STATE_BLEND_ONE, BGFX_STATE_BLEND_INV_SRC_ALPHA);
             constexpr uint64_t INTERPOLATE = BGFX_STATE_BLEND_FUNC_SEPARATE(BGFX_STATE_BLEND_FACTOR, BGFX_STATE_BLEND_INV_FACTOR, BGFX_STATE_BLEND_FACTOR, BGFX_STATE_BLEND_INV_FACTOR);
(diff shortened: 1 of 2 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1912,6 +1917,66 @@ namespace Babylon
         return Napi::Pointer<Graphics::FrameBuffer>::Create(info.Env(), frameBuffer, Napi::NapiPointerDeleter(frameBuffer));
     }
 
+    Napi::Value NativeEngine::CreateMultiFrameBuffer(const Napi::CallbackInfo& info)
```
bghgary, line 1889: Fold this and `CreateFrameBuffer` into one shared private helper (color-attachment init, depth/stencil creation, framebuffer creation, and cleanup are identical) — keep both Napi methods so the protocol is unchanged, just differ in parsing. That also restores the `"Stencil without depth…"` warning and the `isTextureValid` assert this version dropped, and lets the single-RT path stay the 0-or-1-color case (which currently rejects `colorCount == 0`).

bghgary (changes requested):

One refactor comment inline.

Plugins/NativeEngine/Source/NativeEngine.cpp
bkaradzic-microsoft: Done in 04f709eb. Both are now folded into a shared private `CreateFrameBufferImpl(env, colorTextures, width, height, generateStencilBuffer, generateDepth, samples)`. The two N-API methods are unchanged on the wire and only differ in how they parse `info[0]` (single nullable texture vs. array) before calling the helper.

That restores the `"Stencil without depth…"` warning and the `isTextureValid` assert the multi path had dropped, and makes the single-RT path the 0-or-1-color case (the shared helper no longer rejects `colorCount == 0`). The `maxFBAttachments` upper-bound guard is preserved.

bghgary (approved):

LGTM. The description is now stale, though: since #18568 feature-detects the native surface, the "co-dependent / landing order" framing no longer applies — the two can land separately, and this can come out of draft.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1853,12 +1858,55 @@ namespace Babylon
...
+    Napi::Value NativeEngine::CreateFrameBufferImpl(Napi::Env env, gsl::span<Graphics::Texture* const> colorTextures, uint16_t width, uint16_t height, bool generateStencilBuffer, bool generateDepth, uint32_t samples)
+    {
+        const bgfx::Caps* caps = bgfx::getCaps();
+        const uint32_t colorCount = static_cast<uint32_t>(colorTextures.size());
+        // One slot per color attachment plus a single depth/stencil attachment. bgfx caps the total via
+        // maxFBAttachments; reject out-of-range counts up front rather than relying on the validation assert.
+        if (colorCount + 1 > caps->limits.maxFBAttachments)
+        {
+            throw Napi::Error::New(env, "Invalid number of color attachments for frame buffer");
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 05862e2e. The guard now reserves the depth/stencil slot only when one is generated: `const uint32_t depthStencilCount = (generateStencilBuffer || generateDepth) ? 1u : 0u;` then checks `colorCount + depthStencilCount > caps->limits.maxFBAttachments`. This matches the attachment-creation path below (which only appends the depth/stencil attachment inside the same `generateStencilBuffer || generateDepth` branch), so a max-color-count MRT with no depth/stencil is no longer wrongly rejected.

PR #1761, opened 2026-06-13
Re-enable 4 validation tests that now pass on Babylon Native

bkaradzic-microsoft:
## What

Re-enables **4 validation tests** in `Apps/Playground/Scripts/config.json` that now pass on Babylon Native. **config.json-only change** — no engine/source changes.

| Test | Previous exclusion reason |
|---|---|
| NME Multi Build | Test crashes or hangs on Babylon Native |
| Custom render target | Test crashes on Win32 D3D11 plain (STATUS_BREAKPOINT) |
| UV2 Morphing | Test fails locally on Win32 D3D11 sweep; disabled until BabylonNative fixes are made |
| Vertex Pulling - Morph Targets and Bones | Test fails locally on Win32 D3D11 sweep; disabled until BabylonNative fixes are made |

## How this was determined

Ran a full `--include-excluded` sweep of every quarantined test (426 of them) on the latest `master` build (Win32 / D3D11, RelWithDebInfo). These four came back green. Each exclusion reason was a **Win32 crash or a temporary "disabled until BabylonNative fixes" hold** that is now resolved on current `master`.

## Verification (Win32 / D3D11)

- Each test passes **3/3 in isolation** (`--once --include-excluded --test-index=N`).
- All four pass in the **normal (non-excluded) run flow** (`ran=4 passed=4 failed=0`); pixel diffs are small (920 / 1695 / 252 / 493 px, within tolerance).
- Three of the four also validated **in-context during a full-suite run** before an unrelated pre-existing failure halted the (fail-fast) sweep.

## Scope note

Only tests whose exclusion reason was a **Win32/cross-platform** issue (crash or "disabled until BN fix") are included here. Tests excluded specifically for **Linux/OpenGL** pixel diffs, or for order-dependent/OpenGL-only reasons, were intentionally left excluded since this sweep only validates Win32 D3D11.

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -843,8 +843,6 @@
         {
             "title": "NME Multi Build",
             "playgroundId": "#D8AK3Z#113",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes or hangs on Babylon Native",
             "referenceImage": "nme-multi-build.png"
         },
         {
@@ -1265,8 +1263,6 @@
         {
             "title": "Custom render target",
             "playgroundId": "#TQCEBF#5",
-            "excludeFromAutomaticTesting": true,
-            "reason": "Test crashes on Win32 D3D11 plain (STATUS_BREAKPOINT)",
             "referenceImage": "customRTT.png"
         },
         {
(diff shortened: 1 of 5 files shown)


bghgary (approved):

LGTM

PR #1762, opened 2026-06-16
Preserve animation frames during screenshot readiness renders

matthargett:
## Summary

Add an opt-in render-readiness pump for screenshot tests whose material/effect preparation needs rendering before `executeWhenReady` completes. Pump frames use `scene.render(true, true)` so they do not consume Babylon animation frames before the configured screenshot frame count begins.

Enable the pump only for Gaussian Splatting Loading, `needDepthPrePass`, and the two GLTF serializer skinning/animation variants. Other tests remain unchanged unless they set `renderReadinessPump: true`; `globalThis.__nativeValidationRenderReadinessPump = true` remains available as a diagnostic override.

Pre-ready render exceptions are logged and counted while readiness polling continues. Every completion and failure path stops the pump consistently; before an active camera exists, polling is throttled to 100 ms.

Also make top-level macOS builds use one coherent Xcode SDK and classify JavaScriptCore framework headers as system headers. This avoids Xcode 26 SDK nullability diagnostics becoming JsRuntimeHost `-Werror` failures without suppressing project warnings.

## Validation

- `node --check Apps/Playground/Scripts/validation_native.js`
- Config JSON parse and exact opt-in assertion
- NativeWebGPU macOS: Gaussian Splatting Loading passed with 9 differing pixels; `needDepthPrePass` passed with 0 differing pixels
- GLTF Serializer Skinning and Animation, left- and right-handed variants: 0 differing pixels each at threshold 25
- Non-opt-in controls: VLS skeleton passed with 1 differing pixel; Particles passed with 0 differing pixels, with no readiness-pump frames
- Xcode 26.5 Release/LTO Playground build completed successfully


Apps/Playground/Scripts/validation_native.js
```diff
@@ -72,6 +72,58 @@
...
+            if (!scene || scene.isDisposed === true || (typeof scene.isDisposed === "function" && scene.isDisposed())) {
+                stopped = true;
+                return;
+            }
+
+            try {
+                if (scene.activeCamera && typeof scene.render === "function") {
+                    // Readiness renders are only for material/effect compilation.
+                    // Preserve the screenshot test's animation frame count.
+                    scene.render(true, true);
```
bkaradzic-microsoft, line 135: The `ignoreAnimations` argument here only skips Animatable/animation-group stepping — particle systems, procedural textures, and `onBefore/AfterRenderObservable` handlers still advance on every `scene.render()`. Because the number of readiness frames depends on how long the scene takes to become ready (machine/CI-load dependent), this can shift frame-1 state for sensitive tests and reintroduce pixel-diff flakiness — and it's default-on for all tests. Consider making the pump opt-in per test, or suspending particles/procedural updates during readiness (or resetting render state once the pump stops) so frame 1 is deterministic.

Apps/Playground/Scripts/validation_native.js
```diff
@@ -72,6 +72,58 @@
...
+                    frameCount++;
+                }
+            } catch (e) {
```
bkaradzic-microsoft, line 138: A throw from a pre-ready `scene.render()` is routed to `onFatalError` → `failTest(done)`, turning a transient not-yet-ready render error into a hard test failure. Consider logging/counting readiness-phase exceptions and continuing to poll, failing the test only after the scene is actually ready.

bkaradzic-microsoft (changes requested):
Thanks — the readiness-pump approach is a nice way to warm up compilation without spending the animation frame budget, and the additive/opt-out design is good. Two things I'd like addressed before merge (details inline):

**1. Non-deterministic readiness state.** `scene.render(true, true)` skips animation stepping, but it does *not* freeze particle systems, procedural textures, or `onBefore/AfterRenderObservable`-driven state — those advance on every render. Since the number of readiness frames depends on how long the scene takes to become ready (varies by machine/CI load), sensitive screenshot tests can see a variable amount of pre-frame-1 state drift. With the pump default-on for all tests, this risks reintroducing pixel-diff flakiness. Could we either (a) make the pump opt-in per test, or (b) also suspend particles/procedural updates during readiness (or reset render state after the pump stops) so frame 1 is deterministic?

**2. Pre-ready render exceptions become hard failures.** A throw from `scene.render()` in the pump routes to `onFatalError` → `failTest(done)`. Rendering a not-yet-ready scene can transiently throw; treating that as a definitive failure can convert would-pass tests into failures. Suggest tolerating readiness-phase exceptions (log/count, keep polling) and failing only once the scene is actually ready.

Apps/Playground/Scripts/validation_native.js
matthargett: gotcha. The pump is now opt-in rather than catalog-wide: tests must set `renderReadinessPump: true`, with the global `true` override retained only for diagnostics. It is enabled for the four readiness cases that need it. As controls, VLS skeleton and Particles ran without pump frames and still passed at 1 and 0 differing pixels on my M4 MacBook.

Apps/Playground/Scripts/validation_native.js
matthargett: Updated. A pre-ready render exception is now logged and counted without stopping readiness polling. The pump retains the first and last errors (including stacks), reports a recovery count if readiness succeeds, and includes the retained diagnostics if readiness times out. After readiness, the normal validation render remains a hard failure. debugging silent hangs/timeouts due to unpropagated crashes was a big time sink for me on this, so this seems like a good balance.

matthargett:
the Android CI failure is an intermittent toolchain download problem that's unrelated to my changes

bkaradzic-microsoft (approved):

bkaradzic-microsoft:
Actually, we're considering different approach, not to diverge from B.js test harness...

bkaradzic-microsoft:
On the latest master, the readiness render pump is unnecessary and redundant.

Can you explain exactly condition when you need this?

matthargett:
> On the latest master, the readiness render pump is unnecessary and redundant.
> 
> Can you explain exactly condition when you need this?

let me retest with all my local enabled things and I'll report back.

matthargett:
Retested against current BabylonNative master and Babylon.js master, including the merged BabylonJS/Babylon.js#18709 fix.

I can no longer demonstrate a condition that requires the readiness render pump. Without it, Gaussian Splatting Loading, both serializer variants, the original GUI Slate playground, Near Menu, SpherePanel, and standard texture repetition all reached readiness and produced the expected comparisons. Enabling the pump added four pre-validation renders to Gaussian Splatting Loading but did not change its 9-pixel result. The remaining depth-prepass and VLS failures are independent shader-path issues and are not resolved by the pump.

The instability that originally made the pump appear useful was a separate Native renderer screenshot-capture issue: the configured comparison frame could continue advancing while asynchronous screenshot readback was pending. Preserving the comparison frame directly stabilized the animated serializer tests without speculative readiness rendering.

Closing this PR rather than adding redundant test-harness behavior. The unrelated Apple SDK build fix has been split into BabylonJS/BabylonNative#1783.

PR #1763, opened 2026-06-17
Re-enable blur-cube-with-the-effect-renderer validation test

bkaradzic-microsoft:
## What

Re-enables the `blur-cube-with-the-effect-renderer` Playground validation test (`#4C900K#2`) by
dropping its `excludeFromAutomaticTesting` entry in `config.json`, and bumps the bundled `babylonjs`
to **9.15.0** (which contains the paired engine fix) so the test actually passes in CI.

Two commits:
1. Re-enable `blur-cube-with-the-effect-renderer` in `Apps/Playground/Scripts/config.json`.
2. Bump `babylonjs` to 9.15.0 in `Apps/package.json` + `Apps/package-lock.json`.

Broken out of https://github.com/BabylonJS/BabylonNative/pull/1748 so it can merge independently:
#1748 re-enables this same test alongside six single-file-`.dds` cube tests, but those six are still
gated on an unmerged cube-loading TS change (BabylonJS/Babylon.js#18567), whereas this test only
needs the already-merged depth-test fix. (#1748 currently still carries a duplicate `blur-cube`
re-enable; that entry should be dropped there once this lands.)

## Why it passes now

The test previously failed pixel comparison because the native engine ignored
`depthCullingState.depthTest`, so an `EffectRenderer` fullscreen pass produced an all-black frame.
That is fixed by https://github.com/BabylonJS/Babylon.js/pull/18558 ("[Native] Honor
depthCullingState.depthTest on the native engine"), **merged 2026-06-16 and shipped in `babylonjs`
9.15.0** — the version this PR now pins. With that fix present in the bundled engine the pass renders
correctly and the test matches its reference image.

## Status

- **Pin bump included; the earlier gate is resolved.** The bump to 9.15.0 (which contains #18558) is
  part of this PR, so the re-enable no longer depends on a separate future `babylonjs` bump.
- **CI is fully green**, including every Win32 D3D11 and Ubuntu desktop job that runs the validation
  suite — `blur-cube-with-the-effect-renderer` validates. Ready to merge (pending review).

### DIFF
--- a/Apps/Playground/Scripts/config.json
+++ b/Apps/Playground/Scripts/config.json
@@ -2575,8 +2575,6 @@
             "title": "blur-cube-with-the-effect-renderer",
             "playgroundId": "#4C900K#2",
             "renderCount": 20,
-            "excludeFromAutomaticTesting": true,
-            "reason": "Pixel comparison fails (more than 20% pixels differ)",
             "referenceImage": "blur-cube-with-the-effect-renderer.png"
         },
         {
--- a/Apps/package-lock.json
+++ b/Apps/package-lock.json
@@ -11,13 +11,13 @@
         "UnitTests/JavaScript"
       ],
       "dependencies": {
-        "babylonjs": "^9.3.4",
-        "babylonjs-addons": "^9.3.4",
-        "babylonjs-gltf2interface": "^9.3.4",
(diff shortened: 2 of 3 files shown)


bghgary (approved):

LGTM

PR #1764, opened 2026-07-03
QuickJS

CedricGuillemet:
- Validation native script modified to lower pressure on the stack. Overflow happened with QuickJS because of larger than other engine stack entries.

### DIFF
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -371,21 +371,36 @@
                                 }
                             }
 
-                            currentScene = eval(code + "\r\ncreateScene(engine)");
+                            const pgCode = code + "\r\ncreateScene(engine)";
+                            // Defer scene construction to a fresh macrotask so
+                            // eval()/createScene() run at a shallow native-stack
+                            // depth instead of nested inside the native snippet
+                            // load callback. Deep scenes otherwise pile onto the
+                            // native XHR dispatch frames and can overflow engines
+                            // with a small C stack (e.g. QuickJS).
+                            setTimeout(function () {
+                                try {
+                                    currentScene = eval(pgCode);
 
-                            if (currentScene.then) {
-                                // Handle if createScene returns a promise
-                                currentScene.then(function (scene) {
-                                    currentScene = scene;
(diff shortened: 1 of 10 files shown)


Apps/Playground/Scripts/validation_native.js
```diff
@@ -371,21 +371,36 @@
...
-                                    processCurrentScene(test, referenceImage, done, compareFunction);
-                                }).catch(function (e) {
-                                    console.error(e);
+                            const pgCode = code + "\r\ncreateScene(engine)";
+                            // Defer scene construction to a fresh macrotask so
+                            // eval()/createScene() run at a shallow native-stack
+                            // depth instead of nested inside the native snippet
+                            // load callback. Deep scenes otherwise pile onto the
+                            // native XHR dispatch frames and can overflow engines
+                            // with a small C stack (e.g. QuickJS).
```
ryantrem, line 380: Wouldn't Babylon users hit the same problem? This isn't unique to validation tests, right?

CMakeLists.txt
```diff
@@ -53,8 +53,8 @@ FetchContent_Declare(ios-cmake
     GIT_TAG 4.5.0
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(JsRuntimeHost
-    GIT_REPOSITORY https://github.com/BabylonJS/JsRuntimeHost.git
-    GIT_TAG f07d99119887131a89a3580ada9c2a8dbc7782f7)
+    GIT_REPOSITORY https://github.com/CedricGuillemet/JsRuntimeHost.git
```
ryantrem: Just a reminder to switch this before merge.

Apps/Playground/Scripts/validation_native.js
CedricGuillemet: yes, it's a stack overflow issue. This should happen with other JS engines as well. debug/release build also have influence on the number of stack element count. QuickJS stack elements are bigger so the overflow happens sooner.

CMakeLists.txt
CedricGuillemet: fixed

ryantrem (approved):

Issue #1767, opened 2026-07-07
iOS: SIGSEGV entering WebXR/AR session — GetActiveEncoder() null in NativeXr::BeginUpdate after threading rework (#1652)


julapy:
### Summary

On iOS, entering a WebXR/AR session crashes with `EXC_BAD_ACCESS (SIGSEGV)` inside `NativeXr::Impl::BeginUpdate`. The framebuffer-creation continuation dereferences `GraphicsContext.GetActiveEncoder()`, which returns `nullptr` because it runs on the runtime thread with no frame in flight. This reproduces in the stock **iOS Playground** and appears to be a regression from the threading rework (#1652).

### Environment

- Babylon Native `master` @ `64f545de` (also present on every commit since #1652)
- iOS 26.5, iPhone 15 Pro Max (device — ARKit), Xcode 26
- Repro app: `Apps/Playground` (iOS), built via `cmake -B build -G Xcode -D IOS=ON`

### Repro

1. In `Apps/Playground/Scripts/experience.js`, set `const ar = true;`
2. Build & run the **Playground** target on a physical iOS device.
3. On session entry (the `enterXRAsync("immersive-ar", ...)` path, ~5s after load) the app crashes immediately.

### Crash

```
Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x00000000000002d9

Thread (crashed):
  bgfx::EncoderImpl::discard(unsigned char)
  bgfx::Encoder::touch(unsigned short)
  Babylon::Graphics::FrameBuffer::Clear(bgfx::Encoder&, unsigned short, unsigned int, float, unsigned char)
  Babylon::Plugins::NativeXr::Impl::BeginUpdate()::$_0::operator()() const
  ... (arcana task machinery, on Babylon::JsRuntimeScheduler / runtime thread)
```

### Root cause

In `Plugins/NativeXr/Source/NativeXrImpl.cpp`, `BeginUpdate()` creates the per-view framebuffers inside a `.then(m_runtimeScheduler, ...)` continuation and then does the WebXR implicit clear:

```cpp
// NativeXrImpl.cpp (~line 310)
frameBuffer.Clear(*m_sessionState->GraphicsContext.GetActiveEncoder(),
                  BGFX_CLEAR_COLOR | BGFX_CLEAR_DEPTH | BGFX_CLEAR_STENCIL, 0, 1.0f, 0);
```

[...]

PR #1769, opened 2026-07-07
Implement WebXR Raw Camera Access for the ARKit backend

julapy:
### What

Implements the [WebXR Raw Camera Access module](https://immersive-web.github.io/raw-camera-access/) for the ARKit backend, so Babylon.js content can sample the live AR camera image as a texture inside an `immersive-ar` session:

```js
const cameraAccess = xr.baseExperience.featuresManager.enableFeature(BABYLON.WebXRFeatureName.RAW_CAMERA_ACCESS, "latest");
cameraAccess.onTexturesUpdatedObservable.add((textures) => {
    material.emissiveTexture = textures[0]; // live ARKit camera image
});
```

Babylon.js's existing `WebXRRawCameraAccess` feature works against this unmodified — this PR only adds the native hooks it calls (`XRView.camera`, `XRWebGLBinding.getCameraImage`).

### How

**Commit 1 — xr dependency (ARKit backend):**
- Adds `CameraTexturePointer` / `CameraTextureSize` to `System::Session::Frame::View` (null on backends that don't populate them).
- ARKit already converts the camera image YUV→RGB each frame, but only pre-composites it into the color texture — which is the scene render target and can't double as a sampled camera image. This adds a dedicated BGRA8 texture (recreated on viewport size change, released through the same deleted-texture callback as the color/depth textures) and a second camera pass into it, reusing the shared shader via `screenPipelineState` with the `babylonTexture` slot left unbound so the shader takes the camera branch.

**Commit 2 — NativeXr plugin (JS surface):**
- `XRCamera` (new): exposes `width`/`height` and wraps the platform camera texture for bgfx using the same `createTexture2D` + `overrideInternal`-on-render-thread retry pattern as NativeCamera's `CameraDevice`.
- `XRView.camera`: returns the per-view `XRCamera` (`undefined` when the backend provides no camera texture).
- `XRWebGLBinding.getCameraImage(camera)`: returns the camera image as a `Napi::Pointer<Graphics::Texture>` — the object type NativeEngine uses as a hardware texture's `underlyingResource` — so `WebXRRawCameraAccess` can wrap and sample it directly.

The camera texture is viewport-sized and display-aspect-mapped (same image as the passthrough background), which keeps `view.projectionMatrix`-based intrinsics consistent with what Babylon's `_updateCameraIntrinsics` computes.

### Testing

Verified on device (iPhone 15 Pro Max, iOS 26.5, Xcode 26, Babylon.js 9.9.1 UMD) via the iOS Playground with `ar = true` and the feature enabled: the feature attaches, `onTexturesUpdatedObservable` fires with a 1290×2796 texture, and the live camera image renders on scene geometry inside the AR session at a sustained 60 FPS (camera pass adds well under 1 ms GPU time).

Note: testing `immersive-ar` on current master first requires a fix for #1767 (NativeXr crashes on session start due to a null active encoder after the threading rework); I applied the guard suggested there while validating this PR. This PR is independent of that fix.

Non-ARKit backends are unaffected (`CameraTexturePointer` stays null → `XRView.camera` is `undefined`, which is the same behavior as before this PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dependencies/xr/Include/XR.h
+++ b/Dependencies/xr/Include/XR.h
@@ -248,6 +248,12 @@ namespace xr
                     void* DepthTexturePointer{};
                     Size DepthTextureSize;
 
+                    // WebXR raw camera access: the platform camera image for this
+                    // view, rendered into an RGBA texture each frame. Null when unavailable.
+                    // Currently only populated by the ARKit backend.
+                    void* CameraTexturePointer{};
+                    Size CameraTextureSize;
+
                     float DepthNearZ{};
                     float DepthFarZ{};
 
--- a/Plugins/NativeXr/Source/XRCamera.h
+++ b/Plugins/NativeXr/Source/XRCamera.h
@@ -1,0 +1,146 @@
+#pragma once
+
+// WebXR raw camera access: JS XRCamera object, per the WebXR Raw Camera
+// Access module (https://immersive-web.github.io/raw-camera-access/). Exposed via
+// XRView.camera; the camera image is retrieved with XRWebGLBinding.getCameraImage,
+// which returns a Napi::Pointer<Graphics::Texture> — the same object type
+// NativeEngine uses as an InternalTexture's underlyingResource, so Babylon.js can
+// bind it as a regular sampler.
+//
+// The texture pointer originates from the xr backend (View::CameraTexturePointer, an
+// MTLTexture on ARKit) and is wired to bgfx with the same createTexture2D +
+// overrideInternal pattern used by NativeCamera's CameraDevice.
+
+#include <Babylon/Graphics/DeviceContext.h>
+#include <Babylon/Graphics/Texture.h>
+#include <napi/pointer.h>
+#include <arcana/threading/task.h>
+#include <bgfx/bgfx.h>
+
+#include <atomic>
+#include <memory>
+
+namespace Babylon
+{
+    class XRCamera : public Napi::ObjectWrap<XRCamera>
+    {
+        static constexpr auto JS_CLASS_NAME = "XRCamera";
+
+    public:
+        static void Initialize(Napi::Env env)
+        {
+            Napi::HandleScope scope{env};
+
+            Napi::Function func = DefineClass(
+                env,
+                JS_CLASS_NAME,
+                {
+                    InstanceAccessor("width", &XRCamera::GetWidth, nullptr),
+                    InstanceAccessor("height", &XRCamera::GetHeight, nullptr),
+                });
+
+            env.Global().Set(JS_CLASS_NAME, func);
+        }
+
+        static Napi::Object New(const Napi::CallbackInfo& info)
+        {
+            return info.Env().Global().Get(JS_CLASS_NAME).As<Napi::Function>().New({});
+        }
+
+        XRCamera(const Napi::CallbackInfo& info)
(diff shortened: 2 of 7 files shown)


julapy:
@microsoft-github-policy-service agree company="EyeJack"

Plugins/NativeXr/Source/XRCamera.h
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            // bgfx::overrideInternal only succeeds once the handle has been initialized by a
+            // bgfx frame, so run it on the render thread after rendering and retry until it
+            // sticks (same approach as NativeCamera's CameraDevice::UpdateCameraTexture).
+            if (!m_overrideState->done.load() && !m_overrideState->pending.exchange(true))
+            {
+                arcana::make_task(m_deviceContext->AfterRenderScheduler(), arcana::cancellation::none(),
+                    [state = m_overrideState, handle = m_bgfxHandle, pointer = m_texturePointer]() {
+                        state->done.store(bgfx::overrideInternal(handle, reinterpret_cast<uintptr_t>(pointer)) != 0);
```
bghgary, line 98: Please build this on the existing `Babylon::Plugins::ExternalTexture` plugin instead of hand-rolling `createTexture2D` + `overrideInternal`. It's the cross-platform wrapper for exposing a native texture to JS (`Update()` when the platform texture changes, `CreateForJavaScript()` for `getCameraImage`), and it doesn't use `overrideInternal`, which we're removing elsewhere.

`CameraTexturePointer` is produced on the graphics thread (in `UpdateXRView`), so the `ExternalTexture` can be wrapped and `Update()`'d there, leaving only `CreateForJavaScript` on the JS thread. That also removes the readiness gap where a texture is handed to JS before the override lands.

Plugins/NativeXr/Source/XRCamera.h
```diff
@@ -0,0 +1,146 @@
+#pragma once
+
+// WebXR raw camera access: JS XRCamera object, per the WebXR Raw Camera
```
bghgary, line 3: Please mark this feature experimental. WebXR Raw Camera Access is a non-stabilized module and this lands ARKit-only, so following the existing convention would set expectations: a `> ⚠️ experimental and subject to change` banner (see `Plugins/NativeEncoding/README.md`) and/or an `(experimental)` entry in the top-level README feature list.

Dependencies/xr/Include/XR.h
```diff
@@ -248,6 +248,12 @@ namespace xr
                     void* DepthTexturePointer{};
                     Size DepthTextureSize;
 
+                    // WebXR raw camera access: the platform camera image for this
+                    // view, rendered into an RGBA texture each frame. Null when unavailable.
+                    // Currently only populated by the ARKit backend.
```
bghgary, line 253: Non-blocking: in browsers this module only runs on Chrome/Android (ARCore), since iOS Safari has no WebXR AR at all, so Android is where raw camera access is actually exercised on the web. Is ARCore support planned as a follow-up? Landing iOS-first (as experimental) is fine; mainly want the Android gap acknowledged and tracked rather than silently a no-op.

bghgary (changes requested):

Thanks for the contribution. A few concerns inline, one blocking.

PR #207, opened 2020-04-07
Build system and component architecture documentation

syntheticmagus:

PR #1781, opened 2026-07-16
NativeDawn plugin

CedricGuillemet:


matthargett:
can we coordinate a little more on these kinds of changes? I spent months on the wgpu-native backing for this integration, and you all said you wouldn't be switching off of bgfx. I feel like I'm being as transparent and inclusive as I can, but I'm getting a bit of whiplash with this sudden change in policy (even if it favors my goals).

CedricGuillemet:
> can we coordinate a little more on these kinds of changes? I spent months on the wgpu-native backing for this integration, and you all said you wouldn't be switching off of bgfx. I feel like I'm being as transparent and inclusive as I can, but I'm getting a bit of whiplash with this sudden change in policy (even if it favors my goals).

This is only an experiment. Taking a closer look at https://github.com/BabylonJS/BabylonNative/pull/1605 is also on my todolist. We are also investing time on CI testing and harness. No need to worry , we won't forget you :)

bghgary:
We are still in the middle of planning with some prototyping/experiments. See here for some idea of what we are up to: https://github.com/BabylonJS/Babylon-Lite/issues/10.

CedricGuillemet:
## Babylon-Lite bundles on NativeDawn/WebGPU

Ran all **216** Babylon-Lite scene bundles through the Playground on the NativeDawn/WebGPU backend. **Tested with the V8 JS engine** (the WebGPU bootstrap is V8-only; ChakraCore fails).

**~124 scenes render correctly** — PBR/glTF, IBL/HDR, shadows (incl. CSM), skinning/morph/VAT, most NME materials, frame-graph post-processes (DoF/TAA/bloom), gizmos, geospatial camera, `.babylon` (Sponza), grid material, floating-origin, etc.

**~92 don't work**, grouped by reason:

| Reason | Scenes |
|---|---|
| WASM not fetchable — Havok physics | 40-42,44-49,100-106,209 |
| WASM — Recast navigation | 170,172,173 |
| WASM — manifold CSG | 91 |
| WASM — basis/KTX2, meshopt, text-shaper (silent black) | 36,112,211,180,181 |
| Web Worker (Blob URL) — Gaussian Splatting, gizmos | 120-129,222 |
| `Blob.stream`/ReadableStream missing | 66,72,140 |
| Canvas 2D API incomplete (sprites, DynamicTexture) | 18,24,25,50-57,90,92-98,117,118,145,146,179,205,206 |
| Multi-canvas (`getContext` null) | 227,228 |
| Missing local lab assets (fetch) — *fixable, just copy assets* | 58,59,73,81,83,160,171,174,175,210,254,255,257,258,259,260,266 |
| Silent black, not yet isolated (still black post texture-fix — unrelated cause) | 12,30,63,149,153,263,264 |

Physics/nav/gaussian-splatting fail as expected (no web workers + feature `.wasm` can't be fetched). The biggest *fixable* buckets are the Canvas-2D API and the missing-local-assets.

### How to test a lite bundle

1. In the Babylon-Lite repo, build the lib and bundle a scene as a single self-contained IIFE classic script (the default `build:bundle-scenes` output is ESM+code-split, which the Playground's classic `LoadScript` can't load):
   ```
   pnpm build:lib
   # esbuild: bundle:true, splitting:false, format:"iife",
   #   alias babylon-lite -> packages/babylon-lite/build/lib
   # -> lab/public/bundle/sceneN.playground.js
   ```
2. Copy `sceneN.playground.js` into the Playground `Scripts/` dir.
3. Run it with the **V8** build:
   ```
   Playground.exe --headless app:///Scripts/sceneN.playground.js
   ```
   (drop `--headless` to see the window). Any root-relative asset (`/foo.png`) is redirected to `app:///Scripts/foo.png`, so place sibling assets there.

### How to test the glTF texture fix

```
# the exact bug case (renders textured, not black):
Playground.exe --headless --test "BoomBox with Unlit" app:///Scripts/validation_native.js
# -> "No pixel difference!"  /  "GLTF BoomBox with Unlit Material validated"

# full validation suite:
Playground.exe --headless app:///Scripts/validation_native.js
# -> Run complete. ran=259 passed=259 failed=0
```

laurentdiazfr:
@CedricGuillemet What are your thoughts on https://github.com/wcandillon/react-native-webgpu as a Dawn backend? Would being tied to React Native be a limitation for what you're trying to achieve? React native and Expo excosystem is huge nowadays, it will increase the adoption of babylon native.

CedricGuillemet:
@laurentdiazfr  https://github.com/wcandillon/react-native-webgpu looks great and being able to use Lite directly is awesome. BabylonNative corresponds to different needs (same for BRN). For example, usage of XR or tight integration of native libraries with Babylon. The more possibilities, the better, I think. For example, fills the gap when a user wants to integrate 3D in a native android app.

matthargett:
FYI, expanded my wgpu branch suite based on the report above. A few fixes is that branch, but there's several more WHATWG/W3C API compliance issues that need fixing to expand it to the fuller screenshot suite on browsers. 

I submitted those PRs into JsRuntimeHost, and validated them locally on macOS 26, iOS 18, and Android 12 (Pico XR). I tested across JSC and QuickJS in that matrix, but ran out of time to go further today (I spot check a lot of the screenshots etc manually so I can dig deeper into pixel differences.) 

This has been a great driver to make BabylonNative competitive as a 3D-forward app runtime that stands shoulder to shoulder with other peers like deno, bun, NodeJS, and electron! Very exciting!

CedricGuillemet:
## NativeDawn follow-up: workarounds and shims to replace

This is a cleanup checklist, not a list of merge blockers. The current implementations are useful for proving and testing the backend, but these should be replaced with explicit APIs or fixes in the appropriate layer.

### Runtime/bootstrap

- [ ] **Replace the `BABYLON.Tools.LoadScript` monkey patch with explicit native-host integration** (`Plugins/NativeDawn/Source/NativeDawn.cpp:4242-4265`). Babylon.js chooses `_LoadScriptWeb` because `_native` is absent, then waits forever for a synthetic `<script>.onload`. NativeDawn currently overrides `Tools.LoadScript` with `LoadFile + Function(...)`. The proper fix is for Babylon.js/runtime bootstrap to select or inject the native script loader through a supported capability/API rather than detect native execution from the presence of `_native`. This likely needs a Babylon.js change.

- [ ] **Replace the `BABYLON.NativeEngine` alias and private bootstrap globals** (`NativeDawn.cpp:4226-4347`). NativeDawn pre-creates a `WebGPUEngine`, stores it in `__dawnPendingEngine`/`__dawnEngine`, and replaces `BABYLON.NativeEngine` with a function returning that singleton. A proper backend/bootstrap contract should expose the selected engine and its async readiness without overriding an unrelated constructor.

- [ ] **Expose a supported backend identifier** instead of detecting NativeDawn through `_nativeDawnClear` (`validation_native.js:116-130`, `run_snippet.js:26-30`). `_nativeDawnClear` is an implementation/milestone hook, not a capability API. The runtime or `TestUtils` should expose the active graphics backend explicitly.

- [ ] **Replace magic JS globals used by the host frame/resize bridge** (`Embedding/Source/View.cpp:219-234,278-289`; `NativeDawn.cpp:4904+`). `frame` and `__dawnResize` are private string-based contracts. Prefer typed NativeDawn entry points or one documented runtime bridge object.

### JavaScript-host / Web platform gaps

- [ ] **Remove the synchronous `WebAssembly.instantiate` override** (`NativeDawn.cpp:4426-4443`). It exists because the V8 foreground task runner is not pumped, so async compilation promises never settle. The proper fix is to integrate/pump V8 foreground tasks; globally forcing all WASM compilation onto the JS thread changes semantics and can stall it.

- [ ] **Move UTF-16 support into the shared `TextDecoder` polyfill** (`NativeDawn.cpp:4394-4423`). The JS wrapper exists only because the host decoder supports UTF-8 while glslang/twgsl requests `utf-16le`. JsRuntimeHost/shared polyfills should implement the required encoding labels directly.

- [ ] **Consolidate the DOM/canvas shims into shared polyfills** (`NativeDawn.cpp:4446-4649`; validation fallback at `validation_native.js:772-802`). NativeDawn creates or monkey-patches `document`, `getElementById`, `querySelector`, `window`, canvas aliases, and cached `__dawnCanvas` behavior. These should live in one tested Window/DOM/Canvas compatibility layer rather than backend-specific patches layered over another document implementation.

- [ ] **Fix URL/Blob/object-URL behavior in the shared host polyfills** (`NativeDawn.cpp:4675-4714,4736-4781`). NativeDawn wraps `Blob` to stash bytes, converts object URLs to data URLs, and wraps `URL` to coerce URL-object arguments because the current host APIs are not browser-compatible. Those fixes belong in JsRuntimeHost/common URL, Blob, XHR and image-loading implementations.

- [ ] **Replace the global root-relative `fetch` rewrite with an application asset resolver** (`NativeDawn.cpp:4717-4733`). Rewriting `/asset` to `app:///Scripts/asset` is specific to bundled Babylon-Lite lab scenes and should not alter `fetch` semantics for every NativeDawn application.

- [ ] **Move `Image` / `createImageBitmap` compatibility into a shared image polyfill** (`NativeDawn.cpp:4696-4902`). Native image decoding is required functionality, but implementing the browser API surface inside the graphics backend makes it unavailable or divergent for other backends.

### Dawn/platform workaround

- [ ] **Remove the WARP MSAA lazy-clear workaround after the Dawn issue is fixed** (`NativeDawn.cpp:359-422`). The code enables `nonzero_clear_resources_on_creation_for_testing` and detects WARP using adapter type, Microsoft vendor ID and device-name substrings. This should be tracked upstream in Dawn; string/vendor sniffing and a testing-only toggle should not be permanent.

### Test and CI accommodations

- [ ] **Restore actual NativeDawn pixel validation in CI** (`.github/workflows/build-win32.yml:173-184`). The full suite currently runs with `--generate-references` because WARP output differs from hardware-generated references, so it gates load/render/crash behavior but not image regressions. Use WARP-specific references or a stable hardware runner; keep Babylon-Lite's WARP pixel test as the current narrow guard until then.

- [ ] **Pin the Babylon-Lite revision used by CI** (`build-win32.yml:147-160`). Cloning the default branch tip makes this PR's CI depend on unrelated future Babylon-Lite changes. Pin a commit/tag and update it deliberately.

- [ ] **Fix shader-transpiler readiness upstream instead of pre-warming it** (`validation_native.js:907-948`). The first GLSL effect has order-dependent `isReady()` behavior until glslang/twgsl is loaded. The proper Babylon.js fix should make readiness deterministic; then remove the harness pre-warm.

- [ ] **Replace the growing per-test global reset list with real test isolation** (`validation_native.js:658-727`). The harness manually restores `Math.random`, loader observers, reverse depth, snapshot rendering and every `ForceGLSL` static because playgrounds leak process-global state. Prefer a fresh engine/context per test or a supported Babylon.js reset API.

- [ ] **Turn WebGPU exclusions and widened tolerances into tracked issues** (`Apps/Playground/Scripts/config.json`). The reasons document real gaps well (unsupported LINE_LOOP/TRIANGLE_FAN conversion, decoder/runtime issues, platform raster differences), but issue links/owners are needed so exclusions do not become permanent silently.

### Not included in this debt list

The WebGPU constant globals, `OffscreenCanvas`, `Image`, `createImageBitmap`, requestAnimationFrame, backend-specific CMake selection, and platform guards are not inherently hacks: a native WebGPU host needs those capabilities. The debt is where they are implemented today (backend-local patches/private globals), incomplete standards behavior, or temporary CI bypasses—not the existence of the APIs themselves.

bkaradzic-microsoft:
Good list. Four of these are already root-caused and fixed on my `shotgun` branch — the fixes are in JsRuntimeHost, which is where you're asking for them to live, so they turn into deletions here rather than rewrites.

### Already fixed in JsRuntimeHost (not yet PR'd — on [`bkaradzic-microsoft/JsRuntimeHost@shotgun`](https://github.com/bkaradzic-microsoft/JsRuntimeHost/commit/8561861))

**Remove the synchronous `WebAssembly.instantiate` override** (`NativeDawn.cpp:4426-4443`) — your diagnosis is exactly right. `AppRuntime_V8` creates a default `v8::Platform` but nothing ever called `PumpMessageLoop`, so V8 finished the compile on a background thread, posted a foreground task to resolve the promise, and that task was never run. (Sync `new WebAssembly.Module` was unaffected, which made it look like a hang rather than a failure. `AppRuntime_JSI` already did the equivalent via `TaskRunnerAdapter`; the direct-V8 path was simply missing it.) `DrainMicrotasks` now pumps the queue, and the platform is wrapped in a `DispatchingPlatform` whose foreground task runner nudges the app dispatcher so a pump also happens when the app is otherwise idle. The override can just be deleted.

**Move UTF-16 support into the shared `TextDecoder` polyfill** (`NativeDawn.cpp:4394-4423`) — done in `Polyfills/TextDecoder`: `utf-16le`/`utf-16be` including the leading-BOM strip. The JS wrapper can go.

Same commit also fixes a third one you'd have hit next: `XMLHttpRequest.responseText`/`response` truncated at the first NUL byte. `ResponseString()` returns a `string_view` and only `data()` was handed to Napi, so it was measured with `strlen` — Emscripten inlines the `.wasm` payload as a JS string literal, so `spz.js` arrived as 1955 of 805507 characters.

Together those three are what let me re-enable validation tests 100, 104, 157 and 188 (Gaussian Splatting SPZ, CSG2/manifold), which were excluded precisely because loading a script as an ES module didn't work on Native.

### Already fixed on `master`'s current JsRuntimeHost pin

**The Blob stash / object-URL → `data:` URL conversion** (`NativeDawn.cpp:4675-4714`) is redundant as of JsRuntimeHost #207, which is in `cb988ba`. The URL polyfill now registers a `blob:` scheme resolver with UrlLib, so `fetch`, `XMLHttpRequest` and any other UrlLib consumer resolve `blob:` URLs directly — including the `forceBitmapOverHTMLImageElement` XHR path that motivated the hack. `createObjectURL`/`revokeObjectURL` are backed by a real store, and the shim here is already guarded by `if (!url.Get("createObjectURL").IsFunction())`.

**The `URL` coercion wrapper** (`NativeDawn.cpp:4736-4781`) is still needed though — `URL::URL` does `info[0].As<Napi::String>()` with no `USVString` conversion, so a `URL` object as `url` or `base` throws. That's a genuine small JsRuntimeHost fix.

### Agreed, not fixed

The DOM/canvas/`Image`/`createImageBitmap` consolidation — my Batches 77/78 add `Image`, `PointerEvent`, `screen`, DOM event dispatch and `KeyboardEvent` to `validation_native.js`, which is the same debt you're flagging at `validation_native.js:772-802`, not a resolution. Worth doing once in a shared Window/DOM/Canvas layer rather than twice badly.

Happy to split the JsRuntimeHost commit into PRs (V8 pumping / TextDecoder / XHR NUL are independent) so this PR can bump the pin and drop the three shims.

PR #1788, opened 2026-07-23
Resolve #1786 review follow-ups: drop BN workarounds, bump bgfx.cmake, guard useCache

SergioRZMasson:
## Summary

Addresses the three review follow-ups from #1786 ([bghgary's notes](https://github.com/BabylonJS/BabylonNative/pull/1786)), moving each fix to its proper home so BN no longer carries the workarounds.

### 1. Drop the manual `-msse4.2` flag (`Dependencies/CMakeLists.txt`)

The SSE4.2 minspec that bx's x86_64 SIMD path requires now lives in **bgfx.cmake** (BabylonJS/bgfx.cmake#130), applied `PUBLIC` on the `bx` target — including the `-Xarch_x86_64` scoping needed for Apple universal (arm64;x86_64) builds. BN no longer needs to re-add it.

### 2. Drop the per-plugin `_WIN32_WINNT` / `WINVER` pin (`Plugins/NativeEngine/CMakeLists.txt`)

The root cause was fixed in **bx** (bkaradzic/bx#409): `platform.h` now detects MSVC ARM64 as 64-bit and defaults every 32-bit Windows target to `0x0601` instead of the pre-Vista `0x0502`. With that, arcana's ETW `trace_region.h` compiles on Windows x86/ARM64 without pinning an API baseline on the NativeEngine target.

### 3. Guard `useCache` with `#ifdef SHADER_CACHE` (`Plugins/NativeEngine/Source/ShaderProvider.cpp`)

`useCache` is referenced only inside the `#ifdef SHADER_CACHE` blocks, so its declaration moves into that guard (rather than `[[maybe_unused]]`): the variable — and `instancedAttributes.empty()` — is compiled out entirely when the ShaderCache plugin is disabled.

To consume 1 and 2, this bumps the `bgfx.cmake` pin to `8f7e4455` (which carries the SSE minspec on the bx target and advances bx to `cd6720ce` with the platform.h fixes).

## Testing

- Built NativeEngine for **Win32** against the bumped pin with the `_WIN32_WINNT` workaround removed — `NativeEngine.cpp` / `Program.cpp` / `ShaderProvider.cpp` compile and `NativeEngine.lib` links (previously this arch failed with `error C3861` in `TraceLoggingProvider.h`).
- The SSE minspec on the bx target was verified in bgfx.cmake#130 (compiles the `_mm_round_ps` SSE4.1 intrinsic on Clang/x86_64, `-Xarch_x86_64`-scoped on Apple).
- `ShaderProvider` compiles clean under Clang `-Werror -Wunused-variable` with ShaderCache enabled and disabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ FetchContent_Declare(base-n
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(bgfx.cmake
     GIT_REPOSITORY https://github.com/BabylonJS/bgfx.cmake.git
-    GIT_TAG 3b67f0c0c5505a22f6e88640519b5b473c938ad9
+    GIT_TAG 8f7e4455933085fc719913138c78ef61242709e3
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(CMakeExtensions
     GIT_REPOSITORY https://github.com/BabylonJS/CMakeExtensions.git
--- a/Dependencies/CMakeLists.txt
+++ b/Dependencies/CMakeLists.txt
@@ -115,21 +115,6 @@ if(NOT BABYLON_NATIVE_PLUGIN_NATIVEENGINE_WEBP AND TARGET bimg_decode)
     target_compile_definitions(bimg_decode PRIVATE BIMG_CONFIG_PARSE_WEBP=0)
 endif()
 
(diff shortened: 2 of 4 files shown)


bghgary (approved):
LGTM

PR #1795, opened 2026-07-27
Updated bgfx.cmake to latest.

bkaradzic-microsoft:


bkaradzic-microsoft:
CI status on the latest bgfx.cmake bump: everything green except `iOS_Xcode26` and `iOS_Installation`, which both fail with the same single error:

```
clang: error: argument unused during compilation: '-Xarch_x86_64 -msse4.2'
       [-Werror,-Wunused-command-line-argument]
```

This is a gap in the arch-aware SSE minspec that landed in BabylonJS/bgfx.cmake#133. That change uses `-Xarch_x86_64` for universal Apple builds so the flag only reaches the x86_64 slice, which is correct when `CMAKE_OSX_ARCHITECTURES` describes the slices actually compiled -- true for macOS universal (those jobs pass), but not for Xcode in general.

Our iOS configure goes through leetal/ios-cmake with `PLATFORM=OS64COMBINED`, which reports `architecture(s): arm64;x86_64`. Xcode picks slices per SDK at build time, and the `Release-iphoneos` build compiles arm64 only (x86_64 is the *simulator* slice, built separately). So the `-Xarch_x86_64` applies to nothing, and `-Werror` turns the resulting unused-argument warning into a build failure.

Fix is up as BabylonJS/bgfx.cmake#134: pair the flag with `-Wno-unused-command-line-argument`, scoped to the universal-build branch only, so every x86_64 slice that is really compiled still gets the SSE4.2 minspec. I'll bump the `GIT_TAG` here once that merges.

bkaradzic-microsoft:
Update: the run finished at **31/33 green** -- `iOS_Xcode26` and `iOS_Installation` are the only failures, and both are the single `-Xarch_x86_64` error described above. Everything else, including all the macOS, UWP arm64, Linux, Android and Win32 jobs, passed on the new bgfx.cmake tag.

I've verified the fix (BabylonJS/bgfx.cmake#134) end-to-end rather than guessing: I pointed a throwaway branch's `GIT_TAG` at that commit and re-ran just these two jobs on the same runner image and Xcode.

| Job | bgfx.cmake `7062f11` | bgfx.cmake `d8e3ba1` (#134) |
| --- | --- | --- |
| `iOS_Xcode26 / build` | ❌ exit 65 | ✅ 4m4s |
| `iOS_Installation / test-install` | ❌ exit 65 | ✅ 3m14s |

So #134 is the only thing standing between this PR and a fully green run. Once it merges I'll bump the `GIT_TAG` here and re-run.

bghgary (approved):

LGTM

bkaradzic-microsoft:
Green across the board -- **33/33** on `6c55158` (BabylonJS/bgfx.cmake#134). `iOS_Xcode26` and `iOS_Installation`, the only two failures on the previous tag, both pass now, and nothing else regressed.

PR #1796, opened 2026-07-28
ShaderCompiler: preserve original sampler names colliding with HLSL/MSL keywords

bkaradzic-microsoft:
### Problem

SPIRV-Cross's HLSL and MSL backends rename any shader resource whose name collides with a reserved keyword of the target language, prepending an underscore. Babylon.js generates GLSL samplers named `Texture2D` and `Texture2DArray`, which are HLSL built-in object types, so they are emitted into the bgfx uniform table as `_Texture2D` / `_Texture2DArray`.

Both bgfx and Babylon.js look samplers up by their **original GLSL name**, so the renamed uniform never binds. There is no error — the stage just silently samples nothing and renders transparent.

### Fix

`AppendSamplers` recovers the pre-transpile name from the parser's `ParsedIR` via `get_name(id)`. The `Compiler` transpiles a *private copy* of the IR, so the parser still holds the original names. That name is emitted into the bgfx uniform table and the stage map.

- The texture register (`DecorationBinding`) is untouched — only the app-facing name changes.
- Non-colliding samplers resolve to the identical string, so they are bit-for-bit unaffected.
- The existing OpenGL stage-assignment guard is now keyed on the recovered name so the vertex and fragment passes agree on the same identifier.

### Validation

Built `x64 / Release / D3D11` against this branch and ran the full Playground validation suite:

```
Run complete. ran=300 passed=300 failed=0 missingRef=0 skipped=420
```

No regressions. Because every non-colliding sampler keeps its current name, the change is a no-op for all 300 currently-enabled tests.

### Scope

This is intentionally code-only — no `config.json` changes. The test this unblocks (`MultiRenderTarget with different texture types`, `#XSNYAU#22`) also needs a matching Babylon.js-side fix for the mixed-type MRT write path, so its un-exclusion is deliberately left for a later change once both halves are in. Verified: with only this fix applied, that test still fails, so enabling it here would red the build.

### DIFF
--- a/Plugins/ShaderCompiler/Source/ShaderCompilerCommon.cpp
+++ b/Plugins/ShaderCompiler/Source/ShaderCompilerCommon.cpp
@@ -90,12 +90,21 @@ namespace Babylon::ShaderCompilerCommon
         }
     }
 
-    void AppendSamplers(std::vector<uint8_t>& bytes, const spirv_cross::Compiler& compiler, const spirv_cross::SmallVector<spirv_cross::Resource>& samplers, std::map<std::string, uint8_t>& stages)
+    void AppendSamplers(std::vector<uint8_t>& bytes, const spirv_cross::Compiler& compiler, const spirv_cross::ParsedIR& originalIr, const spirv_cross::SmallVector<spirv_cross::Resource>& samplers, std::map<std::string, uint8_t>& stages)
     {
         for (const spirv_cross::Resource& sampler : samplers)
         {
-            AppendBytes(bytes, static_cast<uint8_t>(sampler.name.size()));
-            AppendBytes(bytes, sampler.name);
+            // SPIRV-Cross's HLSL/MSL backends rename resources whose name collides with a reserved
+            // keyword of the target language (e.g. a GLSL sampler named "Texture2D" or "Texture2DArray"
+            // becomes "_Texture2D" because those are HLSL built-in object types). bgfx's uniform table
+            // and Babylon.js look samplers up by their original GLSL name, so the renamed identifier would
+            // never bind (the sampler silently samples nothing). Recover the pre-transpile name from the
+            // parser's ParsedIR (the Compiler transpiles a private copy, leaving the parser's names intact).
+            const std::string& originalName = originalIr.get_name(sampler.id);
+            const std::string& name = originalName.empty() ? sampler.name : originalName;
+
+            AppendBytes(bytes, static_cast<uint8_t>(name.size()));
+            AppendBytes(bytes, name);
             AppendBytes(bytes, static_cast<uint8_t>(bgfx::UniformType::Sampler | BGFX_UNIFORM_SAMPLERBIT));
 
             // TODO : These values (num, regIndex, regCount) are only used by Vulkan and should be set for that API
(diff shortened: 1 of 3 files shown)


Plugins/ShaderCompiler/Source/ShaderCompilerCommon.cpp
```diff
@@ -90,12 +90,21 @@ namespace Babylon::ShaderCompilerCommon
...
+            // keyword of the target language (e.g. a GLSL sampler named "Texture2D" or "Texture2DArray"
+            // becomes "_Texture2D" because those are HLSL built-in object types). bgfx's uniform table
+            // and Babylon.js look samplers up by their original GLSL name, so the renamed identifier would
+            // never bind (the sampler silently samples nothing). Recover the pre-transpile name from the
+            // parser's ParsedIR (the Compiler transpiles a private copy, leaving the parser's names intact).
+            const std::string& originalName = originalIr.get_name(sampler.id);
+            const std::string& name = originalName.empty() ? sampler.name : originalName;
+
+            AppendBytes(bytes, static_cast<uint8_t>(name.size()));
+            AppendBytes(bytes, name);
```
bghgary, line 107: [Reviewed by Copilot on behalf of @bghgary]

This changes the sampler name written into the bgfx blob and the key of `UniformStages`, and `ShaderCacheImpl::Save` serializes both. The cache is keyed on a hash of the GLSL source, which this does not change, and `CACHE_VERSION` in `Plugins/ShaderCache/Source/ShaderCacheImpl.cpp:38` is still `2`, so an existing `shaderCache.bin` keeps serving pre-fix entries with the renamed identifier and the bug persists with no sign of it. Bump it to `3`.

bghgary (changes requested):

One comment inline.

bkaradzic-microsoft:
Good catch, and thank you — you're right, and I'd missed it. Bumped to `3` in c52a0c1b.

I traced it to confirm the exact failure mode you describe:

- `ShaderCacheImpl::Hash(vertexSource, fragmentSource)` keys the cache purely on the GLSL source. This fix changes neither shader source, so a warm cache produces an identical key and hits.
- The *payload* is what changes. `Save` serializes `VertexBytes` / `FragmentBytes` — which embed the bgfx uniform table now carrying the original sampler name — and separately serializes `UniformStages`, whose keys are the sampler names this PR re-keys.
- `Load` returns `0` on a version mismatch, so `3` correctly discards a stale `shaderCache.bin` and forces a recompile.

Net effect without the bump: on any machine with an existing cache the fix would appear to do nothing, with no diagnostic — which is a worse failure than the original bug, since it looks like the fix simply doesn't work.

I added a short comment above the constant recording *why* 3 exists, so the next person changing the serialized shape has a precedent to follow.

Verified: `ShaderCache.SaveAndLoad` passes, and the validation suite is unchanged at `ran=300 passed=300 failed=0 missingRef=0 skipped=420`. The `Win32_x64_D3D11_PrecompiledShaderTest` job covers the ShaderTool -> ShaderCache -> runtime path end to end (that app is gated off in my local config, so I'm leaning on CI for it).

Plugins/ShaderCompiler/Source/ShaderCompilerCommon.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

Filed #1799 for the underlying problem: the cache key covers the shader source but not the toolchain that compiled it, so this is not the first change it has missed. #1795 moved the bgfx.cmake pin and merged without invalidating either. That issue proposes deriving the version from the dependency pins and compiler sources at configure time so it stops depending on the author noticing.

The manual bump here is still the right fix for this PR.

bghgary (approved):

LGTM

PR #1800, opened 2026-07-29
Canvas: implement getImageData and fix drawImage crash on plain ImageBitmap

bkaradzic-microsoft:
Two Canvas 2D fixes that together make the `drawImage` → `getImageData` round trip work, plus input validation for the new entry points.

## 1. `getImageData` returned all zeros

`ImageData::GetData` was a stub:

```cpp
// return a well size array with 0
// TODO: Get datas from context/canvas
```

It allocated a buffer and `memset` it to 0, so every `getImageData()` call read back a fully transparent image no matter what had been drawn. `Context::GetImageData` also discarded the source x/y arguments (`// TODO: support source x and y`).

BN's canvas is GPU-backed (nanovg), and a framebuffer readback is a poor fit for the data-texture callers that rely on `getImageData`: the surface is premultiplied and may be resampled, so the bytes that come back are not the bytes that went in. Instead this keeps a **CPU-side RGBA8 mirror** — `drawImage` copies the decoded image's exact pixels (`Image::GetPixels` → bimg `m_data`) into `Context::m_cpuPixels`, and `getImageData` reads that region back out. `BlitPixelsToCpu` covers all three `drawImage` arities with nearest-neighbour sampling, which is exact for the 1:1 draws these callers use. The mirror is reallocated and cleared on canvas resize.

Also fixes a pre-existing `ImageData` constructor bug that read the height from `info[1]` instead of `info[2]`, so a non-square `ImageData` reported its width as its height.

## 2. `drawImage` access violation on a plain `ImageBitmap`

Babylon Native sets `forceBitmapOverHTMLImageElement`, which routes `LoadImage` through `NativeEngine::CreateImageBitmap`. That returns a plain JS object `{data, width, height, format}` — **not** a wrapped `NativeCanvasImage`. `Context::DrawImage` called `NativeCanvasImage::Unwrap` on whatever it was handed, so `napi_unwrap` dereferenced a never-wrapped object and crashed with an access violation (0xC0000005).

`DrawImage` now detects the plain `ImageBitmap` by its own `data` typed array, converts its pixels to RGBA8 via `bimg::imageConvert` (memcpy fast path when the source is already RGBA8), creates a transient nanovg image, and shares the arity-3/5/9 draw plus CPU-mirror blit with the existing path through a new `DrawImageCommon` helper. The `NativeCanvasImage` path is behaviour-identical.

## 3. Input validation

Both new paths take sizes straight from JS and hand them to bimg and to raw pointer arithmetic, so the third commit validates them at the boundary:

- `drawImage` rejects an out-of-range bimg texture format, dimensions whose pixel count would overflow the destination allocation, and a `data` array shorter than `width*height` for its format. **`bimg::imageConvert` is given neither buffer's real length**, so without these it would read past the end of the source.
- `getImageData` requires its four arguments and rejects a region whose byte size would overflow `size_t` before `ImageData` tries to allocate it.
- `BlitPixelsToCpu` clamps its iteration range to the destination rect's intersection with the canvas up front. The destination size is caller-controlled and arrives unsigned, so a negative width wraps to ~4e9; the previous per-pixel bounds test still walked the whole rect and could stall the JS thread for billions of no-op iterations.

Range checks are computed in `uint64_t` so they stay meaningful where `size_t` is 32-bit (Android armeabi-v7a), and the new arithmetic uses fixed-width types rather than `long`, whose width differs between MSVC and Apple clang.

## Validation

- **"Displacement map" (idx 200)** — `Mesh.applyDisplacementMap` is exactly this `drawImage` → `getImageData` round trip. It crashes on master; with this change it renders at **0.155% pixel difference** against a 2.5% limit. It is still `excludeFromAutomaticTesting` in `config.json`, so reproduce with `--include-excluded --test-index=200`. Un-excluding it is deliberately left to a separate test-enablement PR.
- **Full validation suite: `ran=295 passed=295 failed=0`** — no regressions.

<details>
<summary>Why 295 and not 300</summary>

The 5 tests at indices 53–57 (`Scissor test`, `Scissor test with 0.9 hardware scaling`, `… with 1.5 hardware scaling`, `… with negative x and y`, `… with out of bounds width and height`) are excluded from that run because they **crash on unmodified master**, not because of anything here:

```
--- BN: CRASH ---
Unknown signal? Exception Code 87a
  7: bgfx/src/renderer_d3d11.cpp  6173  bgfx::d3d11::RendererContextD3D11::submit
  8: bgfx/src/bgfx.cpp            2917  bgfx::Context::renderFrame
 13: Core/Graphics/Source/DeviceImpl.cpp  539  Babylon::Graphics::DeviceImpl::Frame
```

Exit code 3, Win32 x64 D3D11 Debug. Verified pre-existing by reverting `Polyfills/Canvas` to master and rebuilding — identical crash at the same test. (`0x87a` is the DXGI `0x887A…` device-removed family.) The duplicate `Scissor test` at idx 311 passes. Worth noting because it truncates any full-suite run at 52 tests; happy to file it separately.
</details>

## Not included

My working branch also un-excluded idx 106 "Gaussian Splatting Part Test" alongside this change. That does **not** hold on stock upstream, so it is left out: Babylon's SOG loader unpacks the `.sog` container and materialises its WebP images through `URL.createObjectURL`, and upstream JsRuntimeHost has no blob-URL support, so the load fails before any canvas work happens. That un-exclusion needs a JsRuntimeHost change first.

There are **no `config.json` changes in this PR** — it is native-only and enables no tests.

### DIFF
--- a/Polyfills/Canvas/Source/Context.cpp
+++ b/Polyfills/Canvas/Source/Context.cpp
@@ -920,6 +920,15 @@ namespace Babylon::Polyfills::Internal
                 throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap dimensions are too large.");
             }
 
+            // imageGetSize takes uint16_t extents, so anything above 65535 would be truncated and
+            // produce a size for a much smaller image. A 65537x1 RGBA8 bitmap would come back as
+            // 1x1 (4 bytes), a 4-byte buffer would pass the check below, and the RGBA8 memcpy
+            // would then read width*height*4 bytes from it. Reject what bimg cannot describe.
+            if (width > std::numeric_limits<uint16_t>::max() || height > std::numeric_limits<uint16_t>::max())
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap dimensions exceed the maximum supported size.");
+            }
+
             // Ask bimg for the size rather than computing width*height*bpp/8: block-compressed
             // formats round up to whole blocks and have a minimum block count, so a 1x1 BC1 image
             // still occupies one 8-byte block. The naive bits-per-pixel math would accept a
@@ -1078,10 +1087,44 @@ namespace Babylon::Polyfills::Internal
             throw Napi::Error::New(info.Env(), "Context2D.getImageData: invalid number of parameters");
         }
 
-        const auto sx = info[0].As<Napi::Number>().Int32Value();
-        const auto sy = info[1].As<Napi::Number>().Int32Value();
-        const auto sw = info[2].As<Napi::Number>().Uint32Value();
-        const auto sh = info[3].As<Napi::Number>().Uint32Value();
+        auto sx = info[0].As<Napi::Number>().Int32Value();
+        auto sy = info[1].As<Napi::Number>().Int32Value();
+        const auto swInt = info[2].As<Napi::Number>().Int32Value();
+        const auto shInt = info[3].As<Napi::Number>().Int32Value();
+
+        // Parse the extents as signed. Read via Uint32Value, a width of -1 wraps to 4294967295,
+        // and the size_t guard below does not catch it on 64-bit because 4294967295 * 1 is far
+        // below SIZE_MAX/4, so ImageData would go on to allocate roughly 17 GB.
+        //
+        // A negative extent is legal: the browser treats it as a rectangle running in the
+        // opposite direction and returns the normalized region, so fold the sign into the
+        // origin rather than rejecting it. INT32_MIN has no positive counterpart, so it is
+        // rejected outright instead of being negated into itself.
+        if (swInt == std::numeric_limits<int32_t>::min() || shInt == std::numeric_limits<int32_t>::min())
+        {
+            throw Napi::RangeError::New(info.Env(), "Context2D.getImageData: requested region is too large.");
+        }
+
+        uint32_t sw, sh;
+        if (swInt < 0)
+        {
+            sw = static_cast<uint32_t>(-swInt);
+            sx -= static_cast<int32_t>(sw);
+        }
+        else
+        {
+            sw = static_cast<uint32_t>(swInt);
+        }
+
+        if (shInt < 0)
+        {
+            sh = static_cast<uint32_t>(-shInt);
+            sy -= static_cast<int32_t>(sh);
+        }
+        else
+        {
+            sh = static_cast<uint32_t>(shInt);
+        }
 
         // The region is caller-supplied and backs a width*height*4 allocation, so reject sizes
         // that would overflow size_t before ImageData tries to allocate them.
--- a/Polyfills/Canvas/Source/Context.cpp
+++ b/Polyfills/Canvas/Source/Context.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cstring>
+#include <limits>
 #include <optional>
 #include <regex>
 
@@ -778,18 +779,22 @@ namespace Babylon::Polyfills::Internal
         {
             m_cpuWidth = width;
             m_cpuHeight = height;
-            m_cpuPixels.assign(static_cast<size_t>(width) * height * 4, 0);
-        }
-    }
 
-    void Context::BlitImageToCpu(const NativeCanvasImage& image, int32_t sx, int32_t sy, uint32_t sw, uint32_t sh, int32_t dx, int32_t dy, uint32_t dw, uint32_t dh)
-    {
-        BlitPixelsToCpu(image.GetPixels(), image.GetWidth(), image.GetHeight(), sx, sy, sw, sh, dx, dy, dw, dh);
+            const uint64_t pixelCount = static_cast<uint64_t>(width) * height;
+            if (pixelCount > std::numeric_limits<size_t>::max() / 4)
+            {
+                // Leave the mirror empty; drawImage and getImageData both no-op safely on it.
+                m_cpuPixels.clear();
+                return;
+            }
+
+            m_cpuPixels.assign(static_cast<size_t>(pixelCount) * 4, 0);
+        }
     }
 
     void Context::BlitPixelsToCpu(const uint8_t* src, uint32_t srcWidth, uint32_t srcHeight, int32_t sx, int32_t sy, uint32_t sw, uint32_t sh, int32_t dx, int32_t dy, uint32_t dw, uint32_t dh)
     {
-        if (src == nullptr || dw == 0 || dh == 0)
+        if (src == nullptr || dw == 0 || dh == 0 || sw == 0 || sh == 0 || srcWidth == 0 || srcHeight == 0)
         {
             return;
         }
@@ -800,39 +805,37 @@ namespace Babylon::Polyfills::Internal
             return;
         }
 
-        for (uint32_t j = 0; j < dh; ++j)
+        // Clamp the iteration range to the destination rect's intersection with the canvas up
+        // front. The destination size is caller-controlled (and reaches us as an unsigned value,
+        // so a negative width wraps to ~4e9), and iterating the full rect just to reject every
+        // pixel would stall the JS thread. int64_t keeps dx/dy + dw/dh from overflowing.
+        const int64_t iBegin = std::max<int64_t>(0, -static_cast<int64_t>(dx));
+        const int64_t iEnd = std::min<int64_t>(dw, static_cast<int64_t>(m_cpuWidth) - dx);
+        const int64_t jBegin = std::max<int64_t>(0, -static_cast<int64_t>(dy));
+        const int64_t jEnd = std::min<int64_t>(dh, static_cast<int64_t>(m_cpuHeight) - dy);
+
+        for (int64_t j = jBegin; j < jEnd; ++j)
         {
-            const int32_t destY = dy + static_cast<int32_t>(j);
-            if (destY < 0 || destY >= static_cast<int32_t>(m_cpuHeight))
-            {
-                continue;
-            }
-
             // Nearest-neighbor sample of the source row (exact when dh == sh).
-            const int32_t srcYRel = static_cast<int32_t>(static_cast<uint64_t>(j) * sh / dh);
-            const int32_t srcY = sy + srcYRel;
-            if (srcY < 0 || srcY >= static_cast<int32_t>(srcHeight))
+            const int64_t srcY = sy + static_cast<int64_t>(j) * sh / dh;
+            if (srcY < 0 || srcY >= static_cast<int64_t>(srcHeight))
             {
                 continue;
             }
 
-            for (uint32_t i = 0; i < dw; ++i)
+            const size_t destRow = static_cast<size_t>(dy + j) * m_cpuWidth;
+            const size_t srcRow = static_cast<size_t>(srcY) * srcWidth;
+
+            for (int64_t i = iBegin; i < iEnd; ++i)
             {
-                const int32_t destX = dx + static_cast<int32_t>(i);
-                if (destX < 0 || destX >= static_cast<int32_t>(m_cpuWidth))
+                const int64_t srcX = sx + static_cast<int64_t>(i) * sw / dw;
+                if (srcX < 0 || srcX >= static_cast<int64_t>(srcWidth))
                 {
                     continue;
                 }
 
-                const int32_t srcXRel = static_cast<int32_t>(static_cast<uint64_t>(i) * sw / dw);
-                const int32_t srcX = sx + srcXRel;
-                if (srcX < 0 || srcX >= static_cast<int32_t>(srcWidth))
-                {
-                    continue;
-                }
-
-                const size_t srcIndex = (static_cast<size_t>(srcY) * srcWidth + srcX) * 4;
-                const size_t destIndex = (static_cast<size_t>(destY) * m_cpuWidth + destX) * 4;
+                const size_t srcIndex = (srcRow + static_cast<size_t>(srcX)) * 4;
+                const size_t destIndex = (destRow + static_cast<size_t>(dx + i)) * 4;
                 m_cpuPixels[destIndex + 0] = src[srcIndex + 0];
                 m_cpuPixels[destIndex + 1] = src[srcIndex + 1];
                 m_cpuPixels[destIndex + 2] = src[srcIndex + 2];
@@ -898,11 +901,32 @@ namespace Babylon::Polyfills::Internal
                 return;
             }
 
+            // Everything below is caller-supplied. bimg::imageConvert reads width*height*bpp bits
+            // from the source and writes width*height*4 bytes to the destination without knowing
+            // either buffer's real length, so validate both before handing it any pointers.
+            if (format >= bimg::TextureFormat::Count)
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap has an out-of-range pixel format.");
+            }
+
+            const uint64_t pixelCount = static_cast<uint64_t>(width) * height;
+            if (pixelCount > std::numeric_limits<size_t>::max() / 4)
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap dimensions are too large.");
+            }
+
+            const uint64_t requiredBits = static_cast<uint64_t>(data.ByteLength()) * 8;
+            const uint8_t bitsPerPixel = bimg::getBitsPerPixel(format);
+            if (bitsPerPixel == 0 || pixelCount > requiredBits / bitsPerPixel)
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap data is smaller than width*height for its format.");
+            }
+
             const uint8_t* srcBytes = data.Data();
-            std::vector<uint8_t> rgba(static_cast<size_t>(width) * height * 4);
+            std::vector<uint8_t> rgba(static_cast<size_t>(pixelCount) * 4);
             if (format == bimg::TextureFormat::RGBA8)
             {
-                std::memcpy(rgba.data(), srcBytes, std::min<size_t>(rgba.size(), data.ByteLength()));
+                std::memcpy(rgba.data(), srcBytes, rgba.size());
             }
             else if (!bimg::imageConvert(&Graphics::DeviceContext::GetDefaultAllocator(), rgba.data(), bimg::TextureFormat::RGBA8, srcBytes, format, width, height, 1))
             {
@@ -1017,10 +1041,22 @@ namespace Babylon::Polyfills::Internal
 
     Napi::Value Context::GetImageData(const Napi::CallbackInfo& info)
     {
+        if (info.Length() < 4)
+        {
+            throw Napi::Error::New(info.Env(), "Context2D.getImageData: invalid number of parameters");
+        }
+
         const auto sx = info[0].As<Napi::Number>().Int32Value();
         const auto sy = info[1].As<Napi::Number>().Int32Value();
         const auto sw = info[2].As<Napi::Number>().Uint32Value();
         const auto sh = info[3].As<Napi::Number>().Uint32Value();
+
+        // The region is caller-supplied and backs a width*height*4 allocation, so reject sizes
+        // that would overflow size_t before ImageData tries to allocate them.
+        if (static_cast<uint64_t>(sw) * sh > std::numeric_limits<size_t>::max() / 4)
+        {
+            throw Napi::RangeError::New(info.Env(), "Context2D.getImageData: requested region is too large.");
+        }
 
         return ImageData::CreateInstance(info.Env(), this, sx, sy, sw, sh);
     }
--- a/Polyfills/Canvas/Source/Context.cpp
+++ b/Polyfills/Canvas/Source/Context.cpp
@@ -32,6 +32,10 @@
 #include "LineCaps.h"
 #include "Gradient.h"
 
+#ifdef BABYLON_NATIVE_PLUGIN_NATIVEENGINE_LOAD_IMAGES
+#include <bimg/bimg.h>
+#endif
+
 namespace Babylon::Polyfills::Internal
 {
     static constexpr auto JS_CONTEXT_CONSTRUCTOR_NAME = "Context";
@@ -780,7 +784,11 @@ namespace Babylon::Polyfills::Internal
 
     void Context::BlitImageToCpu(const NativeCanvasImage& image, int32_t sx, int32_t sy, uint32_t sw, uint32_t sh, int32_t dx, int32_t dy, uint32_t dw, uint32_t dh)
     {
-        const uint8_t* src = image.GetPixels();
+        BlitPixelsToCpu(image.GetPixels(), image.GetWidth(), image.GetHeight(), sx, sy, sw, sh, dx, dy, dw, dh);
+    }
+
+    void Context::BlitPixelsToCpu(const uint8_t* src, uint32_t srcWidth, uint32_t srcHeight, int32_t sx, int32_t sy, uint32_t sw, uint32_t sh, int32_t dx, int32_t dy, uint32_t dw, uint32_t dh)
+    {
         if (src == nullptr || dw == 0 || dh == 0)
         {
             return;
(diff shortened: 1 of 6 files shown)


Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,178 @@ namespace Babylon::Polyfills::Internal
...
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap dimensions are too large.");
+            }
+
+            const uint64_t requiredBits = static_cast<uint64_t>(data.ByteLength()) * 8;
+            const uint8_t bitsPerPixel = bimg::getBitsPerPixel(format);
+            if (bitsPerPixel == 0 || pixelCount > requiredBits / bitsPerPixel)
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap data is smaller than width*height for its format.");
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, fixed in 916595ed. Confirmed by reading `imageGetSize`: it rounds width/height up to whole blocks and applies `minBlockX`/`minBlockY`, so a 1x1 BC1 image really does need a full 8-byte block. The old `pixelCount * bitsPerPixel / 8` math accepted a 1-byte buffer for it. Now calling `bimg::imageGetSize(nullptr, width, height, 1, false, false, 1, format)` and comparing against `data.ByteLength()` directly.

Verified with a negative control: a `drawImage({data: new Uint8Array(1), width: 1, height: 1, format: 0 /* BC1 */})` case is accepted by the old math and rejected by the new one.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -853,13 +1041,24 @@ namespace Babylon::Polyfills::Internal
...
     {
-        // TODO: support source x and y
-        //const auto sx = info[0].As<Napi::Number>().Uint32Value();
-        //const auto sy = info[1].As<Napi::Number>().Uint32Value();
+        if (info.Length() < 4)
+        {
+            throw Napi::Error::New(info.Env(), "Context2D.getImageData: invalid number of parameters");
+        }
+
+        const auto sx = info[0].As<Napi::Number>().Int32Value();
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Correct, fixed in 916595ed. `ReadPixels` now calls `EnsureCpuBuffer()` before reading, which reallocates and zero-fills when the canvas dimensions changed. `ReadPixels` is no longer `const` as a result.

Repro that used to fail: resize the canvas and call `getImageData` with no intervening `drawImage` -- the mirror still held the old buffer and was indexed with the new stride, returning stale pixels.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -830,8 +1017,6 @@ namespace Babylon::Polyfills::Internal
             const auto dy = static_cast<float>(info[6].As<Napi::Number>().Int32Value());
             const auto dWidth = static_cast<float>(info[7].As<Napi::Number>().Uint32Value());
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 916595ed. Both the 5-arg and 9-arg overloads now read their extents with `Int32Value()` and return early when any extent is non-positive, which matches the browser behaviour of drawing nothing. Previously a negative width arrived as roughly 4e9, which queued an enormous `nvgRect` and made the CPU blit iterate for billions of pixels.

Note `BlitPixelsToCpu` already clamps its iteration range in `int64_t`, so it was not reading out of bounds -- but it was still spinning the JS thread, so the guard is worth having at the parse site.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -853,13 +1073,24 @@ namespace Babylon::Polyfills::Internal
...
-        //const auto sy = info[1].As<Napi::Number>().Uint32Value();
+        if (info.Length() < 4)
+        {
+            throw Napi::Error::New(info.Env(), "Context2D.getImageData: invalid number of parameters");
+        }
+
+        const auto sx = info[0].As<Napi::Number>().Int32Value();
+        const auto sy = info[1].As<Napi::Number>().Int32Value();
         const auto sw = info[2].As<Napi::Number>().Uint32Value();
         const auto sh = info[3].As<Napi::Number>().Uint32Value();
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, this was a real out-of-bounds read. `imageGetSize` takes `uint16_t` extents, so the dimensions were truncated *before* the required-size check: a 65537x1 RGBA8 bitmap measured as 1x1, i.e. 4 bytes. A 4-byte buffer therefore passed validation, and the RGBA8 `memcpy` below then read 65537 * 1 * 4 = 262148 bytes from it.

Fixed in 8e5b6fd8 by rejecting dimensions bimg cannot describe.

Negative control: with the guard reverted, the oversized-bitmap assertion does not merely fail, it takes the process down with it.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,186 @@ namespace Babylon::Polyfills::Internal
...
+
+            // Ask bimg for the size rather than computing width*height*bpp/8: block-compressed
+            // formats round up to whole blocks and have a minimum block count, so a 1x1 BC1 image
+            // still occupies one 8-byte block. The naive bits-per-pixel math would accept a
+            // 1-byte buffer for it and imageConvert would then read past the end.
+            const uint64_t requiredBytes = bimg::imageGetSize(nullptr, static_cast<uint16_t>(width), static_cast<uint16_t>(height), 1, false, false, 1, format);
+            if (requiredBytes == 0 || data.ByteLength() < requiredBytes)
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap data is smaller than width*height for its format.");
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Also a real bug, and worse than it first looks. Reading the extents with `Uint32Value` turned a width of `-1` into `4294967295`, and the `SIZE_MAX / 4` guard below does not catch that on 64-bit because `4294967295 * 1` is nowhere near the limit. So `getImageData(0, 0, -1, 1)` sailed through and asked `ImageData` for roughly 17 GB.

Rather than just rejecting negative extents, I matched the browser: per the spec a negative width or height denotes a rectangle running in the opposite direction, and the normalized region is returned. So the sign is folded into the origin (`sx -= sw`). `getImageData(4, 0, -4, 4)` and `getImageData(0, 0, 4, 4)` now return identical pixels, which is asserted in the self-test.

`INT32_MIN` is rejected outright since negating it overflows.

Fixed in 8e5b6fd8. Negative control: with the fix reverted the call stalls for 9.6 s attempting the allocation, then the process dies.

Polyfills/Canvas/Source/Context.h
```diff
@@ -137,6 +142,17 @@ namespace Babylon::Polyfills::Internal
         void PlayPath2D(const NativeCanvasPath2D* path);
         void SetFilterStack();
 
+        // CPU-side RGBA8 mirror of the canvas, sized to the canvas, populated by DrawImage so that
+        // getImageData can return the exact decoded pixels (the GPU nanovg framebuffer is not read back).
+        std::vector<uint8_t> m_cpuPixels;
```
bghgary, line 147: [Reviewed by Copilot on behalf of @bghgary]

The mirror is written only by the three `BlitPixelsToCpu` calls in `DrawImageCommon`, using the raw destination coordinates and a straight byte assignment. So it diverges from what actually gets rendered in more ways than just the missing draw calls:

- **Transform** - `translate`/`rotate`/`scale`/`setTransform` go to nanovg, but `BlitPixelsToCpu` gets the raw `dx`/`dy` arguments. `ctx.translate(10, 10); ctx.drawImage(img, 0, 0)` renders at (10, 10) and mirrors at (0, 0). That breaks the 1:1 `drawImage` -> `getImageData` round trip this is built for, not just the general case.
- **Compositing** - `drawImage` blends source-over (the `NVG_COPY` in `ClearRect` is scoped to that call), while the mirror replaces RGBA including alpha. Overlapping draws, or any image with transparent margins, differ.
- **`globalAlpha`** - applied by nanovg, ignored by the mirror.
- **Clipping** - `nvgScissor` bounds the GPU draw; the mirror writes the full rect.
- **Everything that is not `drawImage`** - `fillRect`, `fillText`, paths and gradients never reach the mirror, and `clearRect` leaves stale pixels in it.

Previously `getImageData` was uniformly zeros, so nobody could mistake it for working. Now it is exact on one narrow path and silently wrong elsewhere, which is harder to catch. Worth deciding whether the mirror tracks the full 2D state, or whether `getImageData` should fail loudly once anything it cannot mirror has touched the canvas.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,195 @@ namespace Babylon::Polyfills::Internal
...
+                std::memcpy(rgba.data(), srcBytes, rgba.size());
+            }
+            else if (!bimg::imageConvert(&Graphics::DeviceContext::GetDefaultAllocator(), rgba.data(), bimg::TextureFormat::RGBA8, srcBytes, format, width, height, 1))
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: unsupported ImageBitmap pixel format.");
+            }
+
+            const int imageIndex = nvgCreateImageRGBA(*m_nvg, static_cast<int>(width), static_cast<int>(height), 0, rgba.data());
+            DrawImageCommon(info, imageIndex, rgba.data(), width, height);
+            nvgDeleteImage(*m_nvg, imageIndex);
```
bghgary, line 955: [Reviewed by Copilot on behalf of @bghgary]

`DrawImageCommon` throws on an unexpected argument count, so `drawImage(bitmap, 1, 2, 3)` skips `nvgDeleteImage` and leaks the nanovg image and its texture on every call. The `NativeCanvasImage` path is unaffected because it caches into `m_nvgImageIndices`. Worth scoping the handle so it is released on the throw path.

Three smaller things in the same block:

- `nvgCreateImageRGBA`'s result is not checked. On failure the draw silently produces nothing while `BlitPixelsToCpu` still writes the mirror, so the two disagree. The `NativeCanvasImage` path at least asserts.
- Every call creates and destroys a GPU texture for the same bitmap. That is fine for a one-shot `applyDisplacementMap`, but anything drawing the same `ImageBitmap` per frame pays a full upload and teardown each time, where the `NativeCanvasImage` path caches.
- `std::vector<uint8_t> rgba(pixelCount * 4)` zero-fills the buffer before it is completely overwritten on both branches. For a 2048x2048 bitmap that is 16 MB of zeroing on top of the 16 MB copy.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -853,13 +1082,58 @@ namespace Babylon::Polyfills::Internal
...
+        if (swInt == std::numeric_limits<int32_t>::min() || shInt == std::numeric_limits<int32_t>::min())
+        {
+            throw Napi::RangeError::New(info.Env(), "Context2D.getImageData: requested region is too large.");
+        }
+
+        uint32_t sw, sh;
+        if (swInt < 0)
+        {
+            sw = static_cast<uint32_t>(-swInt);
+            sx -= static_cast<int32_t>(sw);
```
bghgary, line 1112: [Reviewed by Copilot on behalf of @bghgary]

`sx -= static_cast<int32_t>(sw)` can overflow: `getImageData(-2000000000, 0, -2000000000, 1)` clears the INT32_MIN guard, then subtracts 2e9 from -2e9. `ReadPixels` bounds-checks every pixel so nothing is read out of range, but it is still signed overflow; computing the normalized origin in `int64_t` avoids it.

Separately, the only bound on the region is the `size_t` overflow check, so `getImageData(0, 0, 2000000000, 1)` allocates 8 GB in `m_pixels`, and `GetData` copies it again. Worth a ceiling?

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,195 @@ namespace Babylon::Polyfills::Internal
...
+                {
+                    continue;
+                }
+
+                const size_t srcIndex = (srcRow + static_cast<size_t>(srcX)) * 4;
+                const size_t destIndex = (destRow + static_cast<size_t>(dx + i)) * 4;
+                m_cpuPixels[destIndex + 0] = src[srcIndex + 0];
+                m_cpuPixels[destIndex + 1] = src[srcIndex + 1];
+                m_cpuPixels[destIndex + 2] = src[srcIndex + 2];
+                m_cpuPixels[destIndex + 3] = src[srcIndex + 3];
```
bghgary, line 842: [Reviewed by Copilot on behalf of @bghgary]

This inner loop does a 64-bit integer division per pixel (`i * sw / dw`) and four separate byte stores, which the compiler generally will not merge into one 32-bit store through `std::vector<uint8_t>` indexing.

For the 1:1 draws this is aimed at, none of it is needed: when `dw == sw` and `dh == sh`, `srcX` is just `sx + i`, so each row is a contiguous run and the whole inner loop is one `memcpy` of `runLength * 4`. Worth special-casing that and keeping the sampling loop for the scaled case - and even there, hoisting the division out with an accumulator.

`ReadPixels` has the same shape: clipping the rect up front lets each row be a `memcpy` of `w * 4`, which also drops the two per-pixel bounds checks and makes the full-buffer `memset` unnecessary except for the out-of-range margin.

Polyfills/Canvas/Source/ImageData.cpp
```diff
@@ -57,11 +77,12 @@ namespace Babylon::Polyfills::Internal
...
     {
-        // return a well size array with 0
-        // TODO: Get datas from context/canvas
-        const auto size{m_width * m_height * 4};
+        const auto size{m_pixels.size()};
         auto data{Napi::Uint8Array::New(info.Env(), size)};
-        memset(data.Data(), 0, size);
+        if (size > 0)
+        {
+            std::memcpy(data.Data(), m_pixels.data(), size);
```
bghgary, line 84: [Reviewed by Copilot on behalf of @bghgary]

`m_pixels` only exists to be copied out of. `ReadPixels` fills it, then every `.data` access allocates a fresh `Uint8Array` and copies it again - so one `getImageData` plus one `.data` read walks the pixels three times (memset, mirror -> `m_pixels`, `m_pixels` -> JS) where once would do, and the buffer stays resident for the object's lifetime alongside every array handed out.

Having `ReadPixels` write straight into the typed array, created once, removes the intermediate.

Copying per access also diverges from the browser, where `.data` is a stable live `Uint8ClampedArray` - `id.data === id.data`, and mutating it then calling `putImageData` is the normal round trip. Here each read is a detached copy, so writes are silently dropped. That predates this PR, but it only starts to matter now the data is real rather than always zero.

CedricGuillemet (approved):
LGTM

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,195 @@ namespace Babylon::Polyfills::Internal
...
+                return;
+            }
+
+            // Everything below is caller-supplied. bimg::imageConvert reads the source and writes
+            // width*height*4 bytes to the destination without knowing either buffer's real length,
+            // so validate both before handing it any pointers.
+            if (format >= bimg::TextureFormat::Count)
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap has an out-of-range pixel format.");
+            }
```
bghgary, line 915: [Reviewed by Copilot on behalf of @bghgary]

**This guard does not work on MSVC, and it is the check the whole `ImageBitmap` path rests on.**

`bimg::TextureFormat::Enum` is an unscoped enum with no fixed underlying type and enumerators in `[0, 104]`, so MSVC picks **signed `int`**. `static_cast<Enum>(0xFFFFFFFFu)` is therefore `-1`, and `format >= bimg::TextureFormat::Count` is **false**. Every `format` in `[0x80000000, 0xFFFFFFFF]` becomes an attacker-chosen negative index. Compiled against the repo's toolchain (MSVC 17.0 x64, `/O2 /std:c++17`):

```
sizeof=4 signed=1 Count=96
raw=0xFFFFFFFF as_int=-1           guard rejects? NO
raw=0x80000000 as_int=-2147483648  guard rejects? NO
raw=0x000000C8 as_int=200          guard rejects? YES
```

bimg then indexes its global tables with no bounds check of its own - `s_imageBlockInfo[_format]` (image.cpp:282/287/292) and, in the convert path, `s_packUnpack[_srcFormat].unpack` (image.cpp:1234) which is then **called**. So a negative index gives an out-of-bounds read of `s_imageBlockInfo` at an attacker-chosen offset, an attacker-controlled `srcBpp` (and therefore source stride), and an indirect call through a function pointer read out of bounds. It also defeats the `requiredBytes` check below, since that value is derived from the same table read - a small non-zero result lets `imageConvert` read far past the end of a small `Uint8Array`, and the result reaches JS through the now-working `getImageData`.

GCC and Clang on non-MS ABIs select `unsigned int`, so the comparison holds there. This is Windows/UWP-specific, and both are shipped targets. The cast is formally UB either way.

```js
ctx.drawImage({ data: new Uint8Array(64), width: 1, height: 1, format: 0xFFFFFFFF }, 0, 0);
```

Validate the raw `uint32_t` before it becomes an enum:

```suggestion
            const uint32_t rawFormat = imageObj.Get("format").As<Napi::Number>().Uint32Value();
            if (rawFormat >= static_cast<uint32_t>(bimg::TextureFormat::Count))
            {
                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap has an out-of-range pixel format.");
            }
            const auto format = static_cast<bimg::TextureFormat::Enum>(rawFormat);
```

Worth rejecting `Unknown` and `UnknownDepth` too.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,195 @@ namespace Babylon::Polyfills::Internal
...
+        Napi::Object imageObj = info[0].As<Napi::Object>();
+
+        // NativeEngine.createImageBitmap() returns a plain object carrying the raw decoded pixels
+        // ({data, width, height, format}) rather than a wrapped NativeCanvasImage. Because Babylon
+        // Native sets forceBitmapOverHTMLImageElement, LoadImage delivers these bitmaps to drawImage
+        // (e.g. Mesh.applyDisplacementMap, height/flow maps). Unwrapping such a plain object as a
+        // NativeCanvasImage would dereference garbage and crash, so handle it explicitly by
+        // converting the pixels to RGBA8 and drawing/blitting them directly.
+        if (imageObj.Has("data") && imageObj.Get("data").IsTypedArray())
+        {
```
bghgary, line 898: [Reviewed by Copilot on behalf of @bghgary]

This closes the crash for one object shape and leaves every other shape falling through to `NativeCanvasImage::Unwrap` below - and the unwrap does not validate what it is handed. JsRuntimeHost's napi fork replaced upstream's `IsExternal()` check with a raw internal-field read:

```cpp
// js_native_api_v8.cc:361-367
// [BABYLON-NATIVE-ADDITION]: Increase perf by using internal field instead of private property
Reference* reference =
    static_cast<v8impl::Reference*>(obj->GetAlignedPointerFromInternalField(0));
```

`DefineClass` sets `SetInternalFieldCount(1)` on every napi ObjectWrap class with no type tag, so passing any *other* wrapped object returns a valid pointer to an unrelated C++ object reinterpreted as `NativeCanvasImage*`. That is deterministic type confusion rather than a crash. `ImageData` lines up particularly well: its `m_width`/`m_height` sit at the same offsets and are attacker-set via `getImageData(0, 0, W, H)`, while `m_imageContainer` - the last member of the larger `NativeCanvasImage` - reads past the end of the smaller object into groomable heap. `CreateNVGImageForContext` then does `nvgCreateImageRGBA(nvg, m_width, m_height, 0, m_imageContainer->m_data)`, and `GetPixels()` feeds the same pointer to `BlitPixelsToCpu`.

The reason this matters more in this PR than before it: previously `getImageData` returned zeros, so a wild read was unobservable. The CPU mirror turns it into a read-back channel.

```js
ctx.drawImage(ctx.getImageData(0, 0, 64, 64), 0, 0);
ctx.getImageData(0, 0, 64, 64).data;
```

Worth identifying the argument positively - `InstanceOf` against the persisted `Image` constructor, or a `napi_type_tag` - and throwing a `TypeError` otherwise, rather than duck-typing on `data`. Note `drawImage(canvasElement, ...)` is legitimate Canvas2D, so ordinary content reaches this path.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,195 @@ namespace Babylon::Polyfills::Internal
...
+#ifdef BABYLON_NATIVE_PLUGIN_NATIVEENGINE_LOAD_IMAGES
+            const auto data = imageObj.Get("data").As<Napi::Uint8Array>();
+            const uint32_t width = imageObj.Get("width").As<Napi::Number>().Uint32Value();
+            const uint32_t height = imageObj.Get("height").As<Napi::Number>().Uint32Value();
+            const auto format = static_cast<bimg::TextureFormat::Enum>(imageObj.Get("format").As<Napi::Number>().Uint32Value());
```
bghgary, line 903: [Reviewed by Copilot on behalf of @bghgary]

`Napi::Uint8Array` caches the backing-store pointer and length at construction - `Data()` and `ByteLength()` are plain accessors that never re-query the engine. `data` is captured on the first line here, and the three `Get(...)` calls that follow are full property lookups that run accessors and Proxy traps, so **attacker JS executes after the pointer and length are frozen**. The `memcpy` and `imageConvert` below then use the stale pair.

```js
const buf = new ArrayBuffer(BIG);
const view = new Uint8Array(buf);
ctx.drawImage({
  data: view,
  get width()  { buf.transfer(BIG + 1); return W; },  // detaches, frees old store
  get height() { return H; },
  get format() { return 76; },
}, 0, 0);
ctx.getImageData(0, 0, W, H).data;
```

`ArrayBuffer.prototype.transfer` with a changed length allocates a new store and detaches the old one, dropping the last reference. The stale `_length` still satisfies the `requiredBytes` check while `_data` dangles, so the copy reads freed heap into `rgba` - and out through `getImageData`.

Reading the scalars first and fetching `data` last would close it, with no JS-invoking call between capturing `Data()`/`ByteLength()` and the copy. A `napi_is_detached_arraybuffer` check immediately before use would make it explicit.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -765,9 +771,195 @@ namespace Babylon::Polyfills::Internal
...
+            }
+
+            // imageGetSize takes uint16_t extents, so anything above 65535 would be truncated and
+            // produce a size for a much smaller image. A 65537x1 RGBA8 bitmap would come back as
+            // 1x1 (4 bytes), a 4-byte buffer would pass the check below, and the RGBA8 memcpy
+            // would then read width*height*4 bytes from it. Reject what bimg cannot describe.
+            if (width > std::numeric_limits<uint16_t>::max() || height > std::numeric_limits<uint16_t>::max())
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: ImageBitmap dimensions exceed the maximum supported size.");
+            }
```
bghgary, line 930: [Reviewed by Copilot on behalf of @bghgary]

The rationale here is wrong for the pinned bimg: `imageGetSize` takes `uint32_t _width, uint32_t _height` (bimg.h) and accumulates in `uint64_t` - the `uint16_t` in that signature is `_numLayers`. So nothing is truncated to 16 bits and the "65537x1 comes back as 1x1" example cannot happen.

The cap itself is harmless, but the comment presents it as the load-bearing check against a truncation that does not exist. Worth either dropping it or rewriting the reason.

bghgary (changes requested):

The two fixes are the right ones and the validation work is thorough, but the `ImageBitmap` path has a memory-safety problem that should block: the pixel-format guard does not work on MSVC, so a JS-supplied format in the high half of `uint32_t` becomes a negative index into bimg's global tables, including an indirect call through an out-of-bounds function pointer. Details and a repro inline.

Two related ones on the same path - the plain-object check leaves every other object shape falling through to an unvalidated `Unwrap`, and the typed-array pointer is captured before three property reads that can run attacker JS. Both are made reachable in a new way by the CPU mirror, since `getImageData` now returns real bytes instead of zeros.

The rest is correctness and cost: the mirror ignores transform, compositing, `globalAlpha` and clipping, and several copies look avoidable.

PR #1801, opened 2026-07-29
NativeEngine: route instanced generic attributes landing on TexCoord3..TexCoord15

bkaradzic-microsoft:
Fixes a silent data-loss bug in the divisor-driven instancing reroute in `NativeEngine::Draw`.

## Problem

When Babylon.js declares a *generic* (consumer-named) vertex attribute with `divisor == 1`, the attribute is compiled to an ordinary per-vertex `bgfx::Attrib` slot, but bgfx can only feed per-instance data through the `i_data` slots (the top TEXCOORD semantics). `NativeEngine` therefore reroutes such attributes to the correct `i_data` slot via a lazily-compiled program variant.

That reroute was guarded by `attrib < bgfx::Attrib::TexCoord3`, so any generic instanced attribute that landed on **TexCoord3 or above was silently skipped**. `BuildInstanceDataBuffer` still packed its data into the instance data buffer, but the shader kept reading the attribute as a per-vertex input — i.e. unrelated vertex bytes.

Sprite instancing assigns `cellInfo -> TexCoord3`, so instanced sprites read garbage cell info and rendered collapsed / mis-UVed.

## Fix

Widen the guard to every real bgfx attribute (`attrib < bgfx::Attrib::Count`).

The built-in instanced attributes (`world0-3`, `splatIndex0-3`, `instanceColor`) are assigned *synthetic* locations at/above `INSTANCE_DATA_FIRST_LOCATION - 4`, which is `>= bgfx::Attrib::Count`, so they still compare false and remain correctly skipped — they already arrive as instance data.

## Evidence

Measured on Win32 x64 D3D11 (Debug) with Babylon.js `_features.supportSpriteInstancing` forced to `true`, so the instanced sprite path is actually exercised. Negative control = this commit reverted, everything else identical:

| test | without fix | with fix |
|---|---|---|
| 155 `Sprites` | 4.242% **FAIL** | **1.405% PASS** |
| 293 `Prepass SSAO + sprites` | 1.617% pass | **0.001% pass** |
| 369 `Sprites Pixel Perfect` | 25.532% FAIL | 7.612% FAIL (much closer, still over threshold) |

Full validation suite is unchanged: `ran=295 passed=295 failed=0` (`--test-index=0-52,58-719`; 53-57 are the pre-existing scissor crash on master).

## Notes

- **No `config.json` changes.** Un-excluding the sprite/MSDF tests additionally requires the Babylon.js-side `supportSpriteInstancing` flag; this PR is the Native prerequisite for that work and is a strict improvement on its own.
- **No shader cache version bump needed.** Instanced program variants bypass the shader cache entirely (`ShaderProvider::Get` consults the cache only when `instancedAttributes` is empty), and the non-instanced compilation path is untouched.

### DIFF
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -2431,7 +2431,17 @@ namespace Babylon
                 for (const auto& instance : instances)
                 {
                     const bgfx::Attrib::Enum attrib = instance.first;
-                    if (attrib < bgfx::Attrib::TexCoord3)
+                    // Reroute every consumer-instanced attribute that was compiled to a real
+                    // per-vertex bgfx Attrib slot (Position..TexCoord15, i.e. < Attrib::Count).
+                    // The built-in instanced attributes (world0-3, splatIndex0-3, instanceColor)
+                    // are assigned synthetic locations at/above INSTANCE_DATA_FIRST_LOCATION - 4,
+                    // which is >= Attrib::Count, so they still compare false here and are correctly
+                    // skipped: they already arrive as instance data.
+                    // The previous TexCoord3 boundary silently dropped generic instanced attributes
+                    // landing on TexCoord3..TexCoord15 (e.g. sprite cellInfo -> TexCoord3), leaving
+                    // them reading per-vertex garbage even though BuildInstanceDataBuffer had
+                    // already packed them into the instance data buffer.
+                    if (attrib < bgfx::Attrib::Count)
                     {
                         const size_t rank = count - 1 - ascendingIndex;
                         const uint32_t targetLocation = Babylon::Graphics::INSTANCE_DATA_FIRST_LOCATION - static_cast<uint32_t>(rank);
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -2413,7 +2413,7 @@ namespace Babylon
         }
 
         // Divisor-driven instancing: a consumer-instanced attribute (divisor==1) recorded at a
-        // base bgfx location below TexCoord3 was compiled to a per-vertex slot. bgfx can only feed
+        // real per-vertex bgfx location was compiled to a per-vertex slot. bgfx can only feed
         // per-instance data into i_data slots (the top TEXCOORD semantics), so route those attributes
         // to the correct i_data slot via a lazily-compiled program variant. The target location mirrors
         // BuildInstanceDataBuffer's reverse-attrib packing: highest base attrib -> i_data0 (TEXCOORD31),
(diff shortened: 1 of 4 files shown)


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -2431,7 +2431,17 @@ namespace Babylon
                 for (const auto& instance : instances)
                 {
                     const bgfx::Attrib::Enum attrib = instance.first;
-                    if (attrib < bgfx::Attrib::TexCoord3)
+                    // Reroute every consumer-instanced attribute that was compiled to a real
+                    // per-vertex bgfx Attrib slot (Position..TexCoord15, i.e. < Attrib::Count).
+                    // The built-in instanced attributes (world0-3, splatIndex0-3, instanceColor)
+                    // are assigned synthetic locations at/above INSTANCE_DATA_FIRST_LOCATION - 4,
+                    // which is >= Attrib::Count, so they still compare false here and are correctly
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch — fixed in 670a1d4e.

That block comment was pre-existing text I left untouched, and the widened guard made its "recorded at a base bgfx location below TexCoord3" clause wrong. It now reads "recorded at a real per-vertex bgfx location", and I removed the sentence my inline comment duplicated from it, so the two read as one explanation instead of two overlapping ones.

Comment-only change; no code tokens touched.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -2431,7 +2431,16 @@ namespace Babylon
                 for (const auto& instance : instances)
                 {
                     const bgfx::Attrib::Enum attrib = instance.first;
-                    if (attrib < bgfx::Attrib::TexCoord3)
+                    // "Real per-vertex slot" means Position..TexCoord15, i.e. < Attrib::Count. The
+                    // built-in instanced attributes (world0-3, splatIndex0-3, instanceColor) are
+                    // assigned synthetic locations at/above INSTANCE_DATA_FIRST_LOCATION - 4, which
+                    // is >= Attrib::Count, so they compare false here and are correctly skipped:
```
bghgary, line 2437: [Reviewed by Copilot on behalf of @bghgary]

This is the invariant the widened guard now rests on, and nothing asserts it. `ShaderCompilerCommon.cpp` asserts `INSTANCE_DATA_FIRST_LOCATION >= bgfx::Attrib::Count`, which is location 41 - the *first* instance slot. The value this comment depends on is the lowest built-in one, 41 - 4 = 37.

If `bgfx::Attrib::Count` ever landed in 38..41, that assert would still pass while `instanceColor` began being rerouted - the same silent-garbage failure this PR is fixing. Worth asserting `INSTANCE_DATA_FIRST_LOCATION - 4 >= bgfx::Attrib::Count` next to the existing one.

bghgary (approved):

LGTM. One hardening ask inline.

Plugins/NativeEngine/Source/NativeEngine.cpp
bkaradzic-microsoft: Confirmed and fixed in f92bd3b3.

Checked the numbers against the live enum rather than assuming: `bgfx::Attrib::Count` == 26 (TexCoord0 == 10 ... TexCoord15 == 25), `INSTANCE_DATA_FIRST_LOCATION` == 41, and the lowest built-in is `instanceColor` on i_data4 at 37 - so there are 11 attributes of slack, not the ~0 the existing assert implies.

And your failure mode reproduces exactly. Compiling the two asserts against a forced `Count == 39`:

```
neg.cpp(6): error C2338: static_assert failed: 'NEW ASSERT FIRED'
```

Only the new one fires; `41 >= 39` keeps the old one quiet while `instanceColor` at 37 starts getting rerouted as per-vertex data.

I named the 4 rather than open-coding it, since it was already implicit in `ShaderCompilerTraversers.cpp` and a bare `- 4` next to the existing assert would have been a second undocumented copy:

```cpp
inline constexpr uint32_t BUILTIN_INSTANCE_DATA_SLOT_COUNT{5};
inline constexpr uint32_t BUILTIN_INSTANCE_DATA_LAST_LOCATION{INSTANCE_DATA_FIRST_LOCATION - (BUILTIN_INSTANCE_DATA_SLOT_COUNT - 1)};
```

`static_assert(BUILTIN_INSTANCE_DATA_LAST_LOCATION >= bgfx::Attrib::Count)` goes directly under the existing one, and the `instanceColor` row in the traverser table now points at the constant so adding a built-in on a lower i_data slot has a visible thing to bump.

One limitation worth flagging, inherited from the existing assert rather than introduced here: both live in `ShaderCompilerCommon.cpp`, so neither is compiled in precompiled-shader configurations - while the guard they protect is in `NativeEngine::Draw`, which always is. Happy to mirror the pair into `NativeEngine.cpp` for always-on coverage if you want that; I left it alone to keep this change to the hardening you asked for.

`ShaderCompiler` and `NativeEngine` both build clean.

PR #1808, opened 2026-07-31
NativeEngine: texture upload, cubemap loading, and readback improvements

bkaradzic-microsoft:
## What

Fills in a set of gaps in `NativeEngine`'s texture upload / readback paths so that Babylon.js texture APIs that already work on WebGL behave the same on Native.

Seven self-contained commits:

| Commit | Change |
|---|---|
| `975b63d8` | Load single-file `.dds` / `.ktx` / `.ktx2` cubemaps, including embedded spherical-harmonic coefficients |
| `b1bc011f` | Implement `NativeEngine::updateTextureData` |
| `43dedc29` | Add the `updateTextureDirectly` texture-loader sink |
| `9871bf8e` | Route cube-texture `UpdateTextureData` to `bgfx::updateTextureCube` |
| `138e4577` | Fix a crash loading BC1/DXT1 textures |
| `35f2cdad` | Support cube-map face readback in `NativeEngine.readTexture` |
| `d4d8d438` | Native raw 3D textures, plus a `sampler3D` `texelFetch` coordinate-flip fix |

Each commit builds and runs on its own; they're ordered so the plumbing lands before the callers.

## Validation

Full Playground validation suite, `RelWithDebInfo`, Win32, D3D11:

```
Run complete. ran=301 passed=301 failed=0 missingRef=0 skipped=419
```

No regressions against master's baseline of 300/300.

## About `config.json`

This PR un-excludes **exactly one** test — `Test updateTextureData` — and that one is verified to pass against the **stock npm `babylonjs` 9.15.0** that `Apps/node_modules` resolves to.

I want to flag this explicitly because it bit me: while developing this I had a locally-built Babylon.js fork in `Apps/node_modules` (12.7 MB `babylon.max.js`, same declared version `9.15.0` as the 8.6 MB npm build). Against that fork, 16 tests appeared to pass. Against stock npm, only 1 of the 16 actually does. The other 15 need Babylon.js-side changes that haven't landed yet:

- cubemap tests → need a Babylon.js-side single-file cubemap loader path. There is **no open PR for this today**: BabylonJS/Babylon.js#18560 was the attempt, but it was closed unmerged on 2026-06-17 and the work landed nowhere, so `nativeEngine.cubeTexture.pure.ts` on `master` still throws `Cannot load cubemap because 6 files were not defined` for a single file.
- CDF renderer tests → `CDF renderer is not supported by the current engine`
- `TEXTURE_3D` / `FLOAT` tests → require the raw-3D-texture constants to be plumbed through the engine caps

Those un-exclusions are deliberately **not** in this PR and will follow once the corresponding Babylon.js work is released.

## Note for anyone touching the shader compiler

`d4d8d438` includes a fix in `ShaderCompilerTraversers.cpp` that is worth calling out, because the failure mode is nasty and invisible in the common configuration.

BabylonNative builds SPIRV-Cross with `SPIRV_CROSS_WEBMIN` (see the root `CMakeLists.txt`; `BABYLON_NATIVE_DISABLE_WEBMIN` turns it off). In that configuration a number of opcodes — including **`OpIMul`** — are compiled out to `SPIRV_CROSS_INVALID_CALL()`, which is a bare `assert(false)`. Under `NDEBUG` that is a **no-op**: the instruction is visited, no result id is set, and the failure surfaces much later at the first consumer of that id as `Cannot resolve expression type.` — and since `SPIRV_CROSS_THROW` is also stripped to `throw CompilerError("")` under WEBMIN, the message you actually get is empty.

Concretely: emitting `coord * ivec2(1, -1)` from a traverser produces a silently broken HLSL/MSL/Vulkan shader. The fix here computes the flip as `ivec2(coord.x, textureSize(s, lod).y - 1 - coord.y)`, which only needs `OpCompositeExtract` / `OpCompositeConstruct` / `OpISub` — all of which WEBMIN retains. Because that formulation references the coordinate subtree twice, `EOpTextureFetch` handling also moved from `EvPreVisit` to `EvPostVisit` so the traverser doesn't descend into the duplicated subtree and double-flip nested `texture()` calls.

Short version: **don't emit integer multiply from a shader-compiler traverser.** I'll file this upstream against SPIRV-Cross separately — a stripped opcode should fail loudly rather than emit a broken shader.


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1454,6 +1713,89 @@ namespace Babylon
...
+        if (texture->IsCube())
+        {
+            // bgfx addresses a cube texture by (array layer, side 0-5). Only a single (non-array) cube is
+            // supported here, so the JS "layer" argument is the face/side index.
+            texture->UpdateCube(0, static_cast<uint8_t>(layer), mip, x, targetY, width, height, mem);
+        }
+        else
+        {
+            texture->Update2D(layer, mip, x, targetY, width, height, mem);
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, fixed in 73b33cf9.

You're right that the two ends disagreed. The bounds check admits `layer` up to `6 * numLayers` (six faces per array layer), but the call site passed that value through as the bgfx **side** and hardcoded array layer 0 -- so for a cube array anything above 5 became an out-of-range side and every update landed on layer 0 regardless.

Now decomposed to match the range the bounds check actually admits:

```cpp
texture->UpdateCube(static_cast<uint16_t>(layer / 6), static_cast<uint8_t>(layer % 6), mip, x, targetY, width, height, mem);
```

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1759,6 +2346,10 @@ namespace Babylon
         auto buffer{info[6].As<Napi::ArrayBuffer>()};
         uint32_t bufferOffset{info[7].As<Napi::Number>().Uint32Value()};
         uint32_t bufferLength{info[8].As<Napi::Number>().Uint32Value()};
+        // Optional cube-map face index (0-5). -1 (or absent) means a plain 2D read.
+        const int32_t faceIndex{(info.Length() > 9 && info[9].IsNumber()) ? info[9].As<Napi::Number>().Int32Value() : -1};
+        const bool isCubeFace{faceIndex >= 0};
+        const uint16_t srcZ{isCubeFace ? static_cast<uint16_t>(faceIndex) : static_cast<uint16_t>(0)};
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed that the missing upper bound was a real hole -- fixed in 73b33cf9.

I did **not** implement it as a flat 0-5 check though, because that would have broken 2D-array readback. Babylon.js passes this same argument as a *slice* index for 2D arrays, where values above 5 are perfectly legal (`BaseTexture.readPixels` in `baseTexture.pure.ts`):

```ts
if (this._texture.isCube || this._texture.is2DArray) {
    return engine._readTexturePixels(this._texture, width, height, faceIndex, level, buffer, ...);
}
return engine._readTexturePixels(this._texture, width, height, -1, level, buffer, ...);
```

So the bound is the texture's actual srcZ extent, which also keeps it consistent with the `UpdateTextureData` check:

```cpp
const uint16_t numLayers{texture->NumLayers() > 0 ? texture->NumLayers() : static_cast<uint16_t>(1)};
const uint32_t maxSrcZ{texture->IsCube() ? 6u * numLayers : numLayers};
if (isCubeFace && static_cast<uint32_t>(faceIndex) >= maxSrcZ)
{
    deferred.Reject(Napi::Error::New(env, "readTexture face/layer index is out of range for this texture.").Value());
}
```

Rejecting (rather than throwing) to stay consistent with the other validation failures in this promise-returning path.

Core/Graphics/Source/Texture.cpp
```diff
@@ -64,13 +64,43 @@ namespace Babylon::Graphics
...
+        m_width = width;
+        m_height = height;
+        m_depth = depth;
+        m_hasMips = hasMips;
+        m_numLayers = 1;
+        m_format = format;
+        m_flags = flags;
+        m_isCube = false;
+        m_is3D = false;
+        m_is3D = true;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Yep, copy/paste artifact -- removed the dead `m_is3D = false;` store in 73b33cf9.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -1790,6 +1818,74 @@ namespace Babylon::ShaderCompilerTraversers
...
+                // The sampler and lod operands are still referenced by the original texelFetch node.
+                // Reusing the same node pointers inside a second call (textureSize) would give those
+                // subtrees two parents in the AST, which later traversers (sampler splitting, SPIR-V
+                // generation) corrupt. Clone them so each reference is an independent node. If either
+                // operand is something we cannot safely clone, skip the flip rather than risk it.
+                TIntermTyped* samplerClone{CloneLeaf(sampler)};
+                TIntermTyped* lodClone{CloneLeaf(lod)};
+                if (samplerClone == nullptr || lodClone == nullptr)
+                {
+                    return coordinate;
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

`CloneLeaf` returns `nullptr` for anything that is not a symbol or a constant, and the flip is then skipped silently. `texelFetch(s, coord, someComputedLod)` samples with un-flipped Y and produces a wrong image with no diagnostic. Worth throwing or falling back rather than returning the coordinate unchanged.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -1790,6 +1818,74 @@ namespace Babylon::ShaderCompilerTraversers
...
+
+                // textureSize(sampler, lod).y - 1
+                TIntermTyped* sizeY{m_intermediate->addIndex(EOpIndexDirect, size, m_intermediate->addConstantUnion(1, loc), loc)};
+                sizeY->setType(intType);
+                TIntermTyped* maxY{m_intermediate->addBinaryMath(EOpSub, sizeY, m_intermediate->addConstantUnion(1, loc), loc)};
+
+                // coordinate.x and coordinate.y
+                TIntermTyped* coordinateX{m_intermediate->addIndex(EOpIndexDirect, coordinate, m_intermediate->addConstantUnion(0, loc), loc)};
+                coordinateX->setType(intType);
+                TIntermTyped* coordinateY{m_intermediate->addIndex(EOpIndexDirect, coordinate, m_intermediate->addConstantUnion(1, loc), loc)};
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

`sampler` and `lod` are cloned because reusing a node would give it two parents, but `coordinate` is referenced twice right here — once for `.x` on the line above, once for `.y` here — and is not cloned. By the argument in the comment at the top of this function, that is the same aliasing: the original reference is replaced by the returned aggregate, but two new ones are created in its place.

Is the coordinate safe for a reason the sampler and lod are not, or should it be cloned too?

Core/Graphics/Source/Texture.cpp
```diff
@@ -64,13 +64,42 @@ namespace Babylon::Graphics
...
+        m_handle = bgfx::createTexture3D(width, height, depth, hasMips, format, flags);
+        if (!bgfx::isValid(m_handle))
+        {
+            throw std::runtime_error{"Failed to create 3D texture"};
+        }
+
+        m_ownsHandle = true;
+        m_width = width;
+        m_height = height;
+        m_depth = depth;
```
bghgary, line 102: [Reviewed by Copilot on behalf of @bghgary]

`m_depth` is only ever assigned here, and `Dispose()` clears just the handle — so a `Texture` re-created as 2D or cube after having been 3D keeps the old depth and `Depth()` returns a stale value. The sibling `Create*` methods each reset `m_is3D` but none of them reset `m_depth`.

Each `Create*` hand-assigning its own subset of the metadata is what made the dead `m_is3D` store possible too; a shared reset would close both.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1454,6 +1713,91 @@ namespace Babylon
...
+
+        // Match the vertical orientation the base upload applies (PrepareImage flips the whole image when
+        // originBottomLeft ? invertY : !invertY). To land a sub-rectangle at the same place, flip it to the
+        // mirrored Y origin and reverse its rows so row 0 of the source lines up with the flipped base data.
+        const bool flip{bgfx::getCaps()->originBottomLeft ? invertY : !invertY};
+        const uint16_t targetY{flip ? static_cast<uint16_t>(mipHeight - y - height) : y};
+        const bgfx::Memory* mem{bgfx::alloc(requiredSize)};
+        if (flip)
+        {
+            const uint32_t rowBytes{requiredSize / height};
```
bghgary, line 1824: [Reviewed by Copilot on behalf of @bghgary]

`requiredSize / height` is only a row stride for uncompressed formats. Measured with `bgfx::calcTextureSize` on this branch (Win32 D3D11, RelWithDebInfo):

| format | size | storageSize | `requiredSize / height` | correct block-row stride |
|---|---|---|---|---|
| BC1 | 4x4 | 8 | **2** | 8 |
| BC1 | 16x16 | 128 | **8** | 32 |
| BC3 / BC7 | 4x4 | 16 | **4** | 16 |
| RGBA8 | 4x4 | 64 | 16 | 16 |

So for a 4x4 BC1 the loop makes 4 passes of 2 bytes over a single 8-byte block, reversing the `[color0:2][color1:2][indices:4]` layout into garbage; at 16x16 it reverses 16 8-byte chunks where the real layout is 4 block-rows of 32. Only the uncompressed row matches.

Reachable by default rather than in principle: `bgfx::getCaps()->originBottomLeft` is `0` on D3D11 (bgfx logs `origin top left`), so `flip = originBottomLeft ? invertY : !invertY` is true whenever `invertY` is left at its default.

`bgfx::updateTexture2D` also needs block-aligned `x`/`y`/`width`/`height` for compressed textures, which is not checked either. Rejecting compressed formats here would be enough, given the base upload path already handles them.

bghgary (changes requested):

Concerns inline.

Separately, the deferred cubemap un-exclusions point at a dead link. The description names [Babylon.js#18560](https://github.com/BabylonJS/Babylon.js/pull/18560) as the prerequisite, but that PR was closed on 2026-06-17 without merging, and the work did not land elsewhere: `nativeEngine.cubeTexture.pure.ts` on master was last touched by [#18441](https://github.com/BabylonJS/Babylon.js/pull/18441) (tree-shaking) and still throws `Cannot load cubemap because 6 files were not defined` for a single file. No open Babylon.js PR replaces it.

Core/Graphics/Source/Texture.cpp
bkaradzic-microsoft: Confirmed and fixed in a3241c16. `m_depth` was assigned only by `Create3D`, and `Dispose()` clears just the handle, so a `Texture` re-created as 2D or cube after having been 3D kept reporting the old depth.

I took the shared-reset route you suggested rather than adding one more hand-written assignment. There is now a private `ResetMetadata()` that returns every shape field to its default, and `Create2D`, `Create3D`, `CreateCube` and `Attach` all call it immediately after `Dispose()` before assigning the subset that applies to them. That also let me delete the scattered `m_isCube = false; m_is3D = false;` lines, which is what made the class of bug possible in the first place: any field a given path forgets is now defaulted rather than inherited from the previous, differently shaped texture.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: You were right, and thank you for pushing on it — the answer to "was the coordinate safe for a different reason?" is **no, it was not safe at all**. Fixed in a3241c16: the coordinate is now cloned too, with the original node supplying one reference and the clone the other, so every node ends up with exactly one parent.

The post-visit ordering only masked one symptom (a nested `texelFetch` being flipped twice). The aliased DAG still went on to sampler splitting and SPIR-V generation, which is precisely the situation the comment right above cites as the reason for cloning `sampler` and `lod`. I have updated that comment so it now explains why all three operands are cloned instead of implying the coordinate is special.

For coverage, I extended `ShaderCompilation.CompileComprehensiveGLSL` rather than relying on "the render suite still passes". That distinction turned out to matter: I instrumented the traverser and ran the full 301-test validation suite, and it only hit **three** `texelFetch` flips, all of them with a bare symbol coordinate and a constant lod — i.e. the suite gave the new code **zero** coverage. The test now exercises a constructor, a binary expression, a nested constructor over a float expression, a nested constructor over integer binaries, built-in calls, and a non-constant lod, which cover the unary, binary and aggregate clone paths and a symbol lod. Verified by the same instrumentation that all of those shapes now reach `CloneExpression`.

While writing that test I hit two unrelated constructs that the WEBMIN builds reject — integer multiply (the `OpIMul` case already called out in this PR) and bitwise `&`. I kept them out of the test and left a comment saying why, so nobody re-adds them and blames the flip.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: Confirmed, and fixed in a3241c16 — `CloneLeaf` is gone, replaced by `CloneExpression`, a structural deep clone that handles symbols, constant unions, binary, unary and aggregate nodes. It copies each node's operator, type and source location verbatim rather than rebuilding through `TIntermediate::add*`, so no constant folding or type promotion can make the copy diverge from the original. Anything it genuinely cannot copy now throws `std::runtime_error` instead of returning `nullptr`, so we can never again silently emit an unflipped coordinate.

Two notes from checking how much this actually mattered:

I surveyed the 133 `texelFetch` call sites in the Babylon.js shaders. `lod` is almost always a constant (`0`, `1`) or a bare symbol, so the old leaf-only clone was usually enough for it. But **coordinates are frequently complex** — `ivec2(gl_FragCoord.xy)`, `ivec2(vUV * texSize)`, `coord + ivec2(1,1)`, `ivec2(t0 % w, t0 / w)` are all common. So the `nullptr` path was not an exotic corner; the reason it did not show up as failures is the second point below.

One behaviour I want to flag explicitly rather than have it found later: cloning an `EOpFunctionCall` aggregate duplicates a side-effecting call. That is not a regression — it is exactly what the `ProcessSamplerFlip` macro this code replaced already did, since it expanded the coordinate operand twice — but it is inherent to expressing the flip as `ivec2(c.x, size.y - 1 - c.y)` and worth knowing about.

Plugins/NativeEngine/Source/NativeEngine.cpp
bkaradzic-microsoft: Confirmed, and your arithmetic is right. I checked the block info in bimg (`image.cpp:30`, `{4, 4, 4, 8, ...}` for BC1 — a 4x4 block in 8 bytes): at 4x4 `requiredSize / height` gives `8/4 = 2` against a real block row of 8, and at 16x16 it gives `128/16 = 8` against a correct 32.

It is also reachable by default rather than theoretical: `originBottomLeft` is only ever set true in `renderer_gl.cpp`, so on D3D11 it is 0 and `flip = !invertY` is true whenever `invertY` is left false.

Fixed in a3241c16 by rejecting block-compressed formats outright, as you suggested, rather than fixing the stride. Fixing the stride would not be enough: mirroring block rows cannot flip a texture vertically without re-encoding the texel rows packed inside each block. BC1 happens to keep its 4 index bytes in row order, but BC3 alpha and BC7 do not, so a correct flip means a full decode/re-encode. bgfx also wants block-aligned `x`/`y`/`width`/`height` here, which the code was not checking either; rejecting the formats moots that too.

The base upload path (`loadTexture` → `PrepareImage`) already handles compressed data, so this is not a capability loss, and the error message says so. The check is `texture->Format() < bgfx::TextureFormat::Unknown`, which works because every block-compressed format sorts before `Unknown` in bgfx's enum (`bgfx.h:222`, "Compressed formats above"). That keeps it bimg-free, which matters here since this file has to build in configurations where bimg is not linked.

bkaradzic-microsoft:
Thanks for the review — all four were real, and all four are fixed in a3241c16. Replies are on the individual threads; summary here.

- **`texelFetch` operand cloning.** `CloneLeaf` is replaced by `CloneExpression`, a structural deep clone (symbol / constant union / binary / unary / aggregate) that throws rather than silently returning an unflipped coordinate. The coordinate is now cloned as well as the sampler and lod.
- **`Texture` metadata.** New shared `ResetMetadata()` called from `Create2D`, `Create3D`, `CreateCube` and `Attach`, so `m_depth` (and anything added later) cannot go stale.
- **`updateTextureData`.** Block-compressed formats are now rejected with a clear error instead of being fed through a row-reversal loop with a bogus stride.
- **Description.** Corrected. BabylonJS/Babylon.js#18560 was closed unmerged on 2026-06-17 and the work landed nowhere, so the deferred cubemap un-exclusions have no current prerequisite PR — I have said that plainly rather than pointing at a dead link.

Validation: full render suite 301/301 (no regressions), `Apps/UnitTests` 17/17.

One side-finding worth its own issue, which I am **not** changing here. `Dependencies/CMakeLists.txt` sets `ENABLE_GLSLANG_WEBMIN ON`, and in that configuration glslang compiles out `TParseContextBase::error()` (`ParseHelper.h`, guarded by `#if !defined(GLSLANG_WEB) || defined(GLSLANG_WEB_DEVEL)`). The error *count* still increments, so `getInfoLog()` returns a bare:

```
ERROR: 3 compilation errors.  No code generated.
```

with no file, line, or reason — which is what every BN shader compile failure looks like today. I lost a fair amount of time to this while writing the test above; the failing constructs were only identifiable by bisecting the shader source. This is the same shape of problem as the `SPIRV_CROSS_WEBMIN` opcode stripping already called out in this PR: a size-optimised dependency silently discarding diagnostics. Happy to file it separately, or to send a small follow-up that turns on `GLSLANG_WEB_DEVEL` (or otherwise restores the message text) if you think that is the right trade against binary size.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -1790,6 +1819,125 @@ namespace Babylon::ShaderCompilerTraversers
...
+                    auto* clone = new TIntermUnary{unary->getOp()};
+                    clone->setOperand(CloneExpression(unary->getOperand()));
+                    clone->setType(unary->getType());
+                    clone->setLoc(unary->getLoc());
+                    return clone;
+                }
+
+                if (TIntermAggregate* aggregate = node->getAsAggregate())
+                {
+                    auto* clone = new TIntermAggregate{aggregate->getOp()};
```
bghgary: This deep clone regresses `MacOS_Sanitizers`: it fails on this commit and passes on both `5d8c2e1a` and master `aa244ec9`.

```
intermediate.h:1707: runtime error: load of value 190, which is not a valid value for type 'bool'
  isUserDefined() <- CloneExpression:1931 <- CloneExpression:1906 <- FlipVerticalTexelCoordinate:1846
```

glslang initializes `userDefined` in `TIntermAggregate()` but not in `TIntermAggregate(TOperator)`, so clones built here carry it uninitialized (190 = 0xBE fill). Default-constructing and setting the operator afterwards fixes that:

```suggestion
                    auto* clone = new TIntermAggregate{};
                    clone->setOperator(aggregate->getOp());
```

The read at 1931 still needs handling — it dereferences the same member on the *source* aggregate, so copying it unconditionally would inherit the same garbage.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -288,25 +289,41 @@ namespace Babylon
...
+                            : bimg::TextureFormat::RGBA8;
+
                     bimg::ImageContainer* oldImage{image};
-                    image = bimg::imageConvert(&allocator, bimg::TextureFormat::RGBA32F, *image, false);
+                    image = bimg::imageConvert(&allocator, dstFormat, *image, false);
                     bimg::imageFree(oldImage);
+
+                    if (image == nullptr)
+                    {
+                        return nullptr;
```
bghgary: `PrepareImage` can now return `nullptr`, but only the `loadTexture` call site checks it (L1650). The other three pass the result straight into an unconditional dereference: L1712 into `LoadTextureFromImage` (L334), and L2175 / L2222 into `LoadCubeTextureFromImages` (L368). Guard those three, or have `PrepareImage` throw instead of returning `nullptr`.

The two other conversions in this function are unchecked as well: the sRGB `imageConvert` at L284 falls through to `image->m_format` at L299, and the `imageGenerateMips` result at L323 is returned without a check.

bghgary (changes requested):

Earlier four look good. Two new concerns inline.

bkaradzic-microsoft:
The `MacOS_Sanitizers` failure on the previous push was a real bug in my change, now fixed in e8ac11c5.

UBSAN caught it:

```
glslang/Include/intermediate.h:1707:43: runtime error: load of value 112, which is not a valid value for type 'bool'
    #0 glslang::TIntermAggregate::isUserDefined() intermediate.h:1707
    #1 FlipSamplerCoordinatesTraverser::CloneExpression(...) ShaderCompilerTraversers.cpp:1931
    #2 FlipSamplerCoordinatesTraverser::CloneExpression(...) ShaderCompilerTraversers.cpp:1906
    #3 FlipSamplerCoordinatesTraverser::CloneExpression(...) ShaderCompilerTraversers.cpp:1915
    #4 FlipSamplerCoordinatesTraverser::FlipVerticalTexelCoordinate(...) ShaderCompilerTraversers.cpp:1846
```

`TIntermAggregate` has two constructors, and they don't initialize the same fields:

```cpp
TIntermAggregate() : TIntermOperator(EOpNull), userDefined(false), pragmaTable(nullptr) { }
TIntermAggregate(TOperator o) : TIntermOperator(o), pragmaTable(nullptr) { }   // userDefined left indeterminate
```

`CloneExpression` used the operator-taking overload and then called `isUserDefined()` on the node being copied. glslang gets away with the same pattern because it only ever sets that flag in one place — `ParseHelper.cpp:1480`, on an `EOpFunctionCall` — and only reads it back at one place, `GlslangToSpv.cpp:3001`, also under `EOpFunctionCall`. So for every aggregate glslang builds through `setAggregateOperator` the field is either initialized or never read.

My clone broke both halves of that invariant: it constructed aggregates through the overload that skips the flag, and then read the flag for every operator rather than just function calls. The recursion in the stack trace above is how it became reachable — post-order traversal rewrites an inner `texelFetch` before the outer one, so the outer clone walks a subtree that `CloneExpression` itself produced, and the load hits the uninitialized byte (`112`).

The fix constructs clones with the default constructor plus `setOperator`, so the flag is always well defined, and restricts the `isUserDefined()` read to `EOpFunctionCall`. It also copies `operationPrecision` and the qualifier list, which the clone was previously dropping.

Worth noting why this only showed up on one leg: macOS builds with `-fsanitize=address,undefined` and `halt_on_error=1`, whereas the Windows sanitizer job is ASan-only (`CMakeLists.txt` has the explicit "MSVC only supports AddressSanitizer" comment). This is a pure-UB bug with no memory error, so only the macOS leg could see it — the other 31 checks passed on the bad commit.

Verified locally: `UnitTests` 17/17, Playground validation suite 301/301 (`skipped=419`, `missingRef=0`).

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: Fixed in `e8ac11c5`, which landed after this review.

You had it exactly right about the cause. `TIntermAggregate`'s operator-taking constructor never initializes `userDefined`; only the default constructor does. So `new TIntermAggregate{aggregate->getOp()}` produced a node whose flag was indeterminate, and the subsequent `if (aggregate->isUserDefined())` read it. UBSAN caught it on macOS because that is the only leg building with `-fsanitize=undefined -fno-sanitize-recover=all`.

The clone is now built with the default constructor and the operator assigned afterwards, which is the only ordering that yields a well-defined flag:

```cpp
auto* clone = new TIntermAggregate{};
clone->setOperator(aggregate->getOp());
```

The read is also now gated on the node actually being a function call, since `userDefined` is only meaningful on (and only ever initialized by glslang for) `EOpFunctionCall`:

```cpp
if (aggregate->getOp() == EOpFunctionCall && aggregate->isUserDefined())
```

Worth noting why this was reachable at all: the traversal is post-order, so a `texelFetch` nested inside another `texelFetch`'s coordinate argument is rewritten first, and the outer rewrite then clones a subtree that this same function produced. That is the path that reaches a clone-of-a-clone and reads the flag.

While in here I also fixed three other fields the clone was silently dropping -- `setOperationPrecision` on the binary, unary and aggregate paths, and `getQualifierList()` on the aggregate path. Those were not undefined behavior, but a clone that loses precision qualifiers can change generated code, so they belong in the same fix.

`MacOS_Sanitizers` passes at this head.

Plugins/NativeEngine/Source/NativeEngine.cpp
bkaradzic-microsoft: Fixed in `c77c6fab`. Rather than add the three missing null checks, I made `PrepareImage` throw, so the safe behavior is the default instead of something each new call site has to remember.

It now throws a `std::runtime_error` at all three failure points -- the sRGB `imageConvert`, the mip-prep `imageConvert`, and `imageGenerateMips` -- with messages naming the source and destination formats via `bimg::getName`, since "conversion failed" alone is not much to debug with. The contract is documented on the function.

That works out cleanly for the callers because of where they run. `LoadTexture`, `LoadCubeTexture` and `LoadCubeTextureWithMips` all execute inside `arcana::make_task`, whose continuation already takes `arcana::expected<void, std::exception_ptr>` and forwards to the JavaScript `onError` callback, so a throw is reported to JS through the path that already existed -- and the now-redundant null check at the `LoadTexture` call site is gone.

`LoadRawTexture` is the exception: it runs synchronously on the JS thread, where a raw `std::exception` would not be translated into a JS error. That one catches and rethrows as a `Napi::Error`.

Net effect is that the three previously unchecked conversions can no longer produce a null dereference, and a failed conversion surfaces as a JS error naming the formats involved rather than as a crash.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1449,11 +1724,121 @@ namespace Babylon
...
+        uint32_t mipWidth{static_cast<uint32_t>(texture->Width()) >> mip};
+        uint32_t mipHeight{static_cast<uint32_t>(texture->Height()) >> mip};
+        if (mipWidth == 0)
+        {
+            mipWidth = 1;
+        }
+        if (mipHeight == 0)
+        {
+            mipHeight = 1;
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in `f6af7424`, and you are right that the clamp reads better -- but the clamp alone would not have been enough, so the fix does both.

Clamping the *result* of the shift does not help when the shift itself is the undefined operation: `width >> 40` is UB before any clamp gets to look at the value. So the level is now range-checked first, against the texture's real mip count:

```cpp
bgfx::TextureInfo mipChainInfo;
bgfx::calcTextureSize(mipChainInfo, texture->Width(), texture->Height(), 1, texture->IsCube(), texture->HasMips(), 1, texture->Format());
if (mip >= mipChainInfo.numMips) { /* reject */ }
```

`bgfx::calcTextureSize` rather than `bimg::imageGetNumMips` because bgfx is always linked while bimg is not in builds without image loading, and this file already sizes uploads that way. It also handles the `mip > 0` on a mip-less texture case for free, since `numMips` is then 1.

Once the level is validated the shifts are well defined, and the "floor the extent at 1" logic collapses to the clamp you suggested:

```cpp
const uint32_t mipWidth{std::max(1u, static_cast<uint32_t>(texture->Width()) >> mip)};
const uint32_t mipHeight{std::max(1u, static_cast<uint32_t>(texture->Height()) >> mip)};
```

Applied to `readTexture` on L2453 as well, which had the same unvalidated shift. That one rejects through the existing `deferred.Reject` chain instead of throwing, to match how the rest of that function reports bad arguments, and the blit decision now compares against these mip extents rather than recomputing raw shifts inline.

bkaradzic-microsoft:
@bghgary — following up on your note that `updateTextureDirectly` now lands here rather than in #1778. That pointer led somewhere: **I have removed it from this PR** (`da895eb3`).

It was added to serve a Babylon.js change that parsed single-file DDS/KTX/KTX2 cubemaps in JS via the shared texture loaders. Those loaders upload one `(face, mip)` at a time, WebGL `texImage2D`-style, so they needed a native sink for the individual uploads.

That approach has since been replaced by parsing the container natively in `loadCubeTexture` and returning the diffuse SH directly — commit `49f6357f` in this PR, which stays. The native parse has no per-`(face, mip)` uploads, so `updateTextureDirectly` ended up with no callers at all: none in this repository, and none on the Babylon.js side. The JS PR that would have used it is now closed.

Shipping it unused seemed like the wrong trade. It is a public JS-facing engine method, so once released it becomes surface area we are obliged to keep working, and its lazy create-the-texture-on-first-upload rule is a texture lifetime behavior that exists nowhere else in `NativeEngine` — worth having only if something actually needs it.

The removal is a clean inverse of the commit that added it (137 lines, `NativeEngine.cpp` + `NativeEngine.h`, no residual references). Full validation suite on this branch after the removal: **301 ran, 301 passed, 0 failed**.

This also resolves the sampling-mode divergence you flagged on the JS side, since the `_setCubeMapTextureParams` stub that carried it is gone along with the rest of that approach.

Also pushed since your last review: `c77c6fab` makes `PrepareImage` throw rather than return `nullptr` (your inline note about the unchecked call sites), and `f6af7424` range-checks the JS-controlled mip level before it is used as a shift distance. Ready for another look when you have time.

Plugins/NativeEngine/Source/NativeEngine.cpp
bghgary: Three replies on this PR posted as the literal text `@reply1.txt`, `@reply2.txt` and `@reply3.txt` — `gh api -f body=@file` does not expand the file, only `-F` does, so those bodies were lost.

bghgary (commented):

All findings fixed at this head. One note inline.

Plugins/NativeEngine/Source/NativeEngine.cpp
bkaradzic-microsoft: Thank you for catching that, and sorry for the noise -- you are exactly right about the cause. I used `gh api -f body=@file`, and `-f` always treats its value as a literal string; only `-F` reads `@file` from disk. So three replies here went up as the literal filenames and the actual content was lost.

I have restored all three in place, so the threads now read correctly:

- **`ShaderCompilerTraversers.cpp` / `MacOS_Sanitizers`** — fixed in `e8ac11c5`. Your diagnosis was right: `TIntermAggregate`'s operator-taking constructor never initializes `userDefined`, only the default constructor does, so the clone read indeterminate memory. It is now built with the default constructor plus `setOperator()`, and the read is gated on `getOp() == EOpFunctionCall` since the flag is only meaningful there. I also picked up three fields the clone was silently dropping (`setOperationPrecision` on the binary/unary/aggregate paths, and `getQualifierList()`).

- **`PrepareImage` returning `nullptr`** — fixed in `c77c6fab` by making it throw instead, so safety is the default rather than something each call site has to remember. The three async callers already run inside `arcana::make_task` and report through the existing `onError` path; `LoadRawTexture` runs on the JS thread and catches/rethrows as `Napi::Error`.

- **Unvalidated mip level** — fixed in `f6af7424`. Range-checked against the real mip count before the shift, since clamping the result does not help when the shift itself is the UB.

A fourth reply on the Babylon.js side was lost the same way; that one is restored too.

Also since your review: `da895eb3` removes `updateTextureDirectly`, per the separate comment above.

bghgary (approved):

LGTM. All findings fixed at this head.

PR #1816, opened 2026-08-05
Fix shape conversion for uniforms used as return expressions

VicenteCartas:
> 🤖 *This PR was created by the create-pr skill.*

## Summary

- handle `TIntermBranch` parents in `injectShapeConversion`, so a uniform used directly as a `return` expression can be rewritten
- skip uniforms that are already `vec4` in `UniformTypeChangeTraverser` as a fast path
- add regression tests for a returned uniform and for an existing `vec4` uniform array

## Root cause

`UniformTypeChangeTraverser` widens loose scalar, `vec2`, and `vec3` uniforms to `vec4`, then injects a shape conversion and replaces the original node in its parent. `injectShapeConversion` only handled aggregate, binary, unary, and selection parents, and threw for anything else:

```
Cannot replace symbol: node type handler unimplemented
```

The failing trigger is **the parent node type, not the vector width**. For `return someUniform;` the parent is a `TIntermBranch`, which was unhandled. Adding a branch handler is the actual fix.

An earlier revision of this PR attributed the failure to `vec4` array reshaping. That was wrong: `return someVec3;` fails identically, and a `vec3` must be widened so it cannot be skipped. Thanks to @bghgary for catching this.

## Fast path (not the fix)

The `type.getVectorSize() < 4` guard is retained because widening an existing `vec4` is a no-op — the type is unchanged and `addShapeConversion` returns the node as-is, so the rewrite is pointless AST churn. It is an optimization, not the fix, and it does not resolve the underlying bug on its own.

## Known remaining gaps

The same class of failure still exists for parents that glslang exposes without a setter, so they cannot be fixed here:

| Construct | Parent node | Status |
| --- | --- | --- |
| `return u;` | `TIntermBranch` | fixed by this PR |
| `while (u)` | `TIntermLoop` | still throws — `TIntermLoop` has `getTest()` but no setter |
| `switch (u)` | `TIntermSwitch` | still throws — `TIntermSwitch` has `getCondition()` but no setter |

These need an upstream glslang change to expose setters, so the underlying bug should not be considered fully closed.

## Validation

Built `UnitTests` from a clean worktree based on upstream `master`. Win32 D3D11 Debug.

| Run | `ExistingVec4UniformArrayIsNotReshaped` | `ReturnedUniformIsShapeConverted` |
| --- | --- | --- |
| This PR (branch handler + guard) | pass | pass |
| Guard reverted, branch handler kept | pass | pass |
| Branch handler reverted, guard kept | pass | **fail** — `Vec3 return effect should compile synchronously` |

Row 2 shows the guard is not what fixes the bug. Row 3 shows the branch handler is. `ShaderCompilation.CompileComprehensiveGLSL` passes in all three configurations.

### DIFF
--- a/Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
+++ b/Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
@@ -123,3 +123,57 @@ TEST(ShaderCompilation, ExistingVec4UniformArrayIsNotReshaped)
 
     device.FinishRenderingCurrentFrame();
 }
+
+TEST(ShaderCompilation, ReturnedUniformIsShapeConverted)
+{
+    Babylon::Graphics::Device device{g_deviceConfig};
+
+    device.StartRenderingCurrentFrame();
+
+    Babylon::AppRuntime::Options options{};
+    options.UnhandledExceptionHandler = [](const Napi::Error& error) {
+        std::cerr << "[Uncaught Error] " << Napi::GetErrorString(error) << std::endl;
+        std::quick_exit(1);
+    };
+
+    Babylon::AppRuntime runtime{options};
+    runtime.Dispatch([&device](Napi::Env env) {
+        device.AddToJavaScript(env);
+
+        Babylon::Polyfills::Console::Initialize(env, [](const char* message, auto) {
+            std::cout << message << std::endl;
+        });
+        Babylon::Polyfills::Window::Initialize(env);
+        Babylon::Plugins::NativeEngine::Initialize(env);
+    });
+
+    Babylon::ScriptLoader loader{runtime};
+    loader.LoadScript("app:///Assets/babylon.max.js");
+    loader.Eval(R"(
+        const engine = new BABYLON.NativeEngine();
+        engine.getCaps().parallelShaderCompile = null;
+        const effect = engine.createEffect({
+            vertexSource: `
+                attribute vec2 position;
+                void main() { gl_Position = vec4(position, 0.0, 1.0); }
+            `,
+            fragmentSource: `
+                precision highp float;
+                uniform vec3 tint;
+                vec3 readTint() { return tint; }
+                void main() { gl_FragColor = vec4(readTint(), 1.0); }
+            `
+        }, ["position"], ["tint"], []);
+        if (!effect.isReady()) { throw new Error("Vec3 return effect should compile synchronously"); }
+        effect.dispose();
+        engine.dispose();
+    )", "returned_uniform_shape_conversion_test.js");
+
+    std::promise<void> done{};
+    loader.Dispatch([&done](Napi::Env) {
+        done.set_value();
+    });
+    done.get_future().get();
+
+    device.FinishRenderingCurrentFrame();
+}
--- a/Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
+++ b/Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
@@ -69,3 +69,57 @@ TEST(ShaderCompilation, CompileComprehensiveGLSL)
 
     device.FinishRenderingCurrentFrame();
 }
+
+TEST(ShaderCompilation, ExistingVec4UniformArrayIsNotReshaped)
+{
+    Babylon::Graphics::Device device{g_deviceConfig};
+
+    device.StartRenderingCurrentFrame();
+
+    Babylon::AppRuntime::Options options{};
+    options.UnhandledExceptionHandler = [](const Napi::Error& error) {
+        std::cerr << "[Uncaught Error] " << Napi::GetErrorString(error) << std::endl;
+        std::quick_exit(1);
+    };
+
+    Babylon::AppRuntime runtime{options};
+    runtime.Dispatch([&device](Napi::Env env) {
+        device.AddToJavaScript(env);
+
+        Babylon::Polyfills::Console::Initialize(env, [](const char* message, auto) {
+            std::cout << message << std::endl;
+        });
+        Babylon::Polyfills::Window::Initialize(env);
+        Babylon::Plugins::NativeEngine::Initialize(env);
+    });
+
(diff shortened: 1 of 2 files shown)


VicenteCartas:
Self code review completed by the code-review workflow. No issues found.

Representative inputs were traced through the guard: scalar, vec2, and vec3 uniforms remain widened; matrices, samplers, structs, blocks, and existing vec4/ivec4/uvec4/bvec4 uniforms remain untouched. The focused vec4-array regression and existing comprehensive shader compilation test both pass locally on D3D11/Win32.

bkaradzic-microsoft (approved):

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -319,6 +319,7 @@ namespace Babylon::ShaderCompilerTraversers
                 const auto basic = type.getBasicType();
                 if (type.getQualifier().isUniformOrBuffer()
                     && !type.isMatrix()
+                    && type.getVectorSize() < 4
```
bghgary, line 324: Please say in the description that this is a fast path rather than the fix, so the underlying bug isn't taken as closed.

The trigger is the parent node type, not the vector width. `injectShapeConversion` handles aggregate, binary, unary and selection, then throws; for `return values[0];` the parent is a `TIntermBranch`. Measured on Win32 D3D11 RelWithDebInfo:

| shader | guard | result |
| --- | --- | --- |
| `return values[0];` (`uniform vec4 values[2]`) | removed | fails |
| `gl_FragColor = values[0];` (same uniform) | removed | compiles |
| `return tint;` (`uniform vec3 tint`) | **applied** | **fails** |

Row 3 is the point: a `vec3` must be widened, so it can't be skipped, and it still fails with this PR. `while (flagUniform)` and `switch (modeUniform)` fail the same way. Adding `else if (auto* branch = parent->getAsBranchNode()) { branch->setExpression(shapeConversion); }` makes the new test pass with this guard removed entirely.

Also, the comment above this condition still lists only matrices, samplers, blocks and structs, and no longer matches the code.

Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
```diff
@@ -69,3 +69,57 @@ TEST(ShaderCompilation, CompileComprehensiveGLSL)
 
     device.FinishRenderingCurrentFrame();
 }
+
+TEST(ShaderCompilation, ExistingVec4UniformArrayIsNotReshaped)
```
bghgary, line 73: Please add the negative control to Validation — as written it lists only the passing run, which can't distinguish a real regression test from one that would pass either way. Confirmed here on Win32 D3D11 RelWithDebInfo: passes at head, and reverting only the guard makes it fail with `Vec4 uniform array effect should compile synchronously`.

Note it also passes if `injectShapeConversion` handles `TIntermBranch` with the guard removed, so it pins the symptom rather than this fix.

bghgary (changes requested):

The change is worth keeping, but the description and title attribute the failure to the wrong thing — please update them. The bug is not vec4-array reshaping: `return someVec3;` fails identically, and this PR does not fix that. Details inline.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
VicenteCartas: [Responded by Copilot on behalf of VicenteCartas]

You're right, and thanks for the precise repro — the original root cause was wrong.

I reproduced your row 3 locally on Win32 D3D11: `return tint;` with `uniform vec3 tint` fails **with the guard applied**, since a `vec3` must be widened and cannot be skipped. That rules out vector width as the trigger and confirms it is the parent node type.

Fixed properly by handling the branch parent, as you suggested:

```cpp
else if (auto* branch = parent->getAsBranchNode())
{
    branch->setExpression(shapeConversion);
}
```

Controls, all on Win32 D3D11 Debug:

| Configuration | `ExistingVec4UniformArray...` | `ReturnedUniformIsShapeConverted` |
| --- | --- | --- |
| branch handler + guard | pass | pass |
| guard reverted, handler kept | pass | pass |
| handler reverted, guard kept | pass | **fail** |

So the guard is not the fix. I kept it purely as a fast path (widening an existing `vec4` is a no-op that only churns the AST) and the description now says so explicitly.

On `while` and `switch`: I confirmed those are the same bug but they are **not** fixable in this repo. `TIntermLoop` exposes `getTest()` and `TIntermSwitch` exposes `getCondition()`, but neither has a public setter, so `injectShapeConversion` cannot replace those nodes without an upstream glslang change. I listed both in a "Known remaining gaps" table in the description so the underlying bug isn't taken as closed.

Also updated the stale comment above the condition to match the code.

Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
VicenteCartas: [Responded by Copilot on behalf of VicenteCartas]

Added the negative controls to the Validation section, and they now cover both directions rather than just the passing run:

| Configuration | `ExistingVec4UniformArray...` | `ReturnedUniformIsShapeConverted` |
| --- | --- | --- |
| this PR | pass | pass |
| guard reverted, branch handler kept | pass | pass |
| branch handler reverted, guard kept | pass | **fail** — `Vec3 return effect should compile synchronously` |

You were also right that the original test pinned the symptom rather than the fix — it passes with the guard removed once the branch handler is present. I added `ShaderCompilation.ReturnedUniformIsShapeConverted`, which uses `uniform vec3 tint` with `return tint;`. That case cannot be skipped by any width guard, so it fails without the branch handler and is a real regression test for the actual root cause. The `vec4` array test is retained as coverage for the fast path.

bghgary (approved):

LGTM. Verified locally at 406ac461: both tests pass on Win32 D3D11 RelWithDebInfo.

PR #1823, opened 2026-08-07
Playground: load earcut and procedural textures, and provide `name` to test scripts

bkaradzic-microsoft:
Three test failures on the native Playground turned out to have the same root cause: the native harness provides a smaller browser environment than the web Playground, so scripts that work on babylonjs.com die here on a missing global. None of it is renderer-specific, so this fixes the same scenes on both the bgfx and the WebGPU/Dawn paths.

### `earcut is not defined`

`PolygonMeshBuilder` resolves `earcut` as a *global* at triangulation time rather than importing it, so any scene that builds a polygon throws. The web Playground loads earcut as a separate script; we never did.

Pinned to `^2.2.4` deliberately, for two reasons — 3.x does ship `dist/earcut.min.js`, so "no UMD build" is *not* one of them:

- **The shape of the global changed.** 2.x assigns the function itself (`.earcut = e()`); 3.x assigns a namespace object (`.earcut = {}`). `PolygonMeshBuilder` calls `earcut(...)`, so 3.x throws at triangulation time.
- **It matches the environment these tests replicate.** `cdn.babylonjs.com/earcut.min.js`, which `playground/index.html` loads, is itself 2.x (6,519 bytes, global-is-the-function).

Thanks to @bghgary for catching the original, incorrect rationale.

### `BABYLON.WoodProceduralTexture is not a constructor`

The procedural texture classes live in their own `babylonjs-procedural-textures` package, which was never in the bundle. The existing config even documents this â€” one test is disabled with the reason *"BABYLON.NormalMapProceduralTexture is not included in the native script bundle"*.

Wired into `BABYLON_SCRIPTS`, `PlaygroundScripts.cpp`, the Android asset copy, and `getNightly.js` so it stays in step with the other Babylon packages. Pinned to `9.15.0` to match every other `babylonjs-*` entry in the lock file â€” on `^9.15.0` npm resolves it to 9.20.0 and then nests a second full copy of `babylonjs` under it.

### `name is not defined`

Some scripts reference a bare `name` without declaring it. In a browser that silently resolves to `window.name` (`""`), so the bug is invisible on the web but throws here.

Fixed by declaring `var name = ""` in the function that *encloses* the `eval`. Because those are **direct** evals, the evaluated script sees that scope and resolves `name` exactly as it would on the web â€” without leaking a real global, and without rewriting the script text (so line numbers in stack traces and any `"use strict"` prologue are preserved).

Defining an actual global `name` is **not** an option, and I tried: it breaks the Babylon UMD bundles, which probe for `name` while being evaluated. `babylonjs.loaders.js` then throws `Error: Invalid argument` at load time and *every* test fails.

### Depends on

BabylonJS/Babylon.js#18799 must land first. The nightly's "Update scripts from snapshot" step refills every `babylon*` file in the Native artifact from the CDN snapshot, but the downlevel that follows names only `babylon.max.js` — so a newly added bundle is refilled and then run un-downleveled. #18799 passes the directory instead, which is correct both before and after this PR.

### Validation

Full 680-test sweep on the Dawn/WebGPU backend, before vs. after:

| | before | after |
|---|---|---|
| PASS | 622 | **625** |
| FAIL | 56 | 53 |

**Zero regressions.** `Polygon` flips to PASS outright; `Procedural textures` and `Show all procedural textures` go from hard exceptions to rendering (they still miss the pixel threshold, tracked separately). The other two deltas (`Iridescence NME`, `Sprite maps`) are known order-dependent flakes, not caused by this change.

### DIFF
--- a/Apps/Playground/Shared/PlaygroundScripts.cpp
+++ b/Apps/Playground/Shared/PlaygroundScripts.cpp
@@ -35,10 +35,8 @@ namespace Playground
         runtime.LoadScript("app:///Scripts/ammo.js");
         // Commenting out recast.js for now because v8jsi is incompatible with asm.js.
         // runtime.LoadScript("app:///Scripts/recast.js");
-        // PolygonMeshBuilder resolves `earcut` as a global at triangulation time rather than
-        // importing it, so it has to be present before any polygon scene runs. The browser
-        // Playground loads it the same way; without it those scenes fail with
-        // `ReferenceError: earcut is not defined`.
+        // PolygonMeshBuilder resolves `earcut` as a global at triangulation time, so without this
+        // any polygon scene fails with `ReferenceError: earcut is not defined`.
         runtime.LoadScript("app:///Scripts/earcut.min.js");
         runtime.LoadScript("app:///Scripts/babylon.max.js");
         // Load addons right after babylon.max.js so addons init sees a fully
@@ -49,8 +47,6 @@ namespace Playground
         runtime.LoadScript("app:///Scripts/babylon.gui.js");
         runtime.LoadScript("app:///Scripts/meshwriter.min.js");
         runtime.LoadScript("app:///Scripts/babylonjs.serializers.js");
-        // Procedural texture classes (WoodProceduralTexture, BrickProceduralTexture, ...) live
-        // in their own package, which the browser Playground also loads separately.
         runtime.LoadScript("app:///Scripts/babylonjs.proceduralTextures.js");
     }
 
--- a/Apps/Playground/Shared/PlaygroundScripts.cpp
+++ b/Apps/Playground/Shared/PlaygroundScripts.cpp
@@ -35,6 +35,11 @@ namespace Playground
         runtime.LoadScript("app:///Scripts/ammo.js");
         // Commenting out recast.js for now because v8jsi is incompatible with asm.js.
         // runtime.LoadScript("app:///Scripts/recast.js");
+        // PolygonMeshBuilder resolves `earcut` as a global at triangulation time rather than
+        // importing it, so it has to be present before any polygon scene runs. The browser
+        // Playground loads it the same way; without it those scenes fail with
+        // `ReferenceError: earcut is not defined`.
+        runtime.LoadScript("app:///Scripts/earcut.min.js");
         runtime.LoadScript("app:///Scripts/babylon.max.js");
         // Load addons right after babylon.max.js so addons init sees a fully
         // constructed BABYLON global.
@@ -44,6 +49,9 @@ namespace Playground
         runtime.LoadScript("app:///Scripts/babylon.gui.js");
         runtime.LoadScript("app:///Scripts/meshwriter.min.js");
         runtime.LoadScript("app:///Scripts/babylonjs.serializers.js");
+        // Procedural texture classes (WoodProceduralTexture, BrickProceduralTexture, ...) live
+        // in their own package, which the browser Playground also loads separately.
+        runtime.LoadScript("app:///Scripts/babylonjs.proceduralTextures.js");
     }
 
     std::function<void(Babylon::Embedding::LogLevel, std::string_view)>
--- a/Apps/package.json
+++ b/Apps/package.json
@@ -6,7 +6,7 @@
     "UnitTests/JavaScript"
   ],
   "scripts": {
-    "getNightly": "node scripts/getNightly.js && npm run downlevel:native-scripts -- node_modules/babylonjs/babylon.max.js",
+    "getNightly": "node scripts/getNightly.js",
     "downlevel:native-scripts": "node scripts/downlevelNativeScripts.mjs"
   },
   "devDependencies": {
@@ -19,7 +19,7 @@
     "babylonjs-gui": "^9.15.0",
     "babylonjs-loaders": "^9.15.0",
     "babylonjs-materials": "^9.15.0",
-    "babylonjs-procedural-textures": "^9.15.0",
+    "babylonjs-procedural-textures": "9.15.0",
     "babylonjs-serializers": "^9.15.0",
     "earcut": "^2.2.4",
     "jsc-android": "^241213.1.0",
--- a/Apps/package.json
+++ b/Apps/package.json
@@ -19,7 +19,9 @@
     "babylonjs-gui": "^9.15.0",
     "babylonjs-loaders": "^9.15.0",
     "babylonjs-materials": "^9.15.0",
+    "babylonjs-procedural-textures": "^9.15.0",
     "babylonjs-serializers": "^9.15.0",
+    "earcut": "^2.2.4",
     "jsc-android": "^241213.1.0",
     "v8-android": "^7.8.2"
   }
--- a/Apps/scripts/getNightly.js
+++ b/Apps/scripts/getNightly.js
@@ -1,5 +1,7 @@
 var https = require('https');
 var fs = require('fs');
+var path = require('path');
+var spawnSync = require('child_process').spawnSync;
 
 function download(filename, url) {
   return new Promise(function (resolve, reject) {
@@ -70,7 +72,31 @@ console.log('Downloading babylon.js nightly');
 // file in place, and so the down-level step that runs next cannot race an incomplete write.
 Promise.all(files.map(function (f) { return download(f[0], f[1]); })).then(function () {
   console.log('Downloaded ' + files.length + ' file(s)');
-}, function (err) {
+
+  // Every bundle we just replaced needs the ES5 down-level, not just babylon.max.js: they all end up
+  // in BABYLON_SCRIPTS and all have to run on Chakra. Deriving the list from `files` means adding a
+  // new download above can never again silently skip this step.
+  var scripts = files
+    .map(function (f) { return f[0]; })
+    .filter(function (p) { return /babylon.*\.js$/i.test(p); });
+
+  if (scripts.length === 0) {
+    return;
+  }
+
+  console.log('Down-leveling ' + scripts.length + ' script(s)');
+  var result = spawnSync(
+    process.execPath,
+    [path.join(__dirname, 'downlevelNativeScripts.mjs')].concat(scripts),
+    { stdio: 'inherit' });
+
+  if (result.error) {
+    throw result.error;
+  }
+  if (result.status !== 0) {
+    throw new Error('downlevelNativeScripts.mjs exited with code ' + result.status);
+  }
+}).catch(function (err) {
   console.error(err.message);
   process.exit(1);
 });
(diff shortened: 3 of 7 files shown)


Apps/Playground/Shared/PlaygroundScripts.cpp
```diff
@@ -44,6 +49,9 @@ namespace Playground
         runtime.LoadScript("app:///Scripts/babylon.gui.js");
         runtime.LoadScript("app:///Scripts/meshwriter.min.js");
         runtime.LoadScript("app:///Scripts/babylonjs.serializers.js");
+        // Procedural texture classes (WoodProceduralTexture, BrickProceduralTexture, ...) live
+        // in their own package, which the browser Playground also loads separately.
+        runtime.LoadScript("app:///Scripts/babylonjs.proceduralTextures.js");
```
bghgary: `Ubuntu_Clang_Hermes / build` fails on this head with **two SIGSEGVs**, not a pixel diff — `Playground: Finished in 2m 24.167s. (exit 3)`, [job log](https://github.com/BabylonJS/BabylonNative/actions/runs/31193739891/job/92916748810). No callstack survived (`addr2line: 'a.out': No such file`).

Attributable to this PR: the same job is green on `master` (`04448f29`, which already contains #1822) and on #1817.

The validation table is a Dawn/WebGPU sweep, so it would not have covered this configuration — Hermes on Linux with the GL renderer.

Worth considering that this may be the same class of thing #1822 surfaced rather than a new defect: scenes that previously died at `earcut is not defined` or `WoodProceduralTexture is not a constructor` now run, so a latent crash further along becomes reachable for the first time. That would make it a real bug this PR exposes rather than causes — but it still needs a cause before merge, since it is a hard crash.

bkaradzic-microsoft:
Thanks — all three review comments were valid; pushed in e421e9a9.

**Exact pin for `babylonjs-procedural-textures`.** Correct catch: the PR description said "pinned to 9.15.0" but `package.json` carried `^9.15.0`. The lock file happened to resolve 9.15.0, so the nesting problem was latent rather than visible — regenerating it would have reintroduced 9.20.0 and the second nested `babylonjs` copy. Now exact in both `package.json` and the lock file's root dependency range.

**Down-leveling more than `babylon.max.js`.** Also correct, and the gap was wider than the comment suggested: `getNightly` replaces *seven* `babylon*.js` bundles and only `babylon.max.js` was being down-leveled, so `gui`, `loaders`, `materials`, `serializers`, `addons` — and now `proceduralTextures` — were running un-transpiled on Chakra after a nightly refresh.

Rather than lengthen the hardcoded argument list (which is what let the two drift apart in the first place), the down-level list is now derived from the download table inside `getNightly.js` and invoked from there, so adding a download can't silently skip the step again. Source maps are excluded by the `.js` anchor, and `downlevelNativeScripts.mjs` filters on basename as a second guard. `npm run downlevel:native-scripts` stays available for manual use.

CI note: `Ubuntu_Clang_Hermes` SIGSEGV'd on the first run, ~82 tests in. It was a flake — the re-run completed the same 304 tests as master with `ran=283 passed=283 failed=0`.

Apps/package.json
```diff
@@ -19,7 +19,9 @@
     "babylonjs-gui": "^9.15.0",
     "babylonjs-loaders": "^9.15.0",
     "babylonjs-materials": "^9.15.0",
+    "babylonjs-procedural-textures": "9.15.0",
     "babylonjs-serializers": "^9.15.0",
+    "earcut": "^2.2.4",
```
bghgary, line 24: The `^2.2.4` constraint is right, but not for the reason given: 3.x does ship `dist/earcut.min.js` (9,648 bytes, UMD, unpkg 200), so "ESM-only, no `dist`" invites a bump from anyone who checks.

Two reasons that do hold. In 2.x the global *is* the function (`.earcut = e()`); in 3.x it is a namespace (`.earcut = {}`), and `PolygonMeshBuilder` calls `earcut(...)`, so 3.x throws at triangulation time. And `cdn.babylonjs.com/earcut.min.js`, which `playground/index.html` loads, is itself 2.x (6,519 bytes) -- so this matches the environment these tests replicate rather than deferring an upgrade.

Apps/scripts/getNightly.js
```diff
@@ -68,7 +72,31 @@ console.log('Downloading babylon.js nightly');
 // file in place, and so the down-level step that runs next cannot race an incomplete write.
 Promise.all(files.map(function (f) { return download(f[0], f[1]); })).then(function () {
   console.log('Downloaded ' + files.length + ' file(s)');
-}, function (err) {
+
+  // Every bundle we just replaced needs the ES5 down-level, not just babylon.max.js: they all end up
+  // in BABYLON_SCRIPTS and all have to run on Chakra. Deriving the list from `files` means adding a
+  // new download above can never again silently skip this step.
+  var scripts = files
```
bghgary, line 79: This has to land in Babylon.js before this PR does. `ci-monorepo.yml` refills the Playground `Scripts` directory from the CDN snapshot with a `dir /b %destDir%\babylon*` loop (L993-996), which would match the `babylonjs.proceduralTextures.js` added here, and the only down-level that follows is on `babylon.max.js` (L1023). That bundle is ES2015, so `Playground.exe` would parse it un-down-levelled on Chakra. `NativeTests` takes the Native artifact at `runVersion: latest` and has no `continueOnError`, so it breaks on the first nightly after this merges.

Naming the new file on that side would fail until this lands, since it is not in the artifact yet; globbing the `Scripts` directory is a no-op today and correct afterwards. Worth considering whether the two can share rather than track each other: `downlevelNativeScripts.mjs` is already byte-identical in both repos, with a header asking the next person to diff them by hand.

Apps/Playground/Shared/PlaygroundScripts.cpp
```diff
@@ -35,6 +35,11 @@ namespace Playground
         runtime.LoadScript("app:///Scripts/ammo.js");
         // Commenting out recast.js for now because v8jsi is incompatible with asm.js.
         // runtime.LoadScript("app:///Scripts/recast.js");
+        // PolygonMeshBuilder resolves `earcut` as a global at triangulation time rather than
+        // importing it, so it has to be present before any polygon scene runs. The browser
+        // Playground loads it the same way; without it those scenes fail with
+        // `ReferenceError: earcut is not defined`.
```
bghgary: Trim to the two facts a reader needs: the global lookup, and the failure without it.

```suggestion
        // PolygonMeshBuilder resolves `earcut` as a global at triangulation time, so without this
        // any polygon scene fails with `ReferenceError: earcut is not defined`.
```

Apps/Playground/Shared/PlaygroundScripts.cpp
bghgary: Drop this one. The seven separate-package loads above carry no comment either, and the two that do earn it state something the code cannot show.

```suggestion
        runtime.LoadScript("app:///Scripts/babylonjs.proceduralTextures.js");
```

bghgary (changes requested):

Blocked on a matching Babylon.js change; four inline.

Apps/package.json
bkaradzic-microsoft: You are right on both counts, and I have corrected the PR description -- thank you for checking rather than taking the claim at face value.

Verified: `earcut@3.0.1` does ship `dist/earcut.min.js` (unpkg 200), so "ESM-only, no `dist`" was simply wrong. The difference is the shape of the global:

```js
// 2.2.4  -- the global IS the function
!function(e){ ... ("undefined"!=typeof window?window: ...).earcut=e() }

// 3.0.1  -- the global is a namespace
!function(e,t){ ... t((e=...||self).earcut={}) }
```

`PolygonMeshBuilder` calls `earcut(...)`, so 3.x throws at triangulation time.

And `cdn.babylonjs.com/earcut.min.js`, which `playground/index.html` loads, is 6,519 bytes and has the 2.x `module.exports=e()` / global-is-the-function shape -- so the pin matches the environment these tests replicate rather than deferring an upgrade. Both of those reasons are now in the description in place of the wrong one.

Apps/scripts/getNightly.js
bkaradzic-microsoft: Agreed on the ordering, and that side is now up: BabylonJS/Babylon.js#18799. It does what you suggested -- passes the `Scripts`/`Assets` directory instead of naming a file, so it is correct before and after this PR. `downlevelNativeScripts.mjs` already walks a directory and selects `/^babylon.*\.js$/i` itself, so it needed no change; I verified against a directory laid out like the artifact that `babylon*.js` get transpiled while `earcut.min.js`, `ammo.js` and the `.map` are left alone.

One correction to the severity, since I measured it before writing that PR: this specific bundle would not have broken the nightly. Syntax actually present in each bundle on the CDN today:

| bundle | downleveled | `?.` | `#priv` | `**` | `async` | `function*` |
|---|---|---|---|---|---|---|
| `babylon.max.js` | **yes** | 19 | 2 | 74 | 1 | 631 |
| `babylonjs.loaders.js` | no | 2 | 0 | 0 | 0 | 148 |
| `babylonjs.serializers.js` | no | 2 | 0 | 0 | 0 | 62 |
| `babylonjs.addons.js` | no | 1 | 0 | 6 | 0 | 17 |
| `babylonjs.materials.js` | no | 0 | 0 | 3 | 0 | 14 |
| `babylon.gui.js` | no | 0 | 0 | 0 | 0 | 18 |
| `babylonjs.proceduralTextures.js` | n/a | 0 | 0 | 0 | 0 | 0 |

`babylonjs.proceduralTextures.js` is the most conservative of the seven -- plain ES2015 classes and arrows, nothing past that -- while `loaders`, `serializers` and `addons` already run un-downleveled with optional chaining and generators. (The `accessor` keyword the existing comment cites does not survive into any of the emitted bundles; every apparent match is a string in an error message.)

So the gap you are pointing at is real and worth closing, but it is broader and older than this PR rather than introduced by it, which is why #18799 fixes the mechanism instead of naming the new file.

On sharing rather than tracking: agreed that two byte-identical copies with a "diff them by hand" header is the wrong steady state. It is a bigger change than either PR should carry though -- it needs a decision on direction (publish it from Babylon.js and consume it in Native, or vendor it) -- so I would rather not fold it in here. Happy to open a separate issue for it if you want it tracked.

Apps/Playground/Shared/PlaygroundScripts.cpp
bkaradzic-microsoft: Applied in be0da802.

Apps/Playground/Shared/PlaygroundScripts.cpp
bkaradzic-microsoft: Dropped in be0da802.

bkaradzic-microsoft:
Thanks -- all four addressed.

- **Comment trims** (be0da802): applied both suggestions verbatim.
- **earcut rationale**: you were right, 3.x does ship `dist/earcut.min.js`. The real reasons are the global changing from the function to a namespace, and `cdn.babylonjs.com/earcut.min.js` being 2.x. Description corrected.
- **Babylon.js side**: BabylonJS/Babylon.js#18799 is up, doing what you suggested (pass the directory, not a filename). Marked as a dependency in the description.

One measured correction on that last one: `babylonjs.proceduralTextures.js` would not actually have broken the nightly -- it is the most conservative of the seven bundles (no optional chaining, no private fields, no generators), while `loaders`, `serializers` and `addons` already run un-downleveled with all three. Details and the full inventory are in the thread. The gap is real, just older and broader than this PR, so #18799 fixes the mechanism rather than naming the new file.

Re-requesting review.

bkaradzic-microsoft:
The blocker is cleared: BabylonJS/Babylon.js#18799 is merged, so the nightly now downlevels every `babylon*.js` in the Native artifact directories rather than only `babylon.max.js`.

bghgary (approved):

The ordering blocker is cleared.

PR #1824, opened 2026-08-07
Canvas: fix 21 correctness and robustness bugs in the shared Canvas2D polyfill

### LINKED ISSUE #1782: bgfx Canvas gradients do not preserve Canvas 2D stroke and geometry semantics
### Problem

The bgfx/NanoVG-backed Canvas polyfill accepts `CanvasGradient` for `fillStyle`, but several Canvas 2D gradient semantics are missing:

- [`GUI Gradient Linear`](https://playground.babylonjs.com/#XCPP9Y#17227) throws `TypeError: A string was expected` because `Context::SetStrokeStyle` unconditionally casts the value to a string.
- [`GUI Gradient Radial`](https://playground.babylonjs.com/#4Z7EK3) differs by 56,743/240,000 pixels on Metal/bgfx because the cached radial image is always a centered 256x256 field and ignores the supplied `(x0, y0, r0)` / `(x1, y1, r1)` circles.
- [`GUI Gradient Linear with transparency`](https://playground.babylonjs.com/#PFK1Z5) differs by about 50,000 pixels because the cached gradient strip is mapped to the fill bounds instead of the gradient endpoints/coordinate space.

These are Canvas-wrapper behavior gaps rather than screenshot tolerance issues. BabylonNative#1730 keeps the tests excluded until the rendering semantics are implemented.

### Expected

Match browser Canvas 2D behavior for:

- `CanvasGradient` values assigned to `strokeStyle`, including paths and text
- linear-gradient coordinates independent of the filled/stroked geometry bounds
- radial gradients defined by two independently centered circles
- multi-stop and transparent interpolation on those paths

The three visual tests above should then be enabled with meaningful pixel budgets, backed by focused Canvas unit tests.

bkaradzic-microsoft:
Fixes #1782.

21 Canvas2D bugs in the shared canvas polyfill. All are backend-agnostic (nanovg / polyfill level), so they affect the existing bgfx path exactly as found — I hit them while bringing up a WebGPU canvas backend, but nothing here is WebGPU-specific.

Most are "the method threw `not implemented` but the caller was asking for nothing", or "the paint was built from the wrong geometry". Several have to land together: 1-4 are all required before the GUI ColorPicker renders at all, and 5-8 before a gradient-stroked GUI control renders at all.

## Correctness

| # | Bug |
| --- | --- |
| 1 | `ImageData.data` returned a fresh copy on every access, so `getImageData` → mutate → `putImageData` silently no-op'd. It is also now the specified `Uint8ClampedArray`; a plain `Uint8Array` wrapped `data[i] = v + 40` modulo 256, turning bright pixels dark |
| 2 | `clip()` suppressed `beginPath()` for every later draw, so each fill repainted the union of every rect since the clip |
| 3 | 3-arg `drawImage(img, dx, dy)` anchored the pattern at `(0,0)` while drawing the rect at `(dx,dy)`, so it drew nothing at any non-zero offset |
| 4 | Gradient paint was built from the shape being filled, not the gradient — every gradient came out horizontal, anchored at the canvas origin, with `(x0,y0)→(x1,y1)` discarded |
| 5 | `strokeStyle` accepted only a color string; a gradient threw `TypeError: A string was expected` and took the whole scene down (GUI `Line`, `Button` border) |
| 6 | Radial gradients discarded the inner circle, so every one came out concentric with `r0 == 0`. Now solves the real two-circle equation. The tangent case (leading coefficient vanishes → linear solve) needs its own branch and is not exotic — the repro's `createRadialGradient(100,100,150, 250,100,300)` lands there |
| 7 | Color stops were interpolated in straight alpha, so `#ff0000ff → #00000000` faded through maroon to black. Now premultiplied as specified |
| 8 | `fill()` never bound the fill style, using whatever color nanovg happened to hold — never a gradient. Every path-filled control (`Ellipse`, `Button` background, `Slider`) filled solid white |
| 9 | `createImageData` was missing entirely |
| 10 | `Path2D` was registered only on `_native`, not as a global, so `AbstractEngine.createCanvasPath2D` threw `ReferenceError` on every engine except `ThinNativeEngine` |
| 11 | `loadTTF` failures surfaced as a bare `Error: Invalid argument`, naming neither method nor argument. Now descriptive, and any ArrayBuffer *view* is accepted — so the usual real cause, `Tools.LoadFileAsync(url)` without `useArrayBuffer`, is named directly |
| 12 | `putImageData` was an unconditional throw. Now writes through a transient nanovg image under `nvgSave`/`nvgReset` with `NVG_COPY`, so it correctly ignores transform, clip, `globalAlpha`, compositing, shadows and filters and replaces rather than blends |
| 13 | `setLineDash` threw — and GUI `Line`/`MultiLine` call it every render with `_dash` defaulting to `[]`, so any scene with a GUI line died on an empty dash list asking for nothing. Empty now means "solid", which is what we already draw |
| 14 | The color regex matched components as 1-3 digit integers, so the very common `rgba(0, 0, 0, 0.5)` failed to parse. Now generic CSS numbers with optional `%`, plus the whitespace-separated and `/ alpha` forms and `hsl()`/`hsla()` — **see the behaviour change below** |
| 15 | All four shadow attributes threw from both getter and setter — and GUI resets `shadowBlur`/`shadowOffset*` to `0` right after drawing, so the throw hit the reset path too and any shadowed control failed outright. nanovg has no shadow primitive, but these are ordinary state attributes, so they are stored and reported |
| 16 | `save()`/`restore()` snapshotted only `fillStyle`/`strokeStyle`, but the wrapper mirrors everything it reprojects or that nanovg cannot report back — so `lineCap`, `lineJoin`, `lineWidth`, `miterLimit`, the dash list, `filter`, `direction`, `letterSpacing`, `globalAlpha` and the shadow attributes were never restored. The getters reported the post-`save()` value forever, and the shadow attributes, re-applied from the mirror on every draw, actually rendered wrong. All fifteen fields are now snapshotted |
| 17 | Coincident color stops were dropped — `std::map::insert()` discarded the second stop at an already-present offset, which is exactly how the spec encodes a hard transition, so a stripe pattern came out as a smooth fade. Now a `std::multimap` |
| 18 | `setLineDash()` cleared the list *before* validating, so `setLineDash([-1])` after `setLineDash([5,10])` left `getLineDash()` returning `[]` |

Two smaller gradient bugs were fixed alongside 4-8: the ramp sample buffer was uninitialized (samples outside the stop range read stack garbage), and `gradientSpan` divided by zero when two stops shared an offset.

## Robustness

Three crash classes found by the Copilot review, all reachable from ordinary script rather than only from a misbehaving embedder:

| # | Bug |
| --- | --- |
| 19 | `fillStyle = "rgb(999…999, 0, 0)"`, `font = "18e999px Arial"` and `letterSpacing = "999…999px"` made `std::stof`/`std::stoi` throw `std::out_of_range`. That is not a `Napi::Error`, so node-addon-api's wrapper did not catch it and it unwound out of the N-API boundary and **terminated the process**. Now `strtof`/`strtol`, with infinities folded to finite extremes so the existing clamps stay well defined (`nvgHSLA` takes `fmodf` of the hue, and `fmodf(inf, 1)` is NaN) |
| 20 | A gradient outliving its context: `UpdateCache()` dereferenced `context.lock()` three times unchecked, so the guard in `Paint()` ran after the crash |
| 21 | `putImageData(img, 0, 0, 0, 0, -2147483648, 1)` normalized negative dirty extents in `int32_t`, making `dirtyX += dirtyWidth` and `-dirtyWidth` signed overflow on caller-controlled input. Now `int64_t`, saturating back after clipping |

19 is confirmed by experiment, not inspection: built without the fix, the unit test binary stops dead inside the `Canvas2D` suite with `[Uncaught Error] Unknown failure` and exit 1, taking the remaining ~20 tests with it. The pre-existing font guard covered only `std::invalid_argument` (the `"normal"` keyword), so `out_of_range` still went through.

## One intentional behaviour change — please look at this closely

The `parseColor` unit test asserted `rgba(16,32,48,64)` → alpha `0x40`, i.e. that alpha is a 0-255 channel like r/g/b. CSS Color defines alpha as 0-1 (or a percentage), so a browser clamps `64` to `1` and paints that colour opaque.

That reading was not a deliberate extension — the old regex matched all four components as 1-3 digit integers and ran `std::stoi` over them, so alpha simply inherited the channel treatment, and the test was written to describe what the implementation happened to do (it arrived with the impl in #1051). It is also the same root cause as bug 14: under a 0-255 alpha, `rgba(0, 0, 0, 0.5)` truncates to `0` and the colour vanishes. Alpha cannot be both, so the test now follows CSS, extended to cover fractional and percentage alpha, the `/ alpha` form, whitespace-separated and percentage components, and `hsl()`/`hsla()`. All ten malformed-input cases still throw.

## Deliberate limitations

- **Non-rectangular `clip()` is approximated.** nanovg's scissor is rectangle-only and no stencil path is available, so the choice is between approximating and throwing; this keeps the rectangular case exact and degrades the other rather than failing the scene.
- **`putImageData()` clobbers the current path** — as do `fillRect`, `clearRect`, `drawImage` and `Path2D` playback, none of which may touch it per spec. nanovg offers no local fix, since `nvgSave`/`nvgRestore` only push and pop `NVGstate` while `nvgBeginPath` clears `ctx->ncommands` and the path cache. A real fix means journaling the path wrapper-side and replaying it; self-contained, and better as a follow-up.

## Validation

Full Playground validation sweep **304/304 PASS, no regressions**; `JavaScript.All` unit tests **36/36**.

Three previously-excluded tests are re-enabled and now pass inside the default 2.5% budget:

| Test | Before | After | Needs |
| --- | --- | --- | --- |
| `GUI Gradient Linear` (`#XCPP9Y#17227`) | `TypeError` — never rendered | **1.14%** | 5, then 8 |
| `GUI Gradient Radial` (`#4Z7EK3`) | 23.6% | **0.62%** | 6 (tangent case) |
| `GUI Gradient Linear with transparency` (`#PFK1Z5`) | 33.8% | **pixel-exact** | 7 |

The `GUI` test went **13.5% → 3.9%** (allowed 4%) across bugs 1-4 — 13.5% → 8.2% (clip) → 6.4% (drawImage origin) → 3.9% (gradients) — because its ColorPicker builds its wheel via `getImageData` + mutate + `putImageData` (1), composites it with a 3-arg `drawImage` at a non-zero offset (3), draws it and its saturation square under one `clip()` (2), and paints that square with two overlaid linear gradients (4). Nine further tests newly pass across the same four fixes.

Bugs 10 and 11 are test-neutral and were verified directly, and new `Canvas2D` unit tests cover the contract that used to throw: both style round-trips, a `CanvasGradient` as `fillStyle` and as `strokeStyle`, a radial gradient from two independent circles, save/restore of a gradient `strokeStyle` and of all fifteen state fields, dash retention on a rejected argument, coincident stops, and three regressions for the out-of-range parse crashes.

One implementation note: `DataView` is read through its `buffer`/`byteOffset`/`byteLength` properties rather than `Napi::DataView`, because Chakra's Node-API backs `napi_get_dataview_info` with `JsGetExternalData`, which only succeeds for natively created DataViews — a JS-constructed one fails with `napi_invalid_arg` even though `napi_is_dataview` returns `true`. That shim inconsistency is a separate JsRuntimeHost issue; this change just avoids depending on it.

### DIFF
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -51,13 +51,29 @@ describe("ColorParsing", function () {
   expect(_native.Canvas.parseColor("#12345678")).to.equal(0x78563412);
   expect(_native.Canvas.parseColor("snow")).to.equal(0xfffafaff);
   expect(_native.Canvas.parseColor("rgb(16,32,48)")).to.equal(0xff302010);
-  expect(_native.Canvas.parseColor("rgba(16,32,48,64)")).to.equal(0x40302010);
+  // Alpha is a 0-1 number (or a percentage) per CSS Color, so any value above
+  // 1 clamps to fully opaque. It is not a 0-255 channel like r/g/b.
+  expect(_native.Canvas.parseColor("rgba(16,32,48,64)")).to.equal(0xff302010);
   expect(_native.Canvas.parseColor("rgb(16,     32   ,  48   )")).to.equal(
     0xff302010
   );
   expect(
     _native.Canvas.parseColor("rgba(    16,     32   ,  48 , 64  )")
-  ).to.equal(0x40302010);
+  ).to.equal(0xff302010);
+  expect(_native.Canvas.parseColor("rgba(16,32,48,1)")).to.equal(0xff302010);
+  expect(_native.Canvas.parseColor("rgba(16,32,48,0)")).to.equal(0x00302010);
+  // Fractional and percentage alpha, whitespace-separated components and the
+  // "/ alpha" form all used to fall through to the "unable to parse" throw.
+  expect(_native.Canvas.parseColor("rgba(16,32,48,0.5)")).to.equal(0x80302010);
+  expect(_native.Canvas.parseColor("rgba(16 32 48 / 50%)")).to.equal(
+    0x80302010
+  );
+  expect(_native.Canvas.parseColor("rgb(16 32 48)")).to.equal(0xff302010);
+  expect(_native.Canvas.parseColor("rgb(100%,0%,0%)")).to.equal(0xff0000ff);
+  expect(_native.Canvas.parseColor("hsl(0,100%,50%)")).to.equal(0xff0000ff);
+  expect(_native.Canvas.parseColor("hsla(0,100%,50%,0.5)")).to.equal(
+    0x800000ff
+  );
 
   it("should throw", function () {
     function incorrectColor() {
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -145,6 +145,70 @@ describe("ColorParsing", function () {
   });
 });
 
+describe("Canvas2D", function () {
+  function createContext(): any {
+    const canvas = new _native.Canvas();
+    canvas.width = 64;
+    canvas.height = 64;
+    return canvas.getContext("2d");
+  }
+
+  it("round-trips a string fillStyle and strokeStyle", function () {
+    const ctx = createContext();
+    ctx.fillStyle = "#ff0000";
+    ctx.strokeStyle = "#00ff00";
+    expect(ctx.fillStyle).to.equal("#ff0000");
+    expect(ctx.strokeStyle).to.equal("#00ff00");
+  });
+
+  it("accepts a CanvasGradient as fillStyle", function () {
+    const ctx = createContext();
+    const gradient = ctx.createLinearGradient(0, 0, 64, 64);
+    gradient.addColorStop(0, "red");
+    gradient.addColorStop(1, "blue");
+    expect(function () {
+      ctx.fillStyle = gradient;
+    }).to.not.throw();
+    expect(ctx.fillStyle).to.not.equal("#ff0000");
+  });
+
+  it("accepts a CanvasGradient as strokeStyle", function () {
+    // strokeStyle used to be string-only and threw "A string was expected",
+    // which broke every GUI control that strokes with a gradient (Line, Button border).
+    const ctx = createContext();
+    const gradient = ctx.createLinearGradient(0, 0, 64, 64);
+    gradient.addColorStop(0, "red");
+    gradient.addColorStop(1, "blue");
+    expect(function () {
+      ctx.strokeStyle = gradient;
+    }).to.not.throw();
+  });
+
+  it("accepts a radial CanvasGradient defined by two independent circles", function () {
+    const ctx = createContext();
+    // Neither concentric nor r0 == 0: both circles have to be honored.
+    const gradient = ctx.createRadialGradient(10, 10, 5, 40, 32, 30);
+    gradient.addColorStop(0, "yellow");
+    gradient.addColorStop(0.5, "pink");
+    gradient.addColorStop(1, "green");
+    expect(function () {
+      ctx.fillStyle = gradient;
+      ctx.strokeStyle = gradient;
+    }).to.not.throw();
+  });
+
+  it("restores a gradient strokeStyle across save/restore", function () {
+    const ctx = createContext();
+    const gradient = ctx.createLinearGradient(0, 0, 64, 64);
+    gradient.addColorStop(0, "red");
+    ctx.strokeStyle = "#0000ff";
+    ctx.save();
+    ctx.strokeStyle = gradient;
+    ctx.restore();
+    expect(ctx.strokeStyle).to.equal("#0000ff");
+  });
+});
+
 function createSceneAndWait(callback: (engine: NativeEngine, scene: Scene) => void, done: () => void) {
   const engine = new NativeEngine();
   const scene = new Scene(engine);
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -207,6 +207,42 @@ describe("Canvas2D", function () {
     ctx.restore();
     expect(ctx.strokeStyle).to.equal("#0000ff");
   });
+
+  // The three parsers below all used to reach std::stof/std::stoi with a value the regex
+  // admits but the target type cannot hold. The resulting std::out_of_range is not a
+  // Napi::Error, so it escaped the N-API callback and terminated the host process outright
+  // rather than surfacing as a JS exception. Reaching the assertion at all is the test.
+  it("survives an out-of-range rgb() component", function () {
+    const ctx = createContext();
+    const huge = "9".repeat(400);
+    expect(function () {
+      ctx.fillStyle = `rgb(${huge}, 0, 0)`;
+    }).to.not.throw();
+    expect(function () {
+      ctx.fillStyle = `rgba(0, 0, 0, ${huge})`;
+    }).to.not.throw();
+  });
+
+  it("survives an out-of-range font size and weight", function () {
+    const ctx = createContext();
+    ctx.font = "18px Arial";
+    // The size regex accepts an exponent, so this parses but does not fit a float.
+    expect(function () {
+      ctx.font = "18e999px Arial";
+    }).to.not.throw();
+    expect(function () {
+      ctx.font = `${"9".repeat(400)} 18px Arial`;
+    }).to.not.throw();
+    // An unparseable font is ignored, so the previous one stays in effect.
+    expect(ctx.font).to.contain("18px");
+  });
+
+  it("survives an out-of-range letterSpacing", function () {
+    const ctx = createContext();
+    expect(function () {
+      ctx.letterSpacing = `${"9".repeat(400)}px`;
+    }).to.not.throw();
+  });
 });
 
 function createSceneAndWait(callback: (engine: NativeEngine, scene: Scene) => void, done: () => void) {
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -1,4 +1,4 @@
-﻿import * as Mocha from "mocha";
+import * as Mocha from "mocha";
 import { expect } from "chai";
 import {
   RequestFile,
@@ -273,6 +273,55 @@ describe("Canvas2D", function () {
     expect(ctx.getLineDash()).to.deep.equal([]);
   });
 
+  it("keeps a gradient assigned to fillStyle alive across a collection", function () {
+    const ctx = createContext();
+    const global: any = Function("return this")();
+
+    // Create and assign the gradient inside a scope that keeps no reference to it, so
+    // the only thing left pointing at it is whatever fillStyle stored. CanvasGradient is
+    // an ObjectWrap, so if the style does not root the JavaScript wrapper the finalizer
+    // deletes the native gradient and the next fill dereferences freed memory.
+    (function () {
+      const gradient: any = ctx.createLinearGradient(0, 0, 64, 0);
+      gradient.addColorStop(0, "red");
+      gradient.addColorStop(1, "blue");
+      gradient.tag = "kept";
+      ctx.fillStyle = gradient;
+    })();
+
+    // Not every engine the polyfill runs on exposes a collection hook; the assertions
+    // below are worth making either way.
+    if (typeof global.CollectGarbage === "function") {
+      global.CollectGarbage();
+    }
+
+    // The getter has to hand back the object that was assigned. The expando proves it is
+    // that same JavaScript object rather than a fresh wrapper.
+    expect(ctx.fillStyle.tag).to.equal("kept");
+    // ...and the native gradient behind it has to still be there.
+    ctx.fillStyle.addColorStop(0.5, "green");
+    ctx.fillRect(0, 0, 64, 64);
+  });
+  it("keeps a gradient assigned to strokeStyle alive across a collection", function () {
+    const ctx = createContext();
+    const global: any = Function("return this")();
+
+    (function () {
+      const gradient: any = ctx.createRadialGradient(32, 32, 0, 32, 32, 32);
+      gradient.addColorStop(0, "red");
+      gradient.addColorStop(1, "blue");
+      gradient.tag = "kept";
+      ctx.strokeStyle = gradient;
+    })();
+
+    if (typeof global.CollectGarbage === "function") {
+      global.CollectGarbage();
+    }
+
+    expect(ctx.strokeStyle.tag).to.equal("kept");
+    ctx.strokeStyle.addColorStop(0.5, "green");
+    ctx.strokeRect(0, 0, 64, 64);
+  });
   it("keeps the previous dash list when the argument is not a list", function () {
     // A present-but-rejected argument skipped the parse loop and then committed the
     // empty temporary, so setLineDash("x") cleared the list where setLineDash([-1])
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -255,6 +255,24 @@ describe("Canvas2D", function () {
     expect(ctx.getLineDash()).to.deep.equal([1, 2]);
   });
 
+  it("keeps the previous dash list when a new one is rejected", function () {
+    // The list was cleared before validation, so a rejected argument -- which the
+    // spec says must leave the previous list untouched -- wiped it instead.
+    const ctx = createContext();
+    ctx.setLineDash([5, 10]);
+    ctx.setLineDash([-1]);
+    expect(ctx.getLineDash()).to.deep.equal([5, 10]);
+    ctx.setLineDash([2, "x"]);
+    expect(ctx.getLineDash()).to.deep.equal([5, 10]);
+    ctx.setLineDash([Number.NaN]);
+    expect(ctx.getLineDash()).to.deep.equal([5, 10]);
+    // A valid list still replaces it, and an empty list still means "solid".
+    ctx.setLineDash([3, 4]);
+    expect(ctx.getLineDash()).to.deep.equal([3, 4]);
+    ctx.setLineDash([]);
+    expect(ctx.getLineDash()).to.deep.equal([]);
+  });
+
   it("keeps two color stops at the same offset", function () {
     // Stops were stored in a std::map keyed by offset, so the second stop at an
     // identical offset was silently dropped and the hard transition it encodes -- a
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -273,10 +273,69 @@ describe("Canvas2D", function () {
     expect(ctx.getLineDash()).to.deep.equal([]);
   });
 
-  it("keeps two color stops at the same offset", function () {
-    // Stops were stored in a std::map keyed by offset, so the second stop at an
-    // identical offset was silently dropped and the hard transition it encodes -- a
-    // stripe pattern, for instance -- came out as a smooth fade instead.
+  it("keeps the previous dash list when the argument is not a list", function () {
+    // A present-but-rejected argument skipped the parse loop and then committed the
+    // empty temporary, so setLineDash("x") cleared the list where setLineDash([-1])
+    // correctly kept it. An absent argument still means "solid".
+    const ctx = createContext();
+    ctx.setLineDash([5, 10]);
+    ctx.setLineDash("x" as any);
+    expect(ctx.getLineDash()).to.deep.equal([5, 10]);
+    ctx.setLineDash(null as any);
+    expect(ctx.getLineDash()).to.deep.equal([5, 10]);
+    ctx.setLineDash(7 as any);
+    expect(ctx.getLineDash()).to.deep.equal([5, 10]);
+    (ctx.setLineDash as any)();
+    expect(ctx.getLineDash()).to.deep.equal([]);
+  });
+
+  it("restores font", function () {
+    // font was the one exposed attribute left out of the saved state. It is worse than
+    // the getter-only cases: the font id resolved here is what the text draw path binds,
+    // so the wrong face actually rendered after a restore().
+    const ctx = createContext();
+    ctx.font = "18px Arial";
+    // The getter reports a normalized serialization, so compare against the round-tripped
+    // value rather than the literal that was assigned.
+    const saved = ctx.font;
+    ctx.save();
+    ctx.font = "40px Times";
+    expect(ctx.font).to.not.equal(saved);
+    ctx.restore();
+    expect(ctx.font).to.equal(saved);
+    expect(function () {
+      ctx.fillText("after restore", 0, 20);
+    }).to.not.throw();
+  });
+
+  it("ignores a fillStyle or strokeStyle that is neither a color string nor a gradient", function () {
+    // Both setters reached napi_unwrap for any object, so `ctx.strokeStyle = {}` unwrapped
+    // an object that was never wrapped and dereferenced whatever the slot held. Per spec
+    // an unusable value leaves the attribute unchanged.
+    const ctx = createContext();
+    ctx.fillStyle = "#ff0000";
+    ctx.strokeStyle = "#00ff00";
+    expect(function () {
+      ctx.fillStyle = {} as any;
+      ctx.strokeStyle = {} as any;
+      ctx.fillStyle = [] as any;
+      ctx.strokeStyle = (function () {}) as any;
+    }).to.not.throw();
+    expect(ctx.fillStyle).to.equal("#ff0000");
+    expect(ctx.strokeStyle).to.equal("#00ff00");
+    // A real gradient is still accepted.
+    const gradient = ctx.createLinearGradient(0, 0, 64, 0);
+    gradient.addColorStop(0, "red");
+    gradient.addColorStop(1, "blue");
+    ctx.fillStyle = gradient;
+    expect(ctx.fillStyle).to.not.equal("#ff0000");
+  });
+
+  it("accepts two color stops at the same offset", function () {
+    // Only checks that the insertion path accepts the duplicate offset the spec allows;
+    // std::map::insert() dropped it by returning {it, false} rather than throwing, so this
+    // does not by itself prove the stop survives. Asserting that needs the rendered ramp,
+    // and gradient fills are not read back by getImageData.
     const ctx = createContext();
     const gradient = ctx.createLinearGradient(0, 0, 64, 0);
     gradient.addColorStop(0, "red");
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -208,6 +208,69 @@ describe("Canvas2D", function () {
     expect(ctx.strokeStyle).to.equal("#0000ff");
   });
 
+  it("restores the shadow attributes across save/restore", function () {
+    // These have no nanovg counterpart, so nvgRestore() never rewound them and a
+    // value assigned after save() survived restore().
+    const ctx = createContext();
+    ctx.shadowColor = "#00ff00";
+    ctx.shadowBlur = 1;
+    ctx.shadowOffsetX = 2;
+    ctx.shadowOffsetY = 3;
+    ctx.save();
+    ctx.shadowColor = "#ff0000";
+    ctx.shadowBlur = 40;
+    ctx.shadowOffsetX = 50;
+    ctx.shadowOffsetY = 60;
+    ctx.restore();
+    expect(ctx.shadowColor).to.equal("#00ff00");
+    expect(ctx.shadowBlur).to.equal(1);
+    expect(ctx.shadowOffsetX).to.equal(2);
+    expect(ctx.shadowOffsetY).to.equal(3);
+  });
+
+  it("restores the remaining drawing state across save/restore", function () {
+    // nvgRestore() rewinds nanovg's copy of these, but the wrapper kept its own
+    // mirror, so the getters reported the post-save() value forever after.
+    // globalAlpha is set but not asserted: it is declared write-only (nullptr
+    // getter), so it has no observable value to check.
+    const ctx = createContext();
+    ctx.lineWidth = 1;
+    ctx.globalAlpha = 1;
+    ctx.lineCap = "butt";
+    ctx.lineJoin = "miter";
+    ctx.miterLimit = 10;
+    ctx.setLineDash([1, 2]);
+    ctx.save();
+    ctx.lineWidth = 9;
+    ctx.globalAlpha = 0.25;
+    ctx.lineCap = "round";
+    ctx.lineJoin = "bevel";
+    ctx.miterLimit = 3;
+    ctx.setLineDash([7, 8, 9]);
+    ctx.restore();
+    expect(ctx.lineWidth).to.equal(1);
+    expect(ctx.lineCap).to.equal("butt");
+    expect(ctx.lineJoin).to.equal("miter");
+    expect(ctx.miterLimit).to.equal(10);
+    expect(ctx.getLineDash()).to.deep.equal([1, 2]);
+  });
+
+  it("keeps two color stops at the same offset", function () {
+    // Stops were stored in a std::map keyed by offset, so the second stop at an
+    // identical offset was silently dropped and the hard transition it encodes -- a
+    // stripe pattern, for instance -- came out as a smooth fade instead.
+    const ctx = createContext();
+    const gradient = ctx.createLinearGradient(0, 0, 64, 0);
+    gradient.addColorStop(0, "red");
+    gradient.addColorStop(0.5, "red");
+    gradient.addColorStop(0.5, "blue");
+    gradient.addColorStop(1, "blue");
+    expect(function () {
+      ctx.fillStyle = gradient;
+      ctx.fillRect(0, 0, 64, 64);
+    }).to.not.throw();
+  });
+
   // The three parsers below all used to reach std::stof/std::stoi with a value the regex
   // admits but the target type cannot hold. The resulting std::out_of_range is not a
   // Napi::Error, so it escaped the N-API callback and terminated the host process outright
--- a/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
+++ b/Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
@@ -432,6 +432,32 @@ describe("Canvas2D", function () {
       ctx.letterSpacing = `${"9".repeat(400)}px`;
     }).to.not.throw();
   });
+
+  it("rejects a createImageData source whose dimensions are not valid extents", function () {
+    const ctx = createContext();
+    // The ImageData overload is duck-typed, so these never went through WebIDL's
+    // unsigned long conversion. A negative width used to wrap to 4294967295 and ask
+    // for a ~17 GB allocation instead of being rejected.
+    const bad = [
+      { width: -1, height: 1 },
+      { width: 1, height: -1 },
+      { width: 1.5, height: 1 },
+      { width: 5e9, height: 1 },
+      { width: Infinity, height: 1 },
+    ];
+    bad.forEach(function (source) {
+      expect(function () {
+        ctx.createImageData(source);
+      }, JSON.stringify(source)).to.throw();
+    });
+  });
+
+  it("creates image data from a valid source object", function () {
+    const ctx = createContext();
+    const data = ctx.createImageData({ width: 4, height: 3 });
+    expect(data.width).to.equal(4);
+    expect(data.height).to.equal(3);
+  });
 });
 
 function createSceneAndWait(callback: (engine: NativeEngine, scene: Scene) => void, done: () => void) {
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -8,6 +8,7 @@
 #include <cstring>
 #include "Colors.h"
 #include "Gradient.h"
+#include "Font.h"
 
 namespace
 {
@@ -63,15 +64,17 @@ namespace Babylon::Polyfills::Internal
 
     void NativeCanvas::LoadTTF(const Napi::CallbackInfo& info)
     {
+        if (info.Length() < 1 || !info[0].IsString())
+        {
+            throw Napi::TypeError::New(info.Env(), "Canvas.loadTTF expects the font name as a string in argument 1.");
+        }
+
         // don't allow same font to be loaded more than once
         // why? because Context doesn't update nvgCreateFontMem when old fontBuffer released
         auto fontName = info[0].As<Napi::String>().Utf8Value();
         if (fontsInfos.find(fontName) == fontsInfos.end())
         {
-            const auto buffer = info[1].As<Napi::ArrayBuffer>();
-            std::vector<uint8_t> fontBuffer(buffer.ByteLength());
-            memcpy(fontBuffer.data(), (uint8_t*)buffer.Data(), buffer.ByteLength());
-            fontsInfos[fontName] = std::move(fontBuffer);
+            fontsInfos[fontName] = GetFontDataArgument(info, 1, "Canvas.loadTTF");
         }
     }
 
--- a/Polyfills/Canvas/Source/Canvas.cpp
+++ b/Polyfills/Canvas/Source/Canvas.cpp
@@ -64,9 +64,14 @@ namespace Babylon::Polyfills::Internal
 
     void NativeCanvas::LoadTTF(const Napi::CallbackInfo& info)
     {
+        LoadTTFCore(info, "Canvas.loadTTF");
+    }
+
+    void NativeCanvas::LoadTTFCore(const Napi::CallbackInfo& info, const char* methodName)
+    {
         if (info.Length() < 1 || !info[0].IsString())
         {
-            throw Napi::TypeError::New(info.Env(), "Canvas.loadTTF expects the font name as a string in argument 1.");
+            throw Napi::TypeError::New(info.Env(), std::string{methodName} + " expects the font name as a string in argument 1.");
         }
 
         // don't allow same font to be loaded more than once
@@ -74,14 +79,14 @@ namespace Babylon::Polyfills::Internal
         auto fontName = info[0].As<Napi::String>().Utf8Value();
         if (fontsInfos.find(fontName) == fontsInfos.end())
         {
-            fontsInfos[fontName] = GetFontDataArgument(info, 1, "Canvas.loadTTF");
+            fontsInfos[fontName] = GetFontDataArgument(info, 1, methodName);
         }
     }
 
     // @deprecated: LoadTTFAsync is always synchronous, use LoadTTF instead
     Napi::Value NativeCanvas::LoadTTFAsync(const Napi::CallbackInfo& info)
     {
-        LoadTTF(info);
+        LoadTTFCore(info, "Canvas.loadTTFAsync");
 
         auto deferred{Napi::Promise::Deferred::New(info.Env())};
         deferred.Resolve(info.Env().Undefined());
--- a/Polyfills/Canvas/Source/Colors.h
+++ b/Polyfills/Canvas/Source/Colors.h
@@ -1,5 +1,8 @@
 #pragma once
 #include <algorithm>
+#include <cmath>
+#include <cstdlib>
+#include <limits>
 #include <regex>
 #include <string>
 #include "nanovg/nanovg.h"
@@ -227,17 +230,34 @@ namespace Babylon::Polyfills::Internal
 
             // Component group indices: value/percent pairs at 1-2, 3-4, 5-6 and
             // the optional alpha at 7-8.
-            const auto channel = [](const std::smatch& match, size_t index) {
-                const float value = std::stof(match[index]);
+            //
+            // Parsed with strtof rather than std::stof: the regex above accepts a digit string of
+            // any length, and std::stof throws std::out_of_range for one that does not fit a
+            // float. That exception is not a Napi::Error, so it would escape this callback and
+            // terminate the host instead of producing a color. strtof cannot throw -- it
+            // saturates to +/-HUGE_VALF on overflow and returns zero on underflow -- and the
+            // infinities are folded back to finite extremes so the clamps below stay well
+            // defined (fmodf of an infinity, which nvgHSLA does to the hue, would be NaN).
+            const auto toFloat = [](const std::ssub_match& match) {
+                const std::string text{match.str()};
+                const float value = std::strtof(text.c_str(), nullptr);
+                if (std::isfinite(value))
+                {
+                    return value;
+                }
+                return value < 0.0f ? std::numeric_limits<float>::lowest() : std::numeric_limits<float>::max();
+            };
+            const auto channel = [&toFloat](const std::smatch& match, size_t index) {
+                const float value = toFloat(match[index]);
                 const float scaled = match[index + 1].matched && match[index + 1].length() > 0 ? value * 255.0f / 100.0f : value;
                 return static_cast<unsigned char>(std::clamp(scaled, 0.0f, 255.0f) + 0.5f);
             };
-            const auto alpha = [](const std::smatch& match, size_t index) {
+            const auto alpha = [&toFloat](const std::smatch& match, size_t index) {
                 if (!match[index].matched)
                 {
                     return static_cast<unsigned char>(255);
                 }
-                const float value = std::stof(match[index]);
+                const float value = toFloat(match[index]);
                 const float normalized = match[index + 1].matched && match[index + 1].length() > 0 ? value / 100.0f : value;
                 return static_cast<unsigned char>(std::clamp(normalized, 0.0f, 1.0f) * 255.0f + 0.5f);
             };
@@ -257,9 +277,9 @@ namespace Babylon::Polyfills::Internal
                 // Hue is an angle in degrees; nvgHSLA wraps it internally but
                 // expects turns. Saturation and lightness are percentages
                 // whether or not the '%' is spelled out.
-                const float hue = std::stof(hslMatch[1]) / 360.0f;
-                const float saturation = std::clamp(std::stof(hslMatch[3]) / 100.0f, 0.0f, 1.0f);
-                const float lightness = std::clamp(std::stof(hslMatch[5]) / 100.0f, 0.0f, 1.0f);
+                const float hue = toFloat(hslMatch[1]) / 360.0f;
+                const float saturation = std::clamp(toFloat(hslMatch[3]) / 100.0f, 0.0f, 1.0f);
+                const float lightness = std::clamp(toFloat(hslMatch[5]) / 100.0f, 0.0f, 1.0f);
                 return nvgHSLA(hue, saturation, lightness, alpha(hslMatch, 7));
             }
         }
--- a/Polyfills/Canvas/Source/Colors.h
+++ b/Polyfills/Canvas/Source/Colors.h
@@ -1,5 +1,7 @@
 #pragma once
+#include <algorithm>
 #include <regex>
+#include <string>
 #include "nanovg/nanovg.h"
 
 namespace Babylon::Polyfills::Internal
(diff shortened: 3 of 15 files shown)


Polyfills/Canvas/Source/ImageData.cpp
```diff
@@ -58,11 +58,18 @@ namespace Babylon::Polyfills::Internal
...
+        // a dark pixel instead of a bright one.
+        const auto byteLength{static_cast<size_t>(pixelCount) * 4};
+        auto data{Napi::Uint8Array::New(info.Env(), byteLength, napi_uint8_clamped_array)};
+        if (context != nullptr && byteLength > 0)
         {
-            context->ReadPixels(sx, sy, m_width, m_height, m_pixels.data());
+            context->ReadPixels(sx, sy, m_width, m_height, data.Data());
         }
+
+        m_data = Napi::Persistent(data);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Verified this on a real build rather than by inspection: `Napi::TypedArrayOf<T>::New(env, length, napi_typedarray_type)` is a documented node-addon-api overload, and `Napi::Uint8Array` is `TypedArrayOf<uint8_t>`, so passing `napi_uint8_clamped_array` compiles and produces a genuine clamped array. A probe through the polyfill prints:

```
getImageData: toString=[object Uint8ClampedArray] isU8C=true ctor=Uint8ClampedArray len=64
createImageData: toString=[object Uint8ClampedArray] isU8C=true ctor=Uint8ClampedArray len=64
clamped: 255,0
```

(the last line is `data[0]=300` and `data[1]=-5` round-tripping through the clamp). So the JS-visible type and the clamping semantics are both correct. Note `Napi::Uint8ClampedArray` is not a type that node-addon-api actually defines, so the suggested alternative would not compile.

Polyfills/Canvas/Source/Gradient.cpp
```diff
@@ -341,6 +349,36 @@ namespace Babylon::Polyfills::Internal
...
+            const float dx = x1 - x0;
+            const float dy = y1 - y0;
+            float length = std::sqrt(dx * dx + dy * dy);
+            if (length < 1e-4f)
+            {
+                // A zero-length gradient would divide by zero in the shader; keep it finite so
+                // the whole shape clamps to a single stop instead of rendering garbage.
+                length = 1e-4f;
+            }
+            return nvgImagePattern(*nvg, x0, y0, length, length, std::atan2(dy, dx), cachedImage, 1.f);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: The vertical extent genuinely is a don't-care here rather than an approximation: the linear ramp is baked into a `GRADIENT_SAMPLES_L x 1` image, so every `v` lands on the single texel row regardless. Using `length` rather than a constant is the safer of the two, though -- nanovg divides the fragment position by the extent, so a fixed `1.0f` against a gradient hundreds of pixels long would produce very large `v` values and correspondingly worse float precision, whereas matching the extents keeps `u` and `v` on the same scale. I have expanded the comment above it to say so.

Polyfills/Canvas/Source/Gradient.cpp
```diff
@@ -341,6 +349,36 @@ namespace Babylon::Polyfills::Internal
...
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, fixed in aad6daf0 -- `Paint()` now returns a default-constructed `NVGpaint` if the weak context has expired instead of dereferencing an empty `shared_ptr`.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1099,6 +1088,61 @@ namespace Babylon::Polyfills::Internal
...
+
+                width = static_cast<uint32_t>(wInt < 0 ? -wInt : wInt);
+                height = static_cast<uint32_t>(hInt < 0 ? -hInt : hInt);
+            }
+            else if (info.Length() >= 1 && info[0].IsObject())
+            {
+                const auto source = info[0].As<Napi::Object>();
+                width = source.Get("width").As<Napi::Number>().Uint32Value();
+                height = source.Get("height").As<Napi::Number>().Uint32Value();
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in aad6daf0. It now checks that `width`/`height` are actually numbers and throws a `TypeError` otherwise, so a random object no longer silently produces a garbage-sized buffer. I have deliberately kept it duck-typed rather than an `InstanceOf` check, because the same code services the Dawn context whose `ImageData` is a distinct class, and the spec overload is only ever reached with an `ImageData`-shaped object.

Polyfills/Canvas/Source/ImageData.h
```diff
@@ -20,6 +20,11 @@ namespace Babylon::Polyfills::Internal
 
         uint32_t m_width{};
         uint32_t m_height{};
-        std::vector<uint8_t> m_pixels;
+
+        // The spec requires `data` to be one live buffer that the caller can
+        // mutate in place; handing back a fresh copy per access silently drops
+        // every write, which breaks the standard getImageData/putImageData
+        // round trip. Hold the array itself so each read returns the same one.
+        Napi::Reference<Napi::Uint8Array> m_data;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: `Napi::Reference` owns the napi_ref and releases it in its own destructor, so the reference is dropped when the `ImageData` instance is finalized -- there is no leak and no lifetime extension beyond the wrapper. An explicit `Reset()` would be redundant with that. Keeping the reference alive for exactly the lifetime of the `ImageData` is also the point of the change: per spec `imageData.data` must return the *same* object every time, which is what makes `getImageData -> mutate -> putImageData` work.

Polyfills/Canvas/Source/Gradient.cpp
```diff
@@ -341,6 +349,45 @@ namespace Babylon::Polyfills::Internal
...
+    NVGpaint CanvasGradient::Paint()
+    {
+        UpdateCache();
+
+        auto nvg = context.lock();
+        if (!nvg)
+        {
+            // The owning context is gone; there is nothing sensible to paint with.
+            return NVGpaint{};
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, and correct -- the guard I added to `Paint()` ran after `UpdateCache()`, so it could never help: the crash had already happened inside it. `UpdateCache()` dereferenced `context.lock()` three times unchecked (once to delete the old image, and once more in each of `LinearGradientStops`/`RadialGradientStops` to create the new one).

Fixed in 7bd4ae14. `UpdateCache()` now locks once, checks both the `shared_ptr` and the `NVGcontext*` it holds, and keeps that lock alive for the whole bake; `LinearGradientStops`/`RadialGradientStops` take the context by reference so they cannot re-lock and observe it expiring between steps. If the context is gone the gradient stays dirty and keeps its handle, which `Dispose()` already knows not to delete.

`Dispose()` had a milder version of the same thing -- it called `context.lock()` twice in a row and dereferenced the second result -- so that now locks once too.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -774,9 +789,137 @@ namespace Babylon::Polyfills::Internal
...
+        {
+            dirtyX = info[3].ToNumber().Int32Value();
+            dirtyY = info[4].ToNumber().Int32Value();
+            dirtyWidth = info[5].ToNumber().Int32Value();
+            dirtyHeight = info[6].ToNumber().Int32Value();
+
+            if (dirtyWidth < 0)
+            {
+                dirtyX += dirtyWidth;
+                dirtyWidth = -dirtyWidth;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed, and fixed in 7bd4ae14. With `dirtyWidth = INT32_MIN` both `dirtyX += dirtyWidth` and `dirtyWidth = -dirtyWidth` are signed overflow, and these come straight from JS via `putImageData(img, 0, 0, 0, 0, -2147483648, 1)`.

The normalization and the clipping now run in `int64_t`. Every input is an int32 to begin with, so nothing can overflow a 64-bit intermediate, and by the time the values are used they have been clipped to the source bitmap and are back in range. The destination origin is the one value that is still unbounded (`dx` is arbitrary and gets offset by the clipped origin), so that one is computed in `int64_t` and saturated to `int32_t` -- anything that saturates is far enough off-canvas to clip to nothing downstream regardless.

Polyfills/Canvas/Source/Colors.h
```diff
@@ -211,22 +213,54 @@ namespace Babylon::Polyfills::Internal
...
+            static const std::string components{
+                R"(\(\s*)" + number + separator + number + separator + number +
+                "(?:" + alphaSeparator + number + R"()?\s*\))"};
+
+            // Component group indices: value/percent pairs at 1-2, 3-4, 5-6 and
+            // the optional alpha at 7-8.
+            const auto channel = [](const std::smatch& match, size_t index) {
+                const float value = std::stof(match[index]);
+                const float scaled = match[index + 1].matched && match[index + 1].length() > 0 ? value * 255.0f / 100.0f : value;
+                return static_cast<unsigned char>(std::clamp(scaled, 0.0f, 255.0f) + 0.5f);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Confirmed by experiment, not just inspection -- and it was worse than one call site. Fixed in 7bd4ae14.

I added a test that sets `ctx.fillStyle = "rgb(" + "9".repeat(400) + ", 0, 0)"`. Built **without** the fix, the unit test binary dies mid-suite:

```
[log]   Canvas2D
[log]     OK round-trips a string fillStyle and strokeStyle
...
[log]     OK restores a gradient strokeStyle across save/restore
--- stderr ---
[Uncaught Error] Unknown failure
=== exit code: 1 ===
```

It stops dead there and takes the remaining ~20 tests with it, exactly as you describe: `std::out_of_range` is not a `Napi::Error`, so node-addon-api's callback wrapper does not catch it and it unwinds out of the N-API boundary. With the fix, the same binary is 32 passing / exit 0.

While confirming it I found the same hazard in two more places that the existing font-parsing guard does *not* cover -- it was written for `std::invalid_argument` ("normal") only, so `out_of_range` still went through:

| Input | Site | Why the regex admits it |
|---|---|---|
| `ctx.fillStyle = "rgb(999...999, 0, 0)"` | `Colors.h` | component length is unbounded |
| `ctx.font = "18e999px Arial"` | `Font.cpp` | `SIZE_REGEX` explicitly accepts an exponent |
| `ctx.font = "999...999 18px Arial"` | `Font.cpp` | `WEIGHT_REGEX` is `\d+`, unbounded |
| `ctx.letterSpacing = "999...999px"` | `Context.cpp` | `\d+(\.\d+)?`, unbounded |

All four now parse with `strtof`/`strtol`, which cannot throw -- they saturate to `+/-HUGE_VALF` (or `LONG_MAX`) instead. Two details worth noting:

- The infinities are folded back to finite extremes before use. `std::clamp` on an infinity is fine, but `nvgHSLA` takes `fmodf` of the hue, and `fmodf(inf, 1)` is NaN, which would have quietly turned an overflow into a garbage color instead of a clamped one.
- A non-finite font size or letter spacing is rejected outright rather than clamped, so the previous valid value stays in effect instead of handing an absurd size to nanovg. Font weight clamps to the CSS 1-1000 range.

Covered by three new regression tests (`survives an out-of-range rgb() component`, `... font size and weight`, `... letterSpacing`). Full Playground sweep is unchanged at 304 PASS / 0 fail.

bkaradzic-microsoft:
You're right, and thanks for catching it — that was a genuine miss on my part, not a stale view on your end.

`canvas-shared-fixes` was sitting two commits ahead of the remote locally, so `99ff0b8b` (gradient semantics) and `7bd4ae14` (the putImageData / `std::stof` / null-context hardening) never made it to the PR. The branch is now pushed and the head is `7bd4ae14`; CI is re-running.

The part worth flagging: as you say, the review threads were already resolved, so the PR read as clean and green while the code being described wasn't actually on it. That's the bad failure mode here — the green checkmark was real, it just wasn't covering the fixes under discussion. I've re-verified against the pushed tree that each resolved thread now corresponds to code that's actually there:

- **`Colors.h`** — `std::stof` replaced with `strtof`, with the non-finite results folded back to finite extremes so the downstream clamps (and `nvgHSLA`'s `fmodf` on hue) stay well defined.
- **`Context.cpp` `putImageData`** — negative-extent normalization moved into `int64_t`, so negating `INT32_MIN` and summing two `INT32_MIN`s are no longer signed overflow.
- **`Gradient.cpp`** — `context.lock()` is locked once and checked for both a null `shared_ptr` and a null underlying context before any dereference, including on the `UpdateCache()` path that previously dereferenced it internally before the check.

I also confirmed the branch still merges cleanly onto current master — `config.json` is touched by both this branch and #1830, but it auto-merges without conflict.

Sorry for the round trip. Will wait for CI to come back green before asking for another look.

bkaradzic-microsoft:
CI is back green on `7bd4ae14` — 34/34, no failures. Ready for another look whenever you have time, @bghgary.

matthargett:
Thanks for closing these conformance gaps! If we are consistently beating the 2.5% threshold of incorrect pixels, I'd recommend lowering it to tighten things up a bit now that so much of the screenshot testing stability and accuracy has made such huge strides.

bkaradzic-microsoft:
Thanks! I agree with the direction, so I pulled the actual numbers rather than guess at the headroom.

Data is from this PR's green CI run (`7bd4ae14`), scraped from the `Pixel difference:` lines the harness logs per test. Validation runs on the `Win32_x64_*_D3D11` jobs, and **254** of the comparisons use the default 2.5% (the other 12 tests in `config.json` already carry explicit higher `errorRatio` overrides, 4%–35%; nothing currently overrides *downward*).

**Distribution on the 2.5% default**

| p50 | p75 | p90 | p95 | p99 | max |
|---|---|---|---|---|---|
| 0.352% | 0.709% | 1.134% | 1.428% | 1.846% | **2.249%** |

So the median has a ~7x margin, but the worst test sits at **2.249%** — only 0.25pp under the limit. Lowering the global default is therefore not quite free:

| new default | tests that would start failing |
|---|---|
| 2.0% | 2 |
| 1.5% | 8 |
| 1.0% | 33 |
| 0.5% | 94 |

**The good news is that these numbers are extremely stable**, which is really the property that matters for tightening a threshold. Across two independent CI runs on different commits (`bd3e3e27` → `7bd4ae14`): p50 0.349% → 0.352%, p90 1.116% → 1.134%, p99 identical at 1.846%, max identical at 2.249%. And within a single run the three JS-engine jobs (Chakra / QuickJS / V8) produce *byte-identical* distributions — same n, same p50/p90/max, same counts over every candidate threshold. The rendering is deterministic; the 2.5% isn't absorbing run-to-run noise, it's absorbing a small number of genuinely-large stable diffs.

That means a tightening is low-risk and won't introduce flakiness — it just has to be paired with per-test overrides for the tail, exactly the mechanism `config.json` already uses in the other direction. Concretely I'd suggest default **1.5%** plus explicit overrides on the 8 tests above it, which tightens 246 of 254 tests by 40% while leaving the known-bad ones documented and visible instead of hidden under a blanket allowance.

Two caveats worth stating: this is Win32 D3D11 only (D3D12 CI doesn't run the pixel comparison today — relevant to #1671), and the per-channel `threshold` of 25 is arguably the blunter knob of the two, since it's what lets a pixel be off by up to 25/255 per channel before it's even counted.

I'd rather not fold this into this PR though — it's repo-wide test policy and would churn `config.json` for 8 unrelated tests, whereas this PR is scoped to the Canvas polyfill. Happy to do it as a follow-up PR right after this lands, with the same measurements attached. Sound good?

bkaradzic-microsoft:
Pushed two more commits (`8bbf33bc`, `bd40ea28`) fixing three further issues, and triaged the twelve comments in the collapsed **Suppressed comments** block of the Copilot review — those don't create review threads, so they weren't visible in the resolved/unresolved list and I'd missed them until now.

Most of that block was generated against `bd3e3e27`, before the last two pushes, so a good part of it describes code that is no longer there. Breakdown:

### Fixed in this push

**`SavedStyle` didn't include the shadow attributes.** Correct, and worse than reported. `nvgRestore()` rewinds nanovg's copy of the state, but the wrapper keeps its own C++ mirrors of everything it has to reproject or that nanovg can't report back, and `SavedStyle` only snapshotted `fillStyle`/`strokeStyle`. So `lineCap`, `lineJoin`, `lineWidth`, `miterLimit`, the dash list, `filter`, `direction`, `letterSpacing` and `globalAlpha` had the same bug — the getters reported the post-`save()` value forever after, and the shadow attributes, which are re-applied from the mirror on every draw, actually rendered wrong. `SavedStyle` now snapshots all fifteen fields.

**Coincident color stops were dropped.** Correct — `std::map` + `insert()` silently discarded the second stop at an offset already present, which is exactly how the spec encodes a hard transition, so a stripe pattern came out as a smooth fade. Now a `std::multimap`. Both consumers only iterate in sorted order into a vector, so it's a drop-in, and the existing zero-width-span guard at `Gradient.cpp:51` turns the duplicate offset into the required hard edge.

**`setLineDash()` didn't retain the previous list on a rejected argument.** Correct, and the code contradicted its own comment: `m_lineDash` was cleared *before* validation, so after `setLineDash([5, 10])`, `setLineDash([-1])` left `getLineDash()` returning `[]` instead of `[5, 10]`. Now parsed into a temporary and committed only once every segment validates.

Three unit tests cover the state round-trip and the dash retention; a fourth covers the duplicate stops. 36/36 pass locally.

### Already fixed in `99ff0b8b`/`7bd4ae14` — these were generated against the older head

- *"`Fill()` never calls the new binder"* / *"only a fill-style binder is introduced; stroke still throws"* — both `BindFillStyle` and `BindStrokeStyle` exist and are called from all six draw sites (`Context.cpp:231, 329, 535, 626, 727, 1521`), and `m_fillStyle`/`m_strokeStyle` are both `std::variant<std::string, CanvasGradient*>`, not strings.
- *"premultiplied-alpha interpolation isn't actually implemented"* — it is, on both paths: `gradientSpan()` premultiplies at `Gradient.cpp:75-76` and unpremultiplies the baked sample at `:88`, and `lerpColor()` does the same at `:226-227` and `:232`.
- *"the radial paint still ignores `x0`, `y0`, `r0`"* — it doesn't; `RadialGradientStops()` calls `solveRadialOffset()` (`:362`), which solves the two-circle equation over all six parameters. The image is still mapped onto the outer circle's bounding box, but that's the correct place to put it now that the ramp is baked per-pixel rather than being a concentric texture.

### Known limitations, unchanged and deliberate

**Non-rectangular `clip()`.** Flagged as "not clipping" — that's fair, and it's called out as such in the PR description. nanovg's scissor is rectangle-only and there's no stencil path available here, so the choice is between an approximation and throwing. This keeps the rectangular case exact and degrades the non-rectangular one instead of failing the scene.

**`putImageData()` clobbers the current path.** Correct, and I'd rather not fix it in this PR. It isn't specific to `putImageData` — `fillRect()`, `clearRect()`, `drawImage()` and `Path2D` playback all call `nvgBeginPath()`, and per spec none of them may touch the current path. nanovg offers no way to fix it locally: `nvgSave`/`nvgRestore` only push and pop `NVGstate`, while `nvgBeginPath` clears `ctx->ncommands` and the path cache, which aren't part of that state. A real fix means journaling the current path wrapper-side and replaying it after each of those operations — the `Path2DCommand` machinery in `Path2D.h` is the obvious basis for it, but it means routing every path method on `Context` through the journal. That's a self-contained change and I'd prefer it as a follow-up rather than growing this PR further; happy to open an issue to track it.

CI was 34/34 green on the previous head; I'll confirm it stays green on `bd40ea28`.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -438,7 +504,17 @@ namespace Babylon::Polyfills::Internal
 
     void Context::Clip(const Napi::CallbackInfo& /*info*/)
     {
-        m_isClipped = true;
+        // A non-rectangular clip path cannot be expressed as a scissor rectangle.
+        // Emulate it by leaving the path current so the next fill draws it, and
+        // leave any enclosing scissor untouched rather than clipping to a
+        // rectangle this path never described.
+        if (m_pathHasNonRect)
```
bghgary, line 542: `roundRect()` + `clip()` misses this branch and falls through to the scissor below, clipping to `m_rectangleClipping` — which `RoundRect` sets to the plain bounding box (:502) without ever setting `m_pathHasNonRect`. The radii are dropped and the corners come out square. A rounded rect is exactly the kind of shape `nvgScissor` cannot express, which is what this flag's own comment says it tracks.

Pre-existing rather than a regression, but the flag is new here and this is the case that slips through it.

Polyfills/Canvas/Source/Context.h
```diff
@@ -120,24 +135,47 @@ namespace Babylon::Polyfills::Internal
...
+            std::string shadowColor;
+            double shadowBlur;
+            double shadowOffsetX;
+            double shadowOffsetY;
+            std::string filter;
+            std::string direction;
+            float miterLimit;
+            float lineWidth;
+            float globalAlpha;
+            float letterSpacing;
```
bghgary, line 165: `font` is the one exposed attribute left out. `SetFont` writes `m_font` and `m_currentFontId` (Context.cpp:1691-1698) and `GetFont` reads `m_font`, so it is the same wrapper-side mirror as the fields here, and per spec `restore()` rewinds it. It is also worse than the getter-only cases: `m_currentFontId` is what the text draw path binds, so the wrong font actually renders after a restore.

Holding this state as a single `State` struct on the class — with `save`/`restore` a push/pop of it — would make a forgotten field impossible rather than unlikely.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1157,7 +1442,67 @@ namespace Babylon::Polyfills::Internal
...
+        // stroke, and emulating one means splitting every path into segments.
+        // Draw solid and warn once rather than failing the whole scene, and keep
+        // the list so getLineDash() round-trips as the spec requires.
+        //
+        // Parsed into a temporary and committed only once every segment
+        // validates, because the spec keeps the previous list when the argument
+        // is rejected -- clearing m_lineDash up front would destroy it first.
+        std::vector<double> parsed;
+
+        if (info.Length() > 0 && info[0].IsArray())
```
bghgary: A non-array argument still wipes the list: the loop is skipped and `m_lineDash = std::move(parsed)` commits the empty temporary, so `setLineDash("x")` or `setLineDash(null)` clears where `setLineDash([-1])` now correctly does not. The *absent* case is clearly deliberate per the comment above, so this is only about a present-but-rejected argument.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -310,17 +379,12 @@ namespace Babylon::Polyfills::Internal
         nvgSave(*m_nvg);
         nvgGlobalCompositeOperation(*m_nvg, NVG_COPY);
 
-        if (!m_isClipped)
-        {
-            nvgBeginPath(*m_nvg);
-        }
+        // See FillRect: clipping is a scissor, so the path must always be reset.
+        nvgBeginPath(*m_nvg);
```
bghgary: Resetting the path here without clearing `m_isClipped` leaves the emulated clip pointing at a path that no longer exists. `FillRect` (:224) still skips its own `nvgBeginPath` while that flag is set, so `beginPath(); arc(); clip(); clearRect(a); fillRect(b)` appends `b` to the rect this call left behind and fills `a` union `b` unclipped. The `nvgRestore` below does not put the old path back -- it only pops `ctx->states`.

`m_pathHasNonRect` goes stale the same way, so a later `clip()` takes the emulation branch on what is now a plain rect. `PutImageData` clears it at :1003 for that reason. The three `DrawImage` sites have both issues too.

Polyfills/Canvas/Source/Gradient.cpp
```diff
@@ -38,23 +41,52 @@ namespace Babylon::Polyfills::Internal
 
     float clampf(float a, float mn, float mx) { return a < mn ? mn : (a > mx ? mx : a); }
 
+    // Canvas2D interpolates gradient stops in premultiplied sRGBA space. Interpolating the
+    // straight (unpremultiplied) components instead bleeds a transparent stop's RGB into the
+    // ramp: "#ff0000ff" -> "#00000000" darkens through maroon to black rather than staying red
+    // and only losing alpha.
```
bghgary: This explains why interpolation happens in premultiplied space, which is a decision made in `gradientSpan` and `lerpColor` rather than by this helper — `premultiply()` is `rgb * a` and reads fine without it.

`lerpColor` already says "See gradientSpan", but the text it points at is here, so that reference doesn't resolve today. Moving this block above the `premultiply` calls in `gradientSpan` fixes both.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -239,14 +264,37 @@ namespace Babylon::Polyfills::Internal
...
+            auto string = value.As<Napi::String>().Utf8Value();
+            const auto color = StringToColor(info.Env(), string);
+            m_strokeStyle = string;
+            nvgStrokeColor(*m_nvg, color);
+        }
+        else if (value.IsObject())
+        {
+            CanvasGradient* canvasGradient = CanvasGradient::Unwrap(value.As<Napi::Object>());
+            m_strokeStyle = canvasGradient;
+        }
```
bghgary: `IsObject()` sends any object to `Unwrap`, so `ctx.strokeStyle = {}` unwraps something that was never wrapped — the hazard this PR's own `putImageData` comment names ("napi_unwrap on an object that was never wrapped dereferences garbage"). Per spec the setter takes `(DOMString or CanvasGradient or CanvasPattern)`; anything else stringifies, fails colour parsing, and is ignored. So the test is "is this a `CanvasGradient`", not "is this an object".

`SetFillStyle` below has the same hole with no object gate at all. Pre-existing, but this PR gives the two setters the same input types with different validation — worth closing together.

If a distinct Dawn-side `CanvasGradient` rules out `InstanceOf` the way it did for `ImageData`, a property probe would do.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1099,6 +1322,68 @@ namespace Babylon::Polyfills::Internal
         }
     }
 
+    namespace
+    {
+        // createImageData has two overloads: (width, height) and (imagedata). Both produce a
+        // blank, transparent-black buffer -- only the dimensions differ in how they are sourced.
+        void ParseCreateImageDataArgs(const Napi::CallbackInfo& info, uint32_t& width, uint32_t& height)
```
bghgary: Inline this into `CreateImageData`. One caller, and the split costs the reader context: the caller reads as three lines that don't say what gets validated, and the out-params exist only to carry results back across a boundary that doesn't need to be there.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1157,7 +1442,67 @@ namespace Babylon::Polyfills::Internal
...
+
+        if (!m_lineDash.empty())
+        {
+            static bool warned = false;
+            if (!warned)
+            {
+                warned = true;
+                fprintf(stderr, "Context2D.setLineDash: dashed strokes are not supported; drawing solid.\n");
+            }
+        }
```
bghgary, line 1537: These warnings go to `stderr`, which on Android and iOS reaches nobody by default — the platforms where a silently-degraded canvas is hardest to diagnose.

`JsConsoleLogger::LogWarn(info.Env(), …)` calls `console.warn`, needs nothing but the `Napi::Env` already in hand, and lands where an app developer will see it. It lives in `Plugins/NativeEngine/Source/` today, so this means relocating it; `JsRuntimeHost`'s `Core/JsRuntime` fits — Canvas already links it through `JsRuntimeInternal`, it exposes `napi` publicly, and JsRuntimeHost owns the Console polyfill that provides the `console` being called. That makes it a JsRuntimeHost change plus a `GIT_TAG` bump here.

The shadow warning in `WarnShadowUnsupported` needs the same treatment, plus an `Napi::Env` parameter — it takes none today, and its three callers all have `info`.

Polyfills/Canvas/Source/Font.cpp
```diff
@@ -10,10 +14,95 @@ namespace
...
+            }
+
+            data = static_cast<const uint8_t*>(buffer.Data()) + byteOffset;
+        }
+        else
+        {
+            throw Napi::TypeError::New(env, std::string{methodName} + " expects the font data to be an ArrayBuffer or "
+                "an ArrayBuffer view, but got " + DescribeValue(value) + ". If the font was fetched with "
+                "Tools.LoadFileAsync, request it as binary -- the signature is LoadFileAsync(url, useArrayBuffer), "
+                "and a text response decodes the font to an unusable string.");
```
bghgary: This is the only Babylon.js API name that ships in a string rather than a comment. The Canvas polyfill may be consumed by embedders that don't use Babylon.js, and they get told to fix a `Tools.LoadFileAsync` call they never made.

The diagnosis is worth keeping — a text-decoded font is the usual cause — but stated against what this layer can see: the argument was a string rather than binary data. "expects the font data to be an ArrayBuffer or an ArrayBuffer view, but got a string — a font fetched as text rather than binary will arrive this way" says it without naming a caller.

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -63,15 +64,17 @@ namespace Babylon::Polyfills::Internal
...
         // don't allow same font to be loaded more than once
         // why? because Context doesn't update nvgCreateFontMem when old fontBuffer released
         auto fontName = info[0].As<Napi::String>().Utf8Value();
         if (fontsInfos.find(fontName) == fontsInfos.end())
         {
-            const auto buffer = info[1].As<Napi::ArrayBuffer>();
-            std::vector<uint8_t> fontBuffer(buffer.ByteLength());
-            memcpy(fontBuffer.data(), (uint8_t*)buffer.Data(), buffer.ByteLength());
-            fontsInfos[fontName] = std::move(fontBuffer);
+            fontsInfos[fontName] = GetFontDataArgument(info, 1, "Canvas.loadTTF");
```
bghgary: `loadTTFAsync` delegates straight to `LoadTTF` (line 84), so a bad argument through that entry point reports `Canvas.loadTTF` — the wrong method, which is the thing `methodName` exists to prevent. It is also the only value ever passed, so the parameter never varies.

Thread the real name from each entry point, or drop the parameter.

Polyfills/Canvas/Source/Path2D.cpp
```diff
@@ -43,6 +43,16 @@ namespace Babylon::Polyfills::Internal
...
+
+        // Browsers expose Path2D as a global constructor, and Babylon.js relies on that: the generic
+        // engines call `new Path2D(d)` from AbstractEngine.createCanvasPath2D. Only NativeEngine
+        // overrides that method to use `_native.Path2D`, so without the global any portable browser
+        // code doing `new Path2D(...)` fails with "Path2D is not defined".
+        auto global = env.Global();
+        if (global.Get(JS_PATH2D_CONSTRUCTOR_NAME).IsUndefined())
+        {
+            global.Set(JS_PATH2D_CONSTRUCTOR_NAME, func);
+        }
```
bghgary: When a global `Path2D` already exists this keeps the foreign one, and `Context::Fill` (315-317) unwraps whatever object it is handed as a `NativeCanvasPath2D`. So the branch that preserves someone else's constructor is the branch that guarantees `ctx.fill(new Path2D(d))` unwraps an object that was never wrapped.

`Canvas::Initialize` calls this at Canvas.cpp:311 while the polyfill is still installing, so a global found here came from outside it.

Claiming the global unconditionally is the fix: a foreign `Path2D` cannot be drawn by this `Context` under any circumstances, so deferring to it only guarantees the failure. Verifying the object in `Fill`/`Stroke` is worth doing as well, but that converts undefined behaviour into a clean error rather than making the call work.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -129,6 +145,187 @@ describe("ColorParsing", function () {
...
+    const ctx = createContext();
+    const gradient = ctx.createLinearGradient(0, 0, 64, 0);
+    gradient.addColorStop(0, "red");
+    gradient.addColorStop(0.5, "red");
+    gradient.addColorStop(0.5, "blue");
+    gradient.addColorStop(1, "blue");
+    expect(function () {
+      ctx.fillStyle = gradient;
+      ctx.fillRect(0, 0, 64, 64);
+    }).to.not.throw();
```
bghgary, line 397: This asserts only that nothing throws, which was already true before the `std::map` → `std::multimap` change: `AddColorStop` calls `colors.insert(...)` (Gradient.cpp:182), and `std::map::insert` drops a duplicate key by returning `{it, false}` rather than throwing. So it passes with and without the change and doesn't guard it.

Catching a dropped stop needs the rendered result — colour stops have no getter to assert against, unlike the dash list above.

bghgary (commented):

Twelve inline.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Inlined in d690deb3 -- `ParseCreateImageDataArgs` is gone, its body is in `CreateImageData`, and the now-empty anonymous namespace with it. The out-params went away with the boundary.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in d690deb3. Added `CanvasGradient::IsInstance(env, value)` and both setters now use it, so neither reaches `Unwrap` for a foreign object and an unusable value leaves the attribute unchanged, per spec.

`InstanceOf` works here where it didn't for `ImageData`: `CanvasGradient` is registered on `_native` (`JsRuntime::NativeObject`), so the constructor is reachable and exact. The `ImageData` property probe exists because callers legitimately pass plain `{data, width, height}` objects -- there's no equivalent duck-typing use for gradients, so the strict check is the right one.

New test `ignores a fillStyle or strokeStyle that is neither a color string nor a gradient`. Without the fix it doesn't just mis-assign, it throws out of the unwrap:

```
AssertionError: expected [Function] to not throw an error but 'Error: Invalid argument' was thrown
```

Polyfills/Canvas/Source/Gradient.cpp
bkaradzic-microsoft: Moved in d690deb3 -- the block now sits in `gradientSpan` immediately above the `premultiply` calls, so `lerpColor`'s "See gradientSpan" resolves and `premultiply()` is left to read as the one-liner it is.

Polyfills/Canvas/Source/Canvas.cpp
bkaradzic-microsoft: Fixed in 21ac5229. Both entry points now call a shared `LoadTTFCore(info, methodName)` and pass their own name, so `loadTTFAsync` reports `Canvas.loadTTFAsync`. Threading the real name rather than dropping the parameter, since the parameter was right -- it was just never given a second value.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Agreed on all of it -- `stderr` is the wrong sink here, and `JsConsoleLogger::LogWarn` with the `Napi::Env` already in hand is the right one.

I'd like to do it as a separate PR rather than in this one. As you laid out it isn't a local change: `JsConsoleLogger` has to move from `Plugins/NativeEngine/Source/` into JsRuntimeHost's `Core/JsRuntime`, which is a JsRuntimeHost PR, and then this repo needs a `GIT_TAG` bump before either call site can use it. That sequencing means it can't land atomically with the fixes here, and I'd rather not hold them behind it.

`WarnShadowUnsupported` gaining an `Napi::Env` parameter belongs in that same change -- all three callers have `info`, so it's mechanical once the logger is reachable.

I'll open the JsRuntimeHost PR next and link it here. Say the word if you'd rather this PR wait for it instead.

Polyfills/Canvas/Source/Path2D.cpp
bkaradzic-microsoft: Agreed and fixed in d690deb3 -- the global is now claimed unconditionally, with your reasoning recorded in the comment: a foreign `Path2D` cannot be drawn by this `Context` under any circumstances, so the branch that preserved it was the branch that guaranteed the bad unwrap.

Verifying the object in `Fill`/`Stroke` is worth doing too, and I'd like to take it with the `CanvasGradient::IsInstance` pattern from the other thread rather than bolt on a second one-off check -- happy to add it here if you'd rather not split it.

Polyfills/Canvas/Source/Font.cpp
bkaradzic-microsoft: Fixed in 21ac5229. The message now reads:

> expects the font data to be an ArrayBuffer or an ArrayBuffer view, but got <describe> . A font fetched as text rather than binary decodes to an unusable string and arrives this way.

Keeps the diagnosis, drops the Babylon.js API name an embedder may have no relationship with.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bkaradzic-microsoft: You're right, and thanks for catching it -- I confirmed it passes with and without the `std::map` -> `std::multimap` change, so it guarded nothing.

I couldn't turn it into a real assertion: `getImageData` reads the CPU mirror that `DrawImage` populates, and the GPU nanovg framebuffer isn't read back, so a gradient fill produces nothing to sample. With no getter on the stops either, there's no observable the test can reach.

So rather than leave a test claiming more than it checks, I've retitled it to `accepts two color stops at the same offset` and recorded in the comment exactly what it does and does not prove -- that `std::map::insert` drops the duplicate by returning `{it, false}` instead of throwing, so "does not throw" was never going to catch it, and that asserting the survival of the stop needs the rendered ramp.

Getting that coverage properly means pixel readback for gradient fills, which is a bigger change than this PR should carry. Flagging it as the gap it is rather than papering over it.

Polyfills/Canvas/Source/Context.h
bkaradzic-microsoft: Fixed in d690deb3 -- `SavedStyle` now carries `font` and `currentFontId`, and `Restore` rewinds both. Confirmed it was the real-rendering case you describe and not just a getter: with only `Save` updated the test still failed on the bound face.

New test `restores font`, which fails without the fix:

```
AssertionError: expected '40px "Times"' to equal '18px "Arial"'
```

(It compares against the round-tripped value rather than the assigned literal, since the getter reports a normalized serialization.)

On folding the whole thing into a single `State` struct that `save`/`restore` push and pop -- I agree, and it's the right shape; a forgotten field is exactly how this one got missed. I'd rather not do it in this PR: it touches every getter and setter on `Context`, which would bury the behavior fixes here in a rename diff. Happy to do it as the immediate follow-up, or here if you'd prefer it not wait.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: You're right that `roundRect` never sets the flag, and I set it -- then the visual tests said no.

Routing `roundRect` into the emulation branch makes the output substantially worse. The emulation leaves the path current so the next fill draws it, and nanovg fills the **union** of the subpaths rather than their intersection, so `roundRect(); clip(); fillRect()` paints the whole `fillRect` instead of the rounded region. Measured on the `Native Canvas` visual test:

| `roundRect` sets `m_pathHasNonRect` | pixel difference |
|---|---|
| no (current) | 1.850% |
| yes | 20.980% |

I also gated it on nanovg's own 0.1 radius threshold first, in case near-zero radii were the problem. Same result.

So dropping the radii is wrong, but it is the far smaller of the two errors, and the emulation cannot be made right by feeding more shapes into it -- a correct `clip()` needs real path clipping (a stencil pass in nanovg). I've left `roundRect` on the scissor and put the measurement and the reasoning in a comment at the end of `RoundRect` so the next reader doesn't repeat the experiment. Happy to open a follow-up issue for the stencil-based clip if you'd like it tracked.

d690deb3

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in d690deb3. A present-but-rejected argument now returns before the commit, so the previous list survives; an absent argument still means solid, as the comment intends.

New test `keeps the previous dash list when the argument is not a list` covers `"x"`, `null` and `7`, plus the no-argument case. It fails without the fix:

```
AssertionError: expected [] to deeply equal [ 5, 10 ]
```

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Confirmed and fixed in d690deb3. Both flags did go stale, and the `nvgRestore` does not restore the path.

Added `Context::ResetPathState()` -- clears `m_isClipped`, clears `m_pathHasNonRect`, begins a new path -- and routed `BeginPath`, `ClearRect`, all three `DrawImage` overloads and `PutImageData` through it. `PutImageData`'s manual `m_pathHasNonRect = false` is gone as redundant. Keeping the three steps in one function is what stops the next path reset from getting it half right, which is how this happened.

bkaradzic-microsoft:
Pushed d690deb3 + 21ac5229, rebased onto master (`8bf91bc5`).

All twelve threads answered. Ten are fixed; two are answered without a code change and say why:

- **`roundRect` + `clip`** -- the fix makes it worse. Setting `m_pathHasNonRect` there takes the `Native Canvas` visual test from 1.850% to 20.980%, because the emulation fills the union of the subpaths, not their intersection. Kept on the scissor, measurement recorded in a comment.
- **`console.warn` for the degraded-canvas warnings** -- agreed, but it needs `JsConsoleLogger` moved into JsRuntimeHost plus a `GIT_TAG` bump, so it can't land atomically here. Proposing a follow-up PR.

Four new unit tests, each verified to fail with the fix reverted and pass with it:

| test | failure without the fix |
|---|---|
| `keeps the previous dash list when the argument is not a list` | `expected [] to deeply equal [ 5, 10 ]` |
| `restores font` | `expected '40px "Times"' to equal '18px "Arial"'` |
| `ignores a fillStyle or strokeStyle that is neither a color string nor a gradient` | `'Error: Invalid argument' was thrown` |
| `accepts two color stops at the same offset` | retitled -- see the thread; it does not guard the `multimap` change and now says so |

Validation on Win32 D3D11:

- `UnitTests`: 21/21 suites, 39/39 JavaScript assertions.
- Full Playground visual sweep: `ran=305 passed=305 failed=0 missingRef=0`.

One thing worth flagging that no thread raised: this PR moves `Native Canvas` from 0.391% to 1.850% against a 2.5% allowance. It arrives with the first commit in the branch and is the expected consequence of changing gradient and colour semantics against a reference image captured under the old behaviour -- but the remaining margin is thin, so it's worth knowing about.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -223,30 +248,52 @@ namespace Babylon::Polyfills::Internal
...
+        if (value.IsString())
+        {
+            auto string = value.As<Napi::String>().Utf8Value();
+            const auto color = StringToColor(info.Env(), string);
+            m_strokeStyle = string;
+            nvgStrokeColor(*m_nvg, color);
+        }
+        else if (CanvasGradient::IsInstance(info.Env(), value))
+        {
+            m_strokeStyle = CanvasGradient::Unwrap(value.As<Napi::Object>());
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, and it reproduces. I instrumented `~CanvasGradient` and the bind path and ran a test that creates the gradient in a scope holding no reference to it, then forces a collection:

```
~CanvasGradient      this=0000021D4B9DA240
BindFillStyle deref  this=0000021D4B9DA240
```

Same address, destroyed and then dereferenced. It did not crash, which is why it went unnoticed -- the freed memory simply had not been reused yet. `CanvasGradient` is an `ObjectWrap`, so the native instance is owned by the JS wrapper and deleted from its finalizer, and `ctx.fillStyle = ctx.createLinearGradient(...)` leaves nothing rooting that wrapper.

Fixed in 4201348e by storing the assigned object in a persistent reference rather than a raw pointer. `SavedStyle` copies the styles on `save()`/`restore()` and `Napi::ObjectReference` is move-only, so it is held through a `shared_ptr` and the copies share one strong reference.

I also took your point about the getter. It now returns the gradient object instead of a `Napi::External` wrapping the raw pointer, which is what the spec asks for and makes `otherCtx.fillStyle = ctx.fillStyle` round-trip -- previously the getter handed back an `External` that the setter would then reject.

Two tests cover it, one for `fillStyle` and one for `strokeStyle`. Each tags the gradient with an expando, drops every reference, collects, and then checks the tag survives and the native gradient still accepts `addColorStop`. Both fail with the change reverted and pass with it. Full suite is 21/21 and the visual sweep is 305/305.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1099,6 +1335,62 @@ namespace Babylon::Polyfills::Internal
...
+            const auto source = info[0].As<Napi::Object>();
+            const auto w = source.Get("width");
+            const auto h = source.Get("height");
+            if (!w.IsNumber() || !h.IsNumber())
+            {
+                throw Napi::TypeError::New(info.Env(), "Context2D.createImageData: argument is not an ImageData");
+            }
+
+            width = w.As<Napi::Number>().Uint32Value();
+            height = h.As<Napi::Number>().Uint32Value();
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Confirmed, and it is worse than just `-1`. I instrumented the old code and measured what each source actually did:

| source | old behaviour |
| --- | --- |
| `{width: -1, height: 1}` | wraps to 4294967295, attempts ~17 GB |
| `{width: 5e9, height: 1}` | wraps to 705032704, attempts ~2.8 GB |
| `{width: 1.5, height: 1}` | silently accepted as 1x1 |
| `{width: Infinity, height: 1}` | already rejected -- `ToUint32` gives 0, caught by the non-zero check |

The first two surface as an error only because the allocation happens to fail on this machine, and the engine reports it as a generic `Invalid function argument` rather than anything meaningful. A wrapped extent that *is* allocatable would succeed and hand back a buffer whose dimensions bear no relation to what was asked for, which is the more dangerous outcome.

Fixed in 7955ed16 by reading the dimensions as doubles and requiring a finite, non-negative integer that fits in `uint32_t` before the conversion.

Two related spots I checked while I was in there and deliberately left alone:

- The `(width, height)` overload just above already handles its own wrap -- it reads via `Int32Value()`, takes the magnitude per spec, and rejects `INT32_MIN` explicitly.
- The `drawImage` ImageBitmap path reads `width`/`height` off an equally duck-typed object, but a wrapped value cannot get through: it is rejected by the `> 65535` clamp that exists because `bimg::imageGetSize` takes `uint16_t` extents.

So `createImageData` was the one place left. Covered by a new test over all four inputs above, plus one asserting a valid source still works; it fails on the first non-rejected case with the change reverted. Suite is 21/21 with 43 JS assertions.

Polyfills/Canvas/Source/Context.h
bghgary: [Responded by Copilot on behalf of @bghgary]

Do the `State` struct as the follow-up, landing immediately after this PR rather than queuing behind further Canvas work.

Polyfills/Canvas/Source/Context.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

My premise was wrong: the emulation fills the union of the subpaths, so `roundRect(); clip(); fillRect()` would lose the clip entirely rather than square its corners. The scissor is the smaller error.

Yes, please open the follow-up for stencil-based path clipping.

Polyfills/Canvas/Source/Path2D.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

Take it with the `IsInstance` follow-up, immediately after this PR. `Stroke` is the worse of the two when you get there — it has no argument gate at all, so `ctx.stroke("x")` reaches `Unwrap` with a string, where `Fill` at least stops at non-objects.

Polyfills/Canvas/Source/Context.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

A separate PR is right. Land it immediately after this one rather than letting it queue behind further Canvas changes.

Polyfills/Canvas/Source/Context.h
bkaradzic-microsoft: Will do -- the `State` struct is the next PR out, not queued behind other Canvas work.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Thanks for double-checking it. I will open the follow-up for stencil-based path clipping so `clip()` can honour a real path instead of falling back to the bounding-box scissor.

Worth recording for whoever picks it up: the scissor is not merely *a* smaller error, it is bounded. It can only ever clip too little (the bounding box is a superset of the path), so content leaks into the corners. The union behaviour fails the other way and unbounded -- the clip is silently dropped and the whole subsequent fill lands. Measured on the roundRect test that was 1.850% with the scissor and 20.980% with the union.

Polyfills/Canvas/Source/Path2D.cpp
bkaradzic-microsoft: Agreed, and confirmed in the current code -- `Stroke` is the worse of the two:

```cpp
// Fill (Context.cpp:330) -- gated
const NativeCanvasPath2D* path = info.Length() >= 1 && info[0].IsObject()
    ? NativeCanvasPath2D::Unwrap(info[0].As<Napi::Object>())
    : nullptr;

// Stroke (Context.cpp:651) -- no type gate at all
const NativeCanvasPath2D* path = info.Length() == 1 ? NativeCanvasPath2D::Unwrap(info[0].As<Napi::Object>()) : nullptr;
```

So `ctx.stroke("x")` casts a string through `As<Napi::Object>()` straight into `Unwrap`. `Fill` stops at non-objects, which also happens to cover the legal `ctx.fill("evenodd")` spelling.

I will take both in the `IsInstance` follow-up right after this PR, so the gate is a real instance-of check rather than `IsObject()` -- the latter still lets `ctx.fill({})` through to `Unwrap`.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Understood -- the JsRuntimeHost PR moving `JsConsoleLogger` goes up immediately after this one, and the Canvas side follows with the `GIT_TAG` bump once it lands.

bghgary (approved):

PR #1830, opened 2026-08-12
Playground validation: make a full 720-test sweep possible

bkaradzic-microsoft:
Two changes that together make a complete run of the 720-test validation suite possible and order-independent. Two commits, reviewable separately.

## 1. `--keep-going`, so a sweep can finish

The runner exits on the first failing test, so a run only ever reports one failure and everything after it goes unmeasured. There's no way to answer *"how many of the 720 tests pass today?"* without repeatedly excluding whatever failed and re-running.

`--keep-going` / `-k` records the failure, continues, and lists every failing title in the summary. The run still exits non-zero if anything failed, so **CI semantics are unchanged**. Without the flag, behaviour is exactly as before.

```
Run complete. ran=3 passed=1 failed=2 missingRef=0 skipped=0
Failed tests (2):
  - Gaussian Splatting Compressed ply SH
  - Gaussian Splatting Update Data
```

Verified with two tests pointed at a mismatched reference image plus one control:

| | result | exit |
|---|---|---|
| without `--keep-going` | `ran=1 passed=0 failed=1` | -1 |
| with `--keep-going` | `ran=3 passed=1 failed=2` | -1 |

**Framebuffer pool floor 512 -> 2048.** Running to completion for the first time immediately exhausted the bgfx framebuffer pool around the FrameGraph tests (`bgfx::createFrameBuffer returned invalid handle`). 512 was only ever sufficient *because* the run stopped early. The existing comment in `Dependencies/CMakeLists.txt` already anticipated this ("combined with V8 GC pacing this can exceed the default limit during long playground sweeps"); this raises the floor to cover a full run.

## 2. Cross-test state leaks

The suite reuses **one engine across all tests**, so anything outliving `scene.dispose()` carries into later tests. The browser harness never sees this because it reloads the page per test.

- **Stray scenes.** A test can leave extra scenes registered on the engine (an async load that created its own scene, or one whose creation promise resolved after validation). They stay registered and keep their resources alive.
- **The engine texture cache.** `BaseTexture._getFromCache` keys on `url`/`noMipmap`/`isCube` only -- *not* on load-time options. A test that leaves a reference behind (e.g. assigning one texture to both `scene.environmentTexture` and a material's `reflectionTexture`) keeps its internal texture cached across `scene.dispose()`, so a later test loading the same URL silently reuses the *previous* test's texture along with its prefiltering/irradiance settings.
- **`SceneLoader.OnPluginActivatedObservable`.** Global, outlives the scene. Snippets use it to configure the glTF loader (`animationStartMode`, `compileMaterials`, ...) and never unregister, so every later glTF test inherits those settings.

Plus a hang guard: a `createScene` promise that never resolves currently hangs the *entire* suite, because the existing `onReadyTimeout` net lives inside `processCurrentScene` and only applies once the promise has resolved. This converts it to a single failed test. (Documented caveat: it only fires if the JS event loop keeps running.)

**This second commit changes no current result** -- 302 ran / 299 passed / 3 failed, identical set, before and after. Its value is removing order-dependence that becomes load-bearing as more tests are enabled.

## Why it matters

This is what made a complete 720-test sweep possible at all. Measuring the full suite in one pass turned up 47 failures previously hidden behind the first one -- and spot-checking those in isolation shows they **pass individually**, so they're cross-test state pollution rather than missing functionality. That's a materially different and far more tractable problem than the tally alone suggested, and none of it is visible without this.

### DIFF
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -8,6 +8,7 @@
     const testHeight = 400;
     const generateReferences = !!opts.generateReferences;
     const breakOnFail = !!opts.breakOnFail;
+    const keepGoing = !!opts.keepGoing;
     const listTests = !!opts.listTests;
     const includeExcluded = !!opts.includeExcluded;
     const testFilters = Array.isArray(opts.testFilters) ? opts.testFilters.map(s => String(s).toLowerCase()) : [];
@@ -72,6 +73,7 @@
     let failedCount = 0;
     let skippedCount = 0;
     let missingRefCount = 0;
+    const failedTitles = [];
 
     function getExclusionReason(t) {
         if (t.onlyVisual) {
@@ -99,6 +101,12 @@
                     " failed=" + failedCount +
                     " missingRef=" + missingRefCount +
                     " skipped=" + skippedCount);
+        if (failedTitles.length > 0) {
+            console.log("Failed tests (" + failedTitles.length + "):");
+            for (let n = 0; n < failedTitles.length; n++) {
+                console.log("  - " + failedTitles[n]);
+            }
+        }
     }
 
     const engine = new BABYLON.NativeEngine();
@@ -663,18 +671,22 @@
                     ranCount++;
                     if (!status) {
                         failedCount++;
+                        failedTitles.push(currentTitle);
                         // failTest() already triggered the debugger before
                         // reaching this callback; no second `debugger` here.
-                        logRunSummary();
-                        TestUtils.exit(-1);
-                        return;
+                        if (!keepGoing) {
+                            logRunSummary();
+                            TestUtils.exit(-1);
+                            return;
+                        }
+                    } else {
+                        passedCount++;
                     }
-                    passedCount++;
                     i++;
                     if (justOnce || i >= config.tests.length) {
                         logRunSummary();
                         engine.dispose();
-                        TestUtils.exit(0);
+                        TestUtils.exit(failedCount > 0 ? -1 : 0);
                         return;
                     }
                     // Defer next iteration to avoid blowing Chakra's
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -228,6 +228,15 @@
 
         currentScene.dispose();
         currentScene = null;
+
+        // A test can leave extra scenes behind (an async load that created its own scene, a scene
+        // whose creation promise resolved after validation, ...). They stay registered on the
+        // reused engine and keep their resources alive, so dispose them here.
+        const strayScenes = engine.scenes.slice();
+        for (let i = 0; i < strayScenes.length; ++i) {
+            strayScenes[i].dispose();
+        }
+
         engine.setHardwareScalingLevel(1);
 
         // Reset render state that persists on the reused engine so each test starts fresh.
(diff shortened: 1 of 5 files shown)


Dependencies/CMakeLists.txt
```diff
@@ -60,12 +60,17 @@ set(BGFX_KHRONOS_INCLUDE_DIR "${bgfx.cmake_SOURCE_DIR}/bgfx/3rdparty/khronos" CA
...
+# CI workflows pass -DBGFX_CONFIG_MAX_FRAME_BUFFERS, and any value below 2048
 # (including a smaller CI override) is clamped up so the pool never regresses
 # below the size the Canvas sweeps need.
-if(NOT BGFX_CONFIG_MAX_FRAME_BUFFERS OR BGFX_CONFIG_MAX_FRAME_BUFFERS LESS 512)
-    set(BGFX_CONFIG_MAX_FRAME_BUFFERS 512)
+#
+# 512 was sufficient only because the validation runner exited on the first
+# failing test. With --keep-going a full 720-test sweep keeps allocating past
+# that point and exhausts the pool around the FrameGraph tests, so the floor
+# has to cover a complete uninterrupted run.
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, fixed in b31520b9 -- now reads `a full validation sweep`. The suite size is genuinely in flux (tests get enabled as Native reaches parity), so an embedded count would have gone stale quickly.

Apps/Playground/Scripts/validation_native.js
```diff
@@ -411,11 +447,35 @@
                                     currentScene = eval(pgCode);
 
                                     if (currentScene.then) {
-                                        // Handle if createScene returns a promise
+                                        // Handle if createScene returns a promise. Guard against a
```
ryantrem, line 452: This code would probably be much simpler and easier to read if we just used await. Can we use it, or do we need to run this code in JS engines that don't support await?

ryantrem (approved):

Apps/Playground/Scripts/validation_native.js
bkaradzic-microsoft: Good call -- done in b473eaa7, and yes we can use it.

I checked the engine question first rather than assuming, because it turns out the obvious evidence is misleading. The CI matrix here covers six engines (Chakra, V8, JSC, Hermes, QuickJS, JSI), and the fact that the suite passes today does *not* demonstrate `await` support: the shipped `babylon.max.js` bundle contains zero native `async function` and 643 `__generator` helpers, i.e. Babylon transpiles async away before it ever reaches these engines. So nothing in the existing suite was exercising native async/await.

Checking the engines individually, they all support async functions. The one I expected to be a problem was Hermes, since this repo builds it with `-D HERMES_ALLOW_BOOST_CONTEXT=0` -- but that flag only affects async *generators* (`async function*`), which Hermes doesn't support at all regardless and which this change doesn't use. Plain `async`/`await` needs no Boost.Context.

On the refactor itself: `Promise.race` already gives the settle-once semantics that the `sceneSettled` latch was hand-rolling, so the three duplicated `if (sceneSettled) { return; }` guards, the duplicate `clearTimeout`, and the separate rejection path all collapse into the existing `try`/`catch`. Net 23 insertions / 25 deletions.

Two details worth flagging since they're easy to get wrong:

- **The timer is cleared in a `finally`, not per-settle-path.** Left pending it keeps the event loop alive for the full 10 minutes after a scene that resolved normally.
- **`eval(pgCode)` still runs before the first `await`.** An async function body executes synchronously up to its first suspension point, so the eval keeps happening at the shallow stack depth that the enclosing `setTimeout(..., 0)` exists to provide for engines with a small C stack (QuickJS). That property is preserved, not incidental.

I also hardened the promise test from `currentScene.then` to `currentScene && currentScene.then`, so a snippet returning null now hits the readable "Failed to evaluate playground snippet" path instead of a TypeError.

Verified on Win32 D3D11 with a full headless sweep on the upstream config: `ran=302 passed=299 failed=3`, identical to the pre-refactor baseline including the same three failing titles (those three are artifacts of my local build, not upstream). CI will cover the remaining five engines -- I'll confirm once it's green, and revert to the callback form if any of them disagree.

bkaradzic-microsoft: CI is green across all 34 jobs on b473eaa7, so `await` is confirmed fine on every engine in the matrix -- no revert needed.

The runtime-executing jobs are the meaningful ones here (the Android/UWP jobs only cross-compile), and they all pass:

| engine | job | result |
|---|---|---|
| Hermes | `Ubuntu_Clang_Hermes`, `Win32_x64_Hermes_D3D11` | pass |
| QuickJS | `Ubuntu_Clang_QuickJS`, `MacOS_QuickJS`, `Win32_x64_QuickJS_D3D11` | pass |
| JSC | `Ubuntu_Clang_JSC`, `Ubuntu_GCC_JSC` | pass |
| JSI | `Win32_x64_JSI_D3D11`, `UWP_arm64_JSI` | pass |
| Chakra | `UWP_x64`, `UWP_arm64` | pass |
| V8 | `Win32_x64_V8_D3D11`, `Android_Ubuntu_V8` | pass |

Hermes was the one I actually expected to be at risk, and it passes both the Linux and Win32 D3D11 runs.

PR #1831, opened 2026-08-12
Playground validation: remove cross-test state leaks from the reused engine

bkaradzic-microsoft:
> [!NOTE]
> Stacked on #1830 — review that one first. This PR's diff is the single `validation_native.js` commit on top.

## Problem

The suite reuses **one engine across all 720 tests**, so anything outliving `scene.dispose()` silently carries into later tests. The browser harness never sees this because it reloads the page per test.

## The three leaks

**1. Stray scenes.** A test can leave extra scenes registered on the engine — an async load that created its own scene, or a scene whose creation promise resolved after validation finished. They stay registered and keep their resources alive.

**2. The engine texture cache.** `BaseTexture._getFromCache` keys on `url`/`noMipmap`/`isCube` only — **not** on load-time options. A test that leaves a reference behind (e.g. assigning one texture to both `scene.environmentTexture` and a material's `reflectionTexture`) keeps its internal texture cached across `scene.dispose()`. A later test loading the same URL then silently reuses the *previous* test's texture, along with its prefiltering/irradiance settings.

**3. `SceneLoader.OnPluginActivatedObservable`.** Global, outlives the scene. Snippets use it to configure the glTF loader (`animationStartMode`, `compileMaterials`, …) and never unregister, so every later glTF test inherits those settings.

## Also: a hang guard

A `createScene` promise that never resolves currently hangs the **entire suite**. The existing `onReadyTimeout` safety net lives inside `processCurrentScene` and only applies *after* the promise resolves. This converts it to a single failed test.

(Caveat documented in-code: this only fires if the JS event loop keeps running. A snippet that blocks the JS thread natively isn't rescued by it.)

## Impact — deliberately none today

| | ran | passed | failed |
|---|---|---|---|
| before | 302 | 299 | 3 |
| after | 302 | 299 | 3 |

Same three failures, identical set. **This PR changes no current result.** Its value is removing order-dependence that becomes load-bearing as more tests get enabled — leak #2 in particular produces failures that only reproduce in sequence and pass in isolation, which is exactly the class of bug that's most expensive to debug later.

I'm raising it now rather than bundled with a test-enablement PR so the mechanism can be reviewed on its own.

### DIFF
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -8,6 +8,7 @@
     const testHeight = 400;
     const generateReferences = !!opts.generateReferences;
     const breakOnFail = !!opts.breakOnFail;
+    const keepGoing = !!opts.keepGoing;
     const listTests = !!opts.listTests;
     const includeExcluded = !!opts.includeExcluded;
     const testFilters = Array.isArray(opts.testFilters) ? opts.testFilters.map(s => String(s).toLowerCase()) : [];
@@ -72,6 +73,7 @@
     let failedCount = 0;
     let skippedCount = 0;
     let missingRefCount = 0;
+    const failedTitles = [];
 
     function getExclusionReason(t) {
         if (t.onlyVisual) {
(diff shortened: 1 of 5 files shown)


bkaradzic-microsoft:
Folded into #1830 -- both commits touch the same file and tell one story (make a full sweep possible, then make it order-independent), so reviewing them together is less overhead than a stack. The state-leak commit is unchanged, just carried over as the second commit there.

Issue #223, opened 2020-04-10
Enable V8 debugging


bghgary:

bghgary:
Closing in favor of #463

Issue #225, opened 2020-04-10
Investigate JSI as replacement for NAPI


bghgary:
<strike>Implement a version of NAPI that will support React Native's JSI interface.</strike>
Implementing NAPI against JSI is a nightmare.

Currently investigating implementing JSI directly instead of NAPI.

For reference, JSI already has implementations for multiple JS VMs:
- [Chakra](https://github.com/microsoft/react-native-windows/blob/master/vnext/JSI/Shared/ChakraRuntime.h)
- [V8](https://github.com/microsoft/v8-jsi)
- [JSC](https://github.com/facebook/react-native/tree/ede2c5031f93068c93786a550ef4def9f84e6b16/ReactCommon/jsi)
- [Hermes](https://github.com/facebook/hermes)

TODOs:
- [X] Prototype NativeEngine with JSI instead of NAPI
- [ ] Measure performance of NativeEngine with JSI compared to NAPI

ryantrem:
Some JSI experiments/tests are available here:
- https://github.com/BabylonJS/Babylon.js/blob/62359f77e9ac03081a4c0ffc2e856f6e7a5c9f69/src/Engines/Extensions/index.ts#L16-L18
- https://github.com/BabylonJS/Babylon.js/blob/eecd7be9356caabef655429a2ac0fc2447bdb640/src/index.ts#L12

PR #1835, opened 2026-08-13
Add a native Draco encoder to the NativeDraco plugin

bkaradzic-microsoft:
Adds a native Draco **encoder** to the `NativeDraco` plugin as `_native.DracoCodec.Encode`. The plugin already ships a native decoder, so this completes the pair and lets Babylon.js's `DracoEncoder` use a synchronous native path instead of fetching and instantiating the `draco_encoder` WebAssembly module at runtime.

## Implementation

Deliberately mirrors Draco's own emscripten glue so the native and WASM paths agree:

- `AddAttributeToMesh<T>` mirrors `PointCloudBuilder::AddAttribute<T>`
- `AddTypedAttributeToMesh` mirrors the WASM encoder's `addAttributeMap`
- `ReadIndices` handles `Uint16Array` / `Uint32Array` index upload
- `TypedArrayData<T>` honors the view's `ByteOffset` rather than assuming offset 0

The returned `{ data, attributeIds }` shape matches what the existing WASM worker and module paths produce, so the JavaScript side needs no special-casing beyond its feature probe. Attribute buffers are documented as Babylon.js `VertexDataTypedArray` (not bare `ArrayBufferView`, which also admits `DataView` / BigInt typed arrays this implementation rejects).

## glTF bitstream (`DRACO_GLTF_BITSTREAM=ON`)

NativeDraco keeps Draco's **glTF bitstream subset**. The plugin is intended to produce and consume glTF-compatible Draco data; Babylon.js defaults to the glTF-only decoder. The subset still supports mesh encoding, normals, and standard Edgebreaker — it constrains the output rather than disabling encoding — and avoids features outside the glTF profile (e.g. predictive valence at slower speeds) that the default decoder may reject. Attribute deduplication may be compiled out; `Encode` guards those passes on the feature macros Draco publishes.

## Risk

Purely additive — `DracoCodec.Decode` and `Version` are untouched and nothing existing changes shape, so no Babylon.js change is required and the encoder sits unused until a JS consumer probes for it. `BABYLON_NATIVE_PLUGIN_NATIVEDRACO=ON` is already set by every CI workflow (win32, uwp, linux, macos, ios, android), so this code is compiled and linked on all platforms by this PR's own run.

## Blocked on [JsRuntimeHost#223](https://github.com/BabylonJS/JsRuntimeHost/pull/223)

The sole red job, `Ubuntu_Clang_QuickJS`, is not a defect here: the new tests make a native module throw, which walks `Napi::Error::what()` into a use-after-free in QuickJS's `napi_escape_handle`. With only that dependency changed, this branch goes from `exit 139` to a clean 16/16 across five runs. Needs a JsRuntimeHost pin bump once #223 lands.


Plugins/NativeDraco/Source/NativeDraco.cpp
```diff
@@ -219,6 +221,223 @@ namespace Babylon::Plugins
...
+        int AddAttributeToMesh(draco::Mesh& mesh, draco::GeometryAttribute::Type type, draco::DataType dataType, long numVertices, int8_t numComponents, const T* values)
+        {
+            std::unique_ptr<draco::PointAttribute> att(new draco::PointAttribute());
+            att->Init(type, numComponents, dataType, /* normalized */ false, numVertices);
+            const int attId = mesh.AddAttribute(std::move(att));
+            draco::PointAttribute* attPtr = mesh.attribute(attId);
+            for (draco::PointIndex i(0); i < numVertices; ++i)
+            {
+                attPtr->SetAttributeValue(attPtr->mapped_index(i), &values[static_cast<size_t>(i.value()) * numComponents]);
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Checked this against the Draco source rather than assuming, and I believe this one is already safe.

`PointAttribute::Init` ends by establishing the identity mapping itself (`src/draco/attributes/point_attribute.cc`):

```cpp
void PointAttribute::Init(Type attribute_type, int8_t num_components,
                          DataType data_type, bool normalized,
                          size_t num_attribute_values) {
  attribute_buffer_ = std::unique_ptr<DataBuffer>(new DataBuffer());
  GeometryAttribute::Init(attribute_type, attribute_buffer_.get(),
                          num_components, data_type, normalized,
                          DataTypeLength(data_type) * num_components, 0);
  Reset(num_attribute_values);
  SetIdentityMapping();          // <-- here
}
```

and `mapped_index` short-circuits on exactly that flag (`point_attribute.h`):

```cpp
AttributeValueIndex mapped_index(PointIndex point_index) const {
  if (identity_mapping_) {
    return AttributeValueIndex(point_index.value());
  }
  return indices_map_[point_index];
}
```

So `mapped_index(i)` is already `AttributeValueIndex(i.value())` here — the values are written to distinct, correct slots, and this matches `PointCloudBuilder::AddAttribute<T>`, which also relies on the mapping `Init` sets up rather than configuring it separately.

I did take the underlying point though: this is an invariant we depend on but were not stating. Added an assert so a future Draco change cannot silently start folding distinct points onto the same attribute value:

```cpp
// PointAttribute::Init ends with SetIdentityMapping(), so mapped_index(i) is
// AttributeValueIndex(i) here. Asserted so a future Draco change cannot silently
// start folding distinct points onto the same attribute value.
assert(attPtr->is_mapping_identity());
```

Plugins/NativeDraco/Source/NativeDraco.cpp
```diff
@@ -219,6 +221,223 @@ namespace Babylon::Plugins
...
+            if (data.TypedArrayType() == napi_uint32_array)
+            {
+                const uint32_t* src = TypedArrayData<uint32_t>(data);
+                for (size_t i = 0; i < count; ++i) { out[i] = static_cast<int>(src[i]); }
+            }
+            else
+            {
+                const uint16_t* src = TypedArrayData<uint16_t>(data);
+                for (size_t i = 0; i < count; ++i) { out[i] = static_cast<int>(src[i]); }
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, fixed in c054aef9.

`ReadIndices` now switches explicitly and rejects anything else, instead of treating every non-`Uint32Array` as a `Uint16Array`:

```cpp
switch (data.TypedArrayType())
{
    case napi_uint32_array: /* ... */ break;
    case napi_uint16_array: /* ... */ break;
    default:
        throw Napi::TypeError::New(env, "Draco: Indices must be a Uint16Array or a Uint32Array");
}
```

You were right that `EncodeDracoMesh` only gated on `IsTypedArray()`, so an `Int32Array`/`Float32Array` would previously have been reinterpreted byte-wise and produced a silently corrupt stream — the worst kind of failure here, since it surfaces far away from the cause.

Plugins/NativeDraco/Source/NativeDraco.cpp
```diff
@@ -219,6 +221,223 @@ namespace Babylon::Plugins
...
+            const long numFaces = static_cast<long>(indices.size()) / 3;
+            mesh.SetNumFaces(numFaces);
+            for (draco::FaceIndex f(0); f < numFaces; ++f)
+            {
+                draco::Mesh::Face face;
+                face[0] = draco::PointIndex(indices[f.value() * 3 + 0]);
+                face[1] = draco::PointIndex(indices[f.value() * 3 + 1]);
+                face[2] = draco::PointIndex(indices[f.value() * 3 + 2]);
+                mesh.SetFace(f, face);
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed, fixed in c054aef9. This path is explicitly building triangle faces, so a trailing partial triangle is a caller error rather than something to drop silently:

```cpp
// This path builds triangle faces, so a trailing partial triangle is a caller error
// rather than something to silently drop.
if (indices.size() % 3 != 0)
{
    throw Napi::TypeError::New(env, "Draco: Index count " + std::to_string(indices.size()) +
                                        " is not a multiple of 3");
}
```

The count is included in the message so the caller can see what was actually passed.

Plugins/NativeDraco/Source/NativeDraco.cpp
```diff
@@ -219,6 +221,223 @@ namespace Babylon::Plugins
...
+            std::vector<int> out(count);
+            if (data.TypedArrayType() == napi_uint32_array)
+            {
+                const uint32_t* src = TypedArrayData<uint32_t>(data);
+                for (size_t i = 0; i < count; ++i) { out[i] = static_cast<int>(src[i]); }
+            }
+            else
+            {
+                const uint16_t* src = TypedArrayData<uint16_t>(data);
+                for (size_t i = 0; i < count; ++i) { out[i] = static_cast<int>(src[i]); }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed, fixed in c054aef9. Indices are now `uint32_t` end to end — `ReadIndices` returns `std::vector<uint32_t>`, the synthesized identity list is `uint32_t`, and the only narrowing is at the `draco::PointIndex` construction itself, whose `ValueType` is `uint32_t`:

```cpp
std::vector<uint32_t> ReadIndices(Napi::Env env, const Napi::TypedArray& data);
...
face[0] = draco::PointIndex(indices[f.value() * 3 + 0]);
```

So there is no longer an `int` in the path that a value above `INT_MAX` could wrap negative through.

This also removed the `static_cast<int>` on the way in, which was the actual sign-flip site.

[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed, fixed in c054aef9. Indices are now `uint32_t` end to end — `ReadIndices` returns `std::vector<uint32_t>`, the synthesized identity list is `uint32_t`, and the only narrowing is at the `draco::PointIndex` construction itself, whose `ValueType` is `uint32_t`:

```cpp
std::vector<uint32_t> ReadIndices(Napi::Env env, const Napi::TypedArray& data);
...
face[0] = draco::PointIndex(indices[f.value() * 3 + 0]);
```

So there is no longer an `int` in the path that a value above `INT_MAX` could wrap negative through.

This also removed the `static_cast<int>` on the way in, which was the actual sign-flip site.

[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed, fixed in c054aef9. Indices are now `uint32_t` end to end — `ReadIndices` returns `std::vector<uint32_t>`, the synthesized identity list is `uint32_t`, and the only narrowing is at the `draco::PointIndex` construction itself, whose `ValueType` is `uint32_t`:

```cpp
std::vector<uint32_t> ReadIndices(Napi::Env env, const Napi::TypedArray& data);
...
face[0] = draco::PointIndex(indices[f.value() * 3 + 0]);
```

So there is no longer an `int` in the path that a value above `INT_MAX` could wrap negative through.

This also removed the `static_cast<int>` on the way in, which was the actual sign-flip site.

Plugins/NativeDraco/Source/NativeDraco.cpp
```diff
@@ -219,6 +221,223 @@ namespace Babylon::Plugins
...
+                return -1;
+            }
+            return attId;
+        }
+
+        // Dispatches AddAttributeToMesh on the typed array's element type, mirroring the WASM
+        // encoder's addAttributeMap.
+        int AddTypedAttributeToMesh(Napi::Env env, draco::Mesh& mesh, draco::GeometryAttribute::Type type, const Napi::TypedArray& data, int8_t numComponents)
+        {
+            const long numVertices = static_cast<long>(data.ElementLength()) / numComponents;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed, fixed in c054aef9:

```cpp
if (numComponents <= 0)
{
    throw Napi::TypeError::New(env, "Draco: Attribute component count must be greater than zero");
}
if (data.ElementLength() % static_cast<size_t>(numComponents) != 0)
{
    throw Napi::TypeError::New(env, "Draco: Attribute length " + std::to_string(data.ElementLength()) +
                                        " is not a multiple of its component count " + std::to_string(numComponents));
}
```

The message carries both the element length and the declared component count, which are the two values that disagree.

I also added the `numComponents <= 0` guard while I was here: `size` comes straight from the JS object, so a zero would have divided by zero on the very next line rather than producing a wrong vertex count.

bkaradzic-microsoft:
The macOS failures were the `JavaScript.All` unit test, not a compile error. The `NativeDraco` suite contained an assertion that the plugin does **not** expose an encoder:

```js
it("does not expose an encoder", function () {
    expect(_native.DracoCodec.Encode).to.equal(undefined);
});
```

That was an accurate description of the old behavior, so this PR necessarily invalidates it. I replaced it with actual encoder coverage rather than just deleting it:

- round trips an indexed mesh through `Encode` and back through `Decode`
- encodes an unindexed mesh
- accepts 32 bit indices
- rejects a non 16/32 bit index buffer, an index count that is not a multiple of 3, an attribute length that is not a multiple of its component count, and a mesh with no position attribute

The last four exercise the validation added in response to the review comments above.

`dist/tests.javaScript.all.js` is committed and consumed directly by `Apps/UnitTests/CMakeLists.txt`, so it has to be regenerated. I rebuilt it with the lockfile-pinned toolchain (`npm ci`, webpack 5.105.2) so the bundle diff is exactly the 74 lines of the source change with no incidental churn.

bkaradzic-microsoft:
CI caught two real problems in the new tests — thanks, both now fixed in b4bdc65f.

**1. `Encode` returned an `Int8Array`.** The result buffer was built with `Napi::Int8Array`, so the encoded bytes came back signed:

```
expected Int8Array[ 68, 82, 65, 67, 79, 2, ... ] to be an instance of Uint8Array
```

Draco output is a byte stream, so `Uint8Array` is the correct type, and it matches what `Decode` accepts as input. This was a genuine API bug that only surfaced because the new round-trip test asserts the returned type.

**2. The unindexed encode test was wrong, not the code.** Without an index buffer the vertices are taken as a flat triangle list, so the vertex count itself has to be a multiple of three. The shared fixture is a four-vertex quad, which correctly tripped the new `%3` validation:

```
Draco: Index count 4 is not a multiple of 3
```

The test now uses a single triangle. The validation behaved exactly as intended here.

Verified locally on Win32 (`JavaScript.All`): **49 passing, 0 failing, exit 0**.

bkaradzic-microsoft:
The `Ubuntu_Clang_QuickJS` segfault turned out to be a real bug in the encode path, not a test problem. Fixed in 8f7252a8.

**Finding it.** CI stdout is block buffered, so the trailing output was lost with the crash. Diffing the printed test list against the passing `Ubuntu_Clang_JSC` job narrowed it to `rejects an index count that is not a multiple of 3` — which is the first test that reads typed array data and *then* throws. The rejection test just before it throws before touching the data pointer, which is why it passed.

**Two fixes:**

1. **Indices were never range checked.** Nothing validated index values against the vertex count, so an out of range index was written into a face and later dereferenced by `DeduplicateAttributeValues` / `DeduplicatePointIds` and the encoder, reading past the end of the attribute buffers. This is reachable from script in one line:

   ```js
   _native.DracoCodec.Encode([{ kind: "position", dracoName: "POSITION", size: 3, data: positions }],
                             new Uint16Array([0, 1, 9999]));
   ```

   It now throws, and there is a test for it. Note this hole only became reachable in this PR, since `DRACO_GLTF_BITSTREAM=OFF` enables the deduplication passes. I also rejected a non-positive position component count, which otherwise divided by zero when computing the vertex count.

2. **Typed array reads now go through `TypedArrayOf<T>::Data()`** instead of `ArrayBuffer().Data()` plus the byte offset. `napi_get_typedarray_info` already returns a pointer to the first element, so the manual offset arithmetic was redundant, and this avoids materializing a temporary `ArrayBuffer` handle just to read a pointer.

Verified locally on Win32 against both engines: default **50 passing**, QuickJS **40 passing**, exit 0 in both cases. I built a local QuickJS configuration specifically to reproduce this.

bkaradzic-microsoft:
Update on the `Ubuntu_Clang_QuickJS` failure: it is not caused by this PR. It is a latent heap-use-after-free in JsRuntimeHost's QuickJS Node-API shim that the new tests happen to expose.

`napi_escape_handle` stores the escaped handle at the scope start index so it outlives the scope, but `napi_close_escapable_handle_scope` then resizes the handle stack back to that same index and frees it. `Napi::ObjectReference::Get` uses an `EscapableHandleScope`, and `Napi::Error::Message()` / `what()` are built on it, so reading the message of a native error on QuickJS reads freed memory.

The tests added here make a native module throw, and `ExternalCallback::Callback` calls `e.what()` when there is no pending QuickJS exception, which walks into the freed handle. ASan on Ubuntu:

```
==ERROR: AddressSanitizer: heap-use-after-free
    #0 ToJSValue                     js_native_api_quickjs.cc:302
    #3 Napi::Error::Message
    #4 Napi::Error::what
    #5 ExternalCallback::Callback    js_native_api_quickjs.cc:164
freed by:
    #1 napi_close_escapable_handle_scope  js_native_api_quickjs.cc:1939
    #2 Napi::ObjectReference::Get
```

Fix is up as BabylonJS/JsRuntimeHost#223. With that branch patched in, this PR's `JavaScript.All` is 50 passing / exit 0 / 0 ASan errors in the exact clang + QuickJS config that is red here.

So this PR is blocked on JsRuntimeHost#223 landing plus a submodule bump. Every other job is green. Happy to either wait for the bump, or land this with the known-external red job if you would rather not serialize them.

bkaradzic-microsoft:
CI status update: re-ran the failed jobs, and the `Win32_x64_D3D11` failure was an unrelated infrastructure flake (403 Forbidden from `snippet.babylonjs.com`) which now passes. That leaves **31/32 green**, with `Ubuntu_Clang_QuickJS` the only red job.

That one job is the external JsRuntimeHost use-after-free described above, not anything in this PR. The fix, BabylonJS/JsRuntimeHost#223, is now **24/24 green** and ready for review. Once it lands and the JsRuntimeHost dependency here is bumped, this PR should be fully green — verified locally by building this branch against the fix with clang + QuickJS on Ubuntu, which gives 50 passing, exit 0, and 0 ASan errors.

bkaradzic-microsoft:
Addressed the four comments in the collapsed **Suppressed comments** block of the Copilot review in `8c53b98b`. Those don't create review threads, so they weren't in the resolved/unresolved list and I'd missed them until now. All four were correct.

**`Encode` should return `Int8Array`, not `Uint8Array`.** Agreed, and I had this backwards. I originally moved it to `Uint8Array` because `Int8Array` surfaces the high bytes as negative numbers, which reads wrong in a debugger — but that's cosmetic, the bytes are identical, and it's the wrong thing to optimize for. `IDracoEncodedMeshData` in `packages/dev/core/src/Meshes/Compression/dracoEncoder.types.ts` declares `data: Int8Array`, because the WASM encoder returns a view onto emscripten's signed `HEAP8`. Since this entry point exists to be a drop-in for that path, the view type has to match what callers `instanceof`-check and declare against. Now `Napi::Int8Array`, with the reason in a comment so it doesn't get "fixed" again. The round-trip test still passes: `Decode` takes any `Napi::TypedArray` and reads raw bytes off the backing buffer, so it doesn't care about the view type.

**README was materially wrong.** Correct — it still described a decode-only, glTF-bitstream-only plugin, omitted `Encode` from the interface declaration, and stated `DRACO_GLTF_BITSTREAM=ON` when this PR sets it `OFF`. Rewritten: the encoder is in the summary and the TypeScript declaration, and the bitstream bullet now gives the real reason for `OFF` (the subset drops pre-glTF backwards compatibility, so it rejects streams from older or full encoders with "Unsupported major version", and it compiles out the deduplication passes the encoder needs) rather than the inverted one.

**Stale `DRACO_GLTF_BITSTREAM` comment at the deduplication guards.** Correct. The `#ifdef`s are there for externally supplied draco targets that may enable the subset, not for this build's configuration, so the comment now describes the conditional case.

**No test with a non-zero `byteOffset`.** This was the valuable one. `TypedArrayData<T>()` exists precisely to read through the typed view rather than off the start of the backing `ArrayBuffer`, and nothing exercised it. Added a test that encodes position *and* index subviews sitting partway into larger buffers, with the padding filled with deliberately wrong values, then decodes and compares corners — so a regression that ignored `byteOffset` would encode the padding and fail the comparison instead of passing quietly. Both use sites are covered (`NativeDraco.cpp:294` for attributes, `:325` for 16-bit indices).

41/41 unit tests pass locally.

On the remaining CI failure: `Ubuntu_Clang_QuickJS` still segfaults at teardown *after* all Draco tests report passing, which matches the pre-existing use-after-free in the QuickJS Node-API shim I described earlier in this PR — it isn't specific to this change.

bkaradzic-microsoft:
The `Ubuntu_Clang_QuickJS` failure is not coming from this PR's code. I reproduced it locally and tracked it down: it's a use-after-free in the QuickJS Node-API shim that [JsRuntimeHost#223](https://github.com/BabylonJS/JsRuntimeHost/pull/223) fixes. This PR is just the first thing to exercise the affected path.

**What happens.** Every test passes, then the process segfaults on the way out:

```
✅ encodes typed array views with a non-zero byteOffset
Segmentation fault (core dumped)
```

The backtrace points somewhere quite specific:

```
#0  js_force_tostring                       quickjs.c:4813
#3  napi_get_value_string_utf8              js_native_api_quickjs.cc:696
#4  Napi::String::Utf8Value
#6  Napi::Error::Message                    napi-inl.h:3087
#7  Napi::Error::what
#8  ExternalCallback::Callback              js_native_api_quickjs.cc:164
```

Frame 8 is the shim's catch-all, which calls `e.what()` when a C++ exception escapes without a pending JS exception. `Napi::Error::what()` reads the error's `message`, and `ObjectReference::Get` does that through an escapable handle scope:

```cpp
inline MaybeOrValue<Napi::Value> ObjectReference::Get(const char* utf8name) const {
  EscapableHandleScope scope(_env);
  ...
  return scope.Escape(result);
}
```

The escaped handle is freed when that scope closes, so reading it afterwards dereferences freed memory. At the crash the JSValue has `tag = -7` (string) but `ptr = 0x7ff8dec9a216`, which isn't even pointer-aligned — reused memory.

This PR triggers it because the encoder tests (`rejects malformed input`, `rejects truncated input`, `rejects an empty buffer`) are what throw `Napi::Error` from a native callback. Nothing on master does, which is why master is green.

**Verified by A/B.** Same tree, same flags as CI (clang, QuickJS, RelWithDebInfo, no sanitizers), only the JsRuntimeHost dependency changed:

| JsRuntimeHost | Result |
| --- | --- |
| current master | exit 139, 1, 139 — segfault |
| with #223 | exit 0 × 5 — clean, 16/16 |

So this job should go green once #223 lands; nothing to change here. Happy to rebase once it does.

Dependencies/CMakeLists.txt
```diff
@@ -190,20 +190,20 @@ function(group_directory_targets DIR FOLDER_NAME)
...
-    # Build only the features in the Draco glTF bitstream spec: mesh compression, normal
-    # encoding and the standard edgebreaker. This is the same subset Babylon.js decodes
-    # with, since its default configuration loads draco_decoder_gltf.wasm rather than the
-    # full decoder, so it costs no capability the JavaScript path has. It drops point cloud
-    # compression, the predictive edgebreaker and pre-glTF backwards compatibility.
-    set(DRACO_GLTF_BITSTREAM ON CACHE BOOL "" FORCE)
+    # The glTF bitstream subset (mesh compression, normal encoding, standard edgebreaker)
+    # is not sufficient here. It rejects streams produced by older/full encoders with
+    # "Unsupported major version" because it drops pre-glTF backwards compatibility, and it
+    # compiles out the attribute deduplication passes the encoder below relies on. Build the
```
bghgary: `NativeDraco.cpp:462-463` says skipping the deduplication passes "still produces a valid stream". This line and `README.md:11` both say the encoder relies on them. Which is authoritative?

If it is the C++ comment, pre-glTF backwards compatibility is the only remaining reason for `DRACO_GLTF_BITSTREAM=OFF`, and both docs should say that instead.

bkaradzic-microsoft:
CI is green now (34/34). Rebased on master and added one commit that is not Draco work, so flagging it explicitly:

`Ubuntu_Clang_QuickJS` was segfaulting the moment it entered the `NativeDraco` tests. The bug was not in this PR -- QuickJS is refcounted, and `napi_escape_handle` was dropping the escaped value's last reference when its escapable scope closed, leaving the caller holding a freed `JSValue`. This PR just happens to be the thing that trips it, because the encoder returns a typed array out of an escapable scope.

Fixed upstream in BabylonJS/JsRuntimeHost#223, now merged, so this bumps the `GIT_TAG` pin from `9271f13b` to `dbd4620f`. That is the only commit between the two, so the bump is a fast-forward carrying nothing else.

Verified locally on Linux/Clang/QuickJS rather than relying on CI alone. With the old pin:

```
[log]   NativeDraco
Segmentation fault (core dumped)
```

and with the new one the same build runs 16/16 suites, 51 assertions, no crash.

Happy to split the pin bump into its own PR if you would rather keep this one purely Draco, but on its own it would have nothing to demonstrate.

Dependencies/CMakeLists.txt
bghgary: `NativeDraco.cpp:462-463` says skipping the deduplication passes "still produces a valid stream". This line, `README.md:11` and the PR description all say the encoder relies on them. Which is authoritative?

If it is the C++ comment, pre-glTF backwards compatibility is the only remaining reason for `DRACO_GLTF_BITSTREAM=OFF`, and all three should say that instead.

CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ FetchContent_Declare(ios-cmake
     EXCLUDE_FROM_ALL)
 FetchContent_Declare(JsRuntimeHost
     GIT_REPOSITORY https://github.com/BabylonJS/JsRuntimeHost.git
-    GIT_TAG 9271f13b654665e343aa0e80ad872b4674534163)
+    GIT_TAG dbd4620fbc4019208bb0867f0b3b518b29043555)
```
bghgary: The description still says this is blocked on JsRuntimeHost#223 and needs a pin bump "once #223 lands". This line is that bump.

It also still calls the risk "purely additive". The bumped commit changes `napi_escape_handle` and `napi_close_escapable_handle_scope` on JavaScriptCore and Chakra as well as QuickJS, returning `napi_invalid_arg` and `napi_escape_called_twice` where they previously always returned `napi_ok`.

Dependencies/CMakeLists.txt
bghgary: `NativeDraco.cpp:462-463` says skipping the deduplication passes "still produces a valid stream". This line, `README.md:11` and the PR description all say the encoder relies on them. Which is authoritative?

If it is the C++ comment, pre-glTF backwards compatibility is the only remaining reason for `DRACO_GLTF_BITSTREAM=OFF`, and all three should say that instead.

CMakeLists.txt
bghgary: This pin is now older than master's. #1848 moved master to `cb988ba`, which contains `dbd4620f` plus JsRuntimeHost #225 and #227, so resolving the conflict in favour of this line reverts both. Neither shows up as a failure.

The description also still says this is blocked on JsRuntimeHost#223 and needs a pin bump "once #223 lands".

Dependencies/CMakeLists.txt
```diff
@@ -177,20 +177,20 @@ function(group_directory_targets DIR FOLDER_NAME)
...
```
bghgary: `NativeDraco.cpp:462-463` says skipping the deduplication passes "still produces a valid stream". This line, `README.md:11` and the PR description all say the encoder relies on them. Which is authoritative?

If it is the C++ comment, pre-glTF backwards compatibility is the only remaining reason for `DRACO_GLTF_BITSTREAM=OFF`, and all three should say that instead.

Plugins/NativeDraco/README.md
```diff
@@ -37,6 +36,24 @@ interface INative {
       }>;
       totalVertices: number;
     };
+    Encode: (
+      attributes: Array<{
+        kind: string;
+        dracoName: string;
+        size: number;
+        data: ArrayBufferView;
```
bghgary: `ArrayBufferView` also admits `DataView`, `Float64Array`, and BigInt typed arrays, but the implementation accepts none of them. Please document the supported typed-array union here instead.

Dependencies/CMakeLists.txt
```diff
@@ -177,20 +177,20 @@ function(group_directory_targets DIR FOLDER_NAME)
...
+    # full decoder/encoder instead.
+    set(DRACO_GLTF_BITSTREAM OFF CACHE BOOL "" FORCE)
```
bghgary: Please keep `DRACO_GLTF_BITSTREAM=ON`. NativeDraco is intended to produce glTF-compatible Draco data, and Babylon.js defaults to the glTF-only Draco decoder. With this option disabled, the encoder can select features outside the glTF subset—such as predictive valence encoding at slower speed settings—producing data that the default decoder may reject. The glTF setting still supports mesh encoding, normals, and standard Edgebreaker; it constrains the output rather than disabling encoding. Please restore the README and PR description to describe this glTF-only scope.

Plugins/NativeDraco/README.md
bghgary: Please use Babylon.js's `VertexDataTypedArray` here. `ArrayBufferView` also includes `DataView`, `Float64Array`, and BigInt typed arrays, which this implementation rejects.

bghgary (changes requested):

Two inline.

Dependencies/CMakeLists.txt
bkaradzic-microsoft: Restored `DRACO_GLTF_BITSTREAM=ON` and rewrote the CMake/README/PR description around glTF-only scope: NativeDraco is meant to produce/consume glTF-compatible Draco (Babylon.js default decoder). Encode still guards dedup behind Draco's feature macros when the subset compiles those out.

Pushed in the latest commit.

Plugins/NativeDraco/README.md
bkaradzic-microsoft: Updated the README TypeScript surface to use Babylon.js `VertexDataTypedArray` (`Exclude<TypedArray, Float64Array | BigInt64Array | BigUint64Array>`) for both Decode input and Encode attribute `data`, with a short note that bare `ArrayBufferView` is too wide.

Pushed in the latest commit.

bghgary (approved):

PR #1836, opened 2026-08-13
Fix D3D11/fxc hang on cascaded shadow maps by flattening narrow varying arrays

bkaradzic-microsoft:
Fixes a D3D11/fxc shader-compilation hang on Babylon.js cascaded shadow maps.

## The problem

For an array-typed varying, SPIRV-Cross emits an array-typed member in the HLSL interface struct (`float vDepthMetric0[4] : TEXCOORD5;`). fxc turns that into an indexable input register range and requires every register in the range to use the **same component mask** (not necessarily all four slots — matching `.x` or matching `.xyz` is fine):

```
error X8000: D3D11 Internal Compiler Error: Invalid Bytecode: Masks (and if pixel
shader, also interpolation mode) on all input registers in an index range must be
identical.
```

Observed hang/reject shapes are **float** and **vec2** arrays. `flat int[4]` and `vec3[4]` compile without this pass. fxc does not simply fail on the bad shapes — it then **hangs**, so D3D11 shader compilation never returns.

The trigger is `lightFragmentDeclaration.fx`, which declares `varying float vDepthMetric{X}[SHADOWCSMNUM_CASCADES{X}]`. Its companion `varying vec4 vPositionFromLight{X}[...]` is unaffected.

## The fix

A new glslang traverser rewrites each narrow array-typed inter-stage varying, before SPIR-V generation, into one varying per element plus a plain (non-varying) global array that all existing references are repointed at. Element-wise copies are:

- **prepended** to fragment `main()`
- inserted **immediately before** a trailing top-level `return` in vertex `main()` (or appended when `main` falls off the end)

The interface then contains no array, so fxc emits no indexable range.

Keeping the global array is what preserves **dynamic indexing**, which a naive rewrite to per-element varyings cannot do — [`lightFragment.fx`](https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Shaders/ShadersInclude/lightFragment.fx#L283) selects the cascade with an index computed per fragment. Routing through a global also keeps writes performed by helper functions working, since they observe the global rather than a copy.

**Locations.** `TType{varyingType, 0}` would copy `layout(location=N)` onto every element and pin duplicate `TEXCOORDN`s. Generated element varyings clear `layoutLocation` so SPIRV-Cross assigns vacant TEXCOORDs the same way it does for undecorated inter-stage varyings (Babylon Native sets explicit locations only on vertex inputs and never calls `mapIO`).

**Scope.** DXBC only; dxc accepts the indexable range, so DXIL is untouched. Vertex inputs (owned by `AssignLocationsAndNamesToVertexVaryings*`) and fragment outputs are skipped. Only narrow (`getVectorSize() < 4` — conservative workaround covering the known float/vec2 failures), sized, 1-D, non-struct, non-matrix, non-builtin arrays qualify. Nested or mid-body returns in vertex `main()` reject the transform with a clear error rather than emitting stale varyings; a trailing top-level `return;` is handled by inserting copies before it.

## Testing

The comprehensive GLSL cross-compilation test grows a scalar varying array read back through a runtime-derived index. On Win32 D3D11, **with** the pass 21/21 unit tests pass and `CompileComprehensiveGLSL` completes in ~0.6s; **without** it, the `X8000` above reproduces and then hangs (>180s). The full enabled Playground sweep is unchanged at 301/302, the single `Outline` failure being an unrelated pre-existing `Unsupported alpha mode: -1`.

Supersedes BabylonJS/SPIRV-Cross#14, which fixed the same issue inside SPIRV-Cross; doing it here avoids carrying a patch on a third-party library.


Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -1697,6 +1697,305 @@ namespace Babylon::ShaderCompilerTraversers
...
+                auto& sequence = mainBody->getSequence();
+                for (size_t i = 0; i < sequence.size(); ++i)
+                {
+                    if (sequence[i] == nullptr)
+                    {
+                        continue;
+                    }
+
+                    auto* branch = sequence[i]->getAsBranchNode();
+                    const bool isTrailingReturn = branch != nullptr && branch->getFlowOp() == EOpReturn && i + 1 == sequence.size();
```
bghgary: A final `return;` is equivalent to falling off the end only before this pass changes `main()`. Line 1915 then appends the varying-copy statements after that return, where they cannot execute. Please either treat every return as early, or insert `copyStatements` immediately before a final return.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -1697,6 +1697,305 @@ namespace Babylon::ShaderCompilerTraversers
...
+                // A struct or matrix element has no single write mask to reason about, and
+                // glslang would need a different construction path for each; neither appears
+                // as an array-typed varying in Babylon.js shaders.
+                if (type.isStruct() || type.isMatrix())
+                {
+                    return false;
+                }
+
+                // Four-component elements already give every register an identical full mask,
+                // so they are legal as an indexable range and are left alone.
```
bghgary: Please correct the explanation of fxc's rule. Each input register has four component slots (`x`, `y`, `z`, `w`), and a dynamically indexed range requires every register to use the same slots; it does not require all four slots. `.x` on every register or `.xyz` on every register is a matching mask. I confirmed this backend compiles `flat int v[4]` and `vec3 v[4]` without the pass, even though both are narrower than `vec4`. Please describe the observed failing shapes (`float` and `vec2`) or call `< 4` a conservative workaround, rather than saying every narrow array necessarily violates the mask rule.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -1758,6 +1758,305 @@ namespace Babylon::ShaderCompilerTraversers
...
+                globalType.getQualifier().clearLayout();
+                globalType.getQualifier().clearInterpolation();
+                globalType.getQualifier().storage = EvqGlobal;
+
+                TIntermSymbol globalPrototype{ids.Next(), symbol->getName(), globalType};
+                originalNameToReplacement[name] = intermediate->addSymbol(globalPrototype);
+
+                // Element type for the flattened varyings. The dereference constructor keeps the
+                // original storage and interpolation qualifiers, which is what these need.
+                const TType elementType{varyingType, 0};
```
bghgary: Please either clear and assign consecutive locations to the generated element varyings, or exclude explicitly located arrays. `TType{varyingType, 0}` preserves the original `layout(location=N)`, so every `_i` element receives location N. ES 310 permits explicit locations on vertex outputs and fragment inputs, and this compiler accepts arbitrary shader source; pinned SPIRV-Cross maps each preserved location to the same `TEXCOORDN`, so D3DCompile rejects the duplicate semantics.

bghgary (approved):

Concerns inline.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: Good catch. A trailing top-level `return;` is not early for control-flow, but copies appended after it never execute. Vertex element copies are now inserted **immediately before** a trailing top-level return when one is present (still appended when `main` falls off the end). Nested/mid-body returns still fail the transform.

Pushed in 10378fdf.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: Updated the header/impl comments: fxc requires matching component masks across the indexable range, not a full `.xyzw` mask. Documented the observed hang/reject shapes as **float** and **vec2**, noted that `flat int[4]` / `vec3[4]` compile without the pass, and kept `getVectorSize() < 4` as an explicit **conservative** workaround rather than claiming every narrow array violates the rule.

Pushed in 10378fdf.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: `TType{varyingType, 0}` was preserving `layout(location=N)` on every element. Each generated element varying now clears `layoutLocation` (`TQualifier::layoutLocationEnd`) so SPIRV-Cross assigns vacant TEXCOORDs the same way it does for undecorated inter-stage varyings, instead of pinning duplicates on the same `TEXCOORDN`.

Pushed in 10378fdf.

PR #1838, opened 2026-08-17
Playground validation: finish the sweep instead of stopping at the first failure

bkaradzic-microsoft:
## Problem

The validation runner exits on the first failing test. On the nightly the first
failure is **index 322 of 720**, so **398 tests have not run since 2026-07-30** —
the sweep has been reporting `ran=175 skipped=147` instead of the full 720 for
over two weeks.

That is not just lost coverage. The tests immediately after the stopping point
were failing too, and nobody could see it:

```
--test-index=320-325   ->  Thin instances + dynamic buffer resize    FAILED
                           Instances + render self motion blur       FAILED
                           Thin instances + render self motion blur  FAILED
```

Only the first of those three was visible in the nightly log.

## Change

**Continue-on-failure is now the default.** The run completes every selected
test, prints the full tally and the list of failed titles, and still exits
non-zero if anything failed. `--keep-going`/`-k` is replaced by its inverse,
the opt-in `--stop-on-first-failure`, which is the behaviour you actually want
while bisecting a long sweep rather than the default for CI.

**Softens the pixel-comparison failure note.** It previously stated that a diff
on a snippet-server test was *"often a transient async asset/font-load timing
flake"*. That is sometimes true, but phrased that confidently it reads as a
verdict — a fully reproducible regression was waved off on the strength of it
for two weeks of nightlies. It now names network timing as *one possible* cause
and gives the cheap way to tell the two apart: a flake varies run to run, a real
regression repeats with an identical pixel count. The re-run command is now
printed for every failing test, not just snippet-loaded ones.

## Verification

Win32 x64 D3D11, RelWithDebInfo, Babylon.js 9.21.2:

| command | result | exit |
|---|---|---|
| `--test-index=320-325` | `ran=3 passed=0 failed=3` | -1 |
| `--test-index=320-325 --stop-on-first-failure` | `ran=1 passed=0 failed=1` | -1 |
| `--test-index=0-5` | `ran=6 passed=6 failed=0` | 0 |

`--help` output and `.github/instructions/babylon-native-debugging.instructions.md`
updated to match.

## Note for reviewers

**Expect the nightly to get louder before it gets quieter.** Once this lands the
sweep will run all 720 tests and report every failure, so the failure list will
grow — those failures already exist, they were just unreachable. The three above
are motion-blur regressions exposed by #1754 enabling MultiRenderTarget support;
they are fixed in a separate PR.

### DIFF
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -51,10 +51,10 @@
     // snippet server it also notes that assets/fonts arrive over the network, so
     // async load timing is one possible cause -- but it is only one of several,
     // and it is cheap to rule out: a timing flake varies run to run, while a real
-    // regression reproduces with a byte-identical diff. Say that explicitly. The
-    // previous wording called such diffs "often a transient flake", which led to
-    // a genuine, perfectly reproducible motion-blur regression being waved off
-    // for two weeks of nightlies.
+    // regression reproduces with an identical pixel-difference count. Say that
+    // explicitly. The previous wording called such diffs "often a transient flake",
+    // which led to a genuine, perfectly reproducible motion-blur regression being
+    // waved off for two weeks of nightlies.
     function logFailureDiagnostics(test) {
         const outDir = TestUtils.getOutputDirectory();
         if (test.referenceImage) {
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -8,7 +8,7 @@
     const testHeight = 400;
     const generateReferences = !!opts.generateReferences;
     const breakOnFail = !!opts.breakOnFail;
-    const keepGoing = !!opts.keepGoing;
+    const stopOnFirstFailure = !!opts.stopOnFirstFailure;
     const listTests = !!opts.listTests;
     const includeExcluded = !!opts.includeExcluded;
     const testFilters = Array.isArray(opts.testFilters) ? opts.testFilters.map(s => String(s).toLowerCase()) : [];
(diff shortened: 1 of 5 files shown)


ryantrem (approved):

Apps/Playground/Scripts/validation_native.js
```diff
@@ -47,24 +47,25 @@
...
-    // and web fonts over the network/CDN, so their output depends on async
-    // asset/font-load timing. A pixel diff there is frequently a transient flake
-    // rather than a real regression -- the 'Parse GUI json with unicode' snippet
-    // test is the canonical example (its GUI text renders with the fallback font
-    // until 'droidsans' finishes loading, shifting thousands of glyph pixels).
+    // rendered/diff PNG paths plus a re-run command. For scenes fetched from the
+    // snippet server it also notes that assets/fonts arrive over the network, so
+    // async load timing is one possible cause -- but it is only one of several,
+    // and it is cheap to rule out: a timing flake varies run to run, while a real
+    // regression reproduces with a byte-identical diff. Say that explicitly. The
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch - fixed in 12a52e3e. The comment now says a real regression reproduces with an identical pixel-difference count, matching what the code logs and what the user-facing message says.

Apps/Playground/Scripts/validation_native.js
```diff
@@ -724,7 +725,7 @@
                         failedTitles.push(currentTitle);
                         // failTest() already triggered the debugger before
                         // reaching this callback; no second `debugger` here.
-                        if (!keepGoing) {
+                        if (stopOnFirstFailure) {
```
bghgary, line 728: Please run the inter-test cleanup before advancing after every failure, and guard completion so it runs once. In the render-loop catch (L367-370), `done(false)` runs without `stopRenderLoop()` or the cleanup in `evaluateScreenshot()`. Continue-on-failure is now the default, so the caller starts the next test while the old callback remains registered. Babylon's `runRenderLoop()` appends callbacks and `currentScene` is module-scoped: once the next scene is assigned, the stale callback renders it. If it reaches `stopFrame`, its `engine.stopRenderLoop()` clears the new test's callback before it runs; if rendering throws again, it calls `done(false)` repeatedly, corrupting counts and scheduling multiple test chains.

The `onReadyTimeout`, snippet-evaluation, and script-evaluation failure paths also bypass the same scene/engine cleanup. Move that cleanup to a once-only completion path, including `stopRenderLoop()`, before scheduling the next test.

Apps/Playground/Scripts/validation_native.js
```diff
@@ -47,24 +47,25 @@
...
             console.log(`  Rendered result: ${outDir}/Results/${test.referenceImage}`);
             console.log(`  Diff overlay:    ${outDir}/Errors/${test.referenceImage}`);
         }
         if (test.playgroundId) {
-            console.log(`  Note: this test loads playgroundId ${test.playgroundId} from the snippet server and pulls GUI/assets/fonts over the network, so a pixel diff is often a transient async asset/font-load timing flake.`);
-            console.log("  Re-run in isolation to confirm a real regression:");
-            console.log(`    Playground --headless --once --test "${test.title || ""}" app:///Scripts/validation_native.js`);
+            console.log(`  Note: this test loads playgroundId ${test.playgroundId} from the snippet server and pulls GUI/assets/fonts over the network, so async asset/font-load timing is one possible cause of a pixel diff.`);
         }
+        console.log("  Re-run in isolation; an identical pixel count on repeat runs means a real regression, not a timing flake:");
```
bghgary, line 67: Please don't say an identical count "means" a real regression. The count discards which pixels changed and by how much, and a repeatable timing failure can consistently produce the same fallback image. Treat a stable count as a reason to compare the saved result/diff images, not proof that timing has been ruled out.

bghgary (commented):

Two concerns inline.

PR #1839, opened 2026-08-17
Map previousWorld0-3 so instanced motion vectors are correct

bkaradzic-microsoft:
Fixes the nightly Playground validation failures on the three motion blur tests.

## Problem

A mesh rendered with object based motion blur (or prepass velocity) declares `previousWorld0-3` alongside `world0-3`, so an instanced or thin-instanced mesh has **8** per-instance `vec4` attributes, 9 with `instanceColor`.

`previousWorld0-3` had no built-in mapping in the shader compiler, and `VertexArray::RecordVertexBuffer` rejected anything past the 5th instanced buffer (`Number of vertex buffer instances greater than 4 is not supported`). Those buffers were never recorded, the shader read a zero previous world matrix, and produced a huge bogus velocity that smeared the whole object.

This is what the nightly has been failing on since MRT support (#1754) enabled object based motion blur on Native.

## Why slots are now assigned per shader

The `i_data` slots cannot come from a fixed per-name table. bgfx requires the used slots to form a **contiguous run starting at `i_data0`**:

- the D3D11 input layout declares `TEXCOORD31` down to `TEXCOORD(31 - N + 1)` at 16-byte-dense offsets;
- the GL path compacts the `i_data` locations it finds.

Since the declared set varies (`world0-3` alone, plus `instanceColor`, plus `previousWorld0-3`), a fixed table leaves a hole in the run and every attribute past the hole reads zero.

So slots are assigned per shader from the set that shader actually declares, in reverse name order: alphabetically first gets the highest slot, last gets `i_data0`. That is dense by construction, stable between the base program compile and any instanced variant (identical declared set), and matches `BuildInstanceDataBuffer`, which packs the recorded buffers by descending attribute location.

It also **reproduces the previous fixed D3D table exactly** for every set that existed before `previousWorld0-3` (`world0-3` → `i_data3..0`, `+instanceColor` → `i_data4`, `splatIndex0-3` → `i_data3..0`), so nothing else changes. OpenGL and Metal already assigned densely; they now share the same map instead of each counting attributes themselves.

The instanced buffer limit is raised to bgfx's real `BGFX_CONFIG_MAX_INSTANCE_DATA_COUNT` (16 — its doc comment still claims 5), and the off-by-one is fixed: the check ran *before* the insert, so it rejected the 5th buffer rather than the 6th.

## Validation

Playground validation, Win32 D3D11, Babylon.js 9.21.2:

| Test | before | after |
|---|---|---|
| Thin instances + dynamic buffer resize | 3.346% | **2.478%** |
| Thin instances + render self motion blur | 2.844% | **2.355%** |
| Instances + render self motion blur | 2.531% | **2.061%** |

All now under the 2.5% allowance, and the instance-limit errors go from 9 to 0. The numbers are bit-identical on hardware D3D11 and on forced WARP, so they are not adapter-flaky.

A full 720 test A/B sweep shows **no other test changing by more than 0.001%**, and an identical set of pre-existing hangs on both builds.

## Known remaining difference

A smaller residual diff remains on these three tests, from a **separate** bug: Babylon.js's `bindAttachments` is a no-op on Native, so the scene clear is applied to every MRT attachment and wipes the velocity attachment's alpha-0 background. The motion blur shader multiplies velocity by that alpha, so the background gets a fixed non-zero velocity and moving objects get a faint halo. Fixing it needs a matching Babylon.js change plus per-attachment clear masking here, so it is not addressed in this PR.

The worst test passes at 2.478% against a 2.5% allowance, so there is little headroom until that second bug is fixed.

### DIFF
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
@@ -27,7 +27,7 @@ namespace Babylon::Graphics
     /// the number of 16-byte per-instance slots (i_data0..i_data15) bgfx can bind in one draw.
     inline constexpr uint32_t MAX_INSTANCE_DATA_SLOT_COUNT{16};
 
-
+    /// The built-in per-instance attributes occupy the top BUILTIN_INSTANCE_DATA_SLOT_COUNT i_data
     /// slots. Which slot each one gets is decided per shader, from the set the shader actually
     /// declares (see ShaderCompilerTraversers.cpp), because bgfx requires the used i_data slots to
     /// be a contiguous run starting at i_data0. The count is the size of the largest possible set:
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
@@ -30,20 +30,19 @@ namespace Babylon::Graphics
     inline constexpr uint32_t MAX_INSTANCE_DATA_SLOT_COUNT{16};
 
     /// The built-in per-instance attributes occupy the top BUILTIN_INSTANCE_DATA_SLOT_COUNT i_data
-    /// slots. Which slot each one gets is decided per shader, from the set the shader actually
-    /// declares (see ShaderCompilerTraversers.cpp), because bgfx requires the used i_data slots to
-    /// be a contiguous run starting at i_data0. The count is the size of the largest possible set:
-    /// world0-3 (or splatIndex0-3), instanceColor, and previousWorld0-3 for motion vectors.
-    /// BUILTIN_INSTANCE_DATA_LAST_LOCATION is the lowest synthetic location any of them can occupy;
-    /// it is the boundary NativeEngine::Draw's "< bgfx::Attrib::Count means a real per-vertex
-    /// attribute that needs rerouting" guard rests on, so it -- not just INSTANCE_DATA_FIRST_LOCATION
-    /// -- must stay >= bgfx::Attrib::Count. Keep in sync when adding a built-in per-instance attribute.
+    /// slots. Which slot each gets is decided per shader from the set it declares (see
+    /// ShaderCompilerTraversers.cpp), because bgfx requires the used i_data slots to be a
+    /// contiguous run starting at i_data0. The count is the largest possible set: world0-3 (or
+    /// splatIndex0-3), instanceColor, and previousWorld0-3 for motion vectors.
+    /// BUILTIN_INSTANCE_DATA_LAST_LOCATION is the lowest synthetic location any can occupy, and is
+    /// the boundary NativeEngine::Draw's "< bgfx::Attrib::Count means a real per-vertex attribute"
+    /// guard rests on, so it -- not just INSTANCE_DATA_FIRST_LOCATION -- must stay >= Attrib::Count.
     inline constexpr uint32_t BUILTIN_INSTANCE_DATA_SLOT_COUNT{9};
     inline constexpr uint32_t BUILTIN_INSTANCE_DATA_LAST_LOCATION{INSTANCE_DATA_FIRST_LOCATION - (BUILTIN_INSTANCE_DATA_SLOT_COUNT - 1)};
 
-    /// The names Babylon.js uses for those built-in per-instance attributes. The shader compiler
-    /// recognizes them by name (ShaderCompilerTraversers.cpp) and NativeEngine counts how many of
-    /// them a program declares to size the instance data buffer, so both must read the same table.
+    /// The names Babylon.js uses for those attributes. The shader compiler recognizes them by name
+    /// and NativeEngine counts how many a program declares to size the instance data buffer, so
+    /// both must read the same table.
     inline constexpr std::array<std::string_view, 13> BUILTIN_INSTANCE_ATTRIBUTE_NAMES{
         "world0",
         "world1",
@@ -60,7 +59,6 @@ namespace Babylon::Graphics
         "splatIndex3",
     };
 
-    /// True when `name` is one of BUILTIN_INSTANCE_ATTRIBUTE_NAMES.
     inline constexpr bool IsBuiltInInstanceAttributeName(std::string_view name)
     {
         for (const std::string_view builtIn : BUILTIN_INSTANCE_ATTRIBUTE_NAMES)
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
@@ -1,8 +1,10 @@
 #pragma once
 
+#include <array>
 #include <cstdint>
 #include <string>
 #include <map>
+#include <string_view>
 #include <vector>
 
 namespace Babylon::Graphics
@@ -38,6 +40,38 @@ namespace Babylon::Graphics
     /// -- must stay >= bgfx::Attrib::Count. Keep in sync when adding a built-in per-instance attribute.
     inline constexpr uint32_t BUILTIN_INSTANCE_DATA_SLOT_COUNT{9};
     inline constexpr uint32_t BUILTIN_INSTANCE_DATA_LAST_LOCATION{INSTANCE_DATA_FIRST_LOCATION - (BUILTIN_INSTANCE_DATA_SLOT_COUNT - 1)};
+
+    /// The names Babylon.js uses for those built-in per-instance attributes. The shader compiler
+    /// recognizes them by name (ShaderCompilerTraversers.cpp) and NativeEngine counts how many of
+    /// them a program declares to size the instance data buffer, so both must read the same table.
+    inline constexpr std::array<std::string_view, 13> BUILTIN_INSTANCE_ATTRIBUTE_NAMES{
+        "world0",
+        "world1",
+        "world2",
+        "world3",
+        "previousWorld0",
+        "previousWorld1",
+        "previousWorld2",
+        "previousWorld3",
+        "instanceColor",
+        "splatIndex0",
+        "splatIndex1",
+        "splatIndex2",
+        "splatIndex3",
+    };
+
+    /// True when `name` is one of BUILTIN_INSTANCE_ATTRIBUTE_NAMES.
+    inline constexpr bool IsBuiltInInstanceAttributeName(std::string_view name)
+    {
+        for (const std::string_view builtIn : BUILTIN_INSTANCE_ATTRIBUTE_NAMES)
+        {
+            if (builtIn == name)
+            {
+                return true;
+            }
+        }
+        return false;
+    }
 
     struct BgfxShaderInfo
     {
--- a/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
+++ b/Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
@@ -23,14 +23,20 @@ namespace Babylon::Graphics
     inline constexpr uint32_t TEXCOORD0_ATTRIBUTE_LOCATION{10};
     inline constexpr uint32_t INSTANCE_DATA_FIRST_LOCATION{TEXCOORD0_ATTRIBUTE_LOCATION + INSTANCE_DATA_FIRST_TEXCOORD};
 
-    /// The built-in per-instance attributes occupy the top BUILTIN_INSTANCE_DATA_SLOT_COUNT i_data
-    /// slots: world0-3 and splatIndex0-3 map to i_data0..i_data3, instanceColor to i_data4 (see
-    /// ShaderCompilerTraversers.cpp's attribute table). BUILTIN_INSTANCE_DATA_LAST_LOCATION is the
-    /// lowest synthetic location any of them can occupy; it is the boundary NativeEngine::Draw's
-    /// "< bgfx::Attrib::Count means a real per-vertex attribute that needs rerouting" guard rests
-    /// on, so it -- not just INSTANCE_DATA_FIRST_LOCATION -- must stay >= bgfx::Attrib::Count.
-    /// Keep in sync when adding a built-in per-instance attribute on a lower i_data slot.
-    inline constexpr uint32_t BUILTIN_INSTANCE_DATA_SLOT_COUNT{5};
+    /// Mirrors bgfx's BGFX_CONFIG_MAX_INSTANCE_DATA_COUNT (bgfx/src/config.h, a private header):
+    /// the number of 16-byte per-instance slots (i_data0..i_data15) bgfx can bind in one draw.
+    inline constexpr uint32_t MAX_INSTANCE_DATA_SLOT_COUNT{16};
+
+
+    /// slots. Which slot each one gets is decided per shader, from the set the shader actually
+    /// declares (see ShaderCompilerTraversers.cpp), because bgfx requires the used i_data slots to
+    /// be a contiguous run starting at i_data0. The count is the size of the largest possible set:
+    /// world0-3 (or splatIndex0-3), instanceColor, and previousWorld0-3 for motion vectors.
+    /// BUILTIN_INSTANCE_DATA_LAST_LOCATION is the lowest synthetic location any of them can occupy;
+    /// it is the boundary NativeEngine::Draw's "< bgfx::Attrib::Count means a real per-vertex
+    /// attribute that needs rerouting" guard rests on, so it -- not just INSTANCE_DATA_FIRST_LOCATION
+    /// -- must stay >= bgfx::Attrib::Count. Keep in sync when adding a built-in per-instance attribute.
+    inline constexpr uint32_t BUILTIN_INSTANCE_DATA_SLOT_COUNT{9};
     inline constexpr uint32_t BUILTIN_INSTANCE_DATA_LAST_LOCATION{INSTANCE_DATA_FIRST_LOCATION - (BUILTIN_INSTANCE_DATA_SLOT_COUNT - 1)};
 
     struct BgfxShaderInfo
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -2553,6 +2553,29 @@ namespace Babylon
         m_boundFrameBufferNeedsRebinding.Set(false);
     }
 
+    // The number of contiguous i_data slots the current program's vertex shader reads for its
+    // built-in per-instance attributes. ShaderCompilerTraversers assigns those attributes a dense
+    // run of slots starting at i_data0, one per declared attribute, so the count of built-in names
+    // in the program's attribute table is that run's length. The instance data buffer must cover it
+    // even when the draw supplied fewer attributes, or D3D11 rejects the input layout.
+    uint32_t NativeEngine::GetBuiltInInstanceDataSlotCount() const
+    {
+        if (m_currentProgram == nullptr)
+        {
+            return 0;
+        }
+
+        uint32_t count{};
+        for (const auto& [name, location] : m_currentProgram->VertexAttributeLocations())
+        {
+            if (Babylon::Graphics::IsBuiltInInstanceAttributeName(name))
+            {
+                ++count;
+            }
+        }
+        return count;
+    }
+
     // Note: For legacy reasons JS might call this function for instance drawing.
     // In that case the instanceCount will be calculated inside the SetVertexBuffers method.
     void NativeEngine::DrawIndexed(NativeDataStream::Reader& data)
@@ -2565,7 +2588,7 @@ namespace Babylon
         if (m_boundVertexArray != nullptr)
         {
             m_boundVertexArray->SetIndexBuffer(encoder, indexStart, indexCount);
-            m_boundVertexArray->SetVertexBuffers(encoder, 0, std::numeric_limits<uint32_t>::max());
+            m_boundVertexArray->SetVertexBuffers(encoder, 0, std::numeric_limits<uint32_t>::max(), 0, GetBuiltInInstanceDataSlotCount());
         }
         DrawInternal(encoder, fillMode);
     }
@@ -2581,7 +2604,7 @@ namespace Babylon
         if (m_boundVertexArray != nullptr)
         {
             m_boundVertexArray->SetIndexBuffer(encoder, indexStart, indexCount);
-            m_boundVertexArray->SetVertexBuffers(encoder, 0, std::numeric_limits<uint32_t>::max(), instanceCount);
+            m_boundVertexArray->SetVertexBuffers(encoder, 0, std::numeric_limits<uint32_t>::max(), instanceCount, GetBuiltInInstanceDataSlotCount());
         }
         DrawInternal(encoder, fillMode);
     }
@@ -2597,7 +2620,7 @@ namespace Babylon
         bgfx::Encoder* encoder = GetEncoder();
         if (m_boundVertexArray != nullptr)
         {
-            m_boundVertexArray->SetVertexBuffers(encoder, verticesStart, verticesCount);
+            m_boundVertexArray->SetVertexBuffers(encoder, verticesStart, verticesCount, 0, GetBuiltInInstanceDataSlotCount());
         }
         DrawInternal(encoder, fillMode);
     }
@@ -2612,7 +2635,7 @@ namespace Babylon
         bgfx::Encoder* encoder = GetEncoder();
         if (m_boundVertexArray != nullptr)
         {
-            m_boundVertexArray->SetVertexBuffers(encoder, verticesStart, verticesCount, instanceCount);
+            m_boundVertexArray->SetVertexBuffers(encoder, verticesStart, verticesCount, instanceCount, GetBuiltInInstanceDataSlotCount());
         }
         DrawInternal(encoder, fillMode);
     }
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -2565,15 +2565,7 @@ namespace Babylon
             return 0;
         }
 
-        uint32_t count{};
-        for (const auto& [name, location] : m_currentProgram->VertexAttributeLocations())
-        {
-            if (Babylon::Graphics::IsBuiltInInstanceAttributeName(name))
-            {
-                ++count;
-            }
-        }
-        return count;
+        return m_currentProgram->BuiltInInstanceDataSlotCount();
     }
 
     // Note: For legacy reasons JS might call this function for instance drawing.
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -3025,10 +3025,10 @@ namespace Babylon
                 {
                     const bgfx::Attrib::Enum attrib = instance.first;
                     // "Real per-vertex slot" means Position..TexCoord15, i.e. < Attrib::Count. The
-                    // built-in instanced attributes (world0-3, splatIndex0-3, instanceColor) are
-                    // assigned synthetic locations at/above INSTANCE_DATA_FIRST_LOCATION - 4, which
-                    // is >= Attrib::Count, so they compare false here and are correctly skipped:
-                    // they already arrive as instance data.
+                    // built-in instanced attributes (world0-3, splatIndex0-3, previousWorld0-3,
+                    // instanceColor) are assigned synthetic locations at or above
+                    // BUILTIN_INSTANCE_DATA_LAST_LOCATION, which is >= Attrib::Count, so they compare
+                    // false here and are correctly skipped: they already arrive as instance data.
                     // The previous TexCoord3 boundary silently dropped generic instanced attributes
                     // landing on TexCoord3..TexCoord15 (e.g. sprite cellInfo -> TexCoord3), leaving
                     // them reading per-vertex garbage even though BuildInstanceDataBuffer had
--- a/Plugins/NativeEngine/Source/NativeEngine.cpp
+++ b/Plugins/NativeEngine/Source/NativeEngine.cpp
@@ -2553,11 +2553,9 @@ namespace Babylon
         m_boundFrameBufferNeedsRebinding.Set(false);
     }
 
-    // The number of contiguous i_data slots the current program's vertex shader reads for its
-    // built-in per-instance attributes. ShaderCompilerTraversers assigns those attributes a dense
-    // run of slots starting at i_data0, one per declared attribute, so the count of built-in names
-    // in the program's attribute table is that run's length. The instance data buffer must cover it
-    // even when the draw supplied fewer attributes, or D3D11 rejects the input layout.
+    // The dense run of i_data slots the current program's vertex shader reads for its built-in
+    // per-instance attributes. The instance data buffer must cover it even when the draw supplied
+    // fewer attributes, or D3D11 rejects the input layout.
     uint32_t NativeEngine::GetBuiltInInstanceDataSlotCount() const
     {
         if (m_currentProgram == nullptr)
--- a/Plugins/NativeEngine/Source/VertexArray.cpp
+++ b/Plugins/NativeEngine/Source/VertexArray.cpp
@@ -51,9 +51,12 @@ namespace Babylon
             }
 
             // Instance data is packed into the top i_data slots, of which bgfx has
-            // MAX_INSTANCE_DATA_SLOT_COUNT. The check runs before the insert, so
-            // `size() >= max` is the entry that would overflow.
-            if (m_vertexBufferInstances.size() >= Babylon::Graphics::MAX_INSTANCE_DATA_SLOT_COUNT)
+            // MAX_INSTANCE_DATA_SLOT_COUNT. Only a new attribute can overflow: re-recording
+            // one that is already present overwrites its entry and needs no extra slot.
+            // The check runs before the insert, so `size() >= max` is the entry that would
+            // overflow.
+            if (m_vertexBufferInstances.find(attrib) == m_vertexBufferInstances.end() &&
+                m_vertexBufferInstances.size() >= Babylon::Graphics::MAX_INSTANCE_DATA_SLOT_COUNT)
             {
                 throw std::runtime_error{"Number of vertex buffer instances greater than " + std::to_string(Babylon::Graphics::MAX_INSTANCE_DATA_SLOT_COUNT) + " is not supported"};
             }
--- a/Plugins/NativeEngine/Source/VertexArray.cpp
+++ b/Plugins/NativeEngine/Source/VertexArray.cpp
@@ -1,5 +1,7 @@
 #include "VertexArray.h"
 #include <cassert>
+#include <string>
+#include "Babylon/Graphics/BgfxShaderInfo.h"
 #include "Babylon/Graphics/DeviceContext.h"
 
 namespace Babylon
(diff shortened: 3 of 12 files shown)


Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
```diff
@@ -23,14 +23,20 @@ namespace Babylon::Graphics
...
-    /// "< bgfx::Attrib::Count means a real per-vertex attribute that needs rerouting" guard rests
-    /// on, so it -- not just INSTANCE_DATA_FIRST_LOCATION -- must stay >= bgfx::Attrib::Count.
-    /// Keep in sync when adding a built-in per-instance attribute on a lower i_data slot.
-    inline constexpr uint32_t BUILTIN_INSTANCE_DATA_SLOT_COUNT{5};
+    /// Mirrors bgfx's BGFX_CONFIG_MAX_INSTANCE_DATA_COUNT (bgfx/src/config.h, a private header):
+    /// the number of 16-byte per-instance slots (i_data0..i_data15) bgfx can bind in one draw.
+    inline constexpr uint32_t MAX_INSTANCE_DATA_SLOT_COUNT{16};
+
+
+    /// slots. Which slot each one gets is decided per shader, from the set the shader actually
```
bghgary: The opening line of this comment is missing, so it now starts mid-sentence at "slots.", and there is a stray extra blank line above. The removed version's first line still fits:

```suggestion

    /// The built-in per-instance attributes occupy the top BUILTIN_INSTANCE_DATA_SLOT_COUNT i_data
    /// slots. Which slot each one gets is decided per shader, from the set the shader actually
```

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -48,10 +50,12 @@ namespace Babylon
                 throw std::runtime_error{"Instancing is not supported"};
             }
 
-            // bgfx allows instancing on at most 4 vec4 attributes
-            if (m_vertexBufferInstances.size() > 4)
+            // Instance data is packed into the top i_data slots, of which bgfx has
+            // MAX_INSTANCE_DATA_SLOT_COUNT. The check runs before the insert, so
+            // `size() >= max` is the entry that would overflow.
+            if (m_vertexBufferInstances.size() >= Babylon::Graphics::MAX_INSTANCE_DATA_SLOT_COUNT)
```
bghgary: `bgfx::getCaps()->limits.maxInstanceData` carries this value at runtime, and `bgfx::getCaps()` is already queried on line 47. Every backend clamps it to the device's `maxVertexAttributes` (`renderer_d3d11.cpp:1373`, `renderer_gl.cpp:3006`, `renderer_vk.cpp:1909`, `renderer_d3d12.cpp:1761`), so 16 is an upper bound a device need not honour. It also drops the mirror, which can drift on a bgfx bump or from a `BGFX_CONFIG_MAX_INSTANCE_DATA_COUNT` override in `Dependencies/CMakeLists.txt` (it already sets eight other `BGFX_CONFIG_*` values). Keep the constant for the static_asserts and use the cap for this check.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -589,15 +602,53 @@ namespace Babylon::ShaderCompilerTraversers
...
+            // leave a hole in the run and the attributes past the hole would read zero.
+            //
+            // Assigning in reverse name order -- the alphabetically first name gets the highest
+            // slot, the last gets i_data0 -- is dense by construction, is stable between the base
+            // program and any instanced variant (the declared set is identical), and matches
+            // BuildInstanceDataBuffer, which packs the recorded instance buffers by descending
+            // attribute location. It also reproduces the previous fixed assignment exactly for the
+            // sets that existed before previousWorld0-3 (world0-3 on i_data3..i_data0, instanceColor
+            // on i_data4).
+            void AssignBuiltInInstanceSlots()
```
bghgary: This ordering rule is the load-bearing part of the change and nothing in CI exercises it.

`Apps/package-lock.json` pins `babylonjs` at 9.15.0 and CI runs `npm install`, so master's `Win32_x64_D3D11` log at `8bf91bc` reports `Babylon Native (v9.15.0) launched`, zero `Number of vertex buffer instances greater than 4` errors, and all three tests `validated` (`ran=302 passed=302 failed=0`) with the bug present. The 9 instance-limit errors you measured are on 9.21.2. So green CI here shows no regression against the 9.15.0 test set, not that the fix works, and a later break in the reverse-name-order assignment would land silently.

`getAttributes` is exposed to JS, so a `ShaderMaterial` in `Apps/UnitTests` declaring `world0-3` + `previousWorld0-3` + `instanceColor` could pin the assignment independently of the Babylon.js version.

Core/Graphics/InternalInclude/Babylon/Graphics/BgfxShaderInfo.h
```diff
@@ -23,14 +23,20 @@ namespace Babylon::Graphics
...
+    /// declares (see ShaderCompilerTraversers.cpp), because bgfx requires the used i_data slots to
+    /// be a contiguous run starting at i_data0. The count is the size of the largest possible set:
```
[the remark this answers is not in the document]
bkaradzic-microsoft: You're right, the first line got dropped when I rewrote the block. Restored in f73cad9d - it opens with "The built-in per-instance attributes occupy the top BUILTIN_INSTANCE_DATA_SLOT_COUNT i_data slots." again, and the stray blank line is gone.

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -48,10 +50,12 @@ namespace Babylon
...
             {
-                throw std::runtime_error{"Number of vertex buffer instances greater than 4 is not supported"};
+                throw std::runtime_error{"Number of vertex buffer instances greater than " + std::to_string(Babylon::Graphics::MAX_INSTANCE_DATA_SLOT_COUNT) + " is not supported"};
             }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch, and it was latent in the original `> 4` check too. Re-recording an attribute that is already present overwrites its entry and needs no extra slot, so only a new key can overflow. Fixed in f73cad9d by guarding the limit on `find(attrib) == end()`. Rebuilt and re-ran the three motion blur tests: identical results (2.478% / 2.061% / 2.355%, 3/3 passing).

Plugins/NativeEngine/Source/VertexArray.cpp
```diff
@@ -48,10 +50,15 @@ namespace Babylon
...
 
-            // bgfx allows instancing on at most 4 vec4 attributes
-            if (m_vertexBufferInstances.size() > 4)
+            // Instance data is packed into the top i_data slots, of which bgfx has
+            // MAX_INSTANCE_DATA_SLOT_COUNT. Only a new attribute can overflow: re-recording
+            // one that is already present overwrites its entry and needs no extra slot.
+            // The check runs before the insert, so `size() >= max` is the entry that would
+            // overflow.
+            if (m_vertexBufferInstances.find(attrib) == m_vertexBufferInstances.end() &&
+                m_vertexBufferInstances.size() >= Babylon::Graphics::MAX_INSTANCE_DATA_SLOT_COUNT)
```
bghgary: `bgfx::getCaps()->limits.maxInstanceData` carries this value at runtime, and `bgfx::getCaps()` is already queried on line 47. Every backend clamps it to the device's `maxVertexAttributes` (`renderer_d3d11.cpp:1373`, `renderer_gl.cpp:3006`, `renderer_vk.cpp:1909`, `renderer_d3d12.cpp:1761`), so 16 is an upper bound a device need not honour. It also drops the mirror, which can drift on a bgfx bump or from a `BGFX_CONFIG_MAX_INSTANCE_DATA_COUNT` override in `Dependencies/CMakeLists.txt` (it already sets eight other `BGFX_CONFIG_*` values). Keep the constant for the static_asserts and use the cap for this check.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bghgary: This ordering rule is the load-bearing part of the change and nothing in CI exercises it.

`Apps/package-lock.json` pins `babylonjs` at 9.15.0 and CI runs `npm install`, so master's `Win32_x64_D3D11` log at `8bf91bc` reports `Babylon Native (v9.15.0) launched`, zero `Number of vertex buffer instances greater than 4` errors, and all three tests `validated` (`ran=302 passed=302 failed=0`) with the bug present. The 9 instance-limit errors you measured are on 9.21.2. So green CI here shows no regression against the 9.15.0 test set, not that the fix works, and a later break in the reverse-name-order assignment would land silently.

`getAttributes` is exposed to JS, so a `ShaderMaterial` in `Apps/UnitTests` declaring `world0-3` + `previousWorld0-3` + `instanceColor` could pin the assignment independently of the Babylon.js version.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -589,15 +602,53 @@ namespace Babylon::ShaderCompilerTraversers
...
+            {
+                unsigned int slot{};
+                for (const auto& [name, symbol] : m_varyingNameToSymbol)
+                {
+                    if (IsInstance(name.c_str()) && !IsGenericInstance(name.c_str()))
+                    {
+                        ++slot;
```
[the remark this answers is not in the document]
bghgary: The collision is real, but it predates this PR and the example does not demonstrate a regression: master's pre-count excludes generics the same way (`ShaderCompilerTraversers.cpp:701` for GL and `:782` for Metal), so `world0-3` plus a generic `zOffset` puts `world3` on `i_data0` there too.

The condition is a generic sorting alphabetically *after* a declared built-in — built-in slots are dense over the built-ins alone, while `BuildInstanceDataBuffer` ranks over everything recorded. A generic sorting below them all (`aOffset`) still lines up exactly.

What this PR does change is how wide that window is: with `previousWorld0-3` declared, the lowest built-in moves from `world0` to `previousWorld0`, so more generic names fall above it.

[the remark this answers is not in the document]
bkaradzic-microsoft: Confirmed and fixed in d8628997 - thanks, this was a real latent bug, though it predates this PR (the old code excluded generic attributes from the count in exactly the same way; I only moved that logic into a shared helper).

Working through it:

`NativeEngine::Draw` computes the routed location as `INSTANCE_DATA_FIRST_LOCATION - rank`, where `rank` is the descending rank over **all** of `m_boundVertexArray->GetInstances()` - built-in and consumer-declared alike - which is exactly `BuildInstanceDataBuffer`'s reverse-iteration packing. So the caller-supplied location is the authoritative one.

On D3D the built-ins carry synthetic locations at or above `BUILTIN_INSTANCE_DATA_LAST_LOCATION`, so `attrib < bgfx::Attrib::Count` is false, they are never rerouted, and they never enter `instancedAttributes`. Their descending synthetic order and the reverse-name-order slot assignment agree by construction (`world3` -> highest location -> rank 0 -> `i_data0`), and a generic lands on the rank after them. That path was consistent.

On OpenGL/Metal the built-ins are recorded at their stable, name-sorted locations, i.e. the same range as consumer-declared ones, so they *do* get rerouted and enter the map. With `world0-3` plus a generic sorting after them, the packing is `zOffset` -> `i_data0`, `world3` -> `i_data1`, ..., while the built-in map still said `world3` -> `i_data0`. Collision, exactly as described.

The fix falls out of the above: whenever the caller supplied a location, use it, and exclude that attribute from the built-in slot assignment. `IsGenericInstance` is replaced by `HasCallerSuppliedInstanceLocation`, which drops the built-in exemption. On GL/Metal all of `world0-3` and the generic now come from the same packing-derived source, so they cannot disagree; on D3D built-ins never reach the map, so the built-in slot path is unchanged.

Re-ran the full 720 test sweep on D3D11: identical to before the change (the three motion blur tests still at 2.478% / 2.061% / 2.355%, no other test moving by more than 0.001%).

Plugins/NativeEngine/Source/VertexArray.cpp
bghgary: `bgfx::getCaps()->limits.maxInstanceData` carries this value at runtime, and `bgfx::getCaps()` is already queried on line 47. Every backend clamps it to the device's `maxVertexAttributes` (`renderer_d3d11.cpp:1373`, `renderer_gl.cpp:3006`, `renderer_vk.cpp:1909`, `renderer_d3d12.cpp:1761`), so 16 is an upper bound a device need not honour. It also drops the mirror, which can drift on a bgfx bump or from a `BGFX_CONFIG_MAX_INSTANCE_DATA_COUNT` override in `Dependencies/CMakeLists.txt` (it already sets eight other `BGFX_CONFIG_*` values). Keep the constant for the static_asserts and use the cap for this check.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -570,34 +570,89 @@ namespace Babylon::ShaderCompilerTraversers
...
+            // (world0-3 alone, plus instanceColor, plus previousWorld0-3), so a fixed table would
+            // leave a hole in the run and the attributes past the hole would read zero.
+            //
+            // Assigning in reverse name order -- the alphabetically first name gets the highest
+            // slot, the last gets i_data0 -- is dense by construction, is stable between the base
+            // program and any instanced variant (the declared set is identical), and matches
+            // BuildInstanceDataBuffer, which packs the recorded instance buffers by descending
+            // attribute location. It also reproduces the previous fixed assignment exactly for the
+            // sets that existed before previousWorld0-3 (world0-3 on i_data3..i_data0, instanceColor
+            // on i_data4).
```
bghgary: This paragraph no longer describes what the code does. After d8628997 the run is denominated over the declared built-ins **minus** those the caller routed, and that set is not identical between the two compiles: the base program compiles with an empty `instancedAttributes` (the default argument on `AssignLocationsAndNamesToVertexVaryings*`), so every declared built-in is assigned there, while a GL/Metal variant excludes the recorded ones. The "(the declared set is identical)" justification is what a future reader will rely on, so it needs to say the set is now declared-minus-routed.

The substantive consequence is that two runs now both start at `i_data0` from different denominators: this one over declared-minus-routed, and the caller-supplied locations over every recorded instanced attribute. They stay disjoint only while every declared built-in is also recorded for the draw. If one is declared but not recorded on GL/Metal while others are — say `instanceColor` declared with `world0-3` recorded — it takes `i_data0` here and collides with `world3`'s caller-supplied `i_data0`, and the failure is a duplicate symbol name rather than something diagnosable. I could not construct that case through Babylon.js, since the defines that declare each group are tied to the buffer existing, so this is an invariant worth asserting rather than a bug I can demonstrate. `slot > 0 && m_instancedAttributes != nullptr && !m_instancedAttributes->empty()` is the dangerous state and would throw with a clear message.

The header doc is stale in the same way and is outside this diff: `ShaderCompilerTraversers.h:78-80` still says the built-in names "keep their fixed mapping", which this PR removes, and does not mention `previousWorld0-3`.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -570,34 +570,89 @@ namespace Babylon::ShaderCompilerTraversers
...
+            void AssignBuiltInInstanceSlots()
```
bghgary: This ordering rule is the load-bearing part of the change, it has now been reworked twice in response to review, and nothing in CI exercises it.

`Apps/package-lock.json` pins `babylonjs` at 9.15.0 and CI runs `npm install`, so this PR's own `Win32_x64_D3D11` log at `f73cad9d` reports `Babylon Native (v9.15.0) launched`, zero `Number of vertex buffer instances greater than 4` errors, and all three tests `validated` (`ran=302 passed=302 failed=0`) — the same as master with the bug present. So green CI here shows no regression against the 9.15.0 test set, not that the fix works, and the mixed built-in/generic path d8628997 rewrites has no coverage at all.

`getAttributes` is exposed to JS, so a `ShaderMaterial` in `Apps/UnitTests` declaring `world0-3` + `previousWorld0-3` + `instanceColor`, and a second one adding a generic instanced attribute sorting after `world3`, would pin both orderings independently of the Babylon.js version.

bkaradzic-microsoft:
Pushed 50250290, which fixes a hard crash this PR would otherwise introduce. It is not reachable on the Babylon.js version this branch pins, which is why CI here is green — I only hit it while testing this PR together with #1840's bump to Babylon.js 9.21.2.

### What breaks

Test 323 `Thin instances + render self motion blur` aborts the process:

```
bgfx/src/renderer_d3d11.cpp(2933): ASSERT SUCCEEDED(__hr__) ->
  m_device->CreateInputLayout(vertexElements, num, ...) FAILED 0x80070057
  #0 bgfx::d3d11::RendererContextD3D11::setInputLayout
  #1 bgfx::d3d11::RendererContextD3D11::submit
```

### Why

Traced the offending draw:

```
shader builtin instance 'previousWorld0' -> i_data7
shader builtin instance 'previousWorld1' -> i_data6
shader builtin instance 'previousWorld2' -> i_data5
shader builtin instance 'previousWorld3' -> i_data4
shader builtin instance 'world0'         -> i_data3
shader builtin instance 'world1'         -> i_data2
shader builtin instance 'world2'         -> i_data1
shader builtin instance 'world3'         -> i_data0
draw instances: 38 39 40 41 (count=4)
```

The vertex shader declares **8** i_data slots; the draw supplies **4**. bgfx derives the instance-data half of the D3D11 input layout from the instance data buffer's stride (`renderer_d3d11.cpp`, the `_numInstanceData` loop), so it declares only TEXCOORD31..28 while the VS signature reads TEXCOORD31..24 — and `CreateInputLayout` rejects a layout that does not cover the signature.

Babylon.js gets there legitimately: `Mesh._renderWithThinInstances` creates the `previousWorld` buffer **after** the draw that first needs it.

```js
this._bind(subMesh, effect, fillMode);
this._draw(subMesh, fillMode, instancesCount);
// Write current matrices as previous matrices
if (this._scene.needsPreviousWorldMatrices && !...previousMatrixData && ...matrixData) {
    if (!this._thinInstanceDataStorage.previousMatrixBuffer) {
        this._thinInstanceDataStorage.previousMatrixBuffer = this._thinInstanceCreateMatrixBuffer("previousWorld", ...);
```

So the first draw binds `world0-3` while the effect already declares `previousWorld0-3`. WebGL tolerates this — an attribute whose array is disabled reads the generic attribute value — and so did Babylon Native before this PR, because a *per-vertex* attribute missing from the layout gets a dummy element from bgfx's `fillVertexLayout`. Instance data has no such fallback, so promoting `previousWorld0-3` to built-in instanced attributes turns a tolerated case into an abort.

### Fix

Count the built-in per-instance attributes the current program declares and pad the instance data buffer to that many 16-byte slots. Padding lands on the high slots, after the packed data, since `BuildInstanceDataBuffer` fills from i_data0 up, and the buffer is already zeroed. The built-in attribute name table moved into `BgfxShaderInfo.h` so the shader compiler and NativeEngine cannot disagree about which names count.

### Verification

| | before 50250290 | after |
|---|---|---|
| 323 on 9.21.2, isolated | abort in `setInputLayout` | validated, 2.355% |
| sequential 56-719 on 9.21.2 | aborts at 323 | ran=256 passed=256 failed=0 |
| sequential 0-52 + 56-719 on 9.15.0 (this branch's pin) | 299/299 | 299/299 |

2.355% is bit-identical to what the instance-limit raise alone produced before this PR's slot assignment existed, so the padded first frame does not perturb the final image.

Plugins/NativeEngine/Source/VertexArray.cpp
bghgary: `bgfx::getCaps()->limits.maxInstanceData` carries this value at runtime, and `bgfx::getCaps()` is already queried on line 47. Every backend clamps it to the device's `maxVertexAttributes` during init — D3D11, D3D12, OpenGL, Vulkan and Metal all do — so 16 is a compile-time ceiling a device need not honour.

Using the cap also drops the mirror, which can drift on a bgfx bump or from a `BGFX_CONFIG_MAX_INSTANCE_DATA_COUNT` override in `Dependencies/CMakeLists.txt` (which already sets a dozen other `BGFX_CONFIG_*` values, including `MAX_VERTEX_STREAMS` and `MAX_FRAME_BUFFERS`).

Keep the constant for the static_asserts in `ShaderCompilerCommon.cpp` and use the cap for this runtime check.

Plugins/NativeEngine/Source/VertexBuffer.cpp
```diff
@@ -145,15 +146,25 @@ namespace Babylon
...
+        // The buffer must cover every i_data slot the vertex shader reads, not just the ones the
+        // draw supplied data for: bgfx derives the number of instance-data inputs it declares from
+        // this buffer's stride, and D3D11's CreateInputLayout fails outright when the vertex
+        // shader's input signature reads a semantic the layout does not declare. Babylon.js can
+        // legitimately draw with fewer: _renderWithThinInstances creates the previousWorld buffer
+        // only *after* the first draw, so that draw binds world0-3 while the effect already
+        // declares previousWorld0-3. The padded slots stay zeroed, matching what WebGL feeds a
+        // vertex attribute whose array is disabled.
+        const size_t slotCount = std::max(static_cast<size_t>(minSlotCount), instances.size());
+        const uint16_t instanceStride = static_cast<uint16_t>(slotCount * kSlotSize);
```
bghgary: The padding lands in the **highest** slots, so this is correct only when the attributes the draw omitted are the ones holding those slots — i.e. the alphabetically first ones, since `AssignBuiltInInstanceSlots` gives the highest slot to the alphabetically first name.

That holds for the case in the comment (`previousWorld` sorts before `world`, so it occupies the top slots and is also what is missing) and for `instanceColor`, which sorts first of all. It is a coincidence between alphabetical order and Babylon.js's creation order rather than something either side enforces, so it is worth stating here — the failure if it ever inverts is silent, with data landing in the wrong slots rather than a padded run.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -570,34 +570,77 @@ namespace Babylon::ShaderCompilerTraversers
...
```
bghgary: `50250290` establishes that a built-in can be declared but not recorded. That case is also broken on Metal, differently from the D3D one it fixes.

Measured at this head — macOS 26.5.2, Babylon.js 9.22.0, `Thin instances + render self motion blur`, logging the slot map at every compile. The base program and later draws are dense and correct; only the first thin-instance draw, where 8 are declared and 4 recorded, collides:

```
[IDATA] builtin previousWorld3 -> i_data0
[IDATA] caller  world3         -> i_data0 (loc 41)
[IDATA][COLLISION] previousWorld3 and world3 both i_data0     (all four pairs)
```

It does not throw — the run renders, with `previousWorld` aliasing `world`, so velocity comes out zero instead of the smear this PR fixes.

The two runs have different denominators: the built-in run counts declared-minus-routed, the caller-supplied locations count everything recorded. They stay disjoint only while every declared built-in is also recorded.

`slot > 0 && m_instancedAttributes != nullptr && !m_instancedAttributes->empty()` is exactly that state and would throw instead of emitting a duplicate symbol.

Two doc points: the block above still says the assignment is "stable between the base program and any instanced variant (the declared set is identical)", which `d8628997` made false — the base compiles with an empty `instancedAttributes`, so it assigns every declared built-in while a variant excludes the routed ones. And `ShaderCompilerTraversers.h:78-80`, outside this diff, still says the built-in names "keep their fixed mapping" and omits `previousWorld0-3`.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -570,34 +570,77 @@ namespace Babylon::ShaderCompilerTraversers
...
+                if (slot > Babylon::Graphics::BUILTIN_INSTANCE_DATA_SLOT_COUNT)
+                {
+                    throw std::runtime_error("Shader declares " + std::to_string(slot) + " built-in per-instance attributes, but at most " + std::to_string(Babylon::Graphics::BUILTIN_INSTANCE_DATA_SLOT_COUNT) + " are supported.");
+                }
+
+                for (const auto& [name, symbol] : m_varyingNameToSymbol)
+                {
+                    if (IsBuiltInInstance(name.c_str()) && !HasCallerSuppliedInstanceLocation(name.c_str()))
+                    {
+                        m_builtInInstanceSlots[name] = --slot;
```
bghgary, line 642: Nothing in CI exercises this assignment: `Apps/package-lock.json` pins `babylonjs` at 9.15.0, where these scenes never declare `previousWorld0-3`, so this PR's own D3D11 run is green for the same reason master is.

`getAttributes` is exposed to JS, so a `ShaderMaterial` in `Apps/UnitTests` declaring `world0-3` + `previousWorld0-3` + `instanceColor` would pin the assignment independently of the npm version — and one declaring `previousWorld0-3` while recording only `world0-3` reproduces the collision above directly.

bkaradzic-microsoft:
Some evidence that this fix is load-bearing, which also explains why it does not flip any `config.json` test on its own.

`master` pins `babylonjs` 9.15.0, and at that version nothing declares `previousWorld0-3`, so the bug this PR fixes is unreachable from the validation suite — that is why enabling tests here would not have demonstrated anything.

#1840 happens to bump the bundles to 9.21.2, and at that version three tests start failing on Win32 D3D11 (and the equivalent jobs on Linux — 9 red jobs in total):

```
- Thin instances + dynamic buffer resize
- Instances + render self motion blur
- Thin instances + render self motion blur
```

I A/B'd this locally, RelWithDebInfo, running just those three:

| build | result |
| --- | --- |
| #1840 alone | `ran=3 passed=0 failed=3` — 3.346% / 2.531% / 2.844% pixel diff (allowed 2.5%) |
| #1840 + this PR merged | `ran=3 passed=3 failed=0` — 2.478% / 2.061% / 2.355% |

So this PR is what unblocks the Babylon.js version bump, and those three tests become real regression coverage for it as soon as the bump lands.

Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -2553,6 +2553,29 @@ namespace Babylon
...
+        uint32_t count{};
+        for (const auto& [name, location] : m_currentProgram->VertexAttributeLocations())
+        {
+            if (Babylon::Graphics::IsBuiltInInstanceAttributeName(name))
+            {
+                ++count;
+            }
+        }
+        return count;
+    }
```
bghgary, line 2567: `m_vertexAttributeLocations` is written once in `Program::Initialize` and cleared only in `Dispose`, so this count is invariant for the program's lifetime — but it is recomputed on every draw call, walking a `std::map<std::string, uint32_t>` and testing each key against the 13-name table.

It also runs for non-instanced draws, where the result is discarded: `SetVertexBuffers` only reaches `BuildInstanceDataBuffer` when `m_vertexBufferInstances` is non-empty.

Computing it once in `Program::Initialize` and returning the cached value removes both.

bghgary (commented):

One blocking comment: the scenario 50250290 fixes for D3D takes a different path on Metal, where I measured a duplicate `i_data0`. The rest are non-blocking.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
```diff
@@ -570,34 +570,95 @@ namespace Babylon::ShaderCompilerTraversers
...
+                {
+                    throw std::runtime_error("Shader declares " + std::to_string(slot) + " built-in per-instance attributes, but at most " + std::to_string(Babylon::Graphics::BUILTIN_INSTANCE_DATA_SLOT_COUNT) + " are supported.");
+                }
+
+                // slot > 0 means at least one declared built-in was not routed by the caller. An
+                // empty instancedAttributes map is the base program compile, where nothing is
+                // routed and this run owns the whole range.
+                if (slot > 0 && m_instancedAttributes != nullptr && !m_instancedAttributes->empty())
+                {
+                    throw std::runtime_error("Shader declares " + std::to_string(slot) + " built-in per-instance attribute(s) that the draw did not record, alongside " + std::to_string(m_instancedAttributes->size()) + " recorded instanced attribute(s). The two i_data assignments would overlap.");
```
bghgary: This throws on a state I measured occurring during normal rendering, so on Metal it converts a silent wrong-render into a hard failure.

At `50250290` I instrumented this function on macOS with Babylon.js 9.22.0, running `Thin instances + render self motion blur`. The first thin-instance draw compiles a variant with 8 declared built-ins and 4 recorded — `slot == 4`, `m_instancedAttributes->size() == 4` — which is exactly this condition. Babylon.js's `_renderWithThinInstances` calls `_draw()` and creates the `previousWorld` buffer only afterwards, so every such scene hits it once.

`NativeEngine::Draw` calls `GetOrCreateInstancedVariant` with no `try` (NativeEngine.cpp:3069), and `ShaderProvider::Get` compiles synchronously, so this propagates out of the draw call.

Throwing is right in the sense that no correct rendering exists for a duplicate slot — the alternative is the silent aliasing I measured. But as written the first frame of any thin-instance motion blur scene fails on Metal, where before it rendered wrongly. That trades one defect for a louder one rather than fixing it.

The underlying fix is to make the two numberings share a denominator, so the state cannot arise: rank the built-in run over *all* declared built-ins (routed or not) and let the caller-supplied location win per attribute, rather than excluding routed names from the count. Then the shader-side slot for `world3` is the same whether or not the draw recorded it, and `previousWorld3` never lands on `i_data0`.

If that is more than this PR should carry, the guard is still worth keeping — but it needs the D3D-only scope it currently lacks, or a follow-up issue linked from the throw message, so it does not ship as a first-frame failure on Apple platforms.

Plugins/NativeEngine/Source/VertexArray.cpp
bkaradzic-microsoft: Done in `a1eaf4dc`. `RecordVertexBuffer` and `RecordStorageBuffer` now bound by `bgfx::getCaps()->limits.maxInstanceData`, and I hoisted the `getCaps()` call already on line 47 into a local so the two checks share it instead of querying twice.

`MAX_INSTANCE_DATA_SLOT_COUNT` stays as the compile-time ceiling for the `static_assert` in `ShaderCompilerCommon.cpp:29` (`BUILTIN_INSTANCE_DATA_SLOT_COUNT <= MAX_INSTANCE_DATA_SLOT_COUNT`), which is what it is actually good for — the runtime check now reads the device value.

Plugins/NativeEngine/Source/VertexBuffer.cpp
bkaradzic-microsoft: Stated in `a1eaf4dc` — I extended the comment block here to record that the padding fills the *highest* slots, that this is only correct because the omitted attributes are the alphabetically first names (which `AssignBuiltInInstanceSlots` gives the top slots), that it holds for `previousWorld` and `instanceColor` by coincidence of alphabetical order and Babylon.js's creation order rather than by any enforced invariant, and that the failure mode if it inverts is silent misplacement rather than a padded run.

Plugins/NativeEngine/Source/NativeEngine.cpp
bkaradzic-microsoft: Done in `a1eaf4dc`. `Program` now computes the count once in `Initialize`, right where `m_vertexAttributeLocations` is populated, caches it in `m_builtInInstanceDataSlotCount`, and resets it in `Dispose`. `GetBuiltInInstanceDataSlotCount` is now a single accessor call, so both the per-draw map walk and the wasted work on non-instanced draws are gone.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: Thanks — the diagnosis was right, but I took the suggested condition literally in `a1eaf4dc` and it was wrong. Correcting the record, since CI caught it: `slot > 0 && m_instancedAttributes != nullptr && !m_instancedAttributes->empty()` is **not** the collision state on D3D, and it broke all six Win32 jobs (`50250290` green -> `a1eaf4dc` red) with

```
Shader declares 4 built-in per-instance attribute(s) that the draw did not
record, alongside 1 recorded instanced attribute(s).
```

on ordinary instanced draws. `NativeEngine::Draw` builds `genericInstancedAttributes` only from instances whose attrib is `< Attrib::Count`, so the built-ins — synthetic locations at or above `BUILTIN_INSTANCE_DATA_LAST_LOCATION` — are excluded from the caller map while still counted in the rank. A draw recording `world0-3` plus one generic attribute routes only the generic one, at rank 4 → `i_data4`, sitting directly on top of the built-in run `[0, 4)`. Dense, correct, and no overlap — but "some declared built-in went unrouted, and something was recorded" is exactly that state, so it threw.

The denominators differ, but that alone does not imply a collision; it has to be tested slot by slot. `56e9b6ea` does that: the built-in run owns `[0, builtInCount)`, so a caller-routed attribute collides only when `INSTANCE_DATA_FIRST_LOCATION - location` falls inside that window.

That still catches your Metal case exactly. 8 declared / 4 recorded gives `builtInCount == 4` for `previousWorld0-3`, while `world3` is routed to `i_data0`, `world2` to `i_data1`, and so on — all four below 4, so all four throw rather than aliasing. And it stays quiet for the D3D arrangement above, where the caller slot is 4.

Verified against the full validation suite: the overlap error is gone (0 occurrences), and pixel results are identical to `50250290` test-for-test (52/52), so the guard is now silent where it should be.

Both doc points fixed: the "stable between the base program and any instanced variant" claim is gone, and `ShaderCompilerTraversers.h:78-80` now says the built-ins are assigned a dense run rather than a fixed mapping, and lists `previousWorld0-3`.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bkaradzic-microsoft: Agreed, and this PR just demonstrated the cost of the gap the hard way: `a1eaf4dc` shipped a guard that threw on ordinary instanced draws, and nothing caught it until the Win32 validation run went red. A test at this level would have caught it immediately.

Not yet written — flagging it as still open rather than claiming it. Two notes for when I do:

- It does not need a new `.ts` and a rebuilt `dist`. `Tests.ShaderCompilation.cpp` already has two tests (`ExistingVec4UniformArrayIsNotReshaped`, `ReturnedUniformIsShapeConverted`) that drive `engine.createEffect` through `loader.Eval` with inline GLSL, so the assignment can be pinned from a string with no npm involvement — which also makes it independent of the 9.15.0 pin, as you wanted.
- Worth covering both directions: one effect declaring `world0-3` + `previousWorld0-3` + `instanceColor` asserting the full dense run via `getAttributes`, and one reproducing the collision so the throw is pinned as behaviour rather than left to be re-derived.

Plugins/ShaderCompiler/Source/ShaderCompilerTraversers.cpp
bghgary: [Responded by Copilot on behalf of @bghgary]

My suggested condition was wrong and your correction is right — I proposed `slot > 0 && !empty()` from the Metal case without checking it against the D3D arrangement, where an unrouted built-in alongside a routed generic is the normal, correct state.

Verified `56e9b6ea` by hand on both:

- **Metal, 8 declared / 4 recorded** — `builtInCount == 4` (`previousWorld0-3`), `world3` routed to location 41 → `callerSlot 0`, `world0` → `callerSlot 3`. All four `< 4`, so it throws instead of aliasing.
- **D3D, `world0-3` + one generic** — built-ins never enter `genericInstancedAttributes`, so `builtInCount == 4` and the only caller slot is 4. `4 < 4` is false: silent, as it must be.

Also confirms the CI history — `50250290` 34/34, `a1eaf4dc` 6 failures all Win32 D3D11, `56e9b6ea` 34/34.

One thing the new guard does not cover, worth a comment rather than code: it tests the caller slots against the built-in run, but not the built-in run against `BuildInstanceDataBuffer`'s padding. Those stay consistent only while the omitted built-ins are the alphabetically first names — the coincidence you documented in `VertexBuffer.cpp`. If it ever inverts, this guard is silent and the padding misplaces data.

bghgary (approved):

PR #1842, opened 2026-08-19
Stripped the bgfx config comments from the CMake files.

bkaradzic-microsoft:
Follow-up to #1841. Strips four CMake comments, three of them added by that PR and the earlier bgfx work.

They narrated the change that introduced them rather than documenting anything, and cited things that no longer exist — the removed `4096/256/1024` uniform buffer overrides, what bgfx stopped doing to ESSL 1.00 shaders, and tinyexr and bx described as "updated" and "new".

**`Dependencies/CMakeLists.txt`**
- the 9-line uniform buffer note above `BGFX_PLATFORM_SUPPORTS_WGSL` (which it had run into, so it also read as though it described that setting)
- the 4-line tinyexr `_CRT_SECURE_NO_WARNINGS` note
- the 7-line bx SSE4.2 note

**`Polyfills/Canvas/shaderc.cmake`**
- the 5-line ESSL profile note

25 lines deleted, nothing added. No functional change; verified CMake still configures.

I also swept the other 37 tracked CMake files for the same defect and found no further instances. The remaining long comments (RTTI, `/OPT:REF`, metal-cpp, draco, the OpenGL unit test skips) explain current-state rationale rather than a past change, so they are untouched.

### DIFF
--- a/Dependencies/CMakeLists.txt
+++ b/Dependencies/CMakeLists.txt
@@ -49,15 +49,6 @@ target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_DEFAULT_MAX_ENCODERS=2)
 target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_MAX_VERTEX_STREAMS=18)
 target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_MIN_RESOURCE_COMMAND_BUFFER_SIZE=16)
 
-# The uniform buffer sizing knobs are deliberately left at the bgfx defaults.
-# UniformBuffer::update reserves BGFX_CONFIG_UNIFORM_BUFFER_RESIZE_THRESHOLD_SIZE
-# of head room before every write, so the threshold must be at least as large as
-# the largest single uniform record. The opcode encoding caps that record at
-# 4 + 1023 * sizeof(Mat4) = 65476 bytes, and bgfx now static_asserts both that
-# bound and BGFX_CONFIG_MIN_UNIFORM_BUFFER_SIZE > the threshold. The old
-# 4096 / 256 / 1024 overrides violated the first assert, which means a shader
-# with a large enough uniform array (e.g. bone matrices) could overrun the
-# buffer. Do not re-add them without satisfying both asserts.
 target_compile_definitions(bgfx PUBLIC BGFX_PLATFORM_SUPPORTS_WGSL=0)
 
 # bgfx bundles the Khronos GLES headers but adds them as a PRIVATE include
@@ -107,10 +98,6 @@ set_property(TARGET bimg PROPERTY FOLDER Dependencies/bgfx)
 set_property(TARGET bx PROPERTY FOLDER Dependencies/bgfx)
 target_compile_definitions(bx PRIVATE _CRT_SECURE_NO_WARNINGS)
 
-# The updated tinyexr (vendored by bimg) uses fopen in its file I/O helpers,
-# which MSVC flags as C4996 and fails the warnings-as-errors UWP build. Mirror
-# the bx workaround above and silence the deprecation for the bimg targets that
-# compile the tinyexr implementation.
 foreach(_bimg_target IN ITEMS bimg_decode bimg_encode tinyexr)
     if(TARGET ${_bimg_target})
         target_compile_definitions(${_bimg_target} PRIVATE _CRT_SECURE_NO_WARNINGS)
@@ -125,13 +112,6 @@ if(NOT BABYLON_NATIVE_PLUGIN_NATIVEENGINE_WEBP AND TARGET bimg_decode)
 endif()
 
 if(NOT APPLE AND NOT ANDROID AND CMAKE_SYSTEM_PROCESSOR MATCHES "(x86_64|amd64|AMD64)" AND CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
-    # The new bx requires SSE4.2 on x86_64 (see bx scripts/toolchain.lua and
-    # include/bx/inline/simd128_sse.inl which uses _mm_round_ps / _mm_blendv_ps
-    # under "SSE4.1 - always available with SSE4.2 minspec"). bgfx.cmake does
-    # not propagate this flag, so add it explicitly for any GCC/Clang frontend on
-    # x86_64 (e.g. Linux, or Clang targeting Windows), none of which enable SSE4.x
-    # by default. MSVC (cl.exe) allows the intrinsics regardless and Apple Silicon
-    # is unaffected.
     foreach(_bx_target IN ITEMS bx bgfx bimg bimg_decode bimg_encode)
         if(TARGET ${_bx_target})
             target_compile_options(${_bx_target} PRIVATE -msse4.2)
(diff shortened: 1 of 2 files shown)


Dependencies/CMakeLists.txt
```diff
@@ -49,15 +49,10 @@ target_compile_definitions(bgfx PRIVATE BGFX_CONFIG_DEFAULT_MAX_ENCODERS=2)
...
-# of head room before every write, so the threshold must be at least as large as
-# the largest single uniform record. The opcode encoding caps that record at
-# 4 + 1023 * sizeof(Mat4) = 65476 bytes, and bgfx now static_asserts both that
-# bound and BGFX_CONFIG_MIN_UNIFORM_BUFFER_SIZE > the threshold. The old
-# 4096 / 256 / 1024 overrides violated the first assert, which means a shader
-# with a large enough uniform array (e.g. bone matrices) could overrun the
-# buffer. Do not re-add them without satisfying both asserts.
+# Don't override the uniform buffer sizes: bgfx static_asserts that the resize
+# threshold is at least the largest uniform record (65476 bytes) and that
+# BGFX_CONFIG_MIN_UNIFORM_BUFFER_SIZE is larger still.
```
bghgary: Passing both asserts doesn't make an override safe, so naming only them overstates the guard. `update()` grows once per write by `max(kIncrement, m_size/2)`, so 65536 / 65537 / 1024 satisfies both asserts and still drops records.

The condition that actually holds is `max(increment, MIN_UNIFORM_BUFFER_SIZE/2) >= 65476`, verified by reachability search over the `(size, pos)` states 1649 configurations can reach. The current defaults satisfy it twice over -- increment 65536, and 131072 > 2 x 65476 -- so deleting the overrides was right; only the stated reason is incomplete.

```suggestion
# Don't override the uniform buffer sizes: bgfx static_asserts that the resize
# threshold is at least the largest uniform record (65476 bytes) and that
# BGFX_CONFIG_MIN_UNIFORM_BUFFER_SIZE is larger still. Those asserts are
# necessary but not sufficient -- update() grows by max(increment, size/2) once
# per write, so that must cover a record as well, and nothing checks it.
```

Dependencies/CMakeLists.txt
```diff
@@ -125,13 +118,9 @@ if(NOT BABYLON_NATIVE_PLUGIN_NATIVEENGINE_WEBP AND TARGET bimg_decode)
...
-    # The new bx requires SSE4.2 on x86_64 (see bx scripts/toolchain.lua and
-    # include/bx/inline/simd128_sse.inl which uses _mm_round_ps / _mm_blendv_ps
-    # under "SSE4.1 - always available with SSE4.2 minspec"). bgfx.cmake does
-    # not propagate this flag, so add it explicitly for any GCC/Clang frontend on
-    # x86_64 (e.g. Linux, or Clang targeting Windows), none of which enable SSE4.x
-    # by default. MSVC (cl.exe) allows the intrinsics regardless and Apple Silicon
-    # is unaffected.
+    # bx requires SSE4.2 on x86_64 (simd128_sse.inl uses _mm_round_ps /
+    # _mm_blendv_ps) and bgfx.cmake does not propagate the flag. GCC/Clang do not
+    # enable SSE4.x by default; MSVC allows the intrinsics regardless.
```
bghgary: `_mm_round_ps` and `_mm_blendv_ps` are SSE4.1, not SSE4.2 — bx says so directly above the uses (`simd128_sse.inl:437`: `// SSE4.1 — always available with SSE4.2 minspec`). What requires SSE4.2 is bx's declared minspec in `scripts/toolchain.lua`.

```suggestion
    # bx's x86_64 minspec is SSE4.2 (scripts/toolchain.lua) and bgfx.cmake does not
    # propagate the flag. Without it GCC/Clang reject the SSE4.1 intrinsics in
    # simd128_sse.inl (_mm_round_ps / _mm_blendv_ps); MSVC allows them regardless.
```

Dependencies/CMakeLists.txt
```diff
@@ -125,13 +112,6 @@ if(NOT BABYLON_NATIVE_PLUGIN_NATIVEENGINE_WEBP AND TARGET bimg_decode)
...
 
 if(NOT APPLE AND NOT ANDROID AND CMAKE_SYSTEM_PROCESSOR MATCHES "(x86_64|amd64|AMD64)" AND CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
-    # The new bx requires SSE4.2 on x86_64 (see bx scripts/toolchain.lua and
-    # include/bx/inline/simd128_sse.inl which uses _mm_round_ps / _mm_blendv_ps
-    # under "SSE4.1 - always available with SSE4.2 minspec"). bgfx.cmake does
-    # not propagate this flag, so add it explicitly for any GCC/Clang frontend on
-    # x86_64 (e.g. Linux, or Clang targeting Windows), none of which enable SSE4.x
-    # by default. MSVC (cl.exe) allows the intrinsics regardless and Apple Silicon
-    # is unaffected.
     foreach(_bx_target IN ITEMS bx bgfx bimg bimg_decode bimg_encode)
```
bghgary: Nothing left says why the flag is needed or why bgfx.cmake doesn't supply it, so this block reads as removable. The stale part was "the new bx".

The suggestion is not a verbatim restore: `_mm_round_ps` and `_mm_blendv_ps` are SSE4.1, not SSE4.2 — bx says so directly above the uses (`simd128_sse.inl:437`). What requires SSE4.2 is bx's declared minspec in `scripts/toolchain.lua`.

```suggestion
    # bx's x86_64 minspec is SSE4.2 (scripts/toolchain.lua) and bgfx.cmake does not
    # propagate the flag. Without it GCC/Clang reject the SSE4.1 intrinsics in
    # simd128_sse.inl (_mm_round_ps / _mm_blendv_ps); MSVC allows them regardless.
    foreach(_bx_target IN ITEMS bx bgfx bimg bimg_decode bimg_encode)
```

Polyfills/Canvas/shaderc.cmake
```diff
@@ -235,11 +235,6 @@ function(add_bgfx_shader FILE FOLDER)
         # essl
         if(NOT "${TYPE}" STREQUAL "COMPUTE")
             set(ESSL_OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/Source/Shaders/essl/${FILENAME}.h)
-            # ESSL 1.00 is no longer usable: bgfx dropped glsl-optimizer and its GLSL
-            # fixup pass, and now unconditionally prepends `#version 300 es` to shaders
-            # that don't carry a version directive. Legacy `attribute`/`varying` sources
-            # fail to compile against that header, so target ESSL 3.00 directly, which is
-            # also what bgfx uses for its own embedded shaders.
             _bn_shaderc_parse(ESSL ${COMMON} ANDROID PROFILE 300_es OUTPUT ${ESSL_OUTPUT} BIN2C "${FILENAME}_essl")
```
bghgary: Nothing left says why 300 rather than 100, so lowering it looks safe and breaks Android. bgfx prepending `#version 300 es` is current behaviour, not the history this PR is removing.

```suggestion
            # Keep this at ESSL 3.00: bgfx prepends `#version 300 es` to
            # versionless shaders and doesn't rewrite `attribute`/`varying`.
            _bn_shaderc_parse(ESSL ${COMMON} ANDROID PROFILE 300_es OUTPUT ${ESSL_OUTPUT} BIN2C "${FILENAME}_essl")
```

bghgary (approved):

PR #1843, opened 2026-08-20
Canvas: consolidate the drawing state into a State struct

bkaradzic-microsoft:
Follow-up to #1824, as agreed in review.

## `State` struct

`save()`/`restore()` had to rewind seventeen separate wrapper-side members, and did it by hand: `Save()` built a `SavedStyle` from a positional initializer list, `Restore()` copied each field back out one at a time. The field list was written out three times over -- the members, the `SavedStyle` mirror, and the two copy loops -- so adding an attribute meant editing four places, and missing one silently produced an attribute that never rewound. That is exactly the bug class #1824 spent several commits fixing.

They are now a single `State` struct, pushed and popped whole:

```cpp
void Context::Save(const Napi::CallbackInfo&)
{
    nvgSave(*m_nvg);
    m_savedStates.push_back(m_state);
}

void Context::Restore(const Napi::CallbackInfo&)
{
    nvgRestore(*m_nvg);
    m_isClipped = false;
    if (!m_savedStates.empty())
    {
        m_state = std::move(m_savedStates.back());
        m_savedStates.pop_back();
    }
}
```

`Save`/`Restore` no longer name a single attribute, so a new one is rewound correctly by construction.

No behavior change: the saved set is exactly what `SavedStyle` held. The current path stays outside the stack, since per spec `save()` does not save it, and `m_fonts` stays out too -- it is a context-wide face cache, not an attribute.

## `globalAlpha` (found by the refactor)

Collecting the fields made a dead one obvious: `globalAlpha` was pushed and popped by every `save()`/`restore()`, but had **zero readers**. It was registered write-only --

```cpp
InstanceAccessor("globalAlpha", nullptr, &Context::SetGlobalAlpha),
```

-- so `ctx.globalAlpha` read back `undefined`, and the setter passed the value straight to `nvgGlobalAlpha` without ever mirroring it.

Registered the getter and mirrored the value. Rendering was already correct (`nvgSave`/`nvgRestore` rewinds nanovg's own copy), so this only changes what the attribute reports.

While validating the value I also applied the spec's range rule: a value that is not finite, or outside `[0, 1]`, is *ignored* and leaves the previous value in place, rather than being clamped or thrown. The old code forwarded `NaN` and out-of-range values to nanovg unchecked.

## Validation

- Unit tests: 21/21 suites, **45** JS assertions (was 43).
- Visual sweep: `ran=305 passed=305 failed=0`; "Native Canvas" unchanged at 1.850%.
- A/B: reverting only the C++ change fails exactly the three `globalAlpha` assertions and nothing else, so the tests are actually gated on the fix.

### DIFF
--- a/Polyfills/Canvas/Source/Context.h
+++ b/Polyfills/Canvas/Source/Context.h
@@ -130,8 +130,12 @@ namespace Babylon::Polyfills::Internal
         {
             std::variant<std::string, GradientStyle> fillStyle{};
             std::variant<std::string, GradientStyle> strokeStyle{};
-            std::string lineCap{};  // 'butt', 'round', 'square'
-            std::string lineJoin{}; // 'round', 'bevel', 'miter'
+            // These four mirror state nanovg also holds. Their initial values must match what
+            // nvgReset actually installs (NVG_BUTT, NVG_MITER, strokeWidth 1, miterLimit 10),
+            // which is also what the spec requires, or a fresh context reports a default it is
+            // not drawing with.
+            std::string lineCap{"butt"};   // 'butt', 'round', 'square'
+            std::string lineJoin{"miter"}; // 'round', 'bevel', 'miter'
 
             // Dash pattern from setLineDash. Retained only so getLineDash() round-trips;
             // strokes are always drawn solid (nanovg has no dashed stroke).
@@ -146,8 +150,8 @@ namespace Babylon::Polyfills::Internal
             double shadowOffsetY{0.0};
             std::string filter{};
             std::string direction{"ltr"}; // 'ltr', 'rtl'
-            float miterLimit{0.f};
-            float lineWidth{0.f};
+            float miterLimit{10.f};
+            float lineWidth{1.f};
             float globalAlpha{1.f};
             float letterSpacing{0.f};
 
--- a/Polyfills/Canvas/Source/Context.h
+++ b/Polyfills/Canvas/Source/Context.h
@@ -152,7 +152,7 @@ namespace Babylon::Polyfills::Internal
             std::string direction{"ltr"}; // 'ltr', 'rtl'
             float miterLimit{10.f};
             float lineWidth{1.f};
-            float globalAlpha{1.f};
+            double globalAlpha{1.0};
             float letterSpacing{0.f};
 
             // font is worse than the getter-only cases above: currentFontId is what the text
--- a/Polyfills/Canvas/Source/Context.h
+++ b/Polyfills/Canvas/Source/Context.h
@@ -87,6 +87,7 @@ namespace Babylon::Polyfills::Internal
         void SetFont(const Napi::CallbackInfo&, const Napi::Value& value);
         Napi::Value GetLetterSpacing(const Napi::CallbackInfo&);
         void SetLetterSpacing(const Napi::CallbackInfo&, const Napi::Value& value);
+        Napi::Value GetGlobalAlpha(const Napi::CallbackInfo&);
         void SetGlobalAlpha(const Napi::CallbackInfo&, const Napi::Value& value);
         Napi::Value GetShadowColor(const Napi::CallbackInfo&);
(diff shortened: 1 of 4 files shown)


Polyfills/Canvas/Source/Context.h
```diff
@@ -106,71 +107,62 @@ namespace Babylon::Polyfills::Internal
...
+            // so nothing is ever drawn from them. Defaults are the spec's.
+            std::string shadowColor{"rgba(0, 0, 0, 0)"};
+            double shadowBlur{0.0};
+            double shadowOffsetX{0.0};
+            double shadowOffsetY{0.0};
+            std::string filter{};
+            std::string direction{"ltr"}; // 'ltr', 'rtl'
+            float miterLimit{0.f};
+            float lineWidth{0.f};
+            float globalAlpha{1.f};
```
bghgary: [Reviewed by Copilot on behalf of @bghgary]

`ctx.globalAlpha = 0.1` reads back `0.10000000149011612`. The three new assertions use only 1, 0.5 and 0, all exact in binary float, so they pass either way. The IDL type is `unrestricted double`.

```suggestion
            double globalAlpha{1.0};
```

The `static_cast<float>(alpha)` in `SetGlobalAlpha` has to move to the `nvgGlobalAlpha` call, or the narrowing survives the type change.

Polyfills/Canvas/Source/Context.h
```diff
@@ -106,71 +107,62 @@ namespace Babylon::Polyfills::Internal
...
-            float lineWidth;
-            float globalAlpha;
-            float letterSpacing;
-
-            // font is worse than the getter-only cases above: m_currentFontId is what the
-            // text draw path binds, so without it the wrong face actually renders after a
+            std::variant<std::string, GradientStyle> fillStyle{};
+            std::variant<std::string, GradientStyle> strokeStyle{};
+            std::string lineCap{};  // 'butt', 'round', 'square'
+            std::string lineJoin{}; // 'round', 'bevel', 'miter'
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in a3ca7b64, and it turned out to be wider than the two lines flagged.

`lineWidth` and `miterLimit` were wrong the same way: they defaulted to `0.f`. All four are mirrors of state nanovg also holds, and `nvgReset` installs `NVG_BUTT`, `NVG_MITER`, `strokeWidth 1` and `miterLimit 10` -- which is also what the spec requires -- so a fresh context was reporting four defaults it was not drawing with.

Nothing rendered changes: the mirror is only ever read by the getters, and `restore()` pops it while `nvgRestore` rewinds nanovg's own copy, so the mirror never drives a draw. Added a test asserting all four on a fresh context; verified against the getters, which return the mirror directly.

I left `filter` alone deliberately. Its spec default is `"none"`, but the empty string is load-bearing here -- `SetFilterStack` uses `m_state.filter.length()` as the "no filter" sentinel -- so changing it would mean teaching the filter parser about `"none"`. That is a behaviour change rather than a default fix, so it does not belong in this PR.

bghgary (approved):

Polyfills/Canvas/Source/Context.h
bghgary: [Responded by Copilot on behalf of @bghgary]

At 5c0c4454 `ctx.globalAlpha = 0.1` still reads back `0.10000000149011612`: `SetGlobalAlpha` narrows with `static_cast<float>(alpha)` before assigning into the double.

```cpp
m_state.globalAlpha = alpha;
nvgGlobalAlpha(*m_nvg, static_cast<float>(m_state.globalAlpha));
```

bghgary:

This is still on master at c8a80715 — `SetGlobalAlpha` narrows before assigning into the double, so `ctx.globalAlpha = 0.1` reads back `0.10000000149011612`.

```cpp
m_state.globalAlpha = alpha;
nvgGlobalAlpha(*m_nvg, static_cast<float>(m_state.globalAlpha));
```

The three assertions use 1, 0.5 and 0, all exact in binary float, so they pass either way.

PR #226, opened 2020-04-12
Replace curl implementations with new library UrlLib

bghgary:
This includes some of the changes from this [PR](https://github.com/BabylonJS/BabylonNative/pull/190) for Android support.

- Remove `curl` and `openssl` submodules
- Add new library `UrlLib` with platform specific implementations of `UrlRequest`
  - Support `app:` scheme which redirects to loading resources from app package (except Win32 which doesn't have an app package; may consider pointing to the executable directory for Win32)
  - Support `file:` scheme for loading local files (if permissions allow)
  - (only GET method so far, no progress, no linux support yet)
- Add new `AndroidExtensions` which replaces the `Platform` library
  - `android::global` for global env and application context
  - `JavaWrappers` for calling Android Java classes
- Fix issue in JSC Napi implementation in ArrayBuffer creation.
  - (Will be fully implementing this soon.)
- Move XMLHttpRequest to Polyfill and implement against new UrlLib
- AppRuntime and XMLHttpRequest no longer take/provide a rootUrl
- A bunch of clean up

PR #1844, opened 2026-08-20
Canvas: unforgeable native instance branding (Path2D/Gradient/Image)

bkaradzic-microsoft:
## Summary

Unforgeable type checks for Canvas polyfill natives (`Path2D`, `CanvasGradient`, `Image`).

**Status:** Held pending JsRuntimeHost type tags (per @bghgary). Nothing in Babylon.js deliberately re-prototypes these objects, so this is not on the critical path for Canvas feature work.

**Feature work that used to stack here** now lives in **#1855** (metrics / filters / `drawImage(canvas)` / `toDataURL`) and no longer depends on this PR.

### Problem
`ObjectWrap::Unwrap` on an unchecked JS value is unsafe on JsRuntimeHost Node-API ports (V8 AV; QuickJS wrong pointer). `instanceof` is forgeable via `Object.setPrototypeOf`.

### Approach (temporary registry)
`NativeInstanceRegistry` brands each instance and only accepts live registered addresses via `TryUnwrap`. Replace with `CheckTypeTag` once tags land on all ports (related: [JsRuntimeHost#226](https://github.com/BabylonJS/JsRuntimeHost/issues/226) covers unwrap safety only — tags are still a separate port gap with no issue yet).

### Scope
- Path2D / Gradient / Image register + safe unwrap
- `fill` / `stroke` / `addPath` / `drawImage` reject bad args with `TypeError`
- Unit tests for spoofed prototypes

### Not blocked
Canvas text metrics, filter lifetime, canvas→canvas blit, `toDataURL` → see #1855.


Polyfills/Canvas/Source/Context.cpp
```diff
@@ -647,8 +659,20 @@ namespace Babylon::Polyfills::Internal
 
     void Context::Stroke(const Napi::CallbackInfo& info)
     {
+        // stroke() takes nothing or a Path2D. The argument used to be unwrapped with no type
+        // check at all, so ctx.stroke("x") cast a string straight into Unwrap.
+        const NativeCanvasPath2D* path = nullptr;
+        if (info.Length() >= 1 && !info[0].IsUndefined())
```
bghgary: `stroke(undefined)` should throw. The IDL is `stroke()` and `stroke(Path2D path)`; `path` is not optional, so with one argument the `Path2D` entry is the only candidate and `undefined` fails it — Chromium gives `TypeError: parameter 1 is not of type 'Path2D'`. `fill` differs only because its first argument is `optional CanvasFillRule fillRule = "nonzero"`, which is what makes `fill(undefined)` the no-argument form.

```suggestion
        if (info.Length() >= 1)
```

`still accepts the valid fill and stroke argument forms` asserts `ctx.stroke(undefined)` does not throw, and its comment carries the `fill` reasoning across to `stroke`. Babylon.js passes no argument to `stroke` at any of its six call sites in `babylon.max.js`, so nothing in the bundle relies on the current behaviour.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -161,6 +161,64 @@ describe("Canvas2D", function () {
...
   });
 
+  it("rejects a non-Path2D argument to fill and stroke", function () {
+    // Neither call type-checked the argument before handing it to ObjectWrap::Unwrap,
+    // which does no checking of its own, so an unrelated object was reinterpreted as a
+    // NativeCanvasPath2D. stroke() was the worse of the two: it had no gate at all, so
+    // even a string got there.
+    // The thrown type is deliberately not asserted: a C++ Napi::TypeError surfaces as a
+    // JS TypeError on some engines and as an InternalError on the QuickJS Node-API port,
+    // so only the fact that it throws is portable. Every other throw test here does the same.
```
bghgary: This reads as a property of the engine; it is a one-line bug in the port. At the pinned JsRuntimeHost `9271f13b`, `napi_throw` returns `napi_pending_exception` (`Core/Node-API/Source/js_native_api_quickjs.cc:1589`) where JavaScriptCore, V8 and ChakraCore all return `napi_ok` — JSC asserts the internal status is `napi_pending_exception` before returning `napi_ok`, so the distinction is deliberate there. node-addon-api treats any non-`napi_ok` from `napi_throw` as a failure and rethrows `Error::New(env)`, which calls `napi_get_and_clear_last_exception` and discards the `TypeError` that `JS_Throw` had already set; the trampoline then finds no pending exception and synthesises the `InternalError`.

The consequence is wider than these assertions: on QuickJS every native throw loses its class and its message, so script sees `InternalError: Uncaught C++ exception:` and nothing else.

Polyfills/Canvas/Source/Path2D.cpp
```diff
@@ -146,6 +166,14 @@ namespace Babylon::Polyfills::Internal
 
     void NativeCanvasPath2D::AddPath(const Napi::CallbackInfo& info)
     {
+        // addPath requires a Path2D. There was no type check and no arity check, so
+        // addPath() read a missing argument and addPath("x") cast a string, both of
+        // which reached Unwrap.
+        if (info.Length() < 1 || !NativeCanvasPath2D::IsInstance(info.Env(), info[0]))
+        {
+            throw Napi::TypeError::New(info.Env(), "Path2D.addPath: the first argument is not a Path2D.");
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Good catch -- fixed in ced0712e. The arity check is now separate and reports `Path2D.addPath: requires at least 1 argument (path).`, which matches the message style used for the `putImageData` arity check in #1824.

Polyfills/Canvas/Source/Path2D.cpp
```diff
@@ -58,14 +58,34 @@ namespace Babylon::Polyfills::Internal
...
+        if (!value.IsObject())
+        {
+            return false;
+        }
+
+        // Tested against the constructor kept on the native object rather than the global one:
+        // the global is writable, so script can replace it, and this has to answer whether the
+        // object is safe to Unwrap, not whether it matches whatever Path2D currently names.
+        const auto constructor = JsRuntime::NativeObject::GetFromJavaScript(env).Get(JS_PATH2D_CONSTRUCTOR_NAME);
+        return constructor.IsFunction() && value.As<Napi::Object>().InstanceOf(constructor.As<Napi::Function>());
```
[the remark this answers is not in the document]
bkaradzic-microsoft: You are right, and it is worse than a type-confusion risk -- both directions are a hard crash. Fixed in 4b37ddff.

I reproduced it before changing anything. `Object.setPrototypeOf(gradient, Path2D.prototype)` makes `gradient instanceof Path2D` true, and `ctx.fill(gradient)` then exits with `0xC0000005` on Win32 D3D11. `Object.create(Path2D.prototype)` passes the same check while having no native wrap behind it at all.

Checking the neighbouring type on your hint that a gradient was the vehicle: `CanvasGradient::IsInstance` had the identical gate, and it is reachable the other way round -- `ctx.fillStyle = spoofedPath2D` stores the impostor, and the next fill does `CanvasGradient::Unwrap(...)->Paint()`. Also `0xC0000005`. Fixed both.

On the mechanism, I went through the options you implied:

- `napi_type_tag_object` is the idiomatic answer, but only the V8 port implements it in this tree; the Chakra, JavaScriptCore and QuickJS ports do not, so it is not usable here.
- A brand property is forgeable even under a symbol, since `Object.getOwnPropertySymbols` hands the symbol to script, which can then define it on any object.

So the authority has to live somewhere script cannot reach, which means C++. `NativeInstanceRegistry<T>` records the address of every live instance; a candidate is accepted only when its unwrapped pointer is one of them. The pointer is compared and never dereferenced before it is accepted, so a foreign wrapped object is rejected rather than misread, and a reused address cannot alias because entries are removed in the destructor. It also calls `napi_unwrap` directly instead of `ObjectWrap::Unwrap`, which throws for an object that was never wrapped -- that is what closes the `Object.create` case.

Both crashes now have regression tests, and they are the prototype-spoofed-native-object tests you asked for. Full run is 21/21 gtest suites and 49 JS assertions, up from 47. I also ran the 305-test visual sweep because the gradient draw path changed: 305/305, unchanged.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
```diff
@@ -0,0 +1,78 @@
...
+        // an object that was never wrapped, and this has to answer "no" for any value at all.
+        static T* TryUnwrap(Napi::Env env, const Napi::Value& value)
+        {
+            if (!value.IsObject())
+            {
+                return nullptr;
+            }
+
+            void* unwrapped{};
+            if (napi_unwrap(env, value, &unwrapped) != napi_ok || unwrapped == nullptr)
```
bghgary: `napi_unwrap` cannot carry this. It is not in the JSI backend's API, which is the whole of the `UWP_arm64_JSI` and `Win32_x64_JSI_D3D11` failures (`error C3861: 'napi_unwrap': identifier not found`). And on V8 it does not answer "is this object wrapped" — `js_native_api_v8.cc:345` at the pinned `9271f13b`:

```cpp
// [BABYLON-NATIVE-ADDITION]: Increase perf by using internal field instead of private property
Reference* reference =
    static_cast<v8impl::Reference*>(obj->GetAlignedPointerFromInternalField(0));
if (result) { *result = reference->Data(); }
```

Upstream Node guards that with `RETURN_STATUS_IF_FALSE(env, val->IsExternal(), napi_invalid_arg)`; the internal-field change dropped it. `Object.create(Path2D.prototype)` has no internal fields, so field 0 is unrelated memory and `Data()` faults — `Win32_x64_V8_D3D11` exits `0xC0000005` in `rejects a prototype-spoofed object in place of a Path2D`. The registry itself is fine there: the same test's other impostor, a real gradient wearing `Path2D.prototype`, unwraps to a valid pointer and is correctly rejected. The comparison never dereferences, but the call that produces the pointer does.

Chakra is the one port that already answers this safely — `FindWrapper` (`js_native_api_chakra.cc:252`) uses `JsGetOwnPropertyDescriptor` specifically so a prototype-chain entry cannot spoof an unwrap — and `Win32_x64_D3D11` is the Chakra job that passes.

`Ubuntu_Clang_QuickJS` segfaults separately, in `rejects a non-Path2D argument to Path2D.addPath`. It clears both spoof tests first, so it may not be this cause. Mocha prints a test name only on pass, so in both crashes the failing test is the one after the last name printed.

Polyfills/Canvas/Source/Path2D.cpp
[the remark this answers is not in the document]
bkaradzic-microsoft: Follow-up: the first version of this fix reached the instance pointer through `ObjectWrap::Unwrap`, and that turned out not to be safe to call on an object that might never have been wrapped. Neither JsRuntimeHost Node-API port honours that part of the contract:

- the V8 port dereferences internal field 0 unconditionally (`[BABYLON-NATIVE-ADDITION]` in `js_native_api_v8.cc`), so a plain object access-violates;
- the QuickJS port falls back to walking the prototype chain, so it returns some other object's native pointer.

`Object.create(Path2D.prototype)` hit exactly this and crashed the V8 and QuickJS unit test runs.

So the check no longer calls `Unwrap` at all: each instance brands its own JS object with an `External` holding its address, and a candidate is accepted only if that address is still registered natively. Externals are opaque to script, and the address is compared, never dereferenced, before it is accepted.

An identity check against the instance's own object would also close brand-copying, but `ObjectWrap::Value()` throws on the QuickJS port, so it is not portable. It is not a memory-safety gap: copying a brand off a real instance yields that live instance, which the script already held, and a brand left over from a collected instance is rejected because the destructor unregisters the address.

Verified on Chakra, V8 and QuickJS: 21/21 gtest, 49 assertions each. Reverting only the C++ change reproduces the access violation. Visual sweep 305/305.

I'll file the `napi_unwrap` contract violations against JsRuntimeHost separately, since every `ObjectWrap::Unwrap` call site in BabylonNative has the same exposure.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
```diff
@@ -0,0 +1,102 @@
...
+            if (!value.IsObject())
+            {
+                return nullptr;
+            }
+
+            // `value` is arbitrary, so the read can run a script accessor that throws.
+            Napi::Value brand{env.Undefined()};
+            try
+            {
+                brand = value.As<Napi::Object>().Get(BRAND_NAME);
```
bghgary: `Get` walks the prototype chain, so `Object.create(realPath)` inherits the brand and is accepted; Chromium throws `TypeError` there for `fill`, `stroke` and `addPath`. `Object.prototype.__nativeInstance = realPath.__nativeInstance` extends that to every object, so `ctx.fill({})` fills instead of throwing.

An own-property read is the obvious fix and does not work here. This napi exposes no own-descriptor getter, and `napi_has_own_property` is broken on the JavaScriptCore port — it calls `Object.hasOwnProperty()` with `this` bound to `Object` and no arguments, ignoring both `object` and `key` (`js_native_api_javascriptcore.cc:1064`), so it always returns false. JavaScriptCore is the default engine on Apple and Linux, so that route would reject every legitimate instance there.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
```diff
@@ -0,0 +1,102 @@
...
+    // destructor unregisters the address.
+    template<typename T>
+    class NativeInstanceRegistry
+    {
+    public:
+        // Call at the very end of the constructor: one that throws never reaches the
+        // destructor, which would leave a dangling address registered.
+        static void Add(const Napi::CallbackInfo& info, T* instance)
+        {
+            info.This().As<Napi::Object>().Set(BRAND_NAME, Napi::External<T>::New(info.Env(), instance));
```
bghgary: Define the brand rather than assigning it. `Set` is `[[Set]]`, so `__nativeInstance` lands enumerable, writable and configurable: `Object.getOwnPropertyNames(new Path2D())` returns `["__nativeInstance"]` where browsers return `[]`, and `delete path.__nativeInstance` permanently breaks a legitimate `Path2D` — every later `fill`/`stroke`/`addPath` with it throws.

```suggestion
            info.This().As<Napi::Object>().DefineProperty(
                Napi::PropertyDescriptor::Value(BRAND_NAME, Napi::External<T>::New(info.Env(), instance), napi_default));
```

Polyfills/Canvas/Source/Path2D.h
```diff
@@ -65,8 +65,15 @@ namespace Babylon::Polyfills::Internal
...
+        // True only for objects actually produced by this class's constructor. Never use
+        // ObjectWrap::Unwrap on an unchecked object; see NativeInstanceRegistry.
+        static bool IsInstance(Napi::Env env, const Napi::Value& value);
+
+        // Returns the wrapped path, or nullptr when `value` is not one of ours.
+        static NativeCanvasPath2D* TryUnwrap(Napi::Env env, const Napi::Value& value);
+
         explicit NativeCanvasPath2D(const Napi::CallbackInfo& info);
-        // virtual ~NativeCanvasPath2D(); // TODO: destructor? empty queue?
+        ~NativeCanvasPath2D();
```
bghgary: `Ubuntu_Clang_QuickJS` segfaults in `rejects a non-Path2D argument to Path2D.addPath`, unchanged from the previous head, so it is not the `napi_unwrap` cause this commit removed. It is Linux-only: `Win32_x64_QuickJS_D3D11` passes, and the macOS and Android QuickJS jobs have no `Unit Tests` step, so a QuickJS run on Windows cannot reproduce it.

This destructor is the strongest lead — it did not exist before this branch, so the QuickJS finalizer path now has work it never had. Not reproduced, so treat it as a starting point rather than the cause.

bkaradzic-microsoft:
The remaining `Ubuntu_Clang_QuickJS` failure is not a Canvas bug — it is a Node-API port bug that this PR is the first thing to trip.

**What the core dump says.** The job uploaded a core this time. `gdb` puts the fault in the *error reporting* path, not in the type check:

```
#0  js_dup                       quickjs.c:1628          <-- SIGSEGV
#4  napi_get_value_string_utf8   js_native_api_quickjs.cc:696
#7  Napi::Error::Message         napi-inl.h:3087
#8  Napi::Error::what            napi-inl.h:3157
#9  ExternalCallback::Callback   js_native_api_quickjs.cc:164
```

The failing callback is `InstanceWrap<NativeCanvasPath2D>::InstanceVoidMethodCallbackWrapper` with `argc=0`, i.e. the `path.addPath()` assertion in the new `rejects a non-Path2D argument to Path2D.addPath` test. The `JSValue` being stringified has `JS_TAG_STRING` and an unaligned, freed pointer.

**Root cause.** The QuickJS port's `napi_throw` returns `napi_pending_exception` after a *successful* throw. node-addon-api reads that as "the throw failed" and re-throws `Error::New(env)`, which consumes the pending exception, so the C++ exception escapes `WrapCallback` with no JS exception set. `ExternalCallback::Callback` then rebuilds the error from `e.what()` after the relevant handle scope has closed — a use-after-free.

This affects every native throw on QuickJS, not just Canvas. I instrumented that catch block locally: **all ~50 native throws** in the unit-test run escape `WrapCallback`. Linux faults; Windows happens to survive reading the freed string, which is why my local QuickJS build stayed green.

It also explains the `InternalError: Uncaught C++ exception: ...` messages I hit earlier in this PR — the real error is being replaced.

**Fix:** BabylonJS/JsRuntimeHost#225. Verified there with an A/B on Linux QuickJS: without the change `expected 'InternalError' to equal 'Error'` (1 failing), with it 213 passing.

So this PR is blocked on JsRuntimeHost#225 landing plus a pin bump, the same way #1835 waited on JsRuntimeHost#223. Everything else here is green (31/32), and the Canvas work itself is verified on Chakra, V8 and QuickJS locally (21/21 gtest, 49 assertions each) with a 305/305 visual sweep.

Polyfills/Canvas/Source/Path2D.cpp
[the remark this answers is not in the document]
bkaradzic-microsoft: Filed the promised follow-up: BabylonJS/JsRuntimeHost#226 covers the `napi_unwrap` contract violations (V8 port dereferences internal field 0 unchecked; QuickJS port falls back to walking the prototype chain).

Chasing the remaining CI failure on this PR turned up a second, separate port bug in the same area — BabylonJS/JsRuntimeHost#225, where the `napi_throw` family returns `napi_pending_exception` after a successful throw. That one makes every native throw on QuickJS escape `WrapCallback` and get rebuilt from `e.what()` after its handle scope has closed, which is the segfault in `Ubuntu_Clang_QuickJS`.

bkaradzic-microsoft:
Confirming the loop is closed: BabylonJS/JsRuntimeHost#225 is merged, the pin here now points at `2390c63c34138a05c06c967d22687bb8e199a340`, and CI is **32/32 green** — including `Ubuntu_Clang_QuickJS`, which is the job that was segfaulting.

No change was needed to the Canvas code for that failure; it was the QuickJS Node-API port reporting failure from a successful `napi_throw`, so the new `Path2D.addPath()` arity check unwound into a use-after-free instead of surfacing a `TypeError`.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
```diff
@@ -0,0 +1,102 @@
...
+    // `instanceof` cannot answer it: a prototype is assignable, so
+    // Object.setPrototypeOf(gradient, Path2D.prototype) makes a gradient pass a Path2D test,
+    // and Object.create(Path2D.prototype) passes with no native object behind it at all.
+    //
+    // Neither can ObjectWrap::Unwrap. Both JsRuntimeHost Node-API ports break the contract
+    // that it fails for an object that was never wrapped: the V8 port dereferences internal
+    // field 0 unconditionally, which access-violates, and the QuickJS port falls back to
+    // walking the prototype chain, which returns some other object's pointer.
+    //
+    // So each instance brands its own JS object with an External holding its address, and a
```
bghgary: `napi_type_tag_object` is the Node-API answer to this, and it is unusable here only because three of the four ports do not implement it. V8 does (`js_native_api_v8.cc:2556`), storing the tag under a **private** key — invisible to script, absent from the prototype chain, not copyable onto another object. Implementing it in the QuickJS, Chakra and JavaScriptCore ports is the fix, and this file then deletes: `Napi::Object::CheckTypeTag` before `Unwrap` rejects both spoofs, with no brand property on the object and no process-wide set or mutex.

JsRuntimeHost#226 alone is not enough. It closes `Object.create(Path2D.prototype)`, which was never wrapped, but a correct `napi_unwrap` still returns `void*` — so a `CanvasGradient` wearing `Path2D.prototype` is genuinely wrapped, unwraps successfully, and is used as a path. Only a tag distinguishes one wrapped type from another.

bghgary (changes requested):

The type check belongs in JsRuntimeHost rather than here — comment inline. The call-site gating, the arity check and the tests all stand.

Not a complete review; holding the smaller points until the mechanism settles.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
bghgary: [Responded by Copilot on behalf of @bghgary]

We're picking up the JsRuntimeHost side — type tags in the QuickJS, Chakra and JavaScriptCore ports, plus the `napi_unwrap` guards from [JsRuntimeHost#226](https://github.com/BabylonJS/JsRuntimeHost/issues/226). Flagging it so you don't start the same work; happy to hand it over if you'd rather take it, since you're already in that code.

This PR then waits on that landing and a pin bump.

bkaradzic-microsoft: Agreed — type tags are the right mechanism, and `NativeInstanceRegistry.h` deletes in favour of `CheckTypeTag` before `Unwrap`.

Your correction on #226 is right, and I'd overstated it: it only closes the never-wrapped `Object.create` case. A genuinely wrapped `CanvasGradient` wearing `Path2D.prototype` unwraps successfully no matter how correct `napi_unwrap` is. My registry separates the two types only because it is instantiated per `T`; the general form of that belongs in the port, not in Canvas.

One thing worth knowing before you start, because it makes this **four ports rather than three**: the V8 implementation is currently dead code.

`js_native_api.h:4` carries a `[BABYLON-NATIVE-ADDITION]`:

```c
#ifndef NAPI_VERSION
#define NAPI_VERSION 5
#endif
```

That shadows upstream's default of 8, so the `#if NAPI_VERSION >= 8` guard beginning at `js_native_api_v8.cc:2555` never compiles. Confirmed against the build rather than by reading: in `napi.lib` (V8, RelWithDebInfo) `napi_type_tag_object` and `napi_check_object_type_tag` are both absent, while `napi_unwrap` and `napi_create_external` are present.

Raising `NAPI_VERSION` alone will not build it. `NAPI_PRIVATE_KEY` is commented out at `js_native_api_v8_internals.h:79`:

```c
// [BABYLON-NATIVE-ADDITION]: Increase perf by using internal field instead of private property
//#define NAPI_PRIVATE_KEY(context)                                      \
//  (v8::Private::New(context->GetIsolate()))
```

and no other definition exists in the tree, so the uses at 2565 and 2595 are unresolved. The internal-field route that comment points at needs a free slot and there isn't one: `SetInternalFieldCount(1)` at `js_native_api_v8.cc:929`, with field 0 already holding the `Reference` (363, 598). So it is restore the private key for tags, go to two internal fields, or carry the tag on the `Reference`.

Worth deciding deliberately either way: `NAPI_VERSION 5` gates every other post-5 API as well, so bumping it is a wider change than type tags alone.

On the handoff — happy for you to take it, and I'll stay off it. If you'd rather split, I'll take the QuickJS port, since I was just in that file for #225.

One question on the interim. Both spoofs are a hard `0xC0000005` reachable from script on master today, and the tagging work is now larger than it looked. Would you like me to land the uncontroversial part of this PR first — the call-site gating, the `addPath` arity check and the tests — which closes the original four holes via `InstanceOf`, and leave the two prototype-spoof crashes to be closed properly by the tag? That keeps the spoof tests in the suite (failing-but-known, or skipped) rather than sitting on a branch. Or I'm equally happy to hold the whole PR until the tag lands, if you'd prefer one clean change.

bkaradzic-microsoft:
Rebased onto `master` (`85355d87`), now 5 commits with no merge. Force-pushed as `a1aab5e1`.

The one conflict was against #1843: `BindFillStyle`/`BindStrokeStyle` moved their style members into `m_state`, and this PR changes the same two lines from `Unwrap` to the checked `TryUnwrap`. Resolved by keeping the check and taking the `m_state` members.

Re-verified after the rebase rather than assuming: 21/21 gtest suites, all the new spoof/gradient tests passing, and the visual sweep still `ran=305 passed=305 failed=0` with "Native Canvas" unchanged at 1.850%.

Still blocked on the type-tag work as discussed above — the rebase is just to keep it mergeable.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
bghgary: [Responded by Copilot on behalf of @bghgary]

Hold the whole PR — one clean change once the tag lands. Nothing is blocked on it, and reaching either spoof takes script that deliberately re-prototypes a canvas object; nothing in Babylon.js does that.

We'll take all four ports.

bkaradzic-microsoft: Understood — holding it. Agreed on the risk call: both spoofs need script that deliberately re-prototypes a canvas object, so there's nothing to race here.

I'll leave the branch rebased and green, and when the tags land I'll swap `NativeInstanceRegistry.h` for `CheckTypeTag`, bump the pin, and re-run. The two spoof tests stay in the suite as the regression cover for the port work.

Marking this a draft in the meantime so it doesn't sit in your review queue. Ping me if you want the QuickJS port after all.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -327,10 +335,19 @@ namespace Babylon::Polyfills::Internal
...
+        // fill() takes either a fillRule string or a Path2D (optionally followed by one).
+        // Anything else is rejected rather than unwrapped: Unwrap does no type checking, so
+        // an arbitrary object would be reinterpreted as a NativeCanvasPath2D.
+        const NativeCanvasPath2D* path = nullptr;
+        if (info.Length() >= 1 && !info[0].IsUndefined())
+        {
+            path = NativeCanvasPath2D::TryUnwrap(info.Env(), info[0]);
+            if (path == nullptr && !info[0].IsString())
+            {
+                throw Napi::TypeError::New(info.Env(), "Context2D.fill: the first argument is neither a Path2D nor a fill rule.");
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Implemented in b99b470f, after rebasing onto master at 2caf5447.

`fill()` now distinguishes a real `Path2D` from the fill-rule overload by converting the rule with `ToString()` and validating the resulting enum value (`nonzero` / `evenodd`). This accepts boxed strings and other string-convertible values in either fill-rule position while still rejecting invalid values. Added regression coverage for boxed/custom-converted rules and invalid rules.

Validation: QuickJS Linux `UnitTests` target built with `-j2`; `JavaScript.All` passed (80 passing).

Polyfills/Canvas/Source/NativeInstanceRegistry.h
bkaradzic-microsoft: Rebased onto current master (2caf5447) and pushed b99b470f. The review feedback about fill-rule conversion is implemented and covered by the QuickJS JavaScript unit suite (80 passing).

The requested type-tag replacement remains blocked exactly as discussed: BabylonNative currently pins JsRuntimeHost f47991dd; issue BabylonJS/JsRuntimeHost#226 is still open, `NAPI_VERSION` remains 5, and the pinned QuickJS, Chakra, and JavaScriptCore ports still have no `napi_type_tag_object` / `napi_check_object_type_tag` implementations. I preserved the existing non-enumerable/non-writable/non-configurable registry fallback and kept this PR in draft rather than inventing a second port-level tagging design here.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
```diff
@@ -0,0 +1,102 @@
...
+    public:
+        // Call at the very end of the constructor: one that throws never reaches the
+        // destructor, which would leave a dangling address registered.
+        static void Add(const Napi::CallbackInfo& info, T* instance)
+        {
+            info.This().As<Napi::Object>().Set(BRAND_NAME, Napi::External<T>::New(info.Env(), instance));
+
+            const std::scoped_lock lock{Mutex()};
+            Instances().insert(instance);
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Already addressed in the current branch: `NativeInstanceRegistry::Add` defines `__nativeInstance` with `Object.defineProperty` and a descriptor containing only `value`, so `enumerable`, `writable`, and `configurable` are all false. It no longer uses `.Set(...)` for the brand property. This addresses the descriptor concern, not the larger mechanism question: the PR remains draft pending JsRuntimeHost type tags, as agreed with Gary; those will replace this temporary registry rather than adding another branding mechanism here.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
```diff
@@ -0,0 +1,102 @@
...
+            T* const candidate = brand.As<Napi::External<T>>().Data();
+
+            const std::scoped_lock lock{Mutex()};
+            return Instances().count(candidate) != 0 ? candidate : nullptr;
+        }
+
+    private:
+        // Shared by every T: a brand read as the wrong type is still rejected, because each T
+        // registers into its own set.
+        static constexpr const char* BRAND_NAME{"__nativeInstance"};
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Already addressed in the current branch: `NativeInstanceRegistry::Add` defines `__nativeInstance` with `Object.defineProperty` and a descriptor containing only `value`, so `enumerable`, `writable`, and `configurable` are all false. It no longer uses `.Set(...)` for the brand property. This addresses the descriptor concern, not the larger mechanism question: the PR remains draft pending JsRuntimeHost type tags, as agreed with Gary; those will replace this temporary registry rather than adding another branding mechanism here.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -183,7 +183,11 @@ namespace Babylon::Polyfills::Internal
         }
         else if (std::holds_alternative<GradientStyle>(m_state.fillStyle))
         {
-            CanvasGradient* gradient = CanvasGradient::Unwrap(std::get<GradientStyle>(m_state.fillStyle)->Value());
+            CanvasGradient* gradient = CanvasGradient::TryUnwrap(info.Env(), std::get<GradientStyle>(m_state.fillStyle)->Value());
+            if (gradient == nullptr)
+            {
+                throw Napi::Error::New(info.Env(), "Fillstyle is not a color string or a gradient.");
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Already addressed in the current branch: the diagnostics consistently use the JavaScript property names, `fillStyle must be a string or CanvasGradient.` and `strokeStyle must be a string or CanvasGradient.` Both the failed-unwrapping and wrong-value-type branches use the corrected spelling. No additional change is needed for this previously resolved comment.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -205,7 +209,11 @@ namespace Babylon::Polyfills::Internal
         }
         else if (std::holds_alternative<GradientStyle>(m_state.strokeStyle))
         {
-            CanvasGradient* gradient = CanvasGradient::Unwrap(std::get<GradientStyle>(m_state.strokeStyle)->Value());
+            CanvasGradient* gradient = CanvasGradient::TryUnwrap(info.Env(), std::get<GradientStyle>(m_state.strokeStyle)->Value());
+            if (gradient == nullptr)
+            {
+                throw Napi::Error::New(info.Env(), "Strokestyle is not a color string or a gradient.");
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Already addressed in the current branch: the diagnostics consistently use the JavaScript property names, `fillStyle must be a string or CanvasGradient.` and `strokeStyle must be a string or CanvasGradient.` Both the failed-unwrapping and wrong-value-type branches use the corrected spelling. No additional change is needed for this previously resolved comment.

PR #1855, opened 2026-08-27
Canvas: text metrics, multi-GUI filters, drawImage(canvas), toDataURL

bkaradzic-microsoft:
## Summary

Canvas2D text metrics, multi-GUI filter lifetime, Canvas-source `drawImage`, and PNG readback.

| Area | Change |
|------|--------|
| Text metrics | Em-scaled font sizes, advance and glyph ink bounds, and consistent no-font fallback properties. Ink bounds exclude SDF, interpolation-border, and blur padding. |
| Multi-GUI filters | Reference-count shared blur programs and uniforms across multiple ADTs. |
| `drawImage(canvas)` | Snapshot rendered source pixels with GPU readback. Apply floating-point source/destination rectangles, cropping, and negative-extent normalization to the GPU draw. Retain temporary uploads until queued NanoVG draws are flushed. |
| `toDataURL` | Encode GPU readback as PNG using `bimg::imageWritePng`; empty, case-variant, and unsupported media types fall back to PNG. |
| Regression coverage | Observe destination GPU pixels through a second canvas, rather than asserting only the CPU mirror. Regenerate the committed JavaScript test bundle. |

## Known visual-baseline differences

The existing `native-canvas.png` is a Native regression baseline, not browser parity. It also captures repeating rather than clamped gradient colors, a dark projected-texture perimeter, and Native-only Droid Sans input. These differences and the matched-input follow-up are documented in [`Polyfills/Canvas/Readme.md`](https://github.com/bkaradzic-microsoft/BabylonNative/blob/0be41b62ad848d0f7df5102247ffa8800aa9c5cf/Polyfills/Canvas/Readme.md).

This review update does not regenerate golden images. Future golden changes must isolate the intended fix against Babylon.js with matching font bytes and scene inputs.

## Validation of the review update

- Windows x64 D3D11/Chakra `UnitTests` build succeeded.
- `CanvasTextMetrics.*:JavaScript.All`: 2/2 GTests passed. JavaScript: 57 passing, 27 pending, no failures.
- Five additional `JavaScript.All` repetitions passed with the same counts. The built JavaScript asset exactly matches the committed generated bundle.
- Font-padding regression: 72 scenarios covering both coordinate origins, two sizes, three glyphs, and blur 0/4/20/40. Standalone SDF and non-SDF variants pass; the original implementation fails the regression.

The existing JavaScript webpack command omits the TypeScript Babel preset. For this bundle regeneration, the existing webpack config was used with the already-installed `@babel/preset-typescript` appended to its loader options in memory, without changing dependency manifests. Native execution used the resulting committed bundle.


Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -229,7 +254,52 @@ namespace Babylon::Polyfills::Internal
...
+            bx::MemoryWriter writer{&memoryBlock};
+            bx::Error err{};
+            bimg::imageWritePng(&writer, width, height, width * 4, rgba.data(), bimg::TextureFormat::RGBA8, false, &err);
+            if (!err.isOk() || memoryBlock.getSize() == 0)
+            {
+                throw Napi::Error::New(info.Env(), "Canvas.toDataURL: PNG encode failed.");
+            }
+
+            const char* pngBytes = static_cast<const char*>(memoryBlock.more(0));
+            const size_t pngSize = memoryBlock.getSize();
```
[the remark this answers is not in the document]
bkaradzic-microsoft: bx::MemoryBlock::more(0) returns m_data without growing when size==0 (bx/inline/readerwriter.inl). That is the intended way to get the buffer start after a write. Left a one-line note in code.

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -229,7 +254,52 @@ namespace Babylon::Polyfills::Internal
...
+            if (info.Length() >= 1 && info[0].IsString())
+            {
+                type = info[0].As<Napi::String>().Utf8Value();
+            }
+            if (type != "image/png" && type != "image/PNG")
+            {
+                // Still produce PNG bytes under the requested type prefix only when png;
+                // unknown types get a clear error rather than a lying data URL.
+                throw Napi::TypeError::New(info.Env(), "Canvas.toDataURL: only image/png is supported on Native.");
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed — comment was wrong. Updated: only image/png is supported; other types throw TypeError.

Polyfills/Canvas/Source/nanovg/nanovg_filterstack.cpp
```diff
@@ -48,6 +48,11 @@ nanovg_filterstack::nanovg_filterstack()
 
 void nanovg_filterstack::InitBgfx()
 {
+    if (s_bgfxRefCount++ > 0)
+    {
+        return;
+    }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Canvas/NVG init and dispose run on the JS/graphics thread only (same as the rest of the Canvas polyfill). A mutex here would not buy safety against real cross-thread bgfx use, which is already disallowed. Leaving the plain int refcount.

Polyfills/Canvas/Source/NativeInstanceRegistry.h
```diff
@@ -0,0 +1,102 @@
...
+    public:
+        // Call at the very end of the constructor: one that throws never reaches the
+        // destructor, which would leave a dangling address registered.
+        static void Add(const Napi::CallbackInfo& info, T* instance)
+        {
+            info.This().As<Napi::Object>().Set(BRAND_NAME, Napi::External<T>::New(info.Env(), instance));
+
+            const std::scoped_lock lock{Mutex()};
+            Instances().insert(instance);
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in the branding base (#1844): brand is now DefineProperty + napi_default (non-enumerable). This PR is rebased onto that tip.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -183,7 +187,11 @@ namespace Babylon::Polyfills::Internal
         }
         else if (std::holds_alternative<GradientStyle>(m_state.fillStyle))
         {
-            CanvasGradient* gradient = CanvasGradient::Unwrap(std::get<GradientStyle>(m_state.fillStyle)->Value());
+            CanvasGradient* gradient = CanvasGradient::TryUnwrap(info.Env(), std::get<GradientStyle>(m_state.fillStyle)->Value());
+            if (gradient == nullptr)
+            {
+                throw Napi::Error::New(info.Env(), "Fillstyle is not a color string or a gradient.");
+            }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed: messages are now fillStyle / strokeStyle and throw TypeError.

Polyfills/Canvas/Source/nanovg/fontstash.h
```diff
@@ -196,9 +196,15 @@ void fons__tt_getFontVMetrics(FONSttFontImpl *font, int *ascent, int *descent, i
...
+// CSS/canvas font-size is em size (map em square → pixels), not ascent+descent.
+float fons__tt_getEmUnits(FONSttFontImpl *font)
+{
+	return (float)font->font->units_per_EM;
+}
+
 float fons__tt_getPixelHeightScale(FONSttFontImpl *font, float size)
 {
-	return size / (font->font->ascender - font->font->descender);
+	return size / fons__tt_getEmUnits(font);
```
bghgary: Please use the em mapping when choosing FreeType's rasterization size too. The new advance scale is `size / units_per_EM`, but `FT_Set_Pixel_Sizes` still uses the old ascent/descent mapping, so glyph advances and bitmaps use different size bases when `FONS_USE_FREETYPE` is enabled.

Polyfills/Canvas/Source/Canvas.cpp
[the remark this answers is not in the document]
bghgary: Changing the comment does not fix `toDataURL`'s contract: an empty, case-variant, or unsupported media type must encode PNG rather than throw.

Polyfills/Canvas/Source/nanovg/fontstash.h
```diff
@@ -1511,6 +1528,13 @@ void fonsDrawDebug(FONScontext* stash, float x, float y)
 	fons__flush(stash);
 }
 
+// Atlas SDF padding; subtracted from reported ink bounds.
+#ifdef FONS_SDF_PADDING
+#define FONS_TEXT_BOUNDS_PADDING ((float)FONS_SDF_PADDING)
+#else
+#define FONS_TEXT_BOUNDS_PADDING 0.0f
+#endif
```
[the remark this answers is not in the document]
bghgary: Please include the blur padding in that follow-up and land it immediately after this PR. `fons__getGlyph` adds `iblur + 2` pixels while `fons__getQuad` removes one, so the current bounds retain the existing one-pixel border plus `iblur` additional pixels on every edge.

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -229,6 +243,48 @@ namespace Babylon::Polyfills::Internal
...
+        {
+            throw Napi::TypeError::New(info.Env(), "Canvas.toDataURL: only image/png is supported on Native.");
+        }
+
+        const uint32_t width = m_width;
+        const uint32_t height = m_height;
+        std::vector<uint8_t> rgba(static_cast<size_t>(width) * height * 4, 0);
+        if (m_context != nullptr && width > 0 && height > 0)
+        {
+            m_context->ReadPixels(0, 0, width, height, rgba.data());
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in 1e8738a1: `toDataURL` now calls `Context::CaptureRGBA()`, which flushes pending NanoVG draws, blits the canvas RT to a `BGFX_TEXTURE_READ_BACK` texture, and blocks on `ReadTextureAsync` (completed via mid-frame flushes). The PNG is encoded from that GPU readback, not the CPU mirror.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1211,6 +1208,32 @@ namespace Babylon::Polyfills::Internal
...
+            if (width == 0 || height == 0)
+            {
+                return;
+            }
+
+            Context* const srcContext = srcCanvas->GetBoundContext();
+            std::vector<uint8_t> rgba(static_cast<size_t>(width) * height * 4, 0);
+            if (srcContext != nullptr)
+            {
+                srcContext->ReadPixels(0, 0, width, height, rgba.data());
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in 1e8738a1: `drawImage(canvas)` now sources pixels from `srcContext->CaptureRGBA()` (flush + GPU readback) so fillRect/text/paths are included, then uploads that buffer as the NVG image.

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -49,6 +58,7 @@ namespace Babylon::Polyfills::Internal
 
     NativeCanvas::~NativeCanvas()
     {
+        m_context = nullptr;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in 1e8738a1: `NativeCanvas::~NativeCanvas` calls `Context::DetachCanvas()` before clearing `m_context`, and `Context::~Context` also nulls `m_canvas` after unbinding. Finalizer order is safe either way.

Polyfills/Canvas/Source/nanovg/fontstash.h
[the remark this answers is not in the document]
bkaradzic-microsoft: Acknowledged — leaving the +1px fontstash border adjustment for a follow-up so this PR stays focused on the readback/dtor fixes. Happy to take it in a small metrics follow-up if you want it in this PR instead.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -806,67 +831,203 @@ namespace Babylon::Polyfills::Internal
...
+        const uint32_t width = m_canvas != nullptr ? m_canvas->GetWidth() : 0;
+        const uint32_t height = m_canvas != nullptr ? m_canvas->GetHeight() : 0;
+        std::vector<uint8_t> rgba(static_cast<size_t>(width) * height * 4, 0);
+        if (m_canvas == nullptr || width == 0 || height == 0)
+        {
+            return rgba;
+        }
+
+        // Push pending NanoVG draws into the canvas RT first.
+        FlushCore();
```
bghgary: Acquire the `FrameCompletionScope` before calling `FlushCore()` and keep it through the readback. The unit-test path can have a non-null encoder with no JavaScript-held scope, so `FlushCore()` skips its own scope and can submit NanoVG work while the render thread ends that encoder.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -806,67 +831,203 @@ namespace Babylon::Polyfills::Internal
...
+
+        // Keep the CPU mirror in sync so getImageData after toDataURL/drawImage sees GPU content.
+        EnsureCpuBuffer();
+        if (!m_cpuPixels.empty() && m_cpuPixels.size() == rgba.size())
         {
-            throw Napi::Error::New(info.Env(), ex.what());
+            m_cpuPixels = rgba;
         }
+
+        return rgba;
```
bghgary: Please convert this GPU readback from premultiplied to straight RGBA before returning it. `toDataURL` writes these bytes unchanged to PNG, while `drawImage(canvas)` passes them to `nvgCreateImageRGBA` without `NVG_IMAGE_PREMULTIPLIED`, so semi-transparent colors are darkened. Please add a semi-transparent pixel assertion; the current opaque-red test cannot detect this.

Polyfills/Canvas/Source/nanovg/fontstash.h
```diff
@@ -304,9 +310,16 @@ void fons__tt_getFontVMetrics(FONSttFontImpl *font, int *ascent, int *descent, i
...
+float fons__tt_getEmUnits(FONSttFontImpl *font)
+{
+	const float scale = stbtt_ScaleForMappingEmToPixels(&font->font, 1.0f);
+	return scale > 0.0f ? 1.0f / scale : 0.0f;
+}
+
 float fons__tt_getPixelHeightScale(FONSttFontImpl *font, float size)
 {
-	return stbtt_ScaleForPixelHeight(&font->font, size);
+	return stbtt_ScaleForMappingEmToPixels(&font->font, size);
```
bghgary: Please update `Apps/Playground/ReferenceImages/native-canvas.png` for this intended em-based font-size change. The current head reproducibly changes only the `BabylonNative` and `Gradient Text!` glyph regions, but this PR leaves the old reference in place, so visual CI fails.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -806,67 +831,203 @@ namespace Babylon::Polyfills::Internal
...
             }
         }
-        catch (const std::exception& ex)
+
+        // Keep the CPU mirror in sync so getImageData after toDataURL/drawImage sees GPU content.
+        EnsureCpuBuffer();
+        if (!m_cpuPixels.empty() && m_cpuPixels.size() == rgba.size())
         {
-            throw Napi::Error::New(info.Env(), ex.what());
+            m_cpuPixels = rgba;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed — NanoVG’s RT is premultiplied (ONE / INV_SRC_ALPHA). CaptureRGBA now unpremultiplies after readback/Y-flip so toDataURL, getImageData (via the CPU mirror refresh), and drawImage(canvas) all see straight alpha. drawImage still uploads with nvgCreateImageRGBA without NVG_IMAGE_PREMULTIPLIED, so NanoVG premultiplies once on the way in.

Polyfills/Canvas/Source/nanovg/fontstash.h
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed for the FONS_USE_FREETYPE path: FT_Set_Pixel_Sizes now uses the em `size` directly (same units as fons__tt_getPixelHeightScale). The old units_per_EM/(ascender-descender) scale was the line-height path and desynced raster from advances. BN’s default build uses the stb_truetype path (already on ScaleForMappingEm); this keeps FreeType builds consistent.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -495,7 +495,23 @@ describe("Canvas2D", function () {
...
+    it("accepts another Canvas as drawImage source and supports toDataURL png", function () {
+      const srcCanvas = new _native.Canvas();
+      srcCanvas.width = 32;
+      srcCanvas.height = 32;
+      const srcCtx = srcCanvas.getContext("2d");
+      srcCtx.fillStyle = "#ff0000";
+      srcCtx.fillRect(0, 0, 32, 32);
+
+      const dstCtx = createContext();
+      expect(function () { dstCtx.drawImage(srcCanvas, 0, 0); }).to.not.throw();
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Updated the test: source fill is rgba(255,0,0,0.5), and after drawImage(canvas) we assert destination getImageData channels (R high, G/B low, A ~128). That fails on transparent/stale readback and on double-premultiply.

Polyfills/Canvas/Source/MeasureText.cpp
```diff
@@ -19,20 +19,18 @@ namespace Babylon::Polyfills::Internal
...
-        // slots (e.g. when called before the text atlas has been populated), and uninitialized
-        // stack values turn into garbage widths the size of millions of pixels, which Babylon
-        // GUI then propagates into a wildly negative fillText x coordinate, putting text off-
-        // screen.
-        float bounds[4] = {0, 0, 0, 0};
-        nvgTextBounds(context->GetNVGContext(), 0, 0, text.c_str(), nullptr, bounds);
+        // Zero-init: nvgTextBounds may leave slots untouched when the atlas is empty.
+                float bounds[4] = {0, 0, 0, 0};
+        // width = advance (not ink box).
+                const float advance = nvgTextBounds(context->GetNVGContext(), 0, 0, text.c_str(), nullptr, bounds);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Agreed — nvgTextBounds intentionally replaces ymin/ymax with fonsLineBounds for layout. Added nvgTextBoundsInk (skips the line-box overwrite) and MeasureText now uses it, plus actualBoundingBoxAscent/Descent from the ink extents. nvgTextBounds is unchanged for existing callers.

bkaradzic-microsoft:
CI was red across the board after merging master (#1857): ``Context.cpp`` still called the old 14-arg ``encoder->blit``. Updated ``CaptureRGBA`` to the ``TextureRegion`` pair API (same pattern as NativeEngine/FrameBuffer).

Pushed in the latest commit.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -673,17 +688,10 @@ namespace Babylon::Polyfills::Internal
...
-        // negative x and clip the text off-canvas. Arial-ish synthesised metrics keep the
-        // centering on-canvas, while the actual FillText still substitutes our loaded font.
-        const bool familyAvailable = !m_state.font.Familiy().empty()
-            && m_fonts.find(m_state.font.Familiy()) != m_fonts.end();
-
-        if (!familyAvailable && m_state.font.Size() > 0.f)
+        // Measure against the same face FillText will bind.
+        const bool fontBound = SetFontFaceId();
+        // No face available: synthesize Arial-ish metrics so callers still get a finite width.
+        if (!fontBound && m_state.font.Size() > 0.f)
```
bghgary: Please also set `actualBoundingBoxAscent` and `actualBoundingBoxDescent` from the synthesized values. This fallback currently returns a different `TextMetrics` shape whenever no font face is available.

Apps/UnitTests/JavaScript/dist/tests.javaScript.all.js
```diff
@@ -28726,6 +28726,22 @@ describe("Canvas2D", function () {
     (0,chai__WEBPACK_IMPORTED_MODULE_3__.expect)(data.width).to.equal(4);
     (0,chai__WEBPACK_IMPORTED_MODULE_3__.expect)(data.height).to.equal(3);
   });
+
+  it("accepts another Canvas as drawImage source and supports toDataURL png", function () {
+    var srcCanvas = new _native.Canvas();
+    srcCanvas.width = 32;
+    srcCanvas.height = 32;
+    var srcCtx = srcCanvas.getContext("2d");
+    srcCtx.fillStyle = "#ff0000";
```
bghgary: Please regenerate this bundle from the updated TypeScript. UnitTests executes the committed `dist` file, which still uses opaque red and never asserts the destination pixels, so the readback/unpremultiply regression test does not run.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1211,6 +1389,40 @@ namespace Babylon::Polyfills::Internal
...
+                }
+                catch (const std::exception& ex)
+                {
+                    throw Napi::Error::New(info.Env(), ex.what());
+                }
+            }
+
+            const int imageIndex = nvgCreateImageRGBA(*m_nvg, static_cast<int>(width), static_cast<int>(height), 0, rgba.data());
+            DrawImageCommon(info, imageIndex, rgba.data(), width, height);
+            nvgDeleteImage(*m_nvg, imageIndex);
```
bghgary: Keep this image alive until after `nvgEndFrame()` flushes the queued draw. `nvgFill()` stores its ID and resolves it during that later flush, so deleting it here makes the GPU path bind `texMissing`; the test's `getImageData` assertion only reads the CPU mirror.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -806,67 +831,220 @@ namespace Babylon::Polyfills::Internal
...
+        const uint32_t width = m_canvas != nullptr ? m_canvas->GetWidth() : 0;
+        const uint32_t height = m_canvas != nullptr ? m_canvas->GetHeight() : 0;
+        std::vector<uint8_t> rgba(static_cast<size_t>(width) * height * 4, 0);
+        if (m_canvas == nullptr || width == 0 || height == 0)
+        {
+            return rgba;
+        }
+
+        // Push pending NanoVG draws into the canvas RT first.
+        FlushCore();
```
bghgary, line 909: Acquire the `FrameCompletionScope` before calling `FlushCore()` and keep it through the readback. The unit-test path can have a non-null encoder with no JavaScript-held scope, so `FlushCore()` skips its own scope and can submit NanoVG work while the render thread ends that encoder.

Polyfills/Canvas/Source/Context.cpp
bghgary, line 696: Please also set `actualBoundingBoxAscent` and `actualBoundingBoxDescent` from the synthesized values. This fallback currently returns a different `TextMetrics` shape whenever no font face is available.

Apps/UnitTests/JavaScript/dist/tests.javaScript.all.js
bghgary: Please regenerate this bundle from the updated TypeScript. UnitTests executes the committed `dist` file, which still uses opaque red and never asserts the destination pixels, so the readback/unpremultiply regression test does not run.

Polyfills/Canvas/Source/Context.cpp
bghgary: Keep this image alive until after `nvgEndFrame()` flushes the queued draw. `nvgFill()` stores its ID and resolves it during that later flush, so deleting it here makes the GPU path bind `texMissing`; the test's `getImageData` assertion only reads the CPU mirror.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1211,6 +1389,40 @@ namespace Babylon::Polyfills::Internal
...
+                    rgba = srcContext->CaptureRGBA();
+                }
+                catch (const std::exception& ex)
+                {
+                    throw Napi::Error::New(info.Env(), ex.what());
+                }
+            }
+
+            const int imageIndex = nvgCreateImageRGBA(*m_nvg, static_cast<int>(width), static_cast<int>(height), 0, rgba.data());
+            DrawImageCommon(info, imageIndex, rgba.data(), width, height);
```
bghgary: The new Canvas-source branch delegates to `DrawImageCommon`, but that helper does not preserve Canvas overload semantics: it truncates every coordinate and extent with `Int32Value()`, treats negative source or destination extents as a no-op instead of growing the rectangle in the opposite direction, and the nine-argument GPU path ignores the source rectangle entirely. The CPU mirror can still make positive integer crops look correct through `getImageData()`. Please apply the full source and destination rectangles to the GPU path using floating-point values.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -495,7 +495,36 @@ describe("Canvas2D", function () {
...
+      const dstCtx = dstCanvas.getContext("2d");
+      expect(function () { dstCtx.drawImage(srcCanvas, 0, 0); }).to.not.throw();
+
+      const url = srcCanvas.toDataURL("image/png");
+      expect(url.indexOf("data:image/png;base64,")).to.equal(0);
+      expect(url.length).to.be.greaterThan(32);
+
+      // Sample destination after drawImage(canvas) — must see straight-alpha red, not
+      // transparent black and not double-premultiplied (dark) red.
+      const sample = dstCtx.getImageData(16, 16, 1, 1).data;
```
bghgary: `getImageData()` reads the CPU mirror that `DrawImageCommon` writes directly, so these assertions never observe the NanoVG draw. Please sample the destination through a second canvas or another GPU readback path before asserting pixels, so the test catches upload, crop, alpha, and image-lifetime failures.

Polyfills/Canvas/Source/MeasureText.cpp
```diff
@@ -19,24 +19,28 @@ namespace Babylon::Polyfills::Internal
...
     Napi::Value MeasureText::CreateInstance(Napi::Env env, Context* context, const std::string& text)
     {
-        // IMPORTANT: zero-initialize. nvgTextBounds / nvgTextMetrics do not always fill these
-        // slots (e.g. when called before the text atlas has been populated), and uninitialized
-        // stack values turn into garbage widths the size of millions of pixels, which Babylon
-        // GUI then propagates into a wildly negative fillText x coordinate, putting text off-
-        // screen.
-        float bounds[4] = {0, 0, 0, 0};
-        nvgTextBounds(context->GetNVGContext(), 0, 0, text.c_str(), nullptr, bounds);
+                // Ink extents (not the full line box): nvgTextBounds replaces ymin/ymax with
```
bghgary: Please normalize indentation throughout the added C++ code. Besides this function, extra indentation remains around `ForceMidFrameFlush` in `DeviceContext.h`, `DeviceContext.cpp`, `DeviceImpl.h`, and `DeviceImpl.cpp`; `SetBlitViewId` in `Canvas.h`; the destructor and readback additions in `Context.cpp` and `Context.h`; and initialization/disposal in `nanovg_filterstack.cpp`.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -495,7 +495,36 @@ describe("Canvas2D", function () {
     expect(data.width).to.equal(4);
     expect(data.height).to.equal(3);
   });
-});
+
+    it("accepts another Canvas as drawImage source and supports toDataURL png", function () {
```
bghgary: Please align this `it` block with the neighboring tests. The block, its body, and its closing brace are all shifted two spaces to the right.

Polyfills/Canvas/Source/nanovg/fontstash.h
```diff
@@ -1551,16 +1580,33 @@ float fonsTextBounds(FONScontext* stash,
 			continue;
 		glyph = fons__getGlyph(stash, font, codepoint, isize, iblur, FONS_GLYPH_BITMAP_OPTIONAL);
 		if (glyph != NULL) {
+			float qx0, qx1, qy0, qy1;
 			fons__getQuad(stash, font, prevGlyphIndex, glyph, scale, state->spacing, &x, &y, &q);
-			if (q.x0 < minx) minx = q.x0;
-			if (q.x1 > maxx) maxx = q.x1;
+			// Drop atlas SDF padding from ink extents.
+						qx0 = q.x0 + FONS_TEXT_BOUNDS_PADDING;
```
bghgary: Please normalize indentation throughout the changed fontstash blocks, including `fons__tt_getFontVMetrics` in `fonsAddFontMem`, `minx = maxx = x`, and this `qx0` assignment.

Apps/Playground/ReferenceImages/native-canvas.png
bghgary, line 1: This golden also captures unrelated output changes: the head repeats the linear gradient where Babylon.js clamps its final color, adds a new dark perimeter around the projected texture, and uses Droid Sans only on the Native path. The em-size correction can land incrementally, but please document these known divergences and their follow-up before accepting the new baseline. Future golden updates should isolate intended changes against Babylon.js under matched inputs rather than treating current Native output as the expected result.

bghgary (changes requested):

Concerns inline.

Polyfills/Canvas/Source/Canvas.cpp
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. PNG is our only encoder, so the requested media type no longer causes a rejection: omitted, empty, case-variant, and unsupported types all return PNG. Added regressions for these cases, including `IMAGE/PNG` and `image/jpeg`.

Polyfills/Canvas/Source/nanovg/fontstash.h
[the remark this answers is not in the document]
bkaradzic-microsoft: Included in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf rather than leaving the correction to a later follow-up. Ink bounds now subtract the SDF padding plus `glyph->blur + 1` on each edge, using the glyph's clamped blur value. The native regression compares against the unpadded stb glyph bounds for both coordinate origins, two font sizes, and blur 0/4/20/40. It fails against the previous implementation and passes with the fix; standalone checks also cover SDF and non-SDF builds.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. `CaptureRGBA` acquires its frame-completion scope before `FlushCore()` and retains it through readback and cleanup, including when an encoder already exists without a JS-held scope.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. The no-font fallback now sets `actualBoundingBoxAscent` and `actualBoundingBoxDescent` from the same synthesized ascent/descent values. Added a regression for the missing-font case.

Apps/UnitTests/JavaScript/dist/tests.javaScript.all.js
bkaradzic-microsoft: Regenerated the committed bundle in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. It now contains the translucent-source GPU-readback assertions and the rectangle/lifetime regressions from the TypeScript source; the native run uses the copied committed `dist` asset, not a separate test script.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. Temporary Canvas/ImageData image IDs are retained by the context until after `nvgEndFrame()` has resolved and submitted their queued draws, with cleanup on context disposal as well. Applied the same lifetime correction to `putImageData`. The regressions observe destination GPU pixels, including multiple queued snapshots of a changing source.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. All three overloads use floating-point geometry. Negative source/destination extents grow in the opposite direction without flipping; out-of-bounds source rectangles clip the destination proportionally. The GPU paint maps the full image into the requested source/destination rectangles, while preserving transform, clipping, and alpha. The fractional-crop regressions use samples analytically separated from both integer-truncation and ignored-crop results, rather than relying on the CPU mirror.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bkaradzic-microsoft: Agreed; the old assertion could pass without a correct GPU draw. 0be41b62ad848d0f7df5102247ffa8800aa9c5cf adds a probe canvas: `probe.drawImage(destination, ...)` captures the destination framebuffer first, and only then do the assertions read the probe's mirror. This exercises the actual destination upload/render/readback path for alpha, fractional crops, negative extents, source clipping, and deferred image lifetime.

Polyfills/Canvas/Source/MeasureText.cpp
bkaradzic-microsoft: Normalized in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf, including `MeasureText`, the four `ForceMidFrameFlush` declaration/definition files, `Canvas.h`, the `Context` destructor/readback additions, and filter-stack initialization/disposal.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bkaradzic-microsoft: Fixed in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. The replacement GPU-readback tests are aligned with their neighboring tests, and the committed bundle was regenerated.

Polyfills/Canvas/Source/nanovg/fontstash.h
bkaradzic-microsoft: Fixed in 0be41b62ad848d0f7df5102247ffa8800aa9c5cf. The changed fontstash blocks now use the surrounding tab indentation, including `fons__tt_getFontVMetrics`, bounds initialization, and the quad-edge assignments.

Apps/Playground/ReferenceImages/native-canvas.png
bkaradzic-microsoft: Documented in `Polyfills/Canvas/Readme.md` (0be41b62ad848d0f7df5102247ffa8800aa9c5cf) and the PR description: the repeated rather than clamped gradient, dark projected-texture perimeter, and Native-only Droid Sans are known divergences, not evidence of browser parity. The follow-up explicitly requires isolating gradient extension and edge sampling/alpha, then comparing against Babylon.js with matching font bytes, dimensions, device-pixel ratio, and scene inputs before updating references. This update does not regenerate or change the golden.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -806,67 +834,218 @@ namespace Babylon::Polyfills::Internal
...
+            m_graphicsContext.ForceMidFrameFlush();
+            {
+                std::unique_lock lock{mutex};
+                if (done)
+                {
+                    break;
+                }
+                // Yield briefly so a render thread that is between Finish/Start
+                // (unit-test pump) can enter Finish and service flushRequested.
+                cv.wait_for(lock, std::chrono::milliseconds{1}, [&] { return done; });
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in ba8ec747. `ForceMidFrameFlush()` now reports whether the render thread serviced the request, and `CaptureRGBA()` aborts with an explicit error instead of polling forever when it cannot. The pending pixel buffer and texture are owned by shared continuation state, so aborting cannot leave bgfx writing into dead stack storage; cleanup occurs after completion.

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
```diff
@@ -1465,3 +1516,26 @@ bgfx::TextureHandle nvglImageHandle(NVGcontext* _ctx, int32_t _image)
...
+        return 0;
+    }
+
+    tex->width = _width;
+    tex->height = _height;
+    tex->type = NVG_TEXTURE_RGBA;
+    // Do not destroy the caller's texture when nvgDeleteImage runs.
+    tex->flags = _flags | NVG_IMAGE_NODELETE;
+    tex->id = _handle;
+    return static_cast<int>(tex->id.idx);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in ba8ec747. NanoVG images now receive a nonzero logical image ID from the context's existing `textureId` counter, independent of the bgfx texture handle index. Lookup, deletion, and `nvglImageHandle()` all resolve through that logical ID, so bgfx handle 0 is valid and no longer aliases NanoVG's failure sentinel.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1203,14 +1409,56 @@ namespace Babylon::Polyfills::Internal
...
             return;
 #else
             throw Napi::Error::New(info.Env(), "drawImage: image loading disabled in this build.");
 #endif
         }
 
+        // drawImage(canvas): InstanceOf gate then Unwrap (real canvases only; deliberate
+        // prototype spoofs are out of scope — see #1844 / JsRuntimeHost type tags).
+        const auto canvasCtorVal = JsRuntime::NativeObject::GetFromJavaScript(info.Env()).Get("Canvas");
+        if (canvasCtorVal.IsFunction() && imageObj.InstanceOf(canvasCtorVal.As<Napi::Function>()))
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in ba8ec747. The Canvas `InstanceOf`/unwrap path now runs before structural ImageBitmap detection. I added a GPU regression that attaches a typed-array `data` own property to a real Canvas and verifies `drawImage(canvas)` still renders the Canvas source.

Polyfills/Canvas/Source/MeasureText.cpp
```diff
@@ -19,21 +19,25 @@ namespace Babylon::Polyfills::Internal
...
+        // fonsLineBounds, which would make height/actualBoundingBox* the em line box.
         float bounds[4] = {0, 0, 0, 0};
-        nvgTextBounds(context->GetNVGContext(), 0, 0, text.c_str(), nullptr, bounds);
+        const float advance = nvgTextBoundsInk(context->GetNVGContext(), 0, 0, text.c_str(), nullptr, bounds);
         float textMetrics[3] = {0, 0, 0};
         nvgTextMetrics(context->GetNVGContext(), &textMetrics[0], &textMetrics[1], &textMetrics[2]);
 
+        // CSS TextMetrics: positive ascent is above the alignment baseline (y=0 here).
+        const float inkAscent = bounds[1] < 0.f ? -bounds[1] : 0.f;
+        const float inkDescent = bounds[3] > 0.f ? bounds[3] : 0.f;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in ba8ec747. `actualBoundingBoxAscent` is now `-bounds[1]` and `actualBoundingBoxDescent` is `bounds[3]` without nonnegative clamps, preserving the signed distances for ink entirely below or above the baseline.

Polyfills/Canvas/Source/nanovg/fontstash.h
```diff
@@ -1551,16 +1580,34 @@ float fonsTextBounds(FONScontext* stash,
...
+			float qx0, qx1, qy0, qy1;
+			const float padding = FONS_TEXT_BOUNDS_PADDING + (float)glyph->blur + 1.0f;
 			fons__getQuad(stash, font, prevGlyphIndex, glyph, scale, state->spacing, &x, &y, &q);
-			if (q.x0 < minx) minx = q.x0;
-			if (q.x1 > maxx) maxx = q.x1;
+			// getQuad removes one of getGlyph's (blur + 2) border pixels.
+			qx0 = q.x0 + padding;
+			qx1 = q.x1 - padding;
+			if (qx1 < qx0) qx0 = qx1 = (q.x0 + q.x1) * 0.5f;
+			if (!hasGlyph) { minx = qx0; maxx = qx1; }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in ba8ec747. Padding-stripped quads only seed or update ink bounds when they have positive width and height; zero-ink glyphs still advance the pen. Added native coverage confirming a whitespace-only run advances while retaining zero-area ink bounds for both Fontstash coordinate origins.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -495,6 +539,437 @@ describe("Canvas2D", function () {
     expect(data.width).to.equal(4);
     expect(data.height).to.equal(3);
   });
+
+  it("renders a semi-transparent Canvas source through the destination GPU", function () {
```
bghgary: Please either run these GPU assertions with a real Metal device or skip them in the no-op host. All five macOS jobs set `BABYLON_NATIVE_TESTS_USE_NOOP_METAL_DEVICE=ON`, so the ten new cases read transparent pixels and fail without testing production Metal.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1140,6 +1366,45 @@ namespace Babylon::Polyfills::Internal
...
+                    throw Napi::Error::New(info.Env(), ex.what());
+                }
+            }
+
+            const int imageIndex = nvgCreateImageRGBA(*m_nvg, static_cast<int>(width), static_cast<int>(height), 0, rgba.data());
+            if (imageIndex == 0)
+            {
+                throw Napi::Error::New(info.Env(), "drawImage: failed to create the source image.");
+            }
+            RetainImageUntilFlush(imageIndex);
```
bghgary, line 1295: Please validate arity and geometry before `CaptureRGBA()` or temporary Canvas/ImageBitmap uploads. No-op Canvas calls currently perform synchronous framebuffer readback, and both source paths retain an unused texture until flush or disposal. Uploads referenced by `nvgFill` must remain alive through flush.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1231,103 +1500,189 @@ namespace Babylon::Polyfills::Internal
...
             ResetPathState();
-
-            nvgRect(*m_nvg, dx, dy, dWidth, dHeight);
+            nvgRect(*m_nvg, drawX, drawY, drawWidth, drawHeight);
             nvgFillPaint(*m_nvg, imagePaint);
             SetFilterStack();
             nvgFill(*m_nvg);
+        }
+
+        BlitPixelsToCpu(srcPixels, srcWidth, srcHeight, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight);
```
bghgary: Please make `getImageData()` read the rendered framebuffer instead of this partial CPU mirror. This blit omits the transform, clip, `globalAlpha`, filter, and GPU sampling; after `CaptureRGBA()` snapshots the framebuffer, later NanoVG draws also leave the mirror stale. The probe-canvas tests bypass both mismatches.

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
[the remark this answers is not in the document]
bkaradzic-microsoft: Follow-up in 8b7d19ba after macOS CI exposed a portability issue in the first implementation: changing IDs for every NanoVG-created texture made all image-backed Canvas draws transparent on Metal, while D3D11 happened to retain matching sequences. NanoVG-created textures now preserve their established bgfx-index IDs; only `nvgCreateImageFromHandle()` uses an independent positive namespace above the complete `uint16_t` bgfx handle range (the first external ID is 65536). Thus a valid bgfx handle 0 remains usable without colliding with NanoVG image 0, while existing renderer behavior is preserved.

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
```diff
@@ -449,7 +450,15 @@ namespace
...
         }
 
-        return bgfx::isValid(tex->id) ? tex->id.idx : 0;
+        if (!bgfx::isValid(tex->id))
+        {
+            return 0;
+        }
+
+        // NanoVG-created textures retain their existing bgfx-index image IDs.
+        // nvgCreateImageFromHandle uses a separate negative ID namespace below.
```
bghgary: Please update this comment: `nvgCreateImageFromHandle` now uses positive IDs above the complete `uint16_t` bgfx handle range, not a negative namespace.

Apps/UnitTests/Source/Tests.JavaScript.cpp
```diff
@@ -17,6 +17,7 @@
 #endif
 #include <Babylon/ScriptLoader.h>
 
+#include <bgfx/bgfx.h>
```
bghgary: Please avoid requiring an uninstalled bgfx header here. `Install/Test` compiles this source against the installed Babylon Native headers, and exact-head `Linux_Installation / test-install` fails at `<bgfx/bgfx.h>: No such file or directory`. Derive the capability through an installed API or compile definition so the consumer build remains valid.

Polyfills/Canvas/Source/Canvas.cpp
```diff
@@ -229,6 +249,46 @@ namespace Babylon::Polyfills::Internal
...
+                rgba = m_context->CaptureRGBA();
+            }
+            catch (const std::exception& ex)
+            {
+                throw Napi::Error::New(info.Env(), ex.what());
+            }
+        }
+
+#ifndef BABYLON_NATIVE_PLUGIN_NATIVEENGINE_LOAD_IMAGES
+        throw Napi::Error::New(info.Env(), "Canvas.toDataURL: image encoding is disabled in this build.");
```
bghgary: Please keep PNG encoding independent of `BABYLON_NATIVE_PLUGIN_NATIVEENGINE_LOAD_IMAGES`, or introduce and document a separate encoding capability and gate the unconditional test. This option promises only input-image loading, but it now removes `bimg_encode` and makes every `toDataURL()` call throw; the current ordering also performs GPU readback before that guaranteed error.

bkaradzic-microsoft:
Fixed the Windows and Linux installation failures in 2379eb396f04e516d9927a2b4028ffae5a375ad6.

Both jobs failed while compiling `Apps/UnitTests/Source/Tests.JavaScript.cpp`: `bgfx/bgfx.h` is not part of the installed public headers. The previous no-op-renderer change added that private dependency to a source file also built by `Install/Test`.

The test host now uses the existing `USE_NOOP_METAL_DEVICE` definition, shared with the Apple test host, to set `hasGpuRendering`. Only the explicit no-op Metal configuration skips the GPU pixel cases; Windows, Linux, and real Metal configurations keep them enabled. No bgfx headers were added to the public package.

Verified the exact header regression locally: the previous source fails against installed headers, while the corrected source compiles with and without the no-op definition. Windows/QuickJS `JavaScript.All` also passes all 85 cases with the GPU assertions enabled. All 34 checks now pass on this exact commit, including both Windows and Linux installation jobs and all macOS configurations: [completed CI run](https://github.com/BabylonJS/BabylonNative/actions/runs/35003929715).

bghgary (changes requested):
Concerns inline.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in ea742744f2697737da07b7b91bf78ade1ca6267b. `drawImage` now checks arity first, coerces its numeric arguments once, and normalizes/clips the rectangles before `CaptureRGBA`, pixel conversion, or image creation. The Canvas, ImageBitmap, and cached native-image paths all use the same geometry preflight. Coercion happens before reading source dimensions, since `valueOf()` can resize the source.

Added a native regression that rejects any texture-creation callback during invalid/no-op calls and verifies that the source framebuffer was never created. It covers non-finite coordinates, zero extents, fully clipped source rectangles, float overflow/underflow, and invalid arities for both Canvas and ImageBitmap sources; it fails on the previous implementation and passes with the fix.

Actual uploads still go through `RetainImageUntilFlush` and are released only after `nvgEndFrame` consumes the queued draws. The existing lifetime regressions now assert against the destination's framebuffer directly rather than a probe canvas.

Polyfills/Canvas/Source/nanovg/nanovg_babylon.cpp
bkaradzic-microsoft: Corrected in ea742744f2697737da07b7b91bf78ade1ca6267b: the comment now states that `nvgCreateImageFromHandle` uses positive IDs above the complete `uint16_t` bgfx handle range. No texture-ID behavior was changed.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in ea742744f2697737da07b7b91bf78ade1ca6267b. `getImageData()` now flushes and reads the rendered framebuffer through `CaptureRGBA`, then copies the requested region into straight-alpha ImageData with transparent out-of-bounds pixels. The CPU mirror and its drawImage/putImageData blits are removed, so later NanoVG draws cannot leave a stale readback cache. Readback failures are surfaced as JavaScript errors. Region normalization uses 64-bit coordinates to avoid signed overflow for negative extents near INT32_MIN.

Removed the probe-canvas workaround from the tests: the existing crop, transform, clip, alpha, sampling, and upload-lifetime cases now call `getImageData()` on the actual destination. Added direct cases for new draws after getImageData/toDataURL snapshots, supported blur-filter coverage, normalized/out-of-bounds regions, resize, and source resizing during numeric coercion. The direct-readback regressions expose the old mirror mismatch; the updated Windows/QuickJS JavaScript suite passes all 89 cases. All 34 cross-platform checks now pass on this exact head: [completed CI run](https://github.com/BabylonJS/BabylonNative/actions/runs/35020726251).

Polyfills/Canvas/Source/Canvas.cpp
bkaradzic-microsoft: Fixed in ea742744f2697737da07b7b91bf78ade1ca6267b. Canvas now links `bimg_encode` unconditionally, and `toDataURL()` no longer has an input-image-loading guard. `BABYLON_NATIVE_PLUGIN_NATIVEENGINE_LOAD_IMAGES` continues to control input loading/conversion only; PNG encoding and its MIME fallback remain available without introducing a second capability flag. There is no longer a guaranteed encoding-disabled error after GPU readback.

Added a native PNG-signature/fallback regression and built and ran it with image loading both ON and OFF. Both configurations pass, including the no-op drawImage preflight test with loading disabled. The Canvas README documents the independent encoding contract.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -806,67 +836,230 @@ namespace Babylon::Polyfills::Internal
         // now it shares the frame encoder with NativeEngine.
         encoder->discard(BGFX_DISCARD_ALL);
 
-        try
+        // bgfx framebuffer pool exhaustion can throw both from UpdateRenderTarget() and
+        // from FrameBufferPool::Acquire() reached via nvgEndFrame() below.
+        const bool needClear = m_canvas->UpdateRenderTarget();
+
+        Graphics::FrameBuffer& frameBuffer = m_canvas->GetFrameBuffer();
```
bghgary, line 843: Please ensure the render target exists before dereferencing it when readback flushes an untouched Canvas. A new Canvas starts with dimensions but no framebuffer and is not dirty, so `getImageData()` or `toDataURL()` immediately after `getContext("2d")` reaches this call and crashes; I reproduced it with `new Canvas().getContext("2d").getImageData(0, 0, 1, 1)`. Please cover both readback APIs on an untouched Canvas.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -1023,141 +1212,107 @@ namespace Babylon::Polyfills::Internal
...
+        for (size_t index = 1; index < info.Length(); ++index)
+        {
+            coordinates[index - 1] = info[index].ToNumber().DoubleValue();
+        }
+        const std::span<const double> arguments{coordinates.data(), info.Length() - 1};
+
+        // Check the nominal Canvas type before the structural ImageBitmap shape.
+        // Canvas objects are extensible, so user code may legitimately add a typed
+        // `data` property without changing the object passed to drawImage.
+        const auto canvasCtorVal = JsRuntime::NativeObject::GetFromJavaScript(info.Env()).Get("Canvas");
```
bghgary: Please determine and retain the source kind before coercing the numeric arguments, then read its current dimensions afterward. `ToNumber()` can run user code that changes a Canvas prototype or removes an ImageBitmap's `data`, so this code can reclassify the already-supplied object into another branch; changing a Canvas prototype and adding `data` makes it throw as an ImageBitmap instead of drawing. Please cover source-classification mutation during coordinate coercion.

bghgary (changes requested):
Concerns inline.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in 60c8972e61e1a64ee0e20677a41e4c753c95fb7d. `UpdateRenderTarget()` now allocates whenever the framebuffer is absent, even if the Canvas has never been resized or marked dirty, so `FlushCore()` has a render target before dereferencing it.

Added a native regression that exercises `getImageData()` and `toDataURL()` independently on fresh default-size canvases immediately after `getContext("2d")`, without drawing or setting dimensions. It checks transparent-black pixels, PNG equality with an untouched no-context canvas, and framebuffer creation. CI then exposed an additional `toDataURL()` bug: it encoded the rounded-up buffer capacity instead of the bytes actually written. Fixed that in af07833856f25870d9020360fc4e82e8e1fcf207 without relaxing the PNG equality assertion. A new deterministic regression verifies that PNG output ends exactly at `IEND`; it failed before the fix (73 PNG bytes versus 4096 serialized bytes). The full Windows suite now passes all 36 native tests, including all 92 JavaScript cases. All 34 checks now pass on af07833856f25870d9020360fc4e82e8e1fcf207: [completed CI run](https://github.com/BabylonJS/BabylonNative/actions/runs/35035344219). The V8 job initially failed during NuGet dependency setup, before compilation; its targeted retry passed without additional code changes.

Polyfills/Canvas/Source/Context.cpp
bkaradzic-microsoft: Fixed in 60c8972e61e1a64ee0e20677a41e4c753c95fb7d. `drawImage()` now retains the native Canvas pointer / ImageBitmap classification before numeric coercion, then coerces coordinates once and reads the source's current dimensions afterward. The chosen source branch is not reconsidered after user code runs. ImageBitmap data is revalidated before use, so deleting it produces a JavaScript type error instead of falling through to native-image unwrapping; no-op geometry still returns before readback/upload.

Added regressions for a coordinate `valueOf()` that resizes a Canvas, removes its prototype, and adds typed `data` (the resized Canvas still draws), and for ImageBitmap data removal during both drawing and no-op calls. Both pass, alongside the existing source-resize and no-op allocation regressions.

bghgary (approved):
LGTM

PR #1858, opened 2026-08-31
Make external texture disposal deterministic

bghgary:

## Context

An `InternalTexture` created by `ExternalTexture::CreateForJavaScript` could become usable again after disposal if its `ExternalTexture` was updated before JavaScript finalization. This prevented callers from deterministically releasing native rendering resources.

## Behavior

Disposed JavaScript textures now remain disposed across later external texture updates. The externally supplied platform texture remains caller-owned and can be released after JavaScript disposal without waiting for garbage collection.

### DIFF
--- a/Apps/UnitTests/CMakeLists.txt
+++ b/Apps/UnitTests/CMakeLists.txt
@@ -27,6 +27,7 @@ set(SOURCES
     "Source/App.cpp"
     "Source/Tests.Device.FrameEncoder.cpp"
     "Source/Tests.ExternalTexture.cpp"
+    "Source/Tests.ExternalTexture.Lifecycle.cpp"
     "Source/Tests.ExternalTexture.DeviceLoss.cpp"
     "Source/Tests.ExternalTexture.LayerCount.cpp"
     "Source/Tests.ExternalTexture.Msaa.cpp"
--- a/Apps/UnitTests/Source/Tests.ExternalTexture.Lifecycle.cpp
+++ b/Apps/UnitTests/Source/Tests.ExternalTexture.Lifecycle.cpp
@@ -105,10 +105,10 @@ namespace
                 auto global = env.Global();
                 auto engine = global.Get("_externalEngine").As<Napi::Object>();
                 engine.Get("dispose").As<Napi::Function>().Call(engine, {});
-                global.Delete("_externalTextureOwner");
-                global.Delete("_externalInternalTexture");
-                global.Delete("_externalNativeTexture");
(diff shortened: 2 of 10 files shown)


bkaradzic-microsoft (approved):
## Review (on behalf of @bkaradzic-microsoft)

**Verdict: Approve** — deterministic disposal is the right fix for the “disposed InternalTexture comes back after `ExternalTexture::Update`” bug. CI is green including sanitizers.

### What looks correct

1. **Root cause shape** — `UpdateTextures` now skips/`erase`s wrappers with `!IsValid()` instead of always `Create2D`/`Attach`-ing them again. Dispose sticks across later updates.
2. **Ownership** — external path goes through `Texture::Create2D(..., nativeTextureHandle)` with `m_ownsHandle = true`. `Dispose` → `bgfx::destroy` only tears down the bgfx wrapper; with `BGFX_SAMPLER_INTERNAL_SHARED` the platform texture stays caller-owned. Matches the readme.
3. **`Attach` stays non-owning** — still appropriate for Canvas FB views; owning external path is unified on `Create2D`.
4. **Create path** — `unique_ptr` until `m_textures.insert` succeeds; no leak on insert failure.
5. **BLIT_DST / RT zero-fill** gated on `nativeTextureHandle == 0` — Babylon-owned textures keep web-like behavior; imported handles do not.
6. **Tests** — lifecycle suite drives real JS `InternalTexture`/`ThinTexture.dispose`, then `Update`, and asserts `!IsValid()`. Private-header test kept out of the install consumer; JSI cleanup via `engine.dispose` is sound.

### Non-blocking notes

1. **Concurrent `Update` vs JS `deleteTexture`** — `Texture::Dispose` is outside `ExternalTexture`’s mutex. A dispose that races between `IsValid()` and `Create2D` can still revive a wrapper. Documented host serialization is fine if that is the long-term contract; a disposed bit under the lock would be belt-and-suspenders if you ever want it.
2. **Multi-wrapper** — logic should refresh live wrappers and drop disposed ones, but there is no test for “dispose A, keep B, `Update` only refreshes B.”
3. **Device reset** — `Dispose()` no-ops when `m_deviceID` mismatches, so `IsValid()` can stay true across reset and a later `Update` may touch that wrapper. Pre-existing edge; not introduced here.
4. **“Release after JS disposal”** — true w.r.t. BN/bgfx refs; hosts still need normal GPU completion before destroying the underlying resource if it was recently bound.

No request-changes items from this pass. Optional follow-ups only.

PR #1857, opened 2026-08-31
Bump bgfx.cmake (TextureRegion blit/read, shader bin v12, Vulkan bindings)

bkaradzic-microsoft:
## Summary
Bumps `bgfx.cmake` to `9a569c775917e3dd2a8cfb8836aeddbe790a8deb` (bgfx `0837ead` via BabylonJS/bgfx#78 / bgfx.cmake#144) and adapts Babylon Native to the Graphics/shader packaging changes in that range.

### API adaptations
- **Blit / read**: `encoder.blit` / `bgfx::blit` take `TextureRegion` pairs; `bgfx::readTexture` → `bgfx::read(TextureRegion, data)`. Updated in `FrameBuffer`, `DeviceImpl`, `NativeCapture`, `NativeEngine`.
- **Shader binary v12**: `CreateBgfxShader` writes `BGFX_CHUNK_MAGIC_*S_VERSION` **12**, emits per-uniform texture meta (`AppendUniformTextureMeta`), and carries D3D SRV/UAV masks.
- **Canvas headers**: embedded canvas shader headers bumped v11 → v12.
- **Depth formats**: D3D depth RT prefers D32F → D24 → D24S8 (Android depth-only stays D24S8). Backbuffer `Depth32` maps to `D32F` (D3D has no DSV for plain D32).
- **ShaderCache**: `CACHE_VERSION` 4 → **5** (v12 binaries are not compatible with prior cache entries).

### Vulkan packaging (shaderc / bgfx v12 binding model)
- Shared VS/FS `layoutBinding` per sampler name (no per-stage compact remap that desynced UniformStages).
- Image bindings start at 2; separate samplers at `image + 16`; UBOs stay 0/1.
- `AppendSamplersVulkan`: `separate_images`, strip `Texture` suffix, set `regIndex` from binding.
- Reject shaders with **>16** distinct sampler textures (bgfx texture stage limit) with an explicit error.
- Packaging helpers live in `ShaderCompilerVulkan.cpp` / common `CreateBgfxShader` sampler-set path (not `#if VULKAN` in Common).

### Tests / CI
- Stage-sampler tests: disjoint VS/FS names and partially shared names (D3D).
- `ShaderCompilation.VulkanRejectsMoreThanSixteenSamplers` (compiled only when `GRAPHICS_API=Vulkan`): unexpected success fails **outside** the catch path (no false pass via a thrown `"…>16…"` message).
[...]

PR #1859, opened 2026-09-01
Fix Vulkan >16 sampler test false-pass and add Linux Vulkan CI

bkaradzic-microsoft:
## Summary
Follow-up to #1857 addressing [@bghgary](https://github.com/BabylonJS/BabylonNative/pull/1857#issuecomment-5500039346):

1. **False pass on unexpected success** — `ShaderCompilation.VulkanRejectsMoreThanSixteenSamplers` no longer treats unexpected success as the expected rejection.
2. **Runnable Vulkan CI** — `Ubuntu_Clang_Vulkan` builds the full Vulkan configuration and runs 16-sampler acceptance and 17-sampler rejection under Xvfb with lavapipe. The Xlib crash is fixed by supplying bgfx with a device-owned display connection that survives renderer shutdown. The XML report must contain both completed, passing tests; missing or skipped tests fail the lane.
3. **Minimal Vulkan stubs** so the tree links: `RendererType/Vulkan`, `DeviceImpl_Vulkan`, `ExternalTexture_Vulkan` (throws / not implemented), `Helpers.Vulkan`. ExternalTexture tests skipped via `SKIP_EXTERNAL_TEXTURE_TESTS`.
4. **Compile fixes** surfaced by the new job: SPIRV-Cross GLSL for Vulkan, glslang IR `using namespace`, ExternalTexture Shared.h stub pattern.

## Test plan
- [x] Local full Linux Vulkan build and the exact CI sampler-boundary step pass
- [x] Existing OpenGL shader regressions pass locally
- [x] All 33 GitHub CI jobs passed, including both Vulkan runtime boundary tests: https://github.com/BabylonJS/BabylonNative/actions/runs/34547542730
- [x] Existing Linux OpenGL / Win32 jobs passed
- [x] Unexpected-success mutation correctly fails the rejection test
- [x] Workflow lint and missing/skipped/failed-report guard checks pass


bkaradzic-microsoft:
Addresses [@bghgary’s comment on #1857](https://github.com/BabylonJS/BabylonNative/pull/1857#issuecomment-5500039346):

- **False pass**: `VulkanRejectsMoreThanSixteenSamplers` uses explicit `status` values (`unexpected_success` / `threw` / `not_ready`); unexpected success fails outside the catch path.
- **CI**: `Ubuntu_Clang_Vulkan` with `GRAPHICS_API=Vulkan` + lavapipe runs that test.

.github/workflows/build-linux.yml
```diff
@@ -12,10 +12,14 @@ on:
...
+            graphics-api:
         required: false
-        type: boolean
-        default: false
+              type: string
+              default: OpenGL
+            enable-sanitizers:
+              required: false
+              type: boolean
+              default: false
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed — the cherry-pick had mangled indentation under `inputs:`/`steps:` (YAML would not parse). Rewrote `build-linux.yml` with correct nesting; validated with a YAML load.

.github/workflows/build-linux.yml
```diff
@@ -39,65 +43,90 @@ jobs:
...
-            -D OpenGL_GL_PREFERENCE=GLVND \
-            -D BABYLON_DEBUG_TRACE=ON \
-            -D BABYLON_NATIVE_PLUGIN_NATIVEDRACO=ON \
-            -D BABYLON_NATIVE_PLUGIN_NATIVEMESHOPT=ON \
-            ${{ inputs.js-engine == 'Hermes' && '-D HERMES_UNICODE_LITE=ON -D HERMES_ALLOW_BOOST_CONTEXT=0' || '' }} \
-            -D ENABLE_SANITIZERS=${{ inputs.enable-sanitizers && 'ON' || 'OFF' }} .
-          ninja -C build/Linux
+            - name: Build X11
+              run: |
+                cmake -G Ninja -B build/Linux \
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in the same rewrite — all `- name:` steps are aligned under `steps:` again.

.github/workflows/build-linux.yml
```diff
@@ -39,65 +43,90 @@ jobs:
...
+                  -D JAVASCRIPTCORE_LIBRARY=/usr/lib/x86_64-linux-gnu/libjavascriptcoregtk-4.1.so \
+                  -D NAPI_JAVASCRIPT_ENGINE=${{ inputs.js-engine }} \
+                  -D CMAKE_BUILD_TYPE=RelWithDebInfo \
+                  -D BX_CONFIG_DEBUG=ON \
+                  -D GRAPHICS_API=${{ inputs.graphics-api }} \
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Left as `inputs.graphics-api` to match the existing Win32 workflow (`build-win32.yml` uses the same hyphenated input + dot form). GitHub Actions accepts this; the real failure was the broken YAML indentation.

Apps/UnitTests/Source/Helpers.Vulkan.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Helpers
+{
+    Babylon::Graphics::TextureT CreateTexture(Babylon::Graphics::DeviceT, uint32_t, uint32_t, uint32_t, bool, uint32_t)
+    {
+        throw std::runtime_error{"Helpers::CreateTexture is not implemented for Vulkan"};
+    }
+
+    void DestroyTexture(Babylon::Graphics::TextureT)
+    {
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Updated: `DestroyTexture` now throws on a non-null handle so incorrect usage surfaces early. Null remains a no-op.

Plugins/ExternalTexture/Source/ExternalTexture_Vulkan.cpp
```diff
@@ -0,0 +1,50 @@
+#include <Babylon/Plugins/ExternalTexture.h>
+#include <Babylon/Graphics/DeviceContext.h>
+#include <Babylon/Graphics/Texture.h>
+#include <napi/pointer.h>
+
+#include "ExternalTexture_Base.h"
+
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Added `#include <optional>`.

Apps/UnitTests/Source/Helpers.Vulkan.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+    void DestroyTexture(Babylon::Graphics::TextureT texture)
+    {
+            // Creation helpers always throw, so a non-null handle means a caller
+            // bypassed them (or Vulkan texture support was added without updating this).
+            if (texture != nullptr)
+            {
+                throw std::runtime_error{"Helpers::DestroyTexture is not implemented for Vulkan"};
+            }
+        }
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Reformatted `DestroyTexture` to match surrounding style.

Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
```diff
@@ -345,32 +348,39 @@ TEST(ShaderCompilation, VulkanRejectsMoreThanSixteenSamplers)
...
+    loader.Dispatch([&done, &status, &message](Napi::Env env) {
+        status = env.Global().Get("__status").ToString();
         message = env.Global().Get("__message").ToString();
         done.set_value();
     });
     done.get_future().get();
     device.FinishRenderingCurrentFrame();
 
-    ASSERT_TRUE(failed || !failureMessage.empty()) << "expected compile failure";
+    ASSERT_NE(status, "unexpected_success") << "compiler accepted >16 samplers; expected rejection";
```
bghgary: Please add a 16-sampler success case alongside this 17-sampler rejection. The guard promises “at most 16,” but only the rejection side is tested; changing `>` to `>=` would still pass this CI test while rejecting the supported boundary.

Plugins/ExternalTexture/Source/ExternalTexture_Vulkan.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    uintptr_t NativeTextureHandle(Graphics::TextureT ptr)
+    {
+        return reinterpret_cast<uintptr_t>(ptr);
+    }
+
+    class ExternalTexture::Impl final : public ImplBase
+    {
+    public:
+        Impl(Graphics::TextureT, std::optional<Graphics::TextureFormatT>)
```
bghgary: `ExternalTexture_Shared.h` defines `Impl::Impl` and `Impl::Update` again after this class, so the Vulkan build fails with redefinition errors once it reaches this file. Please leave these as declarations here and let the shared definitions call the throwing `GetInfo`.

Apps/UnitTests/Source/Tests.ShaderCompilation.cpp
bghgary: Please add a 16-sampler success case alongside this 17-sampler rejection. The guard promises “at most 16,” but only the rejection side is tested; changing `>` to `>=` would still pass this CI test while rejecting the supported boundary.

.github/workflows/build-linux.yml
```diff
@@ -36,68 +40,81 @@ jobs:
 
 
       - name: Install packages
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libjavascriptcoregtk-4.1-dev libgl1-mesa-dev libcurl4-openssl-dev libwayland-dev clang ninja-build
+              run: |
```
bghgary: This indentation makes the reusable workflow invalid again: `run` is nested under the scalar `name`, and the following `- name:` entries are nested inside that step. GitHub scheduled only the two repository-level checks and no CI jobs at this head. Please align `run` with `name` and every step entry under `steps:`.

.github/workflows/build-linux.yml
```diff
@@ -36,68 +40,81 @@ jobs:
...
+                ulimit -c unlimited
+                xvfb-run ./Playground app:///Scripts/validation_native.js
 
-      - name: Unit Tests
-        run: |
-          cd build/Linux/Apps/UnitTests
-          ulimit -c unlimited
-          xvfb-run ./UnitTests
+            - name: Unit Tests
+              if: ${{ inputs.graphics-api != 'Vulkan' }}
```
bghgary: This makes the Vulkan lane build-only, so neither this PR’s `VulkanRejectsMoreThanSixteenSamplers` fix nor the supported 16-sampler boundary is exercised in CI. Please keep runnable Vulkan coverage for both 16-sampler acceptance and 17-sampler rejection; otherwise the original false-pass defect can recur while this job stays green.

bghgary (changes requested):

Concerns inline.

.github/workflows/build-linux.yml
bkaradzic-microsoft: Fixed in 5b51a38051cf4f175b86b90944cf1616f720eaf4. `run` is now aligned with `name`, and every step is directly under `steps:`. Both the reusable workflow and its CI caller pass actionlint 1.7.12; the original version reproduces the YAML parse error with the same validator.

.github/workflows/build-linux.yml
bkaradzic-microsoft: Implemented in 5b51a38051cf4f175b86b90944cf1616f720eaf4. The Vulkan lane now runs both `VulkanAcceptsSixteenSamplers` and `VulkanRejectsMoreThanSixteenSamplers` under Xvfb with lavapipe explicitly selected. An XML guard requires both named tests to run and pass; missing or skipped tests cannot leave the lane green.

The startup crash was a null X11 display passed to bgfx, not a lack of hardware GPU support. The device now owns the display connection and keeps it alive through bgfx shutdown. The tests use synchronous native shader-program creation to preserve the actual compiler diagnostic (JavaScriptCore's Effect error text retained only the stack), and the rejection assertions stay outside the JavaScript catch path.

The full Linux Vulkan build and the exact CI runtime step pass locally. Four Vulkan sampler regressions and four existing OpenGL shader regressions also pass. I temporarily fed the supported 16-sampler shader to the rejection test to simulate unexpected success: it correctly failed, then passed again after restoring 17 samplers.

bkaradzic-microsoft: CI confirmation: [run 34547542730](https://github.com/BabylonJS/BabylonNative/actions/runs/34547542730) passed **all 33 jobs** on 5b51a38051cf4f175b86b90944cf1616f720eaf4.

The Vulkan lane completed its runtime step, not just the build. I downloaded its XML artifact and verified both `ShaderCompilation.VulkanAcceptsSixteenSamplers` and `ShaderCompilation.VulkanRejectsMoreThanSixteenSamplers` ran and passed, with neither missing nor skipped. The existing OpenGL and Windows jobs passed as well.

.github/workflows/build-linux.yml
```diff
@@ -59,43 +66,83 @@ jobs:
         run: sudo sysctl -w kernel.core_pattern=core.%p
 
       - name: Validation Tests
+        if: ${{ inputs.graphics-api != 'Vulkan' }}
```
bghgary, line 76: This silently excludes the entire playground validation suite from the new Vulkan configuration. Please enable the compatible Vulkan validation coverage here, or link a follow-up issue documenting the measured blockers and tracking the excluded coverage.

.github/workflows/build-linux.yml
```diff
@@ -59,43 +66,83 @@ jobs:
...
         run: |
           cd build/Linux/Apps/Playground
           ulimit -c unlimited
           xvfb-run ./Playground app:///Scripts/validation_native.js
 
       - name: Unit Tests
+        if: ${{ inputs.graphics-api != 'Vulkan' }}
```
bghgary: This excludes the normal unit-test run and replaces it with only two sampler cases, even though unsupported external-texture cases already carry Vulkan-specific `GTEST_SKIP` guards. Please run the full Vulkan `UnitTests` binary and retain the XML check for the two boundary cases; if any additional groups are genuinely incompatible, link follow-up issues for those measured failures instead of excluding the suite wholesale.

.github/workflows/build-linux.yml
```diff
@@ -59,43 +66,83 @@ jobs:
...
+              if case.get("status") == "run" and case.get("result") == "completed"
+              and case.find("failure") is None and case.find("skipped") is None
+          }
+          if len(cases) != 2 or completed != expected:
+              raise SystemExit("Both Vulkan sampler boundary tests must run and pass")
+          PY
+
       - name: Module Load Test
-        if: ${{ !inputs.enable-sanitizers }}
+        if: ${{ !inputs.enable-sanitizers && inputs.graphics-api != 'Vulkan' }}
```
bghgary: There is no evident Vulkan incompatibility in `ModuleLoadTest`: its Linux path creates an X11 window and its optional-module filter already permits Vulkan ICD libraries. Please run it in this job; if an actual blocker remains, link a follow-up issue that records and tracks it.

bghgary (changes requested):

Concerns inline.

.github/workflows/build-linux.yml
bkaradzic-microsoft: Implemented in 239107e2. The Vulkan lane now runs the existing Playground validation runner through Linux-valid script URL arguments for the measured-compatible `Capsule` and `Multi camera rendering` cases. The step asserts both exact `Running` lines and `Run complete. ran=2 passed=2 failed=0 missingRef=0 skipped=0`, so a skipped or missing case cannot pass silently. The measured full-suite blockers (23 initial pixel failures followed by a `Glow layer and LODs` SIGSEGV) are tracked with reproduction details in #1878; no pixel thresholds or catalog exclusions were changed.

.github/workflows/build-linux.yml
bkaradzic-microsoft: Implemented in 239107e2. The Vulkan job now runs the normal `UnitTests` binary and emits `vulkan-unit-test-results.xml`; locally this executes 29 cases (14 passed and 15 existing Vulkan-specific skips). Only `ShaderCompilation.CompileComprehensiveGLSL` is narrowly filtered because its Vulkan shader failure leaves the test waiting indefinitely; the reproduction and removal criteria are tracked in #1877. The XML verification checks the 29-case count and requires both `VulkanAcceptsSixteenSamplers` and `VulkanRejectsMoreThanSixteenSamplers` to have completed without failure or skip.

.github/workflows/build-linux.yml
bkaradzic-microsoft: Implemented in 239107e2. `ModuleLoadTest` now runs for Vulkan under Xvfb. The measured local run found Mesa's `libVkLayer_MESA_device_select.so` as the only unexpected module, so the Linux optional-module filter now accepts the `libvklayer_` prefix; the rerun reports `ModuleLoadTest: PASS`.

bghgary (approved):

LGTM

PR #1863, opened 2026-09-03
Playground validation: once-only failure cleanup and pixel-count wording

bkaradzic-microsoft:
## Summary
Post-merge follow-up to #1838 ([review](https://github.com/BabylonJS/BabylonNative/pull/1838#pullrequestreview-4953825129)):

1. **Once-only completion + cleanup** — every test completion path (pixel pass/fail, render-loop `catch`, `onReadyTimeout`, snippet/script load/eval errors) now goes through `makeTestDone`: `stopRenderLoop()`, shared `cleanupAfterTest()` (scene/strays, stencil/scissor, effects, texture cache, SceneLoader plugin observable), then the outer callback. Prevents stale `runRenderLoop` callbacks from rendering the next scene or calling `done(false)` repeatedly under continue-on-failure.
2. **Wording** — a stable pixel-difference count is guidance to compare saved result/diff images, not proof that timing flakes are ruled out.

## Test plan
- [ ] Headless multi-test run with a known early throw still advances and prints a full summary
- [ ] Normal pixel pass/fail unchanged
- [ ] CI playground jobs green

### DIFF
--- a/Apps/Playground/Scripts/validation_native.js
+++ b/Apps/Playground/Scripts/validation_native.js
@@ -37,6 +37,8 @@
     }
 
     function failTest(done) {
+        // done is the once-only completion wrapper from runTest; it stops the
+        // render loop, runs inter-test cleanup, then advances the suite.
         if (breakOnFail) {
             // Trigger the JS debugger if attached; on no-debugger runs the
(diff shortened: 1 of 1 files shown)


bghgary (approved):

LGTM

PR #1865, opened 2026-09-09
Add an opt-in minimal build profile

bghgary:

## Context

Babylon Native's established configuration enables most optional components and repository applications. Embedders that need only the core runtime and graphics libraries currently have to disable each unrelated option individually, and newly added options can silently expand their outputs.

## Worth a look

- The minimal profile is opt-in, preserves explicit component selections, and does not change existing defaults when omitted. A fresh CMake build directory is required when switching profiles.
- `BUILDING.md` lists the optional capabilities and the components required by each bundled application and test target.
- A Release install with applications disabled in both configurations measured 117,534,020 bytes with established defaults and 70,818,430 bytes with the minimal profile, a 39.75% reduction.


bghgary:

Closing for now while the minimal profile is expanded to cover optional dependency capabilities. We can reopen this pull request when that work is ready.

PR #1871, opened 2026-09-11
Preserve queued instance data during buffer updates

bkaradzic-microsoft:
## Root cause

`updateDynamicVertexBuffer` mutates native vertex-buffer data immediately, while draw commands can still be queued in the JavaScript command stream. Instance attributes are copied when those commands are decoded, so a later update can overtake an earlier draw and make it use the new values.

The command-stream pointer persists between frames. Empty submissions therefore must not acquire `FrameCompletionScope`, whose contract intentionally waits for the next active frame.

## Fix

Submit pending commands before mutating a dynamic vertex buffer, preserving the data visible when each queued draw was issued. `SubmitCommands` first flushes the JavaScript staging buffer and returns when it is empty; it acquires frame lifetime protection only when commands require decoding.

## Regression coverage

- Queues a thin-instance draw with matrix translation `0.25`, then updates it to `0.75`; captured red must remain `64`.
  - Before fix: red was `191`.
  - After fix: red is `64`.
- Issues an empty-stream dynamic vertex-buffer update between frames.
  - Before follow-up: remained blocked until another frame started.
  - After follow-up: completes without waiting for a frame.
- Focused run: 11/11 passed three consecutive times, including `NativeEngineInstanceData.*`, `ShaderCompilation.FragCoord*`, upstream `NativeEngineClear.*`, and `Device.FrameCompletionScopeProvidesEncoderOutsideFrame`.
- Build: `cmake --build build\Win32_Review --target UnitTests --config RelWithDebInfo -- /m:2` (D3D11/QuickJS).

This branch uses upstream dependency declarations unchanged and imports no shotgun dependency pins, visual-test enables, or tolerance changes.
## Review hardening

- Setup and preparation errors propagate through shared one-shot completion state; startup/render waits are bounded and late callbacks retain valid state. JS-backed errors are copied before runtime destruction, and frame/texture cleanup covers exceptional paths.
- Six fault scenarios cover synchronous setup/preparation throws, Error/non-Error rejections, unresolved preparation, and late preparation.
- The empty-stream test opens its cleanup frame before reporting a dispatch timeout and keeps the gate open throughout runtime teardown. With the old scope ordering and a 200 ms wait / 300 ms delayed callback, the original path hung beyond 5 seconds; the repaired path reports failure and exits in 1.247 seconds.
- Rebased onto master `44a76ef2`, preserving the newly merged clear/binding regressions. Fault injections were removed before the final build/run. CI for `3fd592a6946ed2d20ca681b6ce20c6732d789263` is fully green: **35/35 checks passed** ([run 35152105436](https://github.com/BabylonJS/BabylonNative/actions/runs/35152105436)).


Plugins/NativeEngine/Source/NativeEngine.cpp
```diff
@@ -1263,6 +1263,12 @@ namespace Babylon
 
         try
         {
+            // Queued draws snapshot instance data when decoded. This immediate
+            // update must not overwrite that data before those draws execute.
+            if (m_commandStream)
+            {
+                SubmitCommands(info);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Confirmed—this was valid. `FrameCompletionScopeProvidesEncoderOutsideFrame` verifies that scope acquisition blocks until the next frame, and the persistent stream made an empty `updateDynamicVertexBuffer` do exactly that.

Fixed in `7a00f293`: `SubmitCommands` now flushes the JS staging buffer, returns immediately when the reader is empty, and only acquires `FrameCompletionScope` when commands actually require decoding. Pending commands still retain the frame/lifetime protection needed for ordering.

Added `DynamicVertexBufferUpdateWithEmptyStreamDoesNotWaitForFrame`. Before the fix it remained blocked until the test started another frame; after the fix it completes between frames. Final focused run passed 8/8 tests, including both ordering regressions and `Device.FrameCompletionScopeProvidesEncoderOutsideFrame`.

bghgary: Move this call outside the buffer-update catch so queued-command failures still propagate to JavaScript. `SubmitCommands` converts decoder failures to `Napi::Error`; this outer `std::exception` catch logs and suppresses them, and the later `endCommandScope()` sees an empty stream.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
```diff
@@ -165,12 +167,16 @@ namespace
...
                 globalThis.render = function () {
                     var scene = globalThis.__scene;
-                    return scene.whenReadyAsync().then(function () {
+                    var preparation = globalThis.__prepare ? globalThis.__prepare() : Promise.resolve();
+                    return preparation.then(function () {
+                        return scene.whenReadyAsync();
+                    }).then(function () {
                         scene.render();
                     });
                 };
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in `eaf19586`. The current instance regression's `__prepare` returns `forceCompilationAsync`, but the hook is injected test setup rather than a Promise-only API. Wrapping its result with `Promise.resolve(...)` preserves the existing Promise path and also safely accepts synchronous/undefined setup hooks.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
```diff
@@ -258,6 +265,103 @@ namespace
...
+    std::promise<void> setupDone;
+    loader.Dispatch([&setupDone](Napi::Env env) {
+        auto nativeEngine = env.Global().Get("BABYLON").As<Napi::Object>().Get("NativeEngine").As<Napi::Function>();
+        env.Global().Set("__engine", nativeEngine.New({}));
+        auto engine = env.Global().Get("__engine").As<Napi::Object>();
+        auto data = Napi::Float32Array::New(env, 3);
+        engine.Set("__buffer", engine.Get("createDynamicVertexBuffer").As<Napi::Function>().Call(engine, {data}));
+        setupDone.set_value();
+    });
+    setupDone.get_future().wait();
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in `eaf19586`. The setup dispatch now captures shared completion state, catches and propagates callback exceptions, and uses a 30-second bounded wait. Timeout/error paths close the active frame before failing, and the shared capture prevents a late callback from referencing destroyed stack promises. I applied the same lifetime/error handling to the subsequent update dispatch. Negative control: an injected setup exception failed with `dynamic vertex-buffer setup failed: setup negative control` in 987 ms instead of hanging; after removing the injection, the focused tests pass.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
```diff
@@ -36,7 +37,8 @@ namespace
         uint32_t height,
         const std::string& vertexShader,
         const std::string& fragmentShader,
-        bool withInputTexture)
+        bool withInputTexture,
+        const std::string& setupScript = {})
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Keeping `{}` here. It value-initializes the empty `std::string` unambiguously, matches modern C++ defaults already used throughout this repository (including `std::string_view sourceUrl = {}`, `RuntimeOptions options = {}`, maps, optionals, and callbacks), and preserves the intentionally identical helper signature shared with the companion change. Replacing it with a string literal would be subjective style churn with no behavioral or readability requirement from the project.

bkaradzic-microsoft:
Follow-up on the two suppressed notes from the original Copilot review:

1. The exception concern was valid. A negative control throwing `Napi::Error` at the nested submission site logged the message but let the regression pass, proving the outer vertex-update catch swallowed it. `dd2908ad` moves `SubmitCommands` outside that handler: the same control now terminates through the JS error path (exit 3), while native buffer-update failures retain the existing logging convention.
2. Empty-stream scope acquisition was already addressed by `7a00f293`. `GetReader()` must remain before the empty check because it invokes the JS stream's flush callback, transferring staged words into the native buffer; a native-only pending check before that flush could incorrectly skip real commands. Empty readers return before acquiring `FrameCompletionScope`.

Restored focused validation: 8/8 tests passed.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
```diff
@@ -165,12 +168,16 @@ namespace
...
 
                     quad.material = material;
+                    SETUP_SCRIPT
                     globalThis.__scene = scene;
                 };
 
                 globalThis.render = function () {
                     var scene = globalThis.__scene;
-                    return scene.whenReadyAsync().then(function () {
+                    var preparation = Promise.resolve(globalThis.__prepare ? globalThis.__prepare() : undefined);
```
bghgary: Propagate failures from both new hooks into the test futures and bound their waits. A synchronous `__prepare` exception occurs before `Promise.resolve` receives its argument, so `AppRuntime` only logs it and `renderDone.get()` never returns; an exception from `SETUP_SCRIPT` similarly leaves `startupDone` unset. Invoke preparation from a promise continuation and catch setup-dispatch exceptions.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
```diff
@@ -258,6 +266,140 @@ namespace
...
+        }
+        catch (...)
+        {
+            setupDone->set_exception(std::current_exception());
+        }
+    });
+
+    if (setupFuture.wait_for(std::chrono::seconds{30}) != std::future_status::ready)
+    {
+        device.FinishRenderingCurrentFrame();
```
bghgary: Run this test's device/runtime lifecycle on a worker thread and apply the timeout from outside it. The setup timeout closes the frame and destroys `runtime` while its dispatch may still be in flight, and the update-start assertion can return without opening the frame needed to unblock the regression path. With a 200 ms wait and a 300 ms callback delay, the test reported the timeout but the process remained alive past 5 seconds.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
bghgary: Propagate failures from both new hooks into shared completion state and bound the waits. A synchronous `__prepare` exception occurs before `Promise.resolve` receives its argument, so `AppRuntime` only logs it and `renderDone.get_future().get()` never returns; an exception from `SETUP_SCRIPT` similarly leaves `startupDone` unset. Invoke preparation from a promise continuation, catch setup-dispatch exceptions, and ensure late callbacks cannot reference destroyed promises.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
```diff
@@ -258,6 +266,140 @@ namespace
...
+            engine.Get("updateDynamicVertexBuffer").As<Napi::Function>().Call(engine, {engine.Get("__buffer"), data});
+            updateDone->set_value();
+        }
+        catch (...)
+        {
+            updateDone->set_exception(std::current_exception());
+        }
+    });
+
+    ASSERT_EQ(updateStartedFuture.wait_for(std::chrono::seconds{30}), std::future_status::ready);
```
bghgary: Do not return from this timeout before opening the cleanup frame. Under the scope-before-empty-check ordering this test guards against, a late callback blocks acquiring `FrameCompletionScope`; this fatal assertion begins `runtime` teardown, which then hangs joining the blocked thread because no next frame opens the acquisition gate. With a 200 ms wait and a 300 ms callback delay, the assertion fired but the process remained alive past 5 seconds. Run the lifecycle on a worker thread with an external timeout, or explicitly open the gate and drain the callback before failing.

bghgary (approved):

Concerns inline.

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
bkaradzic-microsoft: Fixed in 3fd592a6946ed2d20ca681b6ce20c6732d789263. Preparation now starts in a promise continuation; setup/render dispatch exceptions and unexpected runtime errors complete shared, idempotent state. Both waits are bounded, callbacks own their completion state, and JS-backed exceptions are copied to native error strings before teardown. Frame/texture cleanup also runs on error paths.

Added six failure-path scenarios: setup throw, synchronous preparation throw, Error/non-Error promise rejection, never-settling preparation, and delayed preparation past its deadline. The focused 11-case run passed three consecutive times in the CI configuration (D3D11/QuickJS, RelWithDebInfo).

Apps/UnitTests/Source/Tests.ShaderCompilation.FragCoord.cpp
bkaradzic-microsoft: Fixed in 3fd592a6946ed2d20ca681b6ce20c6732d789263. The start-wait result is recorded rather than asserted immediately. The cleanup frame opens before any failure is reported, the callback gets a drain wait, and the acquisition gate stays open through runtime teardown, including when completion itself is late.

I reproduced your 200 ms wait / 300 ms callback delay with scope acquisition restored before the empty check: the old fatal path asserted but was still alive after 5 seconds. With the fix, the same injected regression reports failure and exits cleanly with code 1 in 1.247 seconds. All injections were then removed and the normal focused tests passed again.

PR #1872, opened 2026-09-11
Fix nested canvas clipping and clipped clears

bkaradzic-microsoft:
## Summary
- Intersect rectangular Canvas clips with the active NanoVG scissor so nested, disjoint, translated, signed-rectangle, and restored clips retain parent bounds.
- Implement `clearRect` with opaque destination-out coverage, independent of `globalAlpha` and filters, while restoring the prior drawing state.
- Initialize empty NanoVG filter stacks and add GPU-readback regressions with existing render-test skip handling.
- Fix the OpenPBR validation scene's asynchronous mesh-loading race: return the scene only after mesh import and material setup finish, using the existing per-test source replacement support. The reference image, render count, and tolerance are unchanged.
- Fix the inherited Vulkan CI gate to require both sampler-boundary tests to run and pass without assuming a fixed total suite size.

## Base
Rebased onto `master` at `4af724a6`. Includes signed-rectangle normalization before scissor intersection. This includes the already-merged Canvas work from #1855 and does not depend on the open branding PR #1844.

## Validation
- Windows x64 / D3D11 / QuickJS, `RelWithDebInfo`: all 6 selected native cases passed, including all 94 JavaScript cases.
- A delayed mesh import reproduced every pixel of the blank CI screenshot on both the PR and its upstream base (41.223% mismatch). With the scene-loading fix, both delayed and ordinary runs pass at 0.733%, below the unchanged 2.5% limit.
- The actual snippet's transformed scene creation waits for the import, finishes all 16 material assignments, and propagates import failures.
- The Vulkan gate accepts the actual 34-case CI report and rejects missing, skipped, failed, unexecuted, or incomplete required boundary tests.

### Signed-rectangle update
- Normalize negative width/height before intersecting with the parent clip; preserve the existing transform and antialiasing adjustment.
- Cover both signed axes, their combination, and rotated nested clips using byte-identical GPU comparisons with equivalent positive rectangles.
- Windows x64 / D3D11 / Chakra / RelWithDebInfo: all 6 selected native cases pass, including 68 JavaScript cases; 27 optional codec cases are pending because those plugins are disabled in this build. The new regression fails against the previous PR head on the first negative-width case and passes with the fix.
- No Babylon.js changes, dependency updates, reference changes, or validation-gate adjustments in this update.


Polyfills/Canvas/Source/Context.cpp
```diff
@@ -375,7 +375,11 @@ namespace Babylon::Polyfills::Internal
         const float height = info[3].As<Napi::Number>().FloatValue();
 
         nvgSave(*m_nvg);
-        nvgGlobalCompositeOperation(*m_nvg, NVG_COPY);
+        // NanoVG's clip is shader coverage, so COPY would erase even outside the clip.
+        nvgGlobalCompositeOperation(*m_nvg, NVG_DESTINATION_OUT);
+        nvgGlobalAlpha(*m_nvg, 1.f);
+        nanovg_filterstack clearFilters;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in abf1581a. The concern is valid for this implementation: `nvgFilterStack` assigns the complete `nanovg_filterstack` into `NVGstate::m_filterStack`, and each render callback then assigns it again into `GLNVGcall::filterStack`. Although rendering indexes `stackElements` only below `stackElementCount`, those implicit whole-object copies include the array. I added an in-class `{}` initializer to `stackElements`, which works even with the existing user-provided default constructor and makes every copied element deterministic; `stackElementCount{0}` remains the empty-stack discriminator.

Polyfills/Canvas/Source/Context.cpp
```diff
@@ -533,7 +537,8 @@ namespace Babylon::Polyfills::Internal
         auto h = m_rectangleClipping.height != 0 ? m_rectangleClipping.height : m_canvas->GetHeight();
 
         // expand clipping 1pix in each direction because nanovg AA gets cut a bit short.
-        nvgScissor(*m_nvg, m_rectangleClipping.left - 1, m_rectangleClipping.top - 1, w + 1, h + 1);
+        // A nested clip must not expand its parent's clipping region.
+        nvgIntersectScissor(*m_nvg, m_rectangleClipping.left - 1, m_rectangleClipping.top - 1, w + 1, h + 1);
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Addressed in abf1581a by clarifying the comment only. The `left - 1, top - 1, width + 1, height + 1` bounds predate this PR; this change only replaces `nvgScissor` with `nvgIntersectScissor`. I preserved that established asymmetric AA allowance rather than speculatively changing clip coverage, and the comment now accurately says it extends one pixel toward left/top.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -153,6 +155,138 @@ describe("Canvas2D", function () {
...
+          ctx.fillStyle = "blue";
+          ctx.fillRect(40, 0, 8, 64);
+          texture.update(false);
+          const pixels = await texture.readPixels();
+          if (!(pixels instanceof Uint8Array)) {
+            throw new Error("Expected RGBA8 GPU readback for the canvas texture");
+          }
+          const pixel = (x: number) =>
+            Array.from(
+              pixels.subarray(
```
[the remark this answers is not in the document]
bkaradzic-microsoft: I inspected the pinned native readback path and am keeping this assertion unchanged. Babylon.js 9.21.2 `ThinNativeEngine._readTexturePixels` calls native `readTexture`, then—when no caller buffer is supplied, as in this test—constructs and returns `new Uint8Array(rawBuffer)`. NativeEngine also converts the readback to RGBA8 before resolving the ArrayBuffer. The broader `ArrayBufferView` base-texture typing covers other engines/formats, but this regression intentionally verifies the concrete pinned Native RGBA8 contract; accepting hypothetical clamped/other views would weaken it.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -153,6 +155,138 @@ describe("Canvas2D", function () {
     return canvas.getContext("2d");
   }
 
+  (skipCanvasGpuTests ? it.skip : it)(
```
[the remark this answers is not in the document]
bkaradzic-microsoft: No code change: the fallback scenario is not a supported execution path for this bundle. `Apps/UnitTests/JavaScript/package.json` exposes only `build`/`watch`; CMake packages `tests.javaScript.all.js` solely as an asset of the native `UnitTests` executable, and the only loader is `Apps/UnitTests/Source/Tests.JavaScript.cpp`. That host sets `skipCanvasGpuTests` in `runtime.Dispatch` before `loader.LoadScript("app:///Assets/tests.javaScript.all.js")` (lines 76-80 vs. 112). The same bundle also requires host-provided `setExitCode` and `_native`, so it is not independently runnable in a generic JS runner. A default would either silently run GPU tests without the native render contract or silently skip them, masking a broken supported-host injection. Current-head D3D11 evidence remains `48 passing`, `27 pending`, including both GPU tests.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -153,6 +155,138 @@ describe("Canvas2D", function () {
...
+          const ctx = texture.getContext();
+          ctx.fillStyle = "white";
+          ctx.fillRect(0, 0, 64, 64);
+          ctx.save();
+          ctx.beginPath();
+          ctx.rect(8, 0, 24, 64);
+          ctx.clip();
+
+          ctx.save();
+          if (translated) {
```
bghgary: Please cover a transform that changes rotation, not only translation, and fix the resulting leak. With a 45-degree child transform, `nvgIntersectScissor` approximates the parent as an axis-aligned bounding box; a D3D11 probe paints pixel `(32, 8)` outside a parent clip spanning `[16, 48]`.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -153,6 +155,138 @@ describe("Canvas2D", function () {
...
+          ctx.clip();
+          ctx.fillStyle = "red";
+          ctx.fillRect(0, 0, 64, 64);
+          ctx.restore();
+
+          ctx.fillStyle = "#00ff00";
+          ctx.fillRect(24, 0, 16, 64);
+          ctx.save();
+          ctx.beginPath();
+          ctx.rect(40, 0, 16, 64);
```
bghgary: Please add an adjacent disjoint case and fix its leak. A parent `[8, 32]` clip followed by child `[32, 48]` paints pixel 31 red because both rectangles are expanded before intersection, creating a `[31, 32]` overlap.

Apps/UnitTests/JavaScript/dist/tests.javaScript.all.js
```diff
@@ -28384,6 +28384,101 @@ describe("Canvas2D", function () {
     return canvas.getContext("2d");
   }
 
+  (skipCanvasGpuTests ? it.skip : it)("intersects nested clips and restores parent clips on the GPU", /*#__PURE__*/(0,_babel_runtime_helpers_asyncToGenerator__WEBPACK_IMPORTED_MODULE_0__["default"])(/*#__PURE__*/_babel_runtime_regenerator__WEBPACK_IMPORTED_MODULE_1___default().mark(function _callee() {var _loop, _i, _arr;return _babel_runtime_regenerator__WEBPACK_IMPORTED_MODULE_1___default().wrap(function (_context2) {while (1) switch (_context2.prev = _context2.next) {case 0:
```
bghgary: `npm --prefix Apps run build --workspace UnitTests` rewrites this new block, while the same command leaves the base bundle unchanged. Please regenerate and commit the bundle from the checked-in TypeScript.

bkaradzic-microsoft:
The [Windows/Hermes failure](https://github.com/BabylonJS/BabylonNative/actions/runs/34547795129/job/103104060336) occurred after a successful build: snippet requests returned HTTP 403 on that runner, so all 314 Playground cases failed during scene loading (three failed fetch attempts per case), before rendering.

The same commit passed all 314 cases in the Windows D3D11 and V8 jobs, and the failed snippet URLs now return HTTP 200 from a separate probe. No code or test changes are warranted by this evidence. The failed-only retry (attempt 2, job 103128300425) has now passed both Validation Tests and Unit Tests. All 34 checks are green on the same commit; no code or test changes were needed.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bghgary: Please cover a transform that changes rotation, not only translation, and fix the resulting leak. With a 45-degree child transform, `nvgIntersectScissor` approximates the parent as an axis-aligned bounding box; a D3D11 probe paints pixel `(32, 8)` outside a parent clip spanning `[16, 48]`.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bghgary: Please add an adjacent disjoint case and fix its leak. A parent `[8, 32]` clip followed by child `[32, 48]` paints pixel 31 red because both rectangles are expanded before intersection, creating a `[31, 32]` overlap.

Apps/UnitTests/JavaScript/dist/tests.javaScript.all.js
bghgary: `npm --prefix Apps run build --workspace UnitTests` rewrites this new block, while the same command leaves the base bundle unchanged. Please regenerate and commit the bundle from the checked-in TypeScript.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -194,6 +196,138 @@ describe("Canvas2D", function () {
...
+          const ctx = texture.getContext();
+          ctx.fillStyle = "white";
+          ctx.fillRect(0, 0, 64, 64);
+          ctx.save();
+          ctx.beginPath();
+          ctx.rect(8, 0, 24, 64);
+          ctx.clip();
+
+          ctx.save();
+          if (translated) {
```
bghgary: Please cover a transform that changes rotation, not only translation, and fix the resulting leak. With a 45-degree child transform, `nvgIntersectScissor` approximates the parent as an axis-aligned bounding box; a D3D11 probe paints pixel `(32, 8)` outside a parent clip spanning `[16, 48]`.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -194,6 +196,138 @@ describe("Canvas2D", function () {
...
+          ctx.clip();
+          ctx.fillStyle = "red";
+          ctx.fillRect(0, 0, 64, 64);
+          ctx.restore();
+
+          ctx.fillStyle = "#00ff00";
+          ctx.fillRect(24, 0, 16, 64);
+          ctx.save();
+          ctx.beginPath();
+          ctx.rect(40, 0, 16, 64);
```
bghgary: Please add an adjacent disjoint case and fix its leak. A parent `[8, 32]` clip followed by child `[32, 48]` paints pixel 31 red because both rectangles are expanded before intersection, creating a `[31, 32]` overlap.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bghgary: Please cover a transform that changes rotation, not only translation, and fix the resulting leak. With a 45-degree child transform, `nvgIntersectScissor` approximates the parent as an axis-aligned bounding box; a D3D11 probe paints pixel `(32, 8)` outside a parent clip spanning `[16, 48]`.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bghgary: Please add an adjacent disjoint case and fix its leak. A parent `[8, 32]` clip followed by child `[32, 48]` paints pixel 31 red because both rectangles are expanded before intersection, creating a `[31, 32]` overlap.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bghgary: Please cover a transform that changes rotation, not only translation, and fix the resulting leak. With a 45-degree child transform, `nvgIntersectScissor` approximates the parent as an axis-aligned bounding box; a D3D11 probe paints pixel `(32, 8)` outside a parent clip spanning `[16, 48]`.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
bghgary: Please add an adjacent disjoint case and fix its leak. A parent `[8, 32]` clip followed by child `[32, 48]` paints pixel 31 red because both rectangles are expanded before intersection, creating a `[31, 32]` overlap.

Apps/UnitTests/Source/Tests.JavaScript.cpp
```diff
@@ -79,6 +79,11 @@ TEST(JavaScript, All)
 #else
         env.Global().Set("hasGpuRendering", true);
 #endif
+#if defined(SKIP_RENDER_TESTS) || defined(SKIP_EXTERNAL_TEXTURE_TESTS)
```
bghgary: Please do not tie these Canvas regressions to `SKIP_EXTERNAL_TEXTURE_TESTS`. Vulkan defines it for unavailable imported native-texture helpers, but these tests use `DynamicTexture`; exact-head Vulkan CI skips only these two. Use a dedicated capability or direct Canvas readback so Vulkan retains this coverage.

Apps/UnitTests/JavaScript/src/tests.javaScript.all.ts
```diff
@@ -194,6 +196,209 @@ describe("Canvas2D", function () {
...
+            ctx.restore();
+            ctx.restore();
+            texture.update(false);
+
+            const pixels = await texture.readPixels();
+            if (!(pixels instanceof Uint8Array)) {
+              throw new Error("Expected RGBA8 GPU readback for signed clips");
+            }
+            const description =
+              `rotated=${rotated}, flipX=${flipX}, flipY=${flipY}`;
```
bghgary: The rotated signed-clip test compares every variant against the same leaky positive baseline and checks only `(0, 0)` outside the parent. Please add a boundary assertion and preserve the exact parent intersection across differing transforms; a D3D11 probe still paints `(32, 8)` red outside a parent clip spanning `[16, 48]`.

CedricGuillemet (approved):

PR #1879, opened 2026-09-14
Update bgfx and migrate to explicit swap-chain APIs

bkaradzic-microsoft:
## Summary

Update to bgfx API 161 through merged bgfx.cmake commit `2e8efe6186d1bc311674153c8207c19774bc77a9` (https://github.com/BabylonJS/bgfx.cmake/pull/147), including the synchronized upstream CMake scripts and the EGL/D3D12 fixes in https://github.com/BabylonJS/bgfx/pull/80.

- Bootstrap bgfx without a main window and let Graphics own explicit swap-chain framebuffers. Resize and window replacement preserve the graphics device and its resources.
- Import D3D11 caller-owned RTV/DSV resources while retaining their lifetime and view properties; use the active back buffer for screenshots and continuous capture.
- Coalesce capture requests into one readback and remove the obsolete D3D12 extra-frame screenshot workaround.
- Migrate camera/XR native texture imports, attachment flags, and baseline instancing support to the current API.
- Adapt dependency/install wiring to the synchronized scripts, preserve disabled video support, and deploy the matched DXC runtime pair for desktop x64 D3D12 consumers. Unsupported deployment targets warn rather than receiving incompatible binaries.

## Local validation before opening this PR

| Configuration | Result |
| --- | --- |
| Windows x64 D3D11 / Chakra | UnitTests and Playground build; 40 tests passed |
| Windows x64 D3D12 / Chakra | UnitTests and Playground build; 29 passed, eight existing skips |
| Linux OpenGL / JavaScriptCore / software Mesa | UnitTests and Playground build; 32 passed, two existing skips |
| Android arm64-v8a / API 24 | NativeCamera and NativeXr compile successfully |

Native Canvas pixel validation passes on both D3D11 and D3D12 (1/1 each). New regressions cover swap-chain resize/MSAA/depth modes, window replacement, disable/re-enable, screenshot/capture coalescing and normalization, D3D11 external-view imports, and D3D12 resource states. The debug-layer regression executes and reports zero errors.

Apple SDKs and Android device execution were unavailable locally and require CI/device coverage. Local D3D12 ModuleLoadTest reports an unexpected `cryptsp.dll` module; that result is not counted as a pass or hidden by changing its allowlist.

## Latest rebase and Vulkan fix

- Rebased onto `bb01e283d83da8c043000862c76d9315f76c7a67`, including the merged Canvas font packaging fix (#1881). The original three patches remain unchanged by the rebase.
- Fixed the separate Vulkan failure: X11 WSI was clamping requested 32x24 / 64x48 swap chains to the unchanged 640x480 window. Linux Vulkan now resizes and synchronizes the X11 window before creating or updating the explicit swap chain.
- Clang/RelWithDebInfo Vulkan build under WSL passed. All three failing Device tests passed with lavapipe/Xvfb, with the requested swap-chain dimensions confirmed in logs. Existing capability-based skips were unchanged.
- Windows D3D11/Chakra build and 30/30 focused regressions passed. The replacement Linux Vulkan CI job also passed.
- Replacement CI for `7248189a8b5f31cee975e7023b2eedf7da615900` is green: [run 35237002633](https://github.com/BabylonJS/BabylonNative/actions/runs/35237002633) completed successfully. All 35 checks passed, including `Ubuntu_Clang_Vulkan / build`, with no failures or pending checks.

## Review follow-up

[Gary's review](https://github.com/BabylonJS/BabylonNative/pull/1879#pullrequestreview-5204839815) is addressed in `6d7a68a906198e43cc280ba289eb3c9d100d0f9d`:

- Removed `D3D11Format.h` and `ExternalBackBufferD3D11.h/.cpp` entirely from the PR. The existing D3D11 device implementation owns the back-buffer state and conversion code, with the shared query declared in existing internal `DeviceQueries.h`. The original 105-entry table is preserved verbatim.
- Window depth-only input now explicitly rejects read-only, array and nonzero-mip views that `SwapChain::depth` cannot represent. Supported default views and explicit color/depth framebuffer semantics remain intact. The restriction is documented in the public API.
- Replaced the shader-specific DLL copier with imported DXC runtime dependencies and standard `TARGET_RUNTIME_DLLS` deployment, including ShaderTool and installation. Uses the actual SDK dxcompiler import library and a runtime-only dxil dependency, not a placeholder import library.

Validation: D3D11 RelWithDebInfo build and 27 Device/ExternalTexture tests passed in two shuffled runs; the new read-only rejection regression failed before the fix. D3D12 ShaderTool/ModuleLoadTest builds and real ShaderTool compilation passed. Paired DLL hashes, transitive deployment and generated installation entries were verified. CMake runtime validation used 3.31; 3.21 dependency propagation was checked in its implementation source. The previously documented local ModuleLoadTest `cryptsp.dll` allowlist failure remains unchanged.

### Camera lifetime scope

`8bf2c2203cf6c3197d2249bcf6faea02a2141181` removed the broad Texture shared-state/mutex rewrite and deferred-update API. Apple camera conversion/import remains synchronous in the update call; no deferred callback accesses the C++ wrapper.

`bd2a08fb962d6fd1e99fce740a34f48b4baab78f` addresses Gary's follow-up: the Metal backend borrows the imported object, so retaining it only until the first AfterRender was insufficient. Each import now supplies a native-resource owner held by Graphics::Texture. Camera Close/output replacement cannot release an object while the caller still owns its wrapper. On disposal/recreation, only native ownership crosses BeforeRender -> AfterRender, keeping the object alive until a full render boundary processes the queued handle destruction. No mutex or shared wrapper state is added.

The replacement lifecycle regression is enabled for D3D11 and Metal and keeps the wrapper valid across multiple frames after releasing producer ownership. It covers disposal, wrapper destruction and replacement both before and from AfterRender. The old one-frame retention fails this regression; the final D3D11 build and all 28 focused Device/ExternalTexture tests pass twice. This models producer Close/replacement with native resources, not physical camera execution. Apple SDK/device validation is unavailable locally.

Current CI is green: [run 35292522780](https://github.com/BabylonJS/BabylonNative/actions/runs/35292522780) completed successfully for `bd2a08fb962d6fd1e99fce740a34f48b4baab78f`; all 35 PR checks passed with none pending. All macOS/iOS build and installation jobs passed. Windows QuickJS passed all 314 validation scenes and 56 unit tests, including the new native-owner lifecycle regression. The macOS unit-test job uses the no-op Metal device and skips that native-resource regression, so Apple compilation is verified but real-Metal/physical-camera execution remains unverified. The superseded run for `8bf2c220` was canceled; its Windows QuickJS failure was all 314 snippets returning HTTP 403, while Windows Chakra passed all 314 scenes on the same revision. No validation exclusions, thresholds or retry rules were changed for that service failure.


Core/Graphics/InternalInclude/Babylon/Graphics/D3D11Format.h
```diff
@@ -0,0 +1,206 @@
...
+#define DXGI_FORMAT_ASTC_10X8_UNORM_SRGB  DXGI_FORMAT(175)
+#define DXGI_FORMAT_ASTC_10X10_UNORM      DXGI_FORMAT(178)
+#define DXGI_FORMAT_ASTC_10X10_UNORM_SRGB DXGI_FORMAT(179)
+#define DXGI_FORMAT_ASTC_12X10_UNORM      DXGI_FORMAT(182)
+#define DXGI_FORMAT_ASTC_12X10_UNORM_SRGB DXGI_FORMAT(183)
+#define DXGI_FORMAT_ASTC_12X12_UNORM      DXGI_FORMAT(186)
+#define DXGI_FORMAT_ASTC_12X12_UNORM_SRGB DXGI_FORMAT(187)
+
+namespace Babylon::Graphics::D3D11TextureFormats
+{
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 5f32189c. The ASTC fallback values are now namespaced `inline constexpr DXGI_FORMAT` constants, and the mapping table references those constants directly. The shared header no longer defines global `DXGI_FORMAT_ASTC_*` macros or shadows SDK identifiers.

The fresh installed-SDK consumer build and all 41 D3D11 unit tests pass, including the external-backbuffer regressions moved into their own UnitTests-only source to preserve the public installation boundary.

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -251,15 +267,41 @@ namespace Babylon::Graphics
...
+            });
+
+            ConfigureBgfxRenderType(m_state.Bgfx.InitState);
+            auto init{m_state.Bgfx.InitState};
+            // Own the window framebuffer explicitly so reattaching a window does not
+            // recreate the device or leave bgfx presenting to the previous surface.
+            init.swapChain.nwh = nullptr;
+            init.swapChain.width = 0;
+            init.swapChain.height = 0;
+            init.swapChain.depth = BGFX_INVALID_HANDLE;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: The current ordering is intentional. The real renderer is initialized headlessly, then the host window is attached through an explicit framebuffer. Renderer selection must therefore inspect the requested host surface, not the windowless bootstrap copy.

On Android, moving `ConfigureBgfxRenderType` after clearing `nwh` would select `Noop` for window-backed configurations without an external context, disabling rendering. `ndt` also remains intentional: the bootstrap rendering context must use the host surface's native display, even though it has no main window.

I retained this behavior and added explanatory comments in 5f32189c rather than applying the suggested functional change.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -577,11 +579,11 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
     }
 
-    CameraDevice::CameraDimensions CameraDevice::UpdateCameraTexture(bgfx::TextureHandle textureHandle)
+    CameraDevice::CameraDimensions CameraDevice::UpdateCameraTexture(Graphics::Texture& texture)
     {
-        // Hook into AfterRender to copy over the texture, ensuring that the textureHandle has already been initialized by bgfx.
+        // Hook into AfterRender to copy over the texture without overlapping bgfx's frame encoding.
         // Capture the cancellation token so that the shared pointer is kept alive when arcana checks internally for cancellation.
-        arcana::make_task(m_impl->deviceContext->AfterRenderScheduler(), *m_impl->cancellationSource, [this, textureHandle, cancellationSource{m_impl->cancellationSource}] {
+        arcana::make_task(m_impl->deviceContext->AfterRenderScheduler(), *m_impl->cancellationSource, [this, texture{&texture}, cancellationSource{m_impl->cancellationSource}] {
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 98315df7. The Apple callback now captures a generation-checked `Texture::DeferredUpdate` token backed by shared resource state, not a raw wrapper pointer. Disposal, wrapper destruction, and independent resource recreation invalidate older tokens. Import and disposal use the same state mutex, so disposal either rejects the pending update or runs after the import and invalidates/destroys that resource. This handles explicit deletion as well as wrapper finalization without keeping the JS wrapper alive.

Added three portable regressions covering disposal, destruction, and resource-generation replacement. The integrated D3D11 suite passes all 45 tests; D3D12 passes 33 with eight existing skips; the Linux texture/device/callback suite passes all 11. Android Graphics/NativeCamera also compile. Apple ObjC++ compilation and camera-device execution are not available locally; platform CI will validate the Apple build.
Update: the shared-state approach above is superseded by 8bf2c2203cf6c3197d2249bcf6faea02a2141181 to keep this PR narrowly scoped. Camera conversion and texture import now finish synchronously while the caller owns the wrapper. No queued callback accesses a Graphics::Texture; the only after-frame callback retains the native Metal resource until bgfx has processed the import. This removes the deferred wrapper lifetime problem without mutexes or shared state across every Texture operation.

Texture's pre-mutex implementation is restored. A replacement D3D11 regression covers native-resource retention when the wrapper is disposed, destroyed, or recreated before the next frame finishes; all 28 Device/ExternalTexture tests pass twice. Apple build validation is pending CI; physical camera execution is not available locally.
Further correction in bd2a08fb962d6fd1e99fce740a34f48b4baab78f: retaining a borrowed Metal texture only until import was insufficient, as Gary noted. Imports still complete synchronously, but Texture now holds an optional native-resource owner until disposal/recreation; only that owner is retained across a full BeforeRender -> AfterRender boundary while bgfx processes destruction. Camera Close/output replacement therefore cannot release an object still owned by a caller's wrapper. No mutex, generation token, or deferred wrapper access is reintroduced. The replacement regression is enabled for D3D11 and Metal and keeps wrappers valid beyond multiple AfterRender cycles after producer release. See the current review reply: https://github.com/BabylonJS/BabylonNative/pull/1879#discussion_r4042708300

Core/Graphics/Source/FrameBuffer.cpp
```diff
@@ -51,7 +51,7 @@ namespace Babylon::Graphics
 
     bgfx::FrameBufferHandle FrameBuffer::Handle() const
     {
-        return m_handle;
+        return m_defaultBackBuffer ? m_deviceContext.GetBackBufferHandle() : m_handle;
```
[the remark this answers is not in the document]
bkaradzic-microsoft: Fixed in 98315df7. Default-framebuffer semantics are now separate from device-window routing: only a default wrapper constructed without an explicit handle follows the device's current back buffer. XR eye framebuffers retain their supplied targets, and disposing one cannot make it fall back to the window.

Added `Device.ExplicitDefaultFrameBufferPreservesTarget`. It failed before the fix (wrong handle and the XR-style clear changed window pixels), and now passes on D3D11, D3D12, and Linux software OpenGL. It also covers ordinary explicit targets, lazy device-window routing, and disposal.

Core/Graphics/Source/DeviceImpl.cpp
```diff
@@ -686,19 +722,105 @@ namespace Babylon::Graphics
...
+            bgfx::frame(BGFX_FRAME_DISCARD);
+            if (m_state.BackBufferColor || m_state.BackBufferDepthStencil)
+            {
+                m_externalBackBuffer = std::make_unique<ExternalBackBufferD3D11>(
+                    m_state.BackBufferColor.get(), m_state.BackBufferDepthStencil.get(), swapChain);
+                if (bgfx::isValid(m_externalBackBuffer->GetFrameBuffer()))
+                {
+                    return;
+                }
+                swapChain.depth = m_externalBackBuffer->GetDepthTexture();
```
bghgary: Please preserve the supplied depth-stencil view's subresource and access semantics here, or reject unsupported non-default views. With a window and only `BackBufferDepthStencil` supplied, `ExternalBackBufferD3D11` creates no explicit framebuffer, and this path passes only the imported texture as `swapChain.depth`. Pinned bgfx then reconstructs a writable, non-array view at mip 0, which can target the wrong mip, fail for array-backed views, or conflict with simultaneous shader-resource use.

Plugins/ShaderCompiler/CMakeLists.txt
```diff
@@ -45,6 +45,56 @@ if(GRAPHICS_API STREQUAL "D3D11")
         PRIVATE "d3dcompiler.lib")
 endif()
 
+function(deploy_shader_compiler_runtime TARGET_NAME)
```
bghgary: Please reuse the existing target-runtime deployment path instead of adding a parallel shader-specific copier. Model the DXC files as imported runtime dependencies, as V8 does, and add the standard `$<TARGET_RUNTIME_DLLS:ShaderTool>` copy step for ShaderTool.

Core/Graphics/InternalInclude/Babylon/Graphics/D3D11Format.h
```diff
@@ -0,0 +1,207 @@
...
+#include <bx/bx.h>
+#include <d3d11.h>
+
+#include <optional>
+
+// clang-format off
+
+namespace Babylon::Graphics::D3D11TextureFormats
+{
+    // Copied from bgfx's renderer_d3d.h. These values are defined by newer Windows SDKs, but
```
bghgary: We discussed offline that the extra `D3D11Format` and `ExternalBackBufferD3D11` files should not be added, but that change hasn't happened yet.

bghgary (changes requested):

Concern inline.

Core/Graphics/InternalInclude/Babylon/Graphics/D3D11Format.h
bkaradzic-microsoft: Addressed in 6d7a68a906198e43cc280ba289eb3c9d100d0f9d. D3D11Format.h and ExternalBackBufferD3D11.h/.cpp are now removed entirely from the net PR diff, along with their CMake entries.

The implementation is integrated into the existing DeviceImpl_D3D11.cpp, with back-buffer state owned directly by DeviceImpl instead of an additional helper class/PIMPL. The shared format-query declaration uses the existing internal DeviceQueries.h. The original 105-entry table remains unchanged, avoiding a separate table cleanup or duplicate mapping.

Plugins/ShaderCompiler/CMakeLists.txt
bkaradzic-microsoft: Fixed in 6d7a68a906198e43cc280ba289eb3c9d100d0f9d. Removed the shader-specific copier and its call sites. DXC is modeled with imported SHARED runtime targets: dxcompiler uses the SDK's actual import library, and dxil is a runtime-only imported dependency through IMPORTED_LINK_DEPENDENT_LIBRARIES. Existing application TARGET_RUNTIME_DLLS steps now pick up the pair transitively; ShaderTool has the standard copy step and matching installation handling.

D3D12 ShaderTool and ModuleLoadTest builds pass, and ShaderTool compiles a real shader pair successfully. Both deployed DLL hashes match the bundled bgfx files, including through the indirect ModuleLoadTest dependency. ShaderTool still has no static dxcompiler.dll import. Validation used CMake 3.31; the runtime-only dependency propagation was also checked against CMake 3.21's implementation.

Core/Graphics/Source/DeviceImpl.cpp
bkaradzic-microsoft: Fixed in 6d7a68a906198e43cc280ba289eb3c9d100d0f9d. I took the explicit-rejection option: when a window supplies only a depth view, nonzero mips, array views/resources, and read-only depth/stencil flags are rejected before texture import rather than silently reconstructed differently by bgfx. Writable, non-array, mip-0 views remain supported. Explicit external color/depth framebuffers retain their subresource and access semantics.

Added regressions for all three read-only flag combinations, nonzero mip and array views (including a one-layer array view), and successful recovery with a supported view. The read-only regression failed against the previous revision; all 27 Device/ExternalTexture tests now pass in two shuffled D3D11 runs. The restriction is documented on the public configuration/update API.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
```diff
@@ -577,143 +580,153 @@ fragment float4 fragmentShader(RasterizerData in [[stage_in]],
...
+                texture.Format(),
+                texture.Flags(),
+                reinterpret_cast<uintptr_t>(m_impl->textureRGBA));
+            m_impl->bgfxTexture = &texture;
+            m_impl->bgfxTextureHandle = texture.Handle();
+            m_impl->refreshBgfxTexture = false;
+
+            // Keep only the native resource alive until bgfx processes the import.
+            arcana::make_task(m_impl->deviceContext->AfterRenderScheduler(), arcana::cancellation::none(),
+                [textureRGBA = m_impl->textureRGBA] { (void)textureRGBA; });
```
bghgary: `AfterRender` does not end an imported texture's lifetime. The pinned Metal backend stores the external `MTL::Texture*` without retaining it. On a resolution change, assigning a new `textureRGBA` can release the previous object before bgfx has processed destruction of its handle; after `MediaStream::Stop`, the current object can be released while the caller's `Graphics::Texture` remains valid. Please retain each imported Metal texture until bgfx has finished using its handle and processed its destruction, and add coverage that keeps the wrapper valid beyond `AfterRender` and through Close or replacement.

bghgary (changes requested):

One additional concern inline.

Plugins/NativeCamera/Source/Apple/CameraDevice.mm
bkaradzic-microsoft: You're right — the previous callback retained the Metal object only through import, not for the lifetime of the imported handle. Fixed in bd2a08fb962d6fd1e99fce740a34f48b4baab78f.

Each import now gives `Graphics::Texture` a native-resource owner. Dropping the camera's reference during Close or output replacement cannot release a resource still owned by the caller's wrapper. Disposal/recreation queues `bgfx::destroy` and transfers only the native owner through `BeforeRender` -> `AfterRender`, ensuring a full render boundary processes destruction before release, including disposal from an after-render callback. Imports remain synchronous; no wrapper pointer is captured and no mutex/shared wrapper state is reintroduced.

Replaced the one-frame regression with lifecycle coverage enabled for D3D11 and Metal. It keeps the wrapper valid for multiple frames after releasing producer ownership, then covers disposal, wrapper destruction, and replacement before/from AfterRender. The previous one-frame retention fails the new test. All 28 focused Device/ExternalTexture tests pass twice locally on D3D11. The test models producer Close/replacement with real native resources; physical camera execution was not available locally. CI update: [run 35292522780](https://github.com/BabylonJS/BabylonNative/actions/runs/35292522780) is green for this commit; all 35 PR checks passed, including macOS/iOS builds. Windows QuickJS passed 314/314 validation scenes and 56 unit tests, including the ownership regression. The macOS unit-test job uses the no-op Metal device and skips this regression, so this verifies Apple compilation, not real-Metal or physical-camera execution.

bghgary (approved):

PR #1882, opened 2026-09-17
fix: handle Canvas image parse errors

LDMFD:
## Problem

`bimg::imageParse` creates an internal error scope when it receives no error sink. On Android, malformed or unsupported Canvas `Image` data can therefore reach the debug-break path and terminate the process instead of reporting a normal image-load failure.

## Fix

Pass `bx::ErrorIgnore` to the Canvas image decoder in `NativeCanvasImage::SetBuffer`. `bimg::imageParse` now returns `nullptr` for a parse failure, and the existing `SetSrc` path invokes the JavaScript `onerror` callback.

`NativeEngine::ParseImage` and `ParseCubeImage` already use this pattern on `master`; this change covers the remaining Canvas decoder call site.

## Validation

- Reviewed the existing base64 and URL failure paths: a null image container reaches `HandleLoadImageError`.
- Verified the corresponding Android arm64 Babylon React Native native module builds successfully with this error-sink pattern.


LDMFD:
@microsoft-github-policy-service agree

LDMFD:
Opened by Codex after I experienced native crashing in an Expo app. LMK if the verbiage is too obscure.

bkaradzic-microsoft:
Opened #1893 for the separate Canvas Image.src reload/cancellation lifetime bug. It preserves the existing parse call and error-reporting policy, retains #1887 PNG normalization, and does not add SVG support. Your parser error handling remains needed; these changes should compose without competing decode implementations.

